PR #1, opened 2015-07-21
Initial stacks

parente:
I'm submitting this PR not expecting it to get merged as-is, but as a strawman for discussion. It contains:

1. `minimal-notebook` which acts as a base image or a lightweight notebook environment in which the user can use conda to install everything
2. `scipy-notebook` for a Python 2 and 3 stack
3. `r-notebook` for an R stack
4. READMEs for each

Most package versions are pegged to allow patch level increases only to avoid breakage across Docker builds. (I've learned this the hard way.)

If this is inline with what people are thinking for this repo, follow-ons might be:

1. `all-spark-notebook` (Scala, Python, R w/ Spark 1.4.x plus Yarn and Mesos connectivity preconfigured)
2. `data-science-notebook` (needs a better name) with Python, R, Julia, Scala, Spark (Brian's request)
3. `julia-notebook` maybe

I don't want to go too crazy with a ton of different options yet until we know if less is more or not.

Ideally, once these (or alternatives) are merged, the jupyter account on Docker Hub can be configured to build the minimal-notebook stack on commit and webhook build the other stacks on build of that base.

### DIFF
--- a/minimal-notebook/Dockerfile
+++ b/minimal-notebook/Dockerfile
@@ -1,0 +1,71 @@
+FROM debian:jessie
+
+MAINTAINER Jupyter Project <jupyter@googlegroups.com>
+
+# Install all OS dependencies for fully functional notebook server
+ENV DEBIAN_FRONTEND noninteractive
+RUN apt-get update && apt-get install -yq --no-install-recommends \
+    git \
+    vim \
+    wget \
+    build-essential \
+    python-dev \
+    ca-certificates \
+    bzip2 \
+    unzip \
+    libsm6 \
+    pandoc \
+    texlive-latex-base \
+    texlive-latex-extra \
+    texlive-fonts-extra \
+    texlive-fonts-recommended \
+    supervisor \
+    sudo \
+    && apt-get clean
+
+ENV CONDA_DIR /opt/conda
+ENV NB_USER jovyan
+
+# Install conda
+RUN echo 'export PATH=$CONDA_DIR/bin:$PATH' > /etc/profile.d/conda.sh && \
+    wget --quiet https://repo.continuum.io/miniconda/Miniconda3-3.9.1-Linux-x86_64.sh && \
+    /bin/bash /Miniconda3-3.9.1-Linux-x86_64.sh -b -p $CONDA_DIR && \
+    rm Miniconda3-3.9.1-Linux-x86_64.sh && \
+    $CONDA_DIR/bin/conda install --yes conda==3.14.1
+
+# Create non-root user
+RUN useradd -m -s /bin/bash $NB_USER
+RUN chown -R $NB_USER:$NB_USER $CONDA_DIR
+RUN chown $NB_USER:$NB_USER /home/$NB_USER -R
+
+# Configure user environment
+USER $NB_USER
+ENV HOME /home/$NB_USER
+ENV SHELL /bin/bash
+ENV USER $NB_USER
+ENV PATH $CONDA_DIR/bin:$PATH
+
+# Setup a work directory rooted in home for ease of volume mounting
+ENV WORK $HOME/work
+RUN mkdir -p $WORK
+WORKDIR $WORK
+
+# Install Jupyter notebook
+RUN conda install --yes \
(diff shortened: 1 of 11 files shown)


minimal-notebook/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+    texlive-latex-base \
+    texlive-latex-extra \
+    texlive-fonts-extra \
+    texlive-fonts-recommended \
+    supervisor \
+    sudo \
+    && apt-get clean
+
+ENV CONDA_DIR /opt/conda
+ENV NB_USER jovyan
```
rgbkrk: Ok, cool, I'm glad you adopted a non-root convention here as well.

We're going to end up getting reports about file permissions during mounting but we'll cross that bridge (again) later.

minimal-notebook/README.md
```diff
@@ -0,0 +1,26 @@
...
+docker run -d -p 8888:8888 jupyter/minimal-notebook
+```
+
+## Options
+
+You may customize the execution of the Docker container and the Notebook server it contains with the following optional arguments.
+
+* `-e PASSWORD="YOURPASS"` - Configures Jupyter Notebook to require the given password. Should be conbined with `USE_HTTPS` on untrusted networks.
+* `-e USE_HTTPS=yes` - Configures Jupyter Notebook to accept encrypted HTTPS connections. If a `pem` file containing a SSL certificate and key is not found in `/home/jovyan/.ipython/profile_default/security/notebook.pem`, the container will generate a self-signed certificate for you.
+* `-e GRANT_SUDO=yes` - Gives the `jovyan` user passwordless `sudo` capability. Useful for installing OS packages. **You should only enable `sudo` if you trust the user or if the container is running on an isolated host.**
```
rgbkrk: Ah, nice.

minimal-notebook/notebook.conf
```diff
@@ -0,0 +1,24 @@
+# Licensed Internal Code - Property of IBM
+# Package: com.ibm.swget.ka
+# (c) Copyright IBM Corp. 2013, 2015
+# All Rights Reserved
+# US Government Users Restricted Rights - Use, duplication or disclosure
+# restricted by GSA ADP Schedule Contract with IBM Corp.
```
rgbkrk: Did you mean to include this part?

minimal-notebook/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+    terminado \
+    && conda clean -yt
+
+# Configure Jupyter
+RUN ipython profile create
+
+# Configure container startup
+EXPOSE 8888
+USER root
+CMD ["supervisord", "-n", "-c", "/etc/supervisor/supervisord.conf"]
```
rgbkrk: I'm not a fan of supervisord on the inside here as part of a minimal setup, but I'm happy to hear your reasons why.

rgbkrk:
These READMEs and setups are *really* great @parente. :+1: Very thorough and well thought out.

minimal-notebook/notebook.conf
parente: Heh. Copy/paste error. I'll rebase and clean it out.

minimal-notebook/Dockerfile
parente: Yes. I think this is where additional doc or helper scripts can help.

minimal-notebook/Dockerfile
parente: A couple reasons:

1. If I make changes to the default Jupyter config within my container, I want to restart the notebook server. If the server is the entrypoint of the container, I can't do this: the whole container stops.
2. supervisord acts as a [proper init process](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/) whereas I'm sure Jupyter Notebook was designed for that.

parente:
I've force pushed to condense the commits and remove the mistaken header.

rgbkrk:
Cool, I think these look good.

saif920 (approved):

PR #3, opened 2015-07-29
Notebook stack for Python w/ Spark on Mesos

parente:
Plus README explaining how to use pyspark in local mode and with a Mesos cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,14 @@
 
 OWNER:=jupyter
 STACK?=
+ARGS?=
+DARGS?=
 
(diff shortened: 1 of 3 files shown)


parente:
Should not be merged until after PR #4

minrk:
diff looks a little weird on the config file. Maybe rebase this and the other mesos PR now that #4 is merged?

parente:
Guess the merge commit threw it off. Rebased and now the diff looks correct here and in PR #5.

PR #4, opened 2015-08-10
Make NotebookApp.ip and .port configurable

parente:
Necessary when running Docker with --net=host to avoid unknown port jumps and exposure on all host interfaces


minimal-notebook/ipython_notebook_config.py
```diff
@@ -4,9 +4,9 @@
 PEM_FILE = os.path.join(os.path.dirname(__file__), 'security/notebook.pem')
 
 c = get_config()
-c.NotebookApp.ip = '*'
+c.NotebookApp.ip = os.getenv('INTERFACE', '*')
```
minrk: I like to use `os.getenv('Key', '') or 'DEFAULT'` to handle the case of `KEY=`, which should be treated the same as KEY being undefined.

parente: Makes sense. Force pushed with that change.

PR #14, opened 2015-08-23
Upgrade to Jupyter Notebook 4.0

parente:
* Change minimal-notebook to install notebook=4.0*
* Change config and pem file paths for jupyter 
* Install ipywidgets in all containers that have a Python stack
* Update all READMEs to describe v3.2 and v4.0 since Docker Hub only shows one README for all tags

Doing this all in a 4.0.x branch first to check how it builds on Docker Hub under a 4.0 tag. If all goes well, will merge to master here and latest there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,8 @@ Note: Branches / tags reflect the version of the primary process in each contain
 
 | Docker Hub repo | GitHub source | Git Branch &rarr; Docker Tag |
 | --------------- | ------------- | ---------------------------- |
-| [jupyter/minimal-notebook](https://hub.docker.com/r/jupyter/minimal-notebook/) | [minimal-notebook](./minimal-notebook) | master &rarr; latest <br /> 3.2.x &rarr; 3.2 |
-| [jupyter/scipy-notebook](https://hub.docker.com/r/jupyter/scipy-notebook/) | [scipy-notebook](./scipy-notebook) | master &rarr; latest <br /> 3.2.x &rarr; 3.2 |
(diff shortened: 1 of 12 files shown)


rgbkrk:
Ok, sweet. Will this be the right one for us to base `jupyter/demo` off, or one of the other stacks?

parente:
> Will this be the right one for us to base jupyter/demo off, or one of the other stacks?

The jupyter/minimal-notebook? I guess we could rebase the demo image on it, but it lacks the branding, CSP config, etc. that are in the current jupyter/minimal. Those will need to move into the demo image itself.

rgbkrk:
We can always move the config step to *after* the installation steps. Most of that old ordering was about caching order for my own sanity during dev builds.

parente:
That sounds fine. So how about:

1. this goes in as the 4.0 branch
2. we do all the work to get that tag building and then rebase master on it (i.e., complete issue #14)
3. we open / work the issue against docker-demo-images to rebase that repo on jupyter/minimal-notebook:4.0 once it hits Docker Hub.

If PR #13 goes into master before (2) above completes, we'll need to bring it up to 4.0 also. If it goes in after, we'll need to merge it against the 3.2.x branch instead and then bring it up to 4.0. And regardless of order, it needs to go into the 3.2.x branch anyway for stable builds against that version.

rgbkrk:
1. :white_check_mark: 

I'm going to leave #13 decisions up to you.

rgbkrk:
To save yourself headaches, you may find adapting to only 4.0.x to be easiest (since config files should stay in the same place), but not all kernels are adapted. :confused:

Issue #17, opened 2015-08-27
conda irkernel downgrades ipython-notebook to 3.2.1


parente:
Whelp, conda is downgrading the ipython package back to 3.2.1. Everything appeared to be working smoothly and so I missed the fact that the r-notebook and all-spark-notebook images were actually running notebook server v3.2.1 in the 4.x branch and builds. Doh!

Scrolling back through the build logs, I definitely see the message from conda stating that it was downgrading the package. I'll have to see if there's a flag to error on this case instead of proceeding unabated.

@takluyver Is the R kernel compatible with notebook 4.0.x? Is this just a conda packaging update problem?

rgbkrk:
This is probably just a packaging update problem. I'm using the R kernel with 4.x right now. I cc'ed our main poc in your PR.

parente:
cherry-picked into 4.0.x too. Will leave this open to revert when upstream packaging is updated.

parente:
r-irkernel is fixed upstream. I reverted the workaround. Will kick off rebuild.

PR #18, opened 2015-08-27
Prevent r-irkernel from downgrading to ipython 3.2.1

parente:
Workaround for issue #17.  

Force irkernel to install without deps, then install the irkernel deps. Manually built and tested locally with some notebooks to see if irkernel is minimally functional at least.

@rgbkrk OK as a workaround? IMHO, better than having 4.0 tagged images out there that are really 3.2.1. (Surprise! :tada:)

### DIFF
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -79,10 +79,23 @@ RUN $CONDA_DIR/envs/python2/bin/python \
 # R packages
 RUN conda config --add channels r
 RUN conda install --yes \
-    'r-base=3.1*' \
-    'r-irkernel=0.4*' \
+    'r-base=3.2*' \
     'r-ggplot2=1.0*' \
     'r-rcurl=1.95*' && conda clean -yt
+
+# Issue #17: R downgrades ipython to 3.2.1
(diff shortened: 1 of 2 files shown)


rgbkrk:
Man! That's too bad. I think that's a fine workaround.

/cc @asmeurer - R kernel depends on an old version of IPython

rgbkrk:
:shipit:

asmeurer:
Oh, I guess I should change the dependency to `jupyter`.

rgbkrk:
Thank you as always Aaron. :bow:

asmeurer:
I'll rebuilt it tomorrow.

asmeurer:
FWIW a better workaround would be to run `conda install jupyter` after the current conda command, since conda doesn't actually check existing installed packages when installing things. Also for this reason, in general, it's better to put everything into one conda install command, so that the dependency solver can keep everything consistent (this is especially a good idea for R since conda doesn't even check what R version is installed like it does with Python).

asmeurer:
OK `r-irkernel` depends on `jupyter` now.

parente:
@asmeurer Thank you!

Issue #20, opened 2015-08-29
3.2 tagged images with R now run notebook 4.0


parente:
Inverse problem of the other day. The conda R package now prereqs jupyter 4.0 which winds up upgrading the notebook server. Any way to get these images back on 3.2.1?

rgbkrk:
Well that's funny. We could also pin with the old version of the R kernel too right?

parente:
Doh! You're probably right. For some reason I assumed the dependency was updated with a corresponding version bump on irkernel itself. I'll have a look.

parente:
Addressed by pinning the version and build number in the 3.2.1 brach. Good catch.

Issue #15, opened 2015-08-26
Docker Hub automated builds and undesirable rebuilds


parente:
I've been watching how the automated builds execute on Docker Hub as the number of branches and stacks have grown. With jupyter/minimal-notebook autobuild-on-push turned off, autobuilds for all other stacks turned on, and manual rebuilds of minimal triggering autobuild of all other stacks, there are numerous wasted rebuilds. For example:

- If minimal-notebook changes on git branch master, DH rebuilds latest from master, 3.2 from branch 3.2.x, and 4.0 from branch 4.0.x. The latter two should have been left alone.
- When the above builds complete successfully, **each one** triggers a rebuild of tag latest from master, tag 3.2 from branch 3.2.x, and tag 4.0 from branch 4.0.x for every other stack. That's `(# of branches)^2 x (# of stacks)` rebuilds when just one tagged image for each should have been touched.

This appears to be a fact of life on Docker Hub. We can cut down on the `branches^2` noise by removing the automated triggers from minimal-notebook to all the others if we remember to trigger each one manually every time minimal changes. But even this results in wasted builds. 

Worse, every rebuild potentially creates new image layer SHAs from different build machine which means poor users that decide to pull and pull again will wind up downloading the entire image over again, even if the layers are binary equivalent. Worse, differences may sneak in during undesirable rebuilds since we don't pin OS packages and lead to two equivalently tagged images having slightly different contents.

It appears this behavior is a fact of life with automated builds on DH which is unfortunate because they are more transparent than the build-and-push-yourself flavor. The only way I see around these problems completely is automating the builds more intelligently outside of DH using a third party service or rolling our own with a little GitHub listener.

rgbkrk:
Not to mention the fact that builds race each other - if two builds are running for the same image, it's the last to complete that becomes the current version of the image.

parente:
I was starting to toy with a Jenkins server out on a VM in the cloud that can listen for notifications from GitHub (there's a Jenkins service) and run a little custom script that checks:

* What branch is the change on?
* What folder(s) are the changes in?

Based on that info, it can build the tiny dependency graph of what needs to be rebuilt and pushed to Docker Hub.

rgbkrk:
That or we split these out into separate repos.

On that same line, I do have something that may be aged by now for triggering actions based on Docker webhooks (https://github.com/rgbkrk/autodock)

parente:
[...]

Issue #12, opened 2015-08-20
Proposal for versioning, immutable historical builds


parente:
## Current Setup

* Docker Hub (DH) automatically builds git master and tags it with latest
* DH automatically builds version branches like 4.0.x and tags them with 4.0
* We only ever roll the latest master branch and latest version branch forward with fixes.

## Problems

* We don't tag per git commit so that users can easily roll back to prior Docker image versions. This is important when major libraries change (e.g., Spark). Some users want the latest which should go into master, while others want to stay on the current version.
* Changes on any branch in the git repo causes a build storm on Docker Hub. (See issue #15). All tags jump to the latest build even when nothing changed in the build definition.

## Proposed Improvement

* Stop relying on Docker Hub automated builds. (We need more control.)
* Stop relying on the 3.2.x and 4.0.x branches and branches in general. (Too coarse grained.)
* Adopt the mantra that when a PR wants to bump a library version in one of the stacks, we take it without question.
* Setup a build VM that maintainers have access to easily pull from this repo and build images. (Manually for now. We can automate later.)
* Beef up the Makefile so that a `make latest` (or some such) does the following:
  * Builds the necessary stacks (change in minimal = build everything, change in scipy = just that stack, etc.) from master HEAD
  * Tags the latest built images (even if they were not rebuilt just now) with `latest`
  * Tags those same images with the current git commit SHA
  * Pushes all of those tags and new builds to Docker Hub (Note: The client will be smart about only sending deltas / tag metadata for images that did not change.)

## Net Result for End Users

For people that want to walk-up-and-use the latest and greatest:

```
docker run jupyter/some-stack-name
```

For people that want to depend on a specific container image configuration tied to some point in time in the docker-stacks git history:

```
docker run jupyter/some-stack-name:<some-git-sha>
```

where GitHub / git makes the contents of that particular tagged image visible to the user (i.e., find the SHA in git and look at the Dockerfiles).

parente:
I should add, that in order to affect this change, we'll need to **delete** the existing automated rebuild repos and recreate them as manual repos from pushed images. There will be an window during this process during which the images are not available. There's just no way to manually push to an automated build repo or convert it to a manual repo. 

Maybe we could ask Docker support, but I doubt it'll be possible.

rgbkrk:
We've definitely gone through this before for when I moved `jupyter/demo` over to manual builds. Momentary downtime for this is going to happen.

rgbkrk:
[...]

PR #21, opened 2015-08-29
Add user in script + fixes

parente:
Two additional commits to PR #16 for doc updates, subimage Dockerfile updates, retaining $WORK path compatibility with existing images, and avoiding recursive permission setting on restarts.

@rgbkrk could use another set of eyes on this.

/cc @costerwi


all-spark-notebook/Dockerfile
```diff
@@ -3,8 +3,6 @@ FROM jupyter/minimal-notebook
 
 MAINTAINER Jupyter Project <jupyter@googlegroups.com>
 
-USER root
```
rgbkrk: Is this assumed to be root at this point? We are running `apt-get` right after, so we might as well be explicit.

rgbkrk:
These are nice global installation setups. :+1:

minimal-notebook/notebook.conf
```diff
@@ -1,18 +1,10 @@
 [program:notebook]
 user=jovyan
```
rgbkrk: The minimal notebook specifies `NB_USER` and `environment` below makes use of it. I *would normally* suggest that `NB_USER` then gets propagated, but then a user only has to set `NB_USER` to `root` to elevate permissions.

all-spark-notebook/Dockerfile
parente: None of the Dockerfiles do any kind of USER switching anymore after these changes are applied. However, explicit is better than implicit so I'm fine adding a USER root to the beginning of every Dockerfile just to make it clear.

minimal-notebook/notebook.conf
parente: Would be nice except supervisor doesn't respect env vars in the user field as @costerwi pointed out in PR #20.

parente:
Now has USER root explicitly stated.

costerwi:
These are good changes!  I especially like checking for whether the user was already created.  A few comments:

* We use `$NB_USER` and `jovyan` so interchangeably that it really isn't a variable right now; it would break everything if it were changed.  Should we just replace `$NB_USER` with `jovyan` everywhere to avoid misleading people?
* It's too bad dependent projects need to mount into /home/jovyan/work.  It makes the script awkward and it's more typing for everybody.  Maybe it would make things a little easier if we specify a unique home directory for jovyan with `useradd -d /jovyan ...` ?

parente:
> We use $NB_USER and jovyan so interchangeably that it really isn't a variable right now; it would break everything if it were changed. Should we just replace $NB_USER with jovyan everywhere to avoid misleading people?

Agreed.

> It's too bad dependent projects need to mount into /home/jovyan/work. It makes the script awkward and it's more typing for everybody. Maybe it would make things a little easier if we specify a unique home directory for jovyan with useradd -d /jovyan ... ?

My concern here is people already running `docker run -v /jupyter_notebook_config.py:/home/jovyan/.jupyter/ jupyter/some-notebook-stack:4.0` to volume mount the notebook directory on the host. If we change the path to anything else, the next time they bring pull the 4.0 image, they're going to be in for a surprise. I feel that paths, ports, user names, etc. are part of the "API" of the container.

Really, this is a problem with the versioning scheme on the container images. It would be nice to have tags 4.0_0, 4.0_1, etc. indicating the version of the notebook server and the build of the container that might introduce a breaking change. Unfortunately, the only way to make this happen with automated builds on Docker Hub is to create a new branch for each change like this or new subfolders within the repo. I am interested in solving this problem (issue #15, issue #12), but it might take some more time and thinking to do right.

rgbkrk:
Ok, awesome. Certainly happy with this.

rgbkrk:
> My concern here is people already running docker run -v /jupyter_notebook_config.py:/home/jovyan/.jupyter/ jupyter/some-notebook-stack:4.0 to volume mount their config from the host. If we change the path to anything else, the next time they bring pull the 4.0 image, they're going to be in for a surprise. I feel that paths, ports, user names, etc. are part of the "API" of the container.

That is definitely a difficult balance. Once we unintentionally make a standard, that's... the one people are going to rely on. However, these images haven't been around *that* long. It's the kind of thing where you can post to the mailing list and other comms, hoping to reach some folks before making sweeping changes.

parente:
Would it be OK to continue the path change in a separate issue? It's working in the current notebook-start script, only adds one additional if-block, and is independent of the UID work.

rgbkrk:
Seems fine to me. I'm not of strong opinion here.

costerwi:
This is a good compromise.  Thanks for updating!

PR #22, opened 2015-08-29
Pin irkernel build to avoid auto-upgrade to notebook 4.0

parente:
Fixes issue #20


parente:
Seems like a reasonable workaround to put into the 3.2.x branch. @rgbkrk agree?

rgbkrk:
Works for me.

Issue #25, opened 2015-09-09
Unable to automate installation of bash_kernel (on top of jupyter/minimal-notebook)


mjbright:
OK, so this is more of a question ... than an issue.

I wanted to add bash_kernel to jupyter/minimal-notebook.

Not quite sure about how start-notebook.sh is being used I decided to make a modified copy into which I added the following lines before the final exec line.

        CBIN=/opt/conda/bin                                                                                                                            
                                                                                                                                               
        conda install pip                                                                                                                              
                                                                                                                                               
        sudo -u $NB_USER $CBIN/pip install --user bash_kernel                                                                                          
        sudo -u $NB_USER $CBIN/python -m bash_kernel.install                                                                                           

However these commands don't seem to be executed when invoked by the CMD of the Dockerfile.
Connecting to the docker container and rerunning /start-notebook.sh installs the kernel as expected.

Why is this manual step necessary, and how what should I do so that it's not necessary?

parente:
I haven't tried it, but I think I would opt to put all additional install steps in a new Dockerfile that derives FROM the minimal notebook like so:

```
FROM jupyter/docker-stacks:4.0

/opt/conda/bin/pip install --user bash_kernel
/opt/conda/bin/python -m bash_kernel.install
```

I would `docker build -t jupyter/bash-notebook .` and spawn a container from that image.

The start-notebook.sh script is really about late-creation of the `jovyan` user and housekeeping to ensure permissions are correct for that user in the home directory and the conda directory.

mjbright:
Yes, that was the initial approach I'd taken until I realized that this
wasn't working as bash_kernel was installing only for the current, i.e.
root user, so I moved it after the jovyan user creation (so out of the
Dockerfile into start-notebook.sh).

Whilst bash_kernel could install system-wide, that would solve the problem-
 I think would be interesting to be able to do derivations after the
creation of the jovyan user.

I guess there were good reasons to move the jovyan user creation out of the
Dockerfile ...

parente:
> Whilst bash_kernel could install system-wide, that would solve the problem-

Ah, yes, removing `--user` would do the trick. I see it now.

>  I think would be interesting to be able to do derivations after the
creation of the jovyan user.

Do you have a use case in mind? I'm trying to think of oneand have come up short (but it's still early in the day ;)

> I guess there were good reasons to move the jovyan user creation out of the
Dockerfile ...

[...]

PR #28, opened 2015-09-10
Recursively set group permission on $CONDA_DIR

costerwi:
Subdirectories inside $CONDA_DIR had jovyan group ownership but didn't have group write permission so jovyan couldn't install packages.

### DIFF
--- a/minimal-notebook/start-notebook.sh
+++ b/minimal-notebook/start-notebook.sh
@@ -16,7 +16,7 @@ if [ -z "$user_exists" ] ; then
 
     # Allow NB_USER group to update conda root env
     chown -R root.$NB_USER $CONDA_DIR
-    chmod g+w $CONDA_DIR
+    chmod -R g+w $CONDA_DIR
 fi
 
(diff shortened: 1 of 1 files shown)


parente:
Question, not so much about this PR, but how late-user creation is surfacing problems with other kernels (issue #25  and issue #30). How important is matching the group ID within the container to the group ID host mounted in your use case? I'm looking at restoring user creation within the dockerfile and having the start script change the uid/gid. I'm wondering if I can keep the /home/jovyan and /opt/conda group ownership at root and simply switch the user to a new UID / GID without also having to recurse and update all the group ownerships on disk (one of the problems you're trying to work around in your PRs too).

costerwi:
I have the SetGID bit set on my host notebook directory so the GID of jovyan doesn't seem to cause me troubles so far.  I'm sure it would be helpful to set NB_GID if that's possible.

If you set group ownership of $CONDA_DIR to something other than root, you can add that as a supplementary group for jovyan (useradd -G) which allows him to modify $CONDA_DIR without giving full reign to the whole system.

parente:
Subsumed by PR #31

PR #29, opened 2015-09-10
Preserve notebook ownership on host

costerwi:
This is an effort to prevent the container from changing the ownership of my host's notebook directory and files when it is mounted to `/home/jovyan/work` .


parente:
Subsumed by PR #31

Issue #30, opened 2015-09-10
datascience-notebook: Julia kernel not available for selection 


pdmct:
Julia kernel doesn't appear in the list of available kernels in the Jupyter interface. (Only Python2,3 and R are listed)

pdmct:
I tried manually updating the kernelspec by adding to Dockerfile:

`RUN $CONDA_DIR/envs/python2/bin/python     $CONDA_DIR/envs/python2/bin/ipython     kernelspec install --replace --name ijulia /root/.ipython/kernels/julia-0.3/`

Which gets Julia to show up in the available kernel list -- but kernel doesn't run/crashes when I try to do anything.

parente:
Which docker image tag did you pull? latest?

parente:
I see it too. Looks like another problem with the kernel spec installing into the active user directory (root) when the dockerfile is running instead of system wide. This broke when we moved the jovyan user creation out of the Dockerfile and into the start script to give control over the UID (PR #16)

The possible fix for issue #25 will probably fix regression too.

parente:
With PR #31 and this change in the datascience-notebook/Dockerfile everything works again:

```
USER jovyan

# Install IJulia packages as jovyan and then move the kernelspec out
# to the system share location. Avoids problems with runtime UID change not 
# taking effect properly on the .local folder in the jovyan home dir.
RUN julia -e 'Pkg.add("IJulia")' && \
    mv /home/$NB_USER/.local/share/jupyter $CONDA_DIR/share/ && \
    chmod -R go+rx $CONDA_DIR/share/jupyter && \
    rm -rf /home/$NB_USER/.local/share
RUN julia -e 'Pkg.add("Gadfly")' && julia -e 'Pkg.add("RDatasets")'

USER root
```

PR #31 needs more testing before merging however.

parente:
IJulia just had a regression against Julie 0.3 that will prevent builds even after the PR mentioned above goes into master here. 

https://github.com/JuliaLang/IJulia.jl/commit/e01b2e44f18b2f1dbcc167273949dd2ac8258dea#commitcomment-13197892

parente:
Now fixed in the `latest` and `4.0` tags on Docker Hub.

Issue #24, opened 2015-09-04
Don't use supervisor as the init process


parente:
See https://github.com/jupyter/notebook/issues/334#issuecomment-137703722 for background.

Possible fixes ...

1. Switch to phusion/baseimage. Concern: size and change of OS.

  ```
  phusion/baseimage              latest              e9f50c1887ea        7 weeks ago         237.7 MB
  ubuntu                         14.04               d2a0ecffe6fa        8 weeks ago         188.4 MB
  debian                         wheezy              065218d54d7d        11 weeks ago        84.97 MB
  ```

2.  Use runit like phusion/baseimage does (http://www.sourcediver.org/blog/2014/11/17/using-runit-in-a-docker-container/). Concern: env var workaround described therein.

3. Port phusion/baseimage to debian. Concern: yet another thing to maintain.

4. Wait for docker to give reaping "for free" (https://github.com/docker/docker/issues/11529).

krallin:
Hey there,

I apologize for the shameless plug, but I noticed you were considering using `phusion/baseimage` for its zombie reaping capabilities but were concerned about the size and having to port it to Debian. 

I happen to have written a tool just for that. It's called `tini`, and it's used in a few official Docker images (see here: https://github.com/docker-library/official-images#init). You might want to have a look.

If you're curious; you can find it here on Github: https://github.com/krallin/tini

mjbright:
I love the name already, 'tini' !

parente:
@krallin thanks for sharing! I will definitely look into it.

parente:
Started a branch with tini replacing supervisord. Works as advertised. I need to beat up each image a bit more to ensure there's no surprises.

One point worth clarifying: we're still going to have to set tini to run as root in the images, not because it needs to do so but because our start-notebook.sh script supports options that require root privileges, namely the jovyan UID switch and granting sudo.

jakirkham:
Sounds cool. I've been trying out tini, as well. I'd be curious to know what you find out from your stress tests, @parente.

PR #31, opened 2015-09-11
Create jovyan in Dockerfile, reset UID at startup

parente:
* Create user jovyan with UID=1000 in the default users group in the Dockerfile
* Set group ownership of user home and conda to root to avoid 'users' group from host access when mounted
* Set stick bit on both paths so root owns subdirs too
* Change jovyan UID if NB_UID is specified and is not the default 1000

@costerwi Can you give this a shot and see if it retains what you're looking for via the NB_UID runtime env var plus fixes the problems you were correcting in your latest PRs? Meanwhile, I'm going to see if it fixes both the bash and julia kernel install problems people have reported with the current setup.

### DIFF
--- a/minimal-notebook/Dockerfile
+++ b/minimal-notebook/Dockerfile
@@ -34,7 +34,6 @@ ENV NB_UID 1000
 
 # Install conda
 RUN mkdir -p $CONDA_DIR && \
-    chmod g+s $CONDA_DIR && \
     echo export PATH=$CONDA_DIR/bin:'$PATH' > /etc/profile.d/conda.sh && \
     wget --quiet https://repo.continuum.io/miniconda/Miniconda3-3.9.1-Linux-x86_64.sh && \
     /bin/bash /Miniconda3-3.9.1-Linux-x86_64.sh -f -b -p $CONDA_DIR && \
@@ -50,12 +49,11 @@ RUN conda install --yes \
 # Create jovyan user with UID=1000 and in the 'users' group
 # Grant ownership over the conda dir and home dir, but stick the group as root.
 RUN useradd -m -s /bin/bash -N -u $NB_UID $NB_USER && \
-    chmod g+s /home/$NB_USER && \
     mkdir /home/$NB_USER/work && \
     mkdir /home/$NB_USER/.jupyter && \
     mkdir /home/$NB_USER/.local && \
-    chown -R $NB_USER:root $CONDA_DIR && \
-    chown -R $NB_USER:root /home/$NB_USER
+    chown -R $NB_USER:users $CONDA_DIR && \
+    chown -R $NB_USER:users /home/$NB_USER
 
 # Configure container startup
 EXPOSE 8888
@@ -65,4 +63,4 @@ CMD [ "start-notebook.sh" ]
 COPY start-notebook.sh /usr/local/bin/
 COPY notebook.conf /etc/supervisor/conf.d/
 COPY jupyter_notebook_config.py /home/$NB_USER/.jupyter/
-RUN chown -R $NB_USER:root /home/$NB_USER/.jupyter
+RUN chown -R $NB_USER:users /home/$NB_USER/.jupyter
--- a/minimal-notebook/Dockerfile
+++ b/minimal-notebook/Dockerfile
@@ -29,11 +29,15 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
 # Configure environment
 ENV CONDA_DIR /opt/conda
 ENV PATH $CONDA_DIR/bin:$PATH
+ENV NB_USER jovyan
+ENV NB_UID 1000
 
 # Install conda
-RUN echo export PATH=$CONDA_DIR/bin:'$PATH' > /etc/profile.d/conda.sh && \
+RUN mkdir -p $CONDA_DIR && \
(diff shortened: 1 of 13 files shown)


parente:
Argh, docker fun:

```
drwx--S--- 4 jovyan root 4096 Sep 11 02:58 share
jovyan@a7de91967404:~/.local$ cd share/
bash: cd: share/: Permission denied
```

So after installing ijulia, the `/home/jovyan/.local` folder gets created. Later, when the script switches the jovyan uid to 1005 from 1000 (for testing), the jovyan user can no access the .local folder even though that user clearly owns the folder. All other folders (e.g., .jupyter) it has no problems with. Poking around a bit in the docker issues, it looks like this could be a storage driver specific thing (e.g., AUFS).

I tried various workarounds to no avail. Might be at a dead end.

parente:
Found slight workarounds for the docker permission weirdness for both ijulia and bash kernel. Will kick off builds of all the other subimages tonight and see if anything else breaks with this change.

parente:
I've locally rebuilt every stack using the new minimal-notebook image defined by the Dockerfile changes in this PR. I've confirmed the changes fix #30 and make #25 possible without any serious hoop-jumping. I've run each image in the following two ways (i.e., with and without a UID change for jovyan):

```
make dev STACK=scipy-notebook'
make dev STACK=scipy-notebook DARGS='-e NB_UID=1005'
```

In every case, I've confirmed the containers to be working properly, meaning:

* all expected kernels appear and function
* jovyan has full access to its home directory
* jovyan can conda install / upgrade anything in any conda environment

Remaining problem: I don't know what I don't know. For instance, does setting the default group for jovyan to `users` to avoid having to reset groups everywhere when the UID changes cause a problem for certain host mounted use cases?

I'm just looking to be a bit more careful up front this time to avoid having to come back to updating the base again and again with slight tweaks for different user / mount scenarios.

@rgbkrk Can I get a quick sanity check?

rgbkrk:
Looking really good. Only thing I'm wondering is what the reasoning is for ending as root user in the final user bound images.

parente:
Turns out that if you don't switch back to root, the CMD from the base image launches as the last USER. :eyes:

costerwi:
I agree; this has great promise.  Simpler is better!

minimal-notebook/Dockerfile
```diff
@@ -43,6 +47,16 @@ RUN conda install --yes \
...
 
+# Create jovyan user with UID=1000 and in the 'users' group
+# Grant ownership over the conda dir and home dir, but stick the group as root.
+RUN useradd -m -s /bin/bash -N -u $NB_UID $NB_USER && \
+    chmod g+s /home/$NB_USER && \
+    mkdir /home/$NB_USER/work && \
+    mkdir /home/$NB_USER/.jupyter && \
+    mkdir /home/$NB_USER/.local && \
+    chown -R $NB_USER:root $CONDA_DIR && \
+    chown -R $NB_USER:root /home/$NB_USER
```
costerwi: Why set the group ownership to root?

parente: Exactly what I wanted to ask you. :) The `-N` on useradd puts the `jovyan` user in the default `users` group (GID=100). Anything that user touches in a host mount home or work dir will be accessible by GID=100 on the host unless the root of the mount has proper permissions. To head off any "Oops, I didn't think to protect against a container / host group overlap as well as UID overlap", I figured root was a safe bet.

But I agree it's quirky. I'm open to suggestions here.

costerwi: I didn't notice the `-N` switch but I like the effect.  Coincidentally, I am using the `users` group on my host CentOS system where the GID is also 100 so I'd prefer that over root but I respect that it may not satisfy all applications.
This must be a common Docker problem... Some research may be required for a sensible suggestion.

parente: Long running issue with lots of solutions but no clear best practice:
https://github.com/docker/docker/issues/7198

The real solution probably lies in docker + kernel support for user namespace mapping: https://github.com/docker/docker/pull/12648

My gut tells me we should treat whatever we do as a patch for the time being and so simpler is better. We can state the limitations and exactly what's going to happen in host mounts in the READMEs. My vote is for calling out user=jovyan (fixed), primary group = users (fixed), UID=configurable via env var NB_UID (default: 1000) and GID=100 (fixed).

rgbkrk:
> Turns out that if you don't switch back to root, the CMD from the base image launches as the last USER.

I guess I assumed that was the opinionated behavior we wanted to provide people, so they don't run as root by default.

parente:
supervisord runs as root and changes to the jovyan user. If / when we switch away from that (issue #24), it'll need to change, yes.

rgbkrk:
Oh durr. I totally forgot about that. So sorry.

rgbkrk:
:+1:

parente:
Last set of pushes now has:

* `users` group set for /home/jovyan and /opt/conda (KISS)
* READMEs stating defaults for jovyan (uid, group, gid, and what it owns)
* Fix for r-notebook

Going to put this into master and trigger the DH builds. datascience-notebook will fail to rebuild because IJulia has a regression at the moment and I can't find a way to point at a specific commit in Julia.

PR #33, opened 2015-09-15
Change PORT option to NB_PORT to avoid potential conflicts.

### LINKED ISSUE #32: Use "NB_" prefix for environment variables
I ran the jupyter/all-spark-notebook container on a Mesos cluster with the notebook container port (8888) mapped to a random port on the host.  In this configuration, Mesos sets a "PORT" environment variable when running the container to the **host** port (e.g., 31674), and the notebook server ends up starting on http://localhost:31674 within the container.

To avoid potential conflicts with other environment variables, it may be a good idea to prefix the environment variables with "NB_" or similar, like we already do with NB_UID, NB_USER, etc.

jtyberg:
closes #32


minrk:
+1, but needs rebase. Sorry for the dealy, @jtyberg.

parente:
I think we should hold off on merging this until we address the version scheme noted in issue #12 (@jtyberg works with me and I checked that this is not blocking him in any way.)

parente:
Agreed in #32 that command line arg support resolves this issue, avoids env var conflicts, and is more natural anyway. Closing as not needed, but thanks @jtyberg.

PR #41, opened 2015-09-24
remove latex

kbroughton:
and revert to pre-tini


parente:
Is this an accidental MR against the repo? Is there a problem with tini. We certainly want latex in the opinionated stacks so that people can convert notebooks to other formats without additional installs.

kbroughton:
yes, appologies

parente:
No worries. Cheers!

PR #42, opened 2015-10-09
New docker stack for a headless kernel

poplav:
New docker stack for a headless kernel.  It uses the new kernel gateway from the Jupyter incubator project to provide websocket access without exposing the whole notebook UI.  

I created the stack to be able to test the kernel gateway.  Submitting it here for possible future inclusion into the official docker-stacks, and maybe to open up the discussion about kernel containers.


parente:
@poplav Can you put a bit of text in the commit and/or on the PR so that folks know what this is and it can be discussed? Thanks!

parente:
Thanks. I'd like to hold off for a little bit on merging this until the kernel_gateway matures a bit more. It might be good to point out in the README that it's a jupyter incubator project at the moment.

@rgbkrk thoughts?

rgbkrk:
Yeah I think putting this dockerfile on the incubator itself would be good.

parente:
@rgbkrk Just to clarify, I do believe we should have a set of stacks here with kernels eventually (minimal-kernel, python-kernel, r-kernel, ...). I was only commenting on not wanting to send the wrong message that it's stable and ready to use by merging it today. Do you agree?

rgbkrk:
Totally in sync there. Once we prove it out, we can put it here which tends to get interpreted as stable (or should be).

parente:
So kernel_gateway 0.1.0 is now out on pypi and pip installable. It's already runnable with the jupyter/minimal-notebook image. 

@poplav it would be great if you could revisit this PR now and just `pip install jupyter_kernel_gateway==0.1.0`. We can then start to test the container and see how to tidy it up (e.g., I see pandoc remains and probably shouldn't).

minimal-kernel/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+# Install Tini
+RUN wget --quiet https://github.com/krallin/tini/releases/download/v0.6.0/tini && \
+    echo "d5ed732199c36a1189320e6c4859f0169e950692f451c03e7854243b95f4234b *tini" | sha256sum -c - && \
+    mv tini /usr/local/bin/tini && \
+    chmod +x /usr/local/bin/tini
+
+# Configure environment
+ENV SHELL /bin/bash
+ENV NB_USER jovyan
```
parente: Not sure we even need the jovyan user. Try using `nobody` which should exist in the container.

minimal-kernel/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+# Install Kernel Gateway
+RUN pip3 install jupyter_kernel_gateway==0.2.0
+
+# Configure container startup
+USER $NB_USER
+EXPOSE 8888
+WORKDIR /home/$NB_USER/work
+ENTRYPOINT ["tini", "--"]
+CMD ["/usr/local/bin/start-notebook.sh"]
```
parente: Don't want to run start-notebook.sh. Entrypoint should be something like the one over the kernel gateway demos like `tini -- jupyter kernel_gateway` and the CMD is `--ip=0.0.0.0`

minimal-kernel/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+ENV NB_USER nobody
+ENV NB_UID 1000
+
+# Install Kernel Gateway
+RUN pip3 install jupyter_kernel_gateway==0.2.0
+
+# Configure container startup
+USER $NB_USER
+EXPOSE 8888
+WORKDIR /home/$NB_USER/work
```
parente: nobody probably has no home dir. Maybe run in /tmp by default?

minimal-kernel/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+ENTRYPOINT ["tini", "--", "jupyter", "kernelgateway"]
+CMD ["--KernelGatewayApp.ip=0.0.0.0"]
+
+# Add local files as late as possible to avoid cache busting
+USER root
```
parente: Kill from here down. No need for a config file. The IP address is up there in the CMD. (You can add the port to be explicit too if you want but it defaults to 8.8.8.8 anyway.)

minimal-kernel/jupyter_kernel_gateway_config.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Jupyter Development Team.
```
parente: No need for this file.

minimal-kernel/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Basic Use
+
+The following command starts a container with the Kernel Gateway server listening for HTTP connections on port 8888.
+
+```
+docker run -d -p 8888:8888 jupyter/minimal-kernel
+```
+
+## Docker Options
```
parente: Probably kill this whole section. Point people to the kernel gateway repo for valid env vars and command line args. They're right in the README over there now.

minimal-kernel/README.md
```diff
@@ -0,0 +1,24 @@
+# Kernel Gateway Stack
+
+## What it Gives You
+* [Jupyter Kernel Gateway](https://github.com/jupyter-incubator/kernel_gateway) that spawns minimal Jupyter Python kernel
```
parente: "that enables programmatic access to kernels" maybe?

minimal-kernel/README.md
```diff
@@ -0,0 +1,24 @@
...
+* No preinstalled scientific computing packages
+* Unprivileged user `jovyan` (uid=1000, configurable, see options) in group `users` (gid=100) with ownership over `/home/jovyan`
```
parente: jovyan user is not in the Dockerfile anymore.

minimal-kernel/README.md
```diff
@@ -0,0 +1,24 @@
...
+* [tini](https://github.com/krallin/tini) as the container entrypoint
+* Options for Port and Ip address
```
parente: No need to say this. Point to the README in kernel gateway repo. (See below).

minimal-kernel/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+
+## Basic Use
+
+The following command starts a container with the Kernel Gateway server listening for HTTP connections on port 8888.
```
parente: Maybe state that it runs as `nobody` so that people know it's reasonably secure by default.

minimal-kernel/Dockerfile
```diff
@@ -0,0 +1,47 @@
+# Copyright (c) Jupyter Development Team.
+FROM debian:jessie
+
+MAINTAINER Jupyter Project <jupyter@googlegroups.com>
+
+USER root
+
+# Install all OS dependencies for fully functional notebook server
+ENV DEBIAN_FRONTEND noninteractive
+RUN apt-get update && apt-get install -yq --no-install-recommends \
```
parente: Probably can axe wget, bzip2, unzip, sudo. Maybe ca-certificates too but not sure. Try the seed URL option on kernel gateway and see if it can fetch a notebook from an HTTPS endpoint properly.

minimal-kernel/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+# Configure environment
+ENV SHELL /bin/bash
+ENV NB_USER nobody
+ENV NB_UID 1000
+
+# Install Kernel Gateway
+RUN pip3 install jupyter_kernel_gateway==0.2.0
+
+# Configure container startup
+USER $NB_USER
```
parente: Can't use a variable in USER. Just make it USER nobody and kill the NB_USER and NB_UID env vars above. (Make sure you test the docker build. It should warn you about bad lines like this.)

poplav:
The images size at commit b89f25d is 485.3 MB.

rgbkrk:
Pretty slim, all things considered.

parente:
I think this is good to go in. We can iterate over time. I'll merge this in later after setting up a second PR that updates the Makefile to make sure this new stack builds alongside the notebook based ones properly.

parente:
:balloon:

parente:
Tagged `e7c2515ad9b2` and pushed to Docker Hub.

Issue #26, opened 2015-09-09
Update Spark to version 1.5.0


paulochf:
Just passing by to say that, today, I tried to use all-spark-notebook for Spark 1.5.0 and it got errors. I guess the jar repositories don't have all the packages yet.

Got some "not found" during build, like

```
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]  ::              FAILED DOWNLOADS            ::
[warn]  :: ^ see resolution messages for details  ^ ::
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]  :: org.apache.spark#spark-core_2.10;1.5.0!spark-core_2.10.jar
[warn]  :: org.apache.spark#spark-launcher_2.10;1.5.0!spark-launcher_2.10.jar
[warn]  :: org.apache.spark#spark-network-common_2.10;1.5.0!spark-network-common_2.10.jar
[warn]  :: org.apache.spark#spark-network-shuffle_2.10;1.5.0!spark-network-shuffle_2.10.jar
[warn]  :: org.apache.spark#spark-unsafe_2.10;1.5.0!spark-unsafe_2.10.jar
[warn]  :: org.apache.spark#spark-streaming_2.10;1.5.0!spark-streaming_2.10.jar
[warn]  :: org.apache.spark#spark-sql_2.10;1.5.0!spark-sql_2.10.jar
[warn]  :: org.apache.spark#spark-catalyst_2.10;1.5.0!spark-catalyst_2.10.jar
[warn]  :: org.apache.spark#spark-mllib_2.10;1.5.0!spark-mllib_2.10.jar
[warn]  :: org.apache.spark#spark-graphx_2.10;1.5.0!spark-graphx_2.10.jar
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
```

Let's watch [the packages repo](https://jcenter.bintray.com/org/apache/spark/) to see when they become available.

parente:
Yep. And the other thing that still needs solving is how to tag the current image so that folks that still want to use 1.4.1 aren't surprised the next time they docker pull (Issue #12)

parente:
Closed by #56. Build is running and will be on Docker Hub when ready.

PR #44, opened 2015-10-14
Use spark-1.5.1

zaxariah:
Small change to update to Spark 1.5.1. So far, in my own code and testing, this has worked fine.

### DIFF
--- a/pyspark-notebook/Dockerfile
+++ b/pyspark-notebook/Dockerfile
@@ -6,7 +6,7 @@ MAINTAINER Jupyter Project <jupyter@googlegroups.com>
 USER root
 
 # Spark dependencies
-ENV APACHE_SPARK_VERSION 1.4.1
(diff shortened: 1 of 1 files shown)


parente:
Thanks for it. I'll get on figuring out how to version these dependencies so that users of the current 1.4.1 image can continue to pull the old container image. Then we can merge this and make it latest.

parente:
Ref #26

parente:
Can you update the all-spark-notebook with the same change and make a note in both READMEs of the Spark version number?

depend:
I update Dockerfile and README.md in both all-spark-notebook and pyspark-notebook.  https://github.com/jupyter/docker-stacks/pull/52

parente:
Awesome. Thank you. Will give that PR a quick build sniff test later and then merge.

PR #46, opened 2015-10-15
New versioning scheme

parente:
Ref #12 

Would appreciate a review of the README steps which would be the manual flow for building / pushing new images for the time being. If there's a TON of traffic here that makes manually running the few commands there untenable, we can start to look into automating it with a CI system that runs the make target instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,23 +1,44 @@
 .PHONY: build dev help server
 
 OWNER:=jupyter
-STACK?=
-ARGS?=
-DARGS?=
+ALL_STACKS:=$(shell find . -type f -name 'Dockerfile' -exec dirname {} \; | sed "s|^\./||")
+GIT_MASTER_HEAD_SHA:=$(shell git rev-parse --short=12 --verify HEAD)
+
+
 
 help:
 	@echo
-	@echo ' build STACK=<dirname> - build using Dockerfile in named directory'
-	@echo '   dev STACK=<dirname> - run container using stack name'
-	@echo 'server STACK=<dirname> - run stack container in background'
-	@echo
+	@echo '   build/<stack dirname> - builds the latest image for the stack'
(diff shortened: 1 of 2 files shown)


parente:
Oh and please don't merge this. Before it goes in, we need to play the shuffle game between automated repos and the manually pushed kind. I have all the existing tagged images preserved on the build VM so that I can whack the automated repos and push those same images back up to the Hub as quickly as possible.

jtyberg:
I was able to test the steps (up to the `make release-all`, of course).  Seems pretty straightforward and painless.

minrk:
Seems okay. How is the SSH access to the build bot handled? Automatically get public keys of people with write access to this repo via GitHub API?

parente:
I've added @minrk and @rgbkrk keys to the VM to start. For now, we can add other folks when the need arises by hitting https://github.com/username.keys and pasting them in to authorized_keys. Not sure if anything more automatic is needed at the moment.

minrk:
No, the userlist doesn't change enough that something automated is needed. I was just curious how you were seeding it at first.

parente:
OK. Cool.

I'm going to start swapping out the automated repos today for the manual ones and sanity check that everything is OK. If so, we can put in this PR and use it as the first case to do a build. The build will essentially be a no-op since the Dockerfiles have not changed, but it will have the effect of tagging latest with the git SHA of this PR. Seems appropriate given that's the SHA where we cut over to the new scheme.

rgbkrk:
Ok, awesome.

parente:
Cutting over to manual repos is done. It was trivial because even though I had to delete the repo entries on Docker Hub, it retained all the Docker filesystem layers. So a push of the same images back after the delete was near instantaneous at restoring them.

Next step is to merge this PR and do the manual build.

Issue #47, opened 2015-10-19
Document command line arg passthrough


parente:
We need to document that all Jupyter command line args can be passed via the start command for the container. This obviates the need to add env vars for everything.

We might also consider deprecating and/or eventually removing the env vars that can be passed through trivially: INTERFACE, PORT.

Ref issue #45, PR #33

parente:
And ref #32

parente:
Rebuild in progress. Closed by #50.

PR #50, opened 2015-10-21
Remove trivial env vars, doc command line args

parente:
* Remove PORT and INTERFACE env vars which can conflict with other systems (e.g., Mesos)
* Document command line pass-through to start-notebook.sh

Besides documenting how docker run args can be passed to start-notebook.sh and then through to jupyter notebook, I opted for removing, not deprecating, the redundant env vars. Justification: folks can pin to the previous build git SHA if they want to retain the env var support while master marches forward without them.

### DIFF
--- a/minimal-notebook/jupyter_notebook_config.py
+++ b/minimal-notebook/jupyter_notebook_config.py
@@ -8,6 +8,8 @@ import stat
 PEM_FILE = os.path.join(jupyter_data_dir(), 'notebook.pem')
 
 c = get_config()
+c.NotebookApp.ip = '*'
+c.NotebookApp.port = 8888
 c.NotebookApp.open_browser = False
 
 # Set a certificate if USE_HTTPS is set to any value
--- a/minimal-notebook/jupyter_notebook_config.py
+++ b/minimal-notebook/jupyter_notebook_config.py
@@ -8,8 +8,6 @@ import stat
 PEM_FILE = os.path.join(jupyter_data_dir(), 'notebook.pem')
 
 c = get_config()
-c.NotebookApp.ip = os.getenv('INTERFACE', '') or '*'
-c.NotebookApp.port = int(os.getenv('PORT', '') or 8888)
 c.NotebookApp.open_browser = False
 
 # Set a certificate if USE_HTTPS is set to any value
--- a/minimal-notebook/start-notebook.sh
+++ b/minimal-notebook/start-notebook.sh
@@ -13,4 +13,3 @@ fi
 
 # Start the notebook server
 exec su $NB_USER -c "env PATH=$PATH jupyter notebook $@"
-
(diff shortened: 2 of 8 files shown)


parente:
Ref #47

rgbkrk:
Looks good to me.

minimal-notebook/jupyter_notebook_config.py
```diff
@@ -8,8 +8,6 @@
 PEM_FILE = os.path.join(jupyter_data_dir(), 'notebook.pem')
 
 c = get_config()
-c.NotebookApp.ip = os.getenv('INTERFACE', '') or '*'
```
parente: Ooo. Missed this one. Removing this defaults the notebook back to 127.0.0.1. I think we want to retain the * behavior since these are always in docker containers.

I'm guessing that command line opts override a hardcoded '*' here in the config file. But I need to check that to be sure.

parente:
Yep. Command line overrides config. I've pushed a commit that sets port to 8888 and ip to '*' in the config file so that the defaults remain backward compatible.

Along the way, I also noticed and fixed a problem with $@ in the start-notebook.sh script. This syntax required >1 argument to the star-notebook script to be quoted when passed on docker run. Switching to $* removes the need for the quoting, but also still works if the quoting is used (like I think last tmpnb.org deploy required).

@rgbkrk one more look-see if you please?

minimal-notebook/jupyter_notebook_config.py
```diff
@@ -8,8 +8,8 @@
...
-c.NotebookApp.port = int(os.getenv('PORT', '') or 8888)
+c.NotebookApp.ip = '*'
+c.NotebookApp.port = 8888
```
rgbkrk: Cool. Makes sense to just standardize on these here so we can reason about them in the orchestration systems.

minimal-notebook/start-notebook.sh
```diff
@@ -12,5 +12,4 @@ if [ ! -z "$GRANT_SUDO" ]; then
 fi
 
 # Start the notebook server
-exec su $NB_USER -c "env PATH=$PATH jupyter notebook $@"
-
+exec su $NB_USER -c "env PATH=$PATH jupyter notebook $*"
```
rgbkrk: This assumes `IFS` is defined and not null, which is totally sensible. Semantically the difference of before:

```bash
exec su $NB_USER -c "env PATH=$PATH jupyter notebook \"$1\" \"$2\" ..."
```

after:

```bash
exec su $NB_USER -c "env PATH=$PATH jupyter notebook $1 $2..."
```

is that correct?

parente: Yes, though I believe the first form tries to quote each individual item, but perhaps not properly within the context of the quotes already in the start-notebook script. I was seeing any arguments beyond $1 winding up as parameters to `su`, not to `jupyter notebook`.

minimal-notebook/jupyter_notebook_config.py
parente: Yes. Command line overrides config as expected.

Issue #55, opened 2015-10-22
rzmq requires libzmq.so.4, conda installs libzmq.so.5 by default now


parente:
Short term, need to see if we can pin libzmq back to 4 or install both versions of the lib maybe.

This prevents rebuilds of any containers containing R at the moment.

parente:
Filed here IRkernel/IRkernel#205 but the problem is likely with the conda rzmq package.

parente:
Have a temporary workaround that symlinks the older libzmq and libsodium so files back into `/opt/conda/lib` from the `/opt/conda/pkgs` folder. Going to toss this into the all-spark-notebook stack ONLY just to unblock PR #52.

parente:
Had to add temporary workaround to r-notebook and datascience-notebook now too to account for change in minimal image for PR #50 that leads to rebuild of all images.

parente:
Upstream ContinuumIO/anaconda-issues#506

parente:
Pushed to DockerHub with image tag 2d125a7161b5.

PR #52, opened 2015-10-21
update for Spark 1.5.1

depend:

PR #56, opened 2015-10-22
Update to Spark 1.5.1

parente:
Based on PR #52 but with additional workaround commit for upstream pkg problem noted in #55.


parente:
Builds fine now. Thanks @depend !

PR #80, opened 2015-12-10
make-deploy example: run stacks on docker-machines

parente:
@jtyberg and I spent the morning hacking on this KISS automation for deploying singular docker-stack images to docker-machine controlled hosts. The included README should say it all.

This PR comes in response to various requests about how to go beyond "Great, I can run a jupyter in isolated in docker on my local machine." It's definitely not the end-all, be-all (it's based on Makefiles!), but we think it's a reasonably simple, raw example from which people can learn.

### DIFF
--- a/examples/make-deploy/README.md
+++ b/examples/make-deploy/README.md
@@ -45,7 +45,22 @@ make notebook NAME=your-notebook PORT=9001 WORK_VOLUME=our-work
 
 ### How do I run over HTTPS?
 
-Instead of `make notebook` run `make secure-notebook PASSWORD=your_desired_password`.
+Instead of `make notebook`, run `make self-signed-notebook PASSWORD=your_desired_password`. This target gives you a notebook wtih a self-signed certificate.
+
+### That self-signed certificate is a pain. Let's Encrypt?
+
+Yes. Please.
+
+```
+make letsencrypt FQDN=host.mydomain.com EMAIL=myemail@somewhere.com
+make letsencrypt-notebook
+```
+
+The first command creates a Docker volume named after the notebook container with a `-secrets` suffix. It then runs the `letsencrypt` client with a slew of options (one of which has you automatically agreeing to the Let's Encrypt Terms of Service, see the Makefile). The second command mounts the secrets volume and configures Jupyter to use the full-chain certificate and private key.
+
+Be aware: Let's Encrypt has a pretty [low rate limit per domain](https://community.letsencrypt.org/t/public-beta-rate-limits/4772/3) at the moment. Don't exhaust your requests playing around!
+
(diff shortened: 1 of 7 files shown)


parente:
Now has a Let's Encrypt recipe.

parente:
I don't want to merge this one myself without another set of eyes on it with regards to whether it belongs in docker-stacks or not, whether we've gone totally crazy with make, whether the README even makes sense, etc.

@rgbkrk care to comment? From the makefiles / README can you figure out how to use it against Carina? (Seems like it would be really, really simple to do so.)

examples/make-deploy/README.md
```diff
@@ -0,0 +1,112 @@
...
+### How do I run over HTTPS?
+
+Instead of `make notebook`, run `make self-signed-notebook PASSWORD=your_desired_password`. This target gives you a notebook wtih a self-signed certificate.
+
+### That self-signed certificate is a pain. Let's Encrypt?
+
+Yes. Please.
+
+```
+make letsencrypt FQDN=host.mydomain.com EMAIL=myemail@somewhere.com
```
rgbkrk: Nice!

rgbkrk:
Yeah, it wouldn't be too bad to set it up for `carina`. While Rackspace is part of docker-machine, Carina is not (differences in how certs are used/setup). It comes down to `carina create <name>` followed by `eval "$( carina env <name> )"`. Then it's Docker the rest of the way.

examples/make-deploy/softlayer.makefile
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+softlayer-vm: export SOFTLAYER_CPUS?=4
```
jtyberg: I just tried creating a SoftLayer VM.  The SOFTLAYER_CPU variable is singular, not plural.

parente:
@rgbkrk 

> While Rackspace is part of docker-machine, Carina is not (differences in how certs are used/setup). It comes down to carina create <name> followed by eval "$( carina env <name> )". Then it's Docker the rest of the way.

I imagine the Carina bits could be a another carina.makefile that just uses the Carina CLI to do whatever is necessary to set the local env vars for the rest of the docker targets to "just work."

parente:
So my main remaining question is where does this make hack belong? In this repo? Somewhere else? It largely depends on how much we're willing to support it. My stance is that it's an example at best that folks can use as-is or change as they see fit.

rgbkrk:
> I imagine the Carina bits could be a another carina.makefile that just uses the Carina CLI to do whatever is necessary to set the local env vars for the rest of the docker targets to "just work."

Sounds good to me.

rgbkrk:
These stand pretty well in examples, I think this is fine.

PR #84, opened 2015-12-17
Create jovyan user before conda install

### LINKED ISSUE #81: chown very slow
The docker file for the minimal-notebook runs

```
chown -R $NB_USER:users $CONDA_DIR && \
chown -R $NB_USER:users /home/$NB_USER
```

in the file. I don't know why, but this is very slow. If I build a new docker image and pass a different NB_USER value (so this step isn't cached) it takes ~10 minutes to finish running the chown. This may have to do with how aufs is working behind the scenes (I'm running on Debian, using aufs for docker storage). Since the file is cached on rebuilds, this isn't a big deal, but does the jovyan user need to be owner of /opt/conda? If so, maybe you could create that user first and install conda as user jovyan?

parente:
Avoids costly chown, fixes #81.

Unfortunately, we cannot build all containers that depend on minimal at the moment because the #55 workaround is no longer viable.

### DIFF
--- a/minimal-notebook/Dockerfile
+++ b/minimal-notebook/Dockerfile
@@ -38,31 +38,36 @@ ENV SHELL /bin/bash
 ENV NB_USER jovyan
 ENV NB_UID 1000
 
-# Install conda
-RUN mkdir -p $CONDA_DIR && \
-    echo export PATH=$CONDA_DIR/bin:'$PATH' > /etc/profile.d/conda.sh && \
+# Create jovyan user with UID=1000 and in the 'users' group
+RUN useradd -m -s /bin/bash -N -u $NB_UID $NB_USER && \
+    mkdir -p /opt/conda && \
+    chown jovyan /opt/conda
+
+USER jovyan
+
+# Setup jovyan home directory
+RUN mkdir /home/$NB_USER/work && \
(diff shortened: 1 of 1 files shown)


minimal-notebook/Dockerfile
```diff
@@ -38,31 +38,36 @@ ENV SHELL /bin/bash
...
+    mkdir -p $CONDA_DIR && \
     wget --quiet https://repo.continuum.io/miniconda/Miniconda3-3.9.1-Linux-x86_64.sh && \
     echo "6c6b44acdd0bc4229377ee10d52c8ac6160c336d9cdd669db7371aa9344e1ac3 *Miniconda3-3.9.1-Linux-x86_64.sh" | sha256sum -c - && \
-    /bin/bash /Miniconda3-3.9.1-Linux-x86_64.sh -f -b -p $CONDA_DIR && \
+    /bin/bash Miniconda3-3.9.1-Linux-x86_64.sh -f -b -p $CONDA_DIR && \
     rm Miniconda3-3.9.1-Linux-x86_64.sh && \
     $CONDA_DIR/bin/conda install --yes conda==3.14.1
 
-# Install Jupyter notebook
+# Install Jupyter notebook as jovyan
```
rgbkrk: yay!

Issue #82, opened 2015-12-13
UTF-8 Characters (non-ASCII) in plot titles and axes labels in r-notebook are mangled 


kainhofer:
In r-notebook (both a local docker container as well as on try.jupiter.org), Umlaute in plot titles in R plots are mangled. Example:

Input:

plot(stats::rnorm(50)) +
title(main = "Übertragungszeiten") 

The output will be a chart with title "<c3><9c>bertragungszeiten". I.e. the two-byte UTF-8 character Ü is printed out with the ascii char numbers of each byte. This makes r-notebook for German users quite broken.

[Untitled.html.zip](https://github.com/jupyter/docker-stacks/files/60609/Untitled.html.zip)

parente:
It's not just plots, but even simple string output:

```
a <- "Ü"
print(a)
cat(a)
```

I'm no expert on R string encoding. @takluyver any thoughts? Does something need to be configured in the image here?

rgbkrk:
We're setting all the locale environment variables right?

takluyver:
We made a change recently to IRkernel (IRkernel/IRkernel#224) to ensure JSON messages are correctly decoded as UTF-8. Maybe that will help, but assuming $LANG is set to something like `en_GB.UTF-8`, it should already be working.

parente:
We might not be. I swore there was a PR a couple months back about that, but now that I look at minimal, sure doesn't look like it.

parente:
Issue fix but build blocked by #55. The workaround we were using to get R working no longer works.

parente:
Pushed to DockerHub with image tag 2d125a7161b5.

PR #88, opened 2015-12-17
Bump r-irkernel to 0.5 to fix zmq issue

### LINKED ISSUE #55: rzmq requires libzmq.so.4, conda installs libzmq.so.5 by default now

parente:
Requires r-rmarkdown=0.8, r-devtools=1.9, r-tidyr=0.3 in r-notebook

Fixes #55 and makes it possible to build stacks with R again!


parente:
I successfully built and tested all the R images. Merging this in to unblock the builds for #81 and #82.

Issue #67, opened 2015-11-20
Importing matplotlib gives an error


drdwitte:
I basically just have a number of import statements, with matplotlib.pyplot causing problems. I have it in the scipy notebook but not in the all-spark notebook 

import matplotlib.pyplot as plt
%matplotlib inline  

<b>ImportError: libXrender.so.1: cannot open shared object file: No such file or directory</b>

Full error trace:
----> 1 import matplotlib.pyplot

/opt/conda/lib/python3.4/site-packages/matplotlib/pyplot.py in <module>()
    107 
    108 from matplotlib.backends import pylab_setup
--> 109 _backend_mod, new_figure_manager, draw_if_interactive, _show = pylab_setup()
    110 
    111 

/opt/conda/lib/python3.4/site-packages/matplotlib/backends/__init__.py in pylab_setup()
     30     # imports. 0 means only perform absolute imports.
     31     backend_mod = __import__(backend_name,
---> 32                              globals(),locals(),[backend_name],0)
     33 
     34     # Things we pull in from all backends

/opt/conda/lib/python3.4/site-packages/matplotlib/backends/backend_qt4agg.py in <module>()
     15 from matplotlib.figure import Figure
     16 
---> 17 from .backend_qt5agg import NavigationToolbar2QTAgg
     18 from .backend_qt5agg import FigureCanvasQTAggBase
     19 

/opt/conda/lib/python3.4/site-packages/matplotlib/backends/backend_qt5agg.py in <module>()
     16 
     17 from .backend_agg import FigureCanvasAgg
---> 18 from .backend_qt5 import QtCore
     19 from .backend_qt5 import QtGui
     20 from .backend_qt5 import FigureManagerQT

/opt/conda/lib/python3.4/site-packages/matplotlib/backends/backend_qt5.py in <module>()
     29     figureoptions = None
     30 
---> 31 from .qt_compat import QtCore, QtGui, QtWidgets, _getSaveFileName, __version__
     32 from matplotlib.backends.qt_editor.formsubplottool import UiSubplotTool
     33 

/opt/conda/lib/python3.4/site-packages/matplotlib/backends/qt_compat.py in <module>()
     89     if QT_API in [QT_API_PYQT, QT_API_PYQTv2]:  # PyQt4 API
     90 
---> 91         from PyQt4 import QtCore, QtGui
     92 
     93         try:

ImportError: libXrender.so.1: cannot open shared object file: No such file or directory

parente:
[...]

PR #93, opened 2015-12-29
Install libxrender1 to appease matplotlib et al.

parente:
Bite the bullet and preinstall it so that plotting libs that default
to using desktop rendering just work (matplotlib, ggplot, ...)
out of the box without having to get configuration right beforehand
(e.g., %matplotlib inline ahead of matplotlib import)

Only adds ~100k to the minimal-notebook image size


parente:
Fixes #67

PR #102, opened 2016-01-13
Remove build deps from minimal-kernel image

parente:
* Remove SHELL env var and leftover USER nobody
* Install wget for tini then immediately uninstall it
* Install build-essentials and python3-dev for tornado speed up
  then immediately uninstall them

Before these changes:

`jupyter/minimal-kernel         e7c2515ad9b2        9bd715cba589        56 minutes ago      485.3 MB`

After:

`jupyter/minimal-kernel                            latest              206fd7939811        2 minutes ago       240.5 MB`

So the image size is cut in half. The downside is that anyone looking to extend the image (e.g., add more Python libs, add new kernels, etc.) probably needs to add build-essentials and other build deps back in. 

I think minimization is the correct path for kernel images like this one in contrast to what we did in notebook images. Here, I expect the user to be inheriting `FROM` this image to build a new docker image with their specific kernel dependencies. In the notebook stacks, I expect a user to want to pip/conda install things on-demand and just expect them to work from within a running Jupyter notebook server.

@rgbkrk opinion? @poplav?

### DIFF
--- a/minimal-kernel/Dockerfile
+++ b/minimal-kernel/Dockerfile
@@ -8,21 +8,22 @@ USER root
 # Install all OS dependencies for fully functional notebook server
 ENV DEBIAN_FRONTEND noninteractive
 RUN apt-get update && apt-get install -yq --no-install-recommends \
-    wget \
-    python3-dev \
-    build-essential \
     python3-pip \
     python3-zmq \
     && apt-get clean
 
 # Install Tini
-RUN wget --quiet https://github.com/krallin/tini/releases/download/v0.6.0/tini && \
+RUN apt-get update && \
+    apt-get install -yq --no-install-recommends wget && \
+    wget --quiet https://github.com/krallin/tini/releases/download/v0.6.0/tini && \
     echo "d5ed732199c36a1189320e6c4859f0169e950692f451c03e7854243b95f4234b *tini" | sha256sum -c - && \
     mv tini /usr/local/bin/tini && \
-    chmod +x /usr/local/bin/tini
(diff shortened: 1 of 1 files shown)


rgbkrk:
Holy crap that's smaller. You're definitely right that it makes it awkward for extending the image on a full build environment. Super cool to even have this as an example though.

parente:
I guess I'm thinking about a case where I just want the, say, Scala kernel. I don't need build-essentials. I need openjdk and such. Having build-essentials in the base image is just bloat.

The same is probably true in the era of wheels for Python too. If I want pandas in here with my kernel, I don't (think I) need build-essentials in the base image.

I'm sort of mentally drawing the line between someone building their own docker image starting from this stack vs someone that just wants everything to install cleanly without thinking about deps and docker at all over in one of the notebook stacks.

poplav:
I think that makes sense as this would be the base image other specific stacks could inherent from.  It gives the user of this image the maximum amount of flexibility.

parente:
Bit more minimal now:

`jupyter/minimal-kernel                            latest              8411ad4d0169        4 minutes ago       237.5 MB`

Definitely faster building by removing the need for wget install and uninstall.

parente:
@jhedemonte pointed out that I got a little overzealous with the removal of setuptools. pip kinda needs that. 

Putting that back in.

Unless there's objections, would like to merge. If it turns out working with the smaller image is untenable, we can revisit the contents.

parente:
:clap:

parente:
Built, pushed, and tagged `9cb6443840b2`.

PR #104, opened 2016-01-17
Notebook 4.1.x

parente:
Conda rev'ed to Notebook 4.1.0. Updated pinned install to 4.1*. 

Manually tested plotting, different kernels, and Spark across images. Nothing seems amiss. (Would be nice to have a small suite of notebook-based sniff tests that we could run automatically against each stack using Selenium or something ...)


parente:
All images rebuilt and pushed to Docker Hub under tag `761010eda90d` (and latest of course).

PR #108, opened 2016-01-26
Added ubuntu:14.04 locale fix

jakirkham:
Sorry this is way overdue. Added locale fixes for ubuntu:14.04 so that we can get CI running as was mentioned here. ( https://github.com/jupyter/notebook/pull/547#issuecomment-146991682 )

Circle CI seemed like the best choice for doing CI with docker images. However, they did not allow us to change the locale. Not changing the locale caused us build issue with the `jupyter/notebook` image. Changing locale is not a problem when building on Docker Hub or locally. The image described here simply applies the needed locale change so it can act as a base image for `jupyter/notebook` and allow for CI support on that image.

ping: @parente

### DIFF
--- a/ubuntu_14_04_locale_fix/Dockerfile
+++ b/ubuntu_14_04_locale_fix/Dockerfile
@@ -1,0 +1,1 @@
+FROM ubuntu:14.04
--- a/ubuntu_14_04_locale_fix/Dockerfile
+++ b/ubuntu_14_04_locale_fix/Dockerfile
@@ -1,1 +1,6 @@
 FROM ubuntu:14.04
+
+# Setting the language and encoding won't work without this tweak.
+# See bug here ( https://bugs.launchpad.net/ubuntu/+bug/920601 ).
+RUN DEBIAN_FRONTEND=noninteractive locale-gen en_US.UTF-8 && \
+    DEBIAN_FRONTEND=noninteractive dpkg-reconfigure locales


parente:
This is the first image definition that doesn't fall into the category of pull-and-run for end users. If there's no better place, people can ignore it, of course. But putting it down in a subfolder like `internal` or something similar so it's not at the top level like the others would help.

@minrk @rgbkrk any alternatives or is this the spot?

jakirkham:
Putting it in a lower directory is fine. Keeping it as a separate repo could be fine. Simply up to you guys. Just let me know.

rgbkrk:
Is it bad if this is in the main image and people have to overwrite the env var for their locale?

jakirkham:
Shouldn't be. Should be trivial for anyone to change. Though adding UTF8 as the default is probably a good idea anyway or is that why you are asking?

rgbkrk:
Let's kick this off! Thank you @jakirkham for keeping up with these.

parente:
Er, so I'll move it down to a subdir and kick off a built. Maybe a separate make target because we don't want to rebuild it every time we do the other stacks, right?

parente:
@jakirkham what do you want the image called on Dockerhub so that the jupyter/notebook Dockerfile can refer to it?

rgbkrk:
Thanks @parente

jakirkham:
Obviously I am not keeping up enough. ;)

Maybe it can be `ubuntu_14_04_locale_fix`. It's a bit verbose, but it does get the point across. If you can think of a better name, feel free to pick one. It would be nice that the directory and the docker image share the same name though.

parente:
All set: https://hub.docker.com/r/jupyter/ubuntu_14_04_locale_fix/

jakirkham:
Cool, thanks. FWIW, my Docker Hub handle is the same as my GitHub one.

PR #112, opened 2016-02-02
Set PYSPARK_PYTHON path in python2 kernelspec

### LINKED ISSUE #109: all-spark-notebook: MLlib doesn't find numpy
I'm using the kernel Python 2 and trying to train a KMeans model, but it fails for not finding `numpy`.

```
Py4JJavaError: An error occurred while calling o36.trainKMeansModel.
: org.apache.spark.SparkException: Job aborted due to stage failure: Task 1 in stage 0.0 failed 1 times, most recent failure: Lost task 1.0 in stage 0.0 (TID 1, localhost): org.apache.spark.api.python.PythonException: Traceback (most recent call last):
  File "/usr/local/spark/python/lib/pyspark.zip/pyspark/worker.py", line 98, in main
    command = pickleSer._read_with_length(infile)
  File "/usr/local/spark/python/lib/pyspark.zip/pyspark/serializers.py", line 164, in _read_with_length
    return self.loads(obj)
  File "/usr/local/spark/python/lib/pyspark.zip/pyspark/serializers.py", line 422, in loads
    return pickle.loads(obj)
  File "/usr/local/spark/python/lib/pyspark.zip/pyspark/mllib/__init__.py", line 25, in <module>
ImportError: No module named numpy
```

The code I used for this test is very simple and it's posted here:

```py
from pyspark import SparkContext, SQLContext
from pyspark.mllib.clustering import KMeans, KMeansModel
import pandas as pd
import numpy as np
import os

os.environ["PYSPARK_PYTHON"] = "python2"
sc = SparkContext("local[*]")
sqlc = SQLContext(sc)

rawDf = sqlc.createDataFrame(pd.read_csv("somefile.csv"))
avgClusters = KMeans.train(rawDf,
                           5,
                           maxIterations=10,
                           runs=10,
                           initializationMode="random")
```

parente:
Fixes #109 

/cc @jakirkham @boechat107

### DIFF
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -6,6 +6,9 @@ MAINTAINER Jupyter Project <jupyter@googlegroups.com>
 
 USER root
 
+# Util to help with kernel spec later
+RUN apt-get -y update && apt-get -y install jq
+
 # Spark dependencies
 ENV APACHE_SPARK_VERSION 1.5.1
 RUN apt-get -y update && \
@@ -96,5 +99,7 @@ RUN bash -c '. activate python2 && \
     python -m ipykernel.kernelspec --prefix=$CONDA_DIR && \
     . deactivate'
 # Set PYSPARK_HOME in the python2 spec
-RUN sed -i '$ d' $CONDA_DIR/share/jupyter/kernels/python2/kernel.json && \
-    echo ', "env": {"PYSPARK_PYTHON": "/opt/conda/envs/python2/bin/python"} }' >> $CONDA_DIR/share/jupyter/kernels/python2/kernel.json
+RUN jq --arg v "$CONDA_DIR/envs/python2/bin/python" \
+        '.["env"]["PYSPARK_PYTHON"]=$v' \
+        $CONDA_DIR/share/jupyter/kernels/python2/kernel.json > /tmp/kernel.json && \
+        mv /tmp/kernel.json $CONDA_DIR/share/jupyter/kernels/python2/kernel.json
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -90,12 +90,11 @@ RUN conda install --yes \
 RUN mkdir -p /opt/conda/share/jupyter/kernels/scala
 COPY kernel.json /opt/conda/share/jupyter/kernels/scala/
 
-USER root
-
-# Install Python 2 kernel spec globally to avoid permission problems when NB_UID
-# switching at runtime.
-RUN $CONDA_DIR/envs/python2/bin/python \
-    $CONDA_DIR/envs/python2/bin/ipython \
-    kernelspec install-self
-
-USER jovyan
+# Install Python 2 kernel spec into the Python 3 conda environment which
+# runs the notebook server
+RUN bash -c '. activate python2 && \
+    python -m ipykernel.kernelspec --prefix=$CONDA_DIR && \
+    . deactivate'
+# Set PYSPARK_HOME in the python2 spec
+RUN sed -i '$ d' $CONDA_DIR/share/jupyter/kernels/python2/kernel.json && \
+    echo ', "env": {"PYSPARK_PYTHON": "/opt/conda/envs/python2/bin/python"} }' >> $CONDA_DIR/share/jupyter/kernels/python2/kernel.json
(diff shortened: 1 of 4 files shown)


all-spark-notebook/Dockerfile
```diff
@@ -90,12 +90,11 @@ RUN conda install --yes \
...
-
-USER jovyan
+# Install Python 2 kernel spec into the Python 3 conda environment which
+# runs the notebook server
+RUN bash -c '. activate python2 && \
+    python -m ipykernel.kernelspec --prefix=$CONDA_DIR && \
+    . deactivate'
+# Set PYSPARK_HOME in the python2 spec
+RUN sed -i '$ d' $CONDA_DIR/share/jupyter/kernels/python2/kernel.json && \
+    echo ', "env": {"PYSPARK_PYTHON": "/opt/conda/envs/python2/bin/python"} }' >> $CONDA_DIR/share/jupyter/kernels/python2/kernel.json
```
jakirkham: Something to consider might be to use [`jq`]( https://stedolan.github.io/jq/ ) instead. Might not be a big deal given the simplicity of the file, but it would be more robust to changes of the file contents.

pyspark-notebook/Dockerfile
```diff
@@ -52,13 +52,11 @@ RUN conda create -p $CONDA_DIR/envs/python2 python=2.7 \
...
-USER jovyan
-
+# Install Python 2 kernel spec into the Python 3 conda environment which
+# runs the notebook server
+RUN bash -c '. activate python2 && \
+    python -m ipykernel.kernelspec --prefix=$CONDA_DIR && \
+    . deactivate'
+# Set PYSPARK_HOME in the python2 spec
+RUN sed -i '$ d' $CONDA_DIR/share/jupyter/kernels/python2/kernel.json && \
+    echo ', "env": {"PYSPARK_PYTHON": "/opt/conda/envs/python2/bin/python"} }' >> $CONDA_DIR/share/jupyter/kernels/python2/kernel.json
```
jakirkham: Same as [above]( https://github.com/jupyter/docker-stacks/pull/112/files#r51575188 ).

jakirkham:
Looks like a good change. Made a comment about a better way to modify json files that is worth checking out if you are not already familiar.

Should we be doing this for Python 3, as well?

all-spark-notebook/Dockerfile
parente: I thought about doing it with a Python and json snippet first, but it got to wordy. I've heard of jq but haven't used it. When I've got time, I'll see how much better/shorter it makes things than dear old sed.

parente:
> Should we be doing this for Python 3, as well?

Nothing to do for Python 3. Works as-is because the python path is inherited from the environment of the notebook server which is using the default conda environment based on Python 3. (Confirmed by using the original code example from the issue in a python3 notebook and noting that numpy is found without error.)

all-spark-notebook/Dockerfile
parente: Yes. Win.

```bash
jq --arg v "$CONDA_DIR/envs/python2/bin/python" \
  '.["env"]["PYSPARK_PYTHON"]=$v' \
  $CONDA_DIR/share/jupyter/kernels/python2/kernel.json
```

jakirkham: Personally, that reads as nicer and clearer to me, but I leave it up to you.

parente: I agree with you. Good recommendation. The only quirky part is apt-get installing jq in the dockerfile, doing this, then taking jq back out. So in the end it becomes:

```
RUN apt-get -y update && \
    apt-get -y install jq && \
    jq --arg v "$CONDA_DIR/envs/python2/bin/python" \
        '.["env"]["PYSPARK_PYTHON"]=$v' \
        $CONDA_DIR/share/jupyter/kernels/python2/kernel.json
    apt-get remove -y jq && \
    apt-get clean
```

We could leave jq there, but it's a bit odd to only have it in the pyspark and all-spark notebooks. If anything, it belongs in minimal-notebook, but I see that as a step to take separate from this PR.

jakirkham: Yeah. So, one thought is just to leave `jq` in there. The binary from `apt-get` only takes up ~199K. It doesn't avoid the normal docker ugliness, but we could separate these into two steps at least, which would be an improvement. Namely install `jq` and then fix `kernel.json`

parente: And after thinking about how the apt-get has to run as root and how jq should really run as jovyan, it's just easier to get jq installed in the `USER root` part of the Dockerfile at the top and use it down below to make the edit.

jakirkham: Sounds good to me.

Issue #62, opened 2015-11-12
Reduce image sizes


nspaeth:
The minimal-notebook installs texlive, which pulls in x11 for a hefty increase in image size. It would be nice to install a version that doesn't require those extra dependencies, or to manually remove them after install, if that won't break anything.

parente:
I've thought about this. On the one hand, minimal notebook is the minimal image that enables *all* of the notebook server functions, including the Download As menu items. Without texlive, some of those don't work (according to my experiments at least). On the other hand, I recognize that getting a minimal image that enables all-but-notebook-conversion might be desirable for some use cases.

nspaeth:
Sorry, I think I misspoke. I assumed that texlive was necessary for something, so what I intended to say was to install a version of texlive that doesn't require x11 (ie, not compiled with support for it).

parente:
Ah. I'm not aware of a different version that doesn't have the x11 bits, but I'm definitely open to finding one.

nspaeth:
Unfortunately I think it might be somewhat involved. The following stackexchange questions give some possibilities, but I haven't had a chance to try any of it yet:

http://tex.stackexchange.com/questions/1092/how-to-install-vanilla-texlive-on-debian-or-ubuntu
(which links to: http://www.tug.org/texlive/debian.html )
http://askubuntu.com/questions/74523/how-can-i-install-a-package-without-installing-some-dependencies

parente:
I'm playing a bit with https://github.com/jwilder/docker-squash against the minimal-notebook to see if/how it helps.

parente:
Comparisons of before/after squash:

```
$ docker images
jupyter/minimal-notebook                          latest-squashed     56c057ff3066        About an hour ago   1.777 GB
jupyter/minimal-notebook                          latest              bb9372ddc21e        3 weeks ago         2.162 GB
```

```
$ docker history jupyter/minimal-notebook:latest
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
bb9372ddc21e        3 weeks ago         /bin/sh -c chown -R $NB_USER:users /home/$NB_   1.353 kB
59be73705b0e        3 weeks ago         /bin/sh -c #(nop) COPY file:92ab115ce14297458   1.353 kB
ebb4a093127c        3 weeks ago         /bin/sh -c #(nop) COPY file:2b820ac933e1e163a   398 B
[...]

PR #121, opened 2016-02-13
Update stacks to current conda, python3.5, and add h5py

fperez:
Refresh the images with more current versions of all packages as well as using current conda and Python3.5 as the default py3 environment.

I also added h5py, which is quite useful in general and necessary to be able to run [the LIGO gravitational wave analysis notebooks](http://nbviewer.jupyter.org/urls/losc.ligo.org/s/events/GW150914/GW150914_tutorial.ipynb) out of the box.

Is it correct that this repo will automatically build the images on [docker hub](https://hub.docker.com/u/jupyter)? I did test the builds locally.

### DIFF
--- a/datascience-notebook/Dockerfile
+++ b/datascience-notebook/Dockerfile
@@ -23,42 +23,44 @@ USER jovyan
 
 # Install Python 3 packages
 RUN conda install --yes \
-    'ipywidgets=4.0*' \
+    'ipywidgets=4.1*' \
     'pandas=0.17*' \
-    'matplotlib=1.4*' \
-    'scipy=0.16*' \
-    'seaborn=0.6*' \
-    'scikit-learn=0.16*' \
+    'matplotlib=1.5*' \
+    'scipy=0.17*' \
+    'seaborn=0.7*' \
+    'scikit-learn=0.17*' \
     'scikit-image=0.11*' \
     'sympy=0.7*' \
-    'cython=0.22*' \
+    'cython=0.23*' \
     'patsy=0.4*' \
     'statsmodels=0.6*' \
     'cloudpickle=0.1*' \
     'dill=0.2*' \
-    'numba=0.22*' \
-    'bokeh=0.10*' \
+    'numba=0.23*' \
+    'bokeh=0.11*' \
+    'h5py=2.5*' \
     && conda clean -yt
 
 # Install Python 2 packages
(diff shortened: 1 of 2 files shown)


parente:
> Is it correct that this repo will automatically build the images on docker hub? I did test the builds locally.

No, we're still on a manual build process because of the problems we had with the Docker HUB process. The README in the root of the repo has the maintainer steps. Issue #7 has some ideas about trying to automate with CircleCI or something similar.

I've started the manual build. Will post back when the images are tagged on dockerhub.

> jed/emacs editors

We might want to think about having a split between minimal-notebook and really-minimal-notebook. If folks are using minimal-notebook as a base for their own stacks where they don't use one or more of vim, emacs, wget, pandoc, etc., then we're bloating things unnecessarily for their use case. But of course I don't know if that's how folks are using minimal-notebook or not.

parente:
For ref, issue #62 has a convo about making minimal-notebook *smaller* rather than more feature-filled.

parente:
Fails to build with a conflict between python 3.5 and cython=0.22 in the scipy-notebook. I'll submit a follow on PR.

fperez:
Oops, sorry, I checked the build with the ds stack and didn't rerun the
scipy one at the end...

I just pushed the cython version fix.

fperez:
Re. `emacs`, I just took it off, good point, as it's pretty large. `jed` OTOH is very small, and it provides an emacs-like experience that should be enough for all uses.  But I think that if we offer vim in the base emacs, having at least a more emacs-like editor is useful (it's pretty much the first thing I need on a unix box :)

parente:
I'm fine with emacs being there for now. I was just thinking aloud about splitting the minimal-notebook into:


- tiny-notebook - bare minimum prereqs to get just the notebook server going with bare minimum functionality (e.g., no pandoc, just a python3 kernel); a small image you can start from to build your own stack
- minimal-notebook - tiny-notebook + a few more prereqs to get a notebook environment in which you can actually start to do some work

The images are slowly pushing to Docker Hub ... 500 errors galore from the hub. Retry, retry, retry.

parente:
Tagged `latest` and `9f9907cf1df8` on Docker Hub.

fperez:
Thanks, @parente! Note that I did remove emacs for now, jed can do the job of a non-vim, emacs-like editor for the terminal with a smaller footprint.

But I do like the tiny/minimal split idea, I think it can be a useful split.

PR #122, opened 2016-02-14
Bump python libs to match other images

parente:
Includes changes from #121 in other stacks so that everything builds successfully off of the minimal-notebook image.

### DIFF
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -66,23 +66,23 @@ USER jovyan
 
 # Install Python 3 packages
 RUN conda install --yes \
-    'ipywidgets=4.0*' \
+    'ipywidgets=4.1*' \
     'pandas=0.17*' \
-    'matplotlib=1.4*' \
-    'scipy=0.16*' \
-    'seaborn=0.6*' \
-    'scikit-learn=0.16*' \
+    'matplotlib=1.5*' \
+    'scipy=0.17*' \
+    'seaborn=0.7*' \
+    'scikit-learn=0.17*' \
     && conda clean -yt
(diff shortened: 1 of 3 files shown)


rgbkrk:
Looking good.

parente:
Tagged `latest` and `9f9907cf1df8` on Docker Hub.

willingc:
@parente At some point, we may wish to enter the short description for the Docker Hub files. It would likely be helpful for the end users especially those that may be using the beta Kitematic on OS X (https://kitematic.com/).

Also, it may make sense at some point to have the images become "official" on Docker Hub. https://docs.docker.com/docker-hub/official_repos/

parente:
@willingc Is there a spot where the descriptions are missing? I thought each one had a short description plus a long description containing a link to the README of each stack here on GitHub.

At one point in the past, I had the README content of each stack duplicated in the long description on Docker Hub. But keeping them in sync with the proper latest version from GitHub turned into a chore. (See issue #12.) I opted for having the link from the long description to the README with the latest, correct documentation over two copies with one out of sync.

willingc:
@parente Ah... I think the ones that you have revised do have the short description in Docker Hub. I noticed a few missing the description such as JupyterHub and configurable-http-proxy. 

On Kitematic (which is supposed to get info from Docker Hub), it looks like the Notebook description isn't updated (weird - caching perhaps).
screen shot 2016-02-14 at 7 08 31 pm

screen shot 2016-02-14 at 7 03 38 pm

jakirkham:
Looks like there is a bug in kitematic ( https://github.com/docker/kitematic/issues/1532 ).

PR #125, opened 2016-02-15
Add example for deploying docker-stacks using Docker Compose

jtyberg:
This PR includes an example for deploying docker-stacks using the latest versions of Docker (1.10.0+) Docker Compose (1.6.0+) and Docker Machine (0.6.0+).

This basic example deploys just a stand-alone notebook server, so it's just a start.  I'm thinking about adding examples of more complex deployments.


examples/docker-compose/.gitignore
```diff
@@ -0,0 +1 @@
+.DS_Store
```
parente: We should add this to the root .gitignore if it's not already there.

examples/docker-compose/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Pre-requisites
+
+* [Docker Engine](https://docs.docker.com/engine/) 1.10.0+
+* [Docker Machine](https://docs.docker.com/machine/) 0.6.0+
+* [Docker Compose](https://docs.docker.com/compose/) 1.6.0+
+
+See the [installation instructions](https://docs.docker.com/engine/installation/) for your environment.
+
+## Provision Docker machines
```
parente: Maybe add a note here about how this will work on **any** docker-machine supported host. The two below are just examples. Scripts welcome as PRs!

examples/docker-compose/README.md
```diff
@@ -0,0 +1,85 @@
...
+# Create virtual device
+bin/softlayer.sh myhost
+
+# Add DNS entry (SoftLayer DNS zone must exist for SOFTLAYER_DOMAIN)
+bin/sl-dns.sh myhost
+```
+
+## Deploy stand-alone Jupyter Notebook
+
+Build and run a `jupyter/minimal-notebook` container on an existing Docker machine.
```
parente: Where does the user specify an alternative stack image?

examples/docker-compose/notebook/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+* [Docker Engine](https://docs.docker.com/engine/) 1.10.0+
+* [Docker Machine](https://docs.docker.com/machine/) 0.6.0+
+* [Docker Compose](https://docs.docker.com/compose/) 1.6.0+
+
+See the [installation instructions](https://docs.docker.com/engine/installation/) for your environment.
+
+## Quickstart
+
+Here's how to build and run a `jupyter/minimal-notebook` container on an existing Docker machine.
```
parente: Hmmm. I had to go back and forth between this README and the one a folder up to understand what I'm supposed to do. Can the parent README have a quickstart section? As a new user, I want to come here, see commands to run, and understand what I'm going to get as a result.

jtyberg: The parent README does have the same quick start.  At least the basic case.

parente:
Thanks for submitting this. I think, the more examples we have on how to deploy, the better, as long as each one has an at-a-glance quickstart for people that want something running with minimal effort.

examples/docker-compose/README.md
jtyberg: I can add an environment variable to allow users to specify a different image or Dockerfile.  Of course, there are limitations.  Changing the image will only work if it's a notebook image; otherwise, the compose files (commands, env vars) would be broken.

parente: Sorry, I meant something simpler: Just add a note in the README that the user can swap out the FROM line in the Dockerfile. (Right?)

examples/docker-compose/notebook/README.md
parente: Can it say "Quickstart" instead of what it says (provisioning docker machines and running notebook servers). I didn't read it as a quickstart at all the first time through.

jtyberg:
I consolidated to a single README and added more detail to address all comments thus far.

parente:
I tested the quickstart, multiple notebooks, multiple notebooks sharing a volume, notebook with a self-signed cert, ... everything except the let's encrypt because I ran things locally. 

Very cool. :shipit:

Issue #113, opened 2016-02-02
Put a monotonically increasing number in the docker image tag


parente:
It's really hard on Docker Hub to find out the image tag with the git SHA associated with the latest image.  It's also hard to figure out the SHA order without some back and forth between github and dockerhub.

Proposal: Add a monotonically increasing number to the docker image tag on every build like so:

```
jupyter/minimal-notebook:1-0017b56d93c9
jupyter/minimal-notebook:2-39432abcef31
jupyter/minimal-notebook:3-deadbeef1234
# etc.
```

This makes the history of images clearer for humans and retains the git-sha link back to the exact state of the repository when the images were built.

jakirkham:
So, we currently are using SHA only? Would it make sense to start having an explicit semantic or other version with a SHA attached for development? Sorry, I am just trying to get my bearings.

parente:
Brief summary: there's tons of libs in these images with tons of different versioning schemes. There's no way to capture the version of Spark (1.5.1), Notebok (4.1), pandas (0.17.1), matplotlib (1.5.1), etc. all in a succinct Docker image tag. So, instead, the git SHA is used as the Docker image version: it's a pointer back to the state of all of the Dockerfiles that were built. (The full history is in #12 and the high level is documented in the README.)

This issue is about adding one additional bit to that tag for human consumption and to give the image builds a visible ordering.

jakirkham:
> Brief summary: there's tons of libs in these images with tons of different versioning schemes.

Definitely. Though we could somehow note when we have introduced a breaking change where at least one of our dependencies has a breaking change (e.g. matplotlib 1.5.x -> 2.0.0) or we break something ourselves. We could also note when there are no breaking changes, but at least one of our dependencies adds new features (e.g. matplotlib 1.5.x -> 1.6.x). Similarly for when there are no breaking changes or new features, but only bug fixes in our dependencies or our image.

That all being said, this may still be too challenging to pursue because not all dependencies may comply with semantic versioning nor may it be appropriate for all of them to do so. The only reason I suggest it is if we are considering changing the versioning to temporal order maybe we should think a little bit more about what versioning means for the stacks themselves even if we decide this ends up being inappropriate.

> The full history is in #12 and the high level is documented in the README.

I think I saw this before recently. It is coming back to me now. :)

> This issue is about adding one additional bit to that tag for human consumption and to give the image builds a visible ordering.

Sure, the proposal sounds reasonable to me.

[...]

PR #126, opened 2016-02-19
Webhook to update docker builds page

parente:
Addresses #113 in a different fashion. Updates the https://github.com/jupyter/docker-stacks/wiki/Docker-build-history page every time a new build tag is pushed to Docker Hub.

Webhook service is implemented in a notebook and turned into an API using jupyter_kernel_gateway for Friday fun.

Not ready to merge yet. I'd like to ensure the copy I've pushed to https://docker-stacks-webhook.mybluemix.net works a few times first. It doesn't have to run here: it can run anywhere kernel gateway can. It was just easiest for me.


rgbkrk:
Heh, this is neat Friday fun! Will you be writing a blog post somewhere?

rgbkrk:
Nice notebook too.

parente:
Worked fine on the last build. I'll update the README to point to the builds page before this goes in.

parente:
All set. If we can't find a way to automate the builds using Circle CI or another hosted service, we can do something similar to this on the build VM we're already using.

PR #136, opened 2016-03-05
Update old distro codename

jakirkham:
Update from `wheezy` to `jessie`. It's possible there is a reason this is still `wheezy`, in which case it would be nice to know what it is. If not, we should probably update them.

Maybe we should hard code these as environment variables in the base image and those variable will carry through to here. Thoughts?

### DIFF
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -24,7 +24,7 @@ RUN cd /usr/local && ln -s spark-${APACHE_SPARK_VERSION}-bin-hadoop2.6 spark
 # Mesos dependencies
 RUN apt-key adv --keyserver keyserver.ubuntu.com --recv E56151BF && \
     DISTRO=debian && \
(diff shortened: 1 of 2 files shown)


parente:
The mesos packages were not available in jessie when I last tried. That's why I had to pull them from wheezy.

jakirkham:
Ah, ok. Let me check.

jakirkham:
Yeah, still not present. I'll close this out.

parente:
If you want to put in a PR to add a comment in the Dockerfile so I / we don't forget again, we can merge that.

jakirkham:
How about an issue? That way it will be a little more visible. Though we can of course have the comment too.

parente:
Sure. Issue is fine also.

jakirkham:
Issue is opened here ( https://github.com/jupyter/docker-stacks/issues/137 ). PR is opened here ( https://github.com/jupyter/docker-stacks/pull/138 ).

Issue #137, opened 2016-03-05
Install Mesos for Debian Jessie (when available)


jakirkham:
This is a reminder that we want to install Mesos for Debian Jessie once available. Currently, we are installing a version from Debian Wheezy as that is where it is available at present.

PR #138, opened 2016-03-05
Comment to get Mesos for Debian Jessie

jakirkham:
Currently, we are installing Mesos for Debian Wheezy as it is not available from Debian Jessie. This adds a comment to remind us to switch to Debian Jessie once it is available.

Related: https://github.com/jupyter/docker-stacks/issues/137


parente:
Thanks!

PR #141, opened 2016-03-07
README: Add badge for Travis CI

jakirkham:
This adds a status badge for Travis CI. Also, it will be a nice test of the build system. As a release was just completed, this should use the cache the entire way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # docker-stacks
 
+[![Build Status](https://travis-ci.org/jupyter/docker-stacks.svg?branch=master)](https://travis-ci.org/jupyter/docker-stacks)
 [![Join the chat at https://gitter.im/jupyter/jupyter](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/jupyter/jupyter?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
 Opinionated stacks of ready-to-run Jupyter applications in Docker.


jakirkham:
Cache is busted by the [`COPY` command]( https://travis-ci.org/jupyter/docker-stacks/builds/114169252#L262-L264 ). Think we need to add a `.dockerignore` file to exclude unrelated files as it does not seem to distinguish between the two. ( https://github.com/jupyter/docker-stacks/issues/142 )

parente:
Do you want to hold off on merging this until there's a fix for the COPY so we have it as a test case? Do you think we need to call out in the README that just because travis is green, it doesn't really have any relation to the "production" image on Docker Hub (yet)?

jakirkham:
> Do you want to hold off on merging this until there's a fix for the COPY so we have it as a test case?

Sounds like a good plan.

> Do you think we need to call out in the README that just because travis is green, it doesn't really have any relation to the "production" image on Docker Hub (yet)?

It does in the sense that the current code in master was able to build images. While we aren't deploying straight from Travis, there are many Travis builds that are configured similarly. There is certainly some value in advertising that we are using CI. Thoughts?

parente:
I'm cool with advertising we're using travis. Just suggesting that elsewhere in the README (the tips section) that we call out that we're not deploying straight from Travis (yet).

jakirkham:
Ok, sure. That sounds reasonable.

jakirkham:
Checked the SHA256 via `openssl` of `minimal-notebook/start-notebook.sh` and it is the same on Travis and on the VM. This is making me think the cache is invalidated for other reasons.

jakirkham:
This appears to be happening even when there is no change to the repo as can be seen in this [build]( https://travis-ci.org/jakirkham/docker-stacks/builds/114333784#L257-L259 ).

parente:
I think we can send this in now. We're tracking the Travis problem over in #150 now.

jakirkham:
...and it's in.

PR #144, opened 2016-03-07
Modified to install Apache Toree using pip

lbustelo:
I've modified the Dockerfile to use `pip` to install Apache Toree. 

It also is set to move to the latest 0.1.0.dev release. @parente Let me know how you feel about this. I'm glad with putting a specific version.

### DIFF
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -36,25 +36,6 @@ RUN apt-key adv --keyserver keyserver.ubuntu.com --recv E56151BF && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-# Scala Spark kernel (build and cleanup)
-RUN cd /tmp && \
-    echo deb http://dl.bintray.com/sbt/debian / > /etc/apt/sources.list.d/sbt.list && \
-    apt-key adv --keyserver keyserver.ubuntu.com --recv 99E82A75642AC823 && \
-    apt-get update && \
-    git clone https://github.com/apache/incubator-toree.git && \
-    apt-get install -yq --force-yes --no-install-recommends sbt && \
-    cd incubator-toree && \
-    git checkout 846292233c && \
-    make dist SHELL=/bin/bash && \
-    mv dist/toree-kernel /opt/toree-kernel && \
-    chmod +x /opt/toree-kernel && \
-    rm -rf ~/.ivy2 && \
-    rm -rf ~/.sbt && \
-    rm -rf /tmp/incubator-toree && \
-    apt-get remove -y sbt && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
-
 # Spark and Mesos pointers
 ENV SPARK_HOME /usr/local/spark
 ENV R_LIBS_USER $SPARK_HOME/R/lib
@@ -102,9 +83,9 @@ RUN conda install --quiet --yes \
     'r-ggplot2=1.0*' \
     'r-rcurl=1.95*' && conda clean -tipsy
 
-# Scala Spark kernel spec
-RUN mkdir -p /opt/conda/share/jupyter/kernels/scala
-COPY kernel.json /opt/conda/share/jupyter/kernels/scala/
+# Toree kernel
+RUN pip install toree==0.1.0.dev4
+RUN jupyter toree install --user --kernel_name='Apache_Toree'
 
 # Install Python 2 kernel spec into the Python 3 conda environment which
 # runs the notebook server


all-spark-notebook/Dockerfile
```diff
@@ -98,9 +80,10 @@ RUN conda install --quiet --yes \
     'r-ggplot2=1.0*' \
     'r-rcurl=1.95*' && conda clean -tipsy
 
-# Scala Spark kernel spec
-RUN mkdir -p /opt/conda/share/jupyter/kernels/scala
-COPY kernel.json /opt/conda/share/jupyter/kernels/scala/
+# Toree kernel
+ENV TOREE_VERSION ~=0.1.0.dev
+RUN pip install toree$TOREE_VERSION
```
parente: Does it need to be an env var? Is it used at runtime?

parente:
One question about the env var inline above. Regarding the version, we're letting patch releases float across the images but holding major.minor constant. I think `~=` let's the minor float.

DrPaulBrewer:
Good idea not to pin the version. Previous public release pinned a github
hex commit from last year.

My forked iopy-spark-notebook in drpaulbrewer/docker-stacks is behaving OK.
No more heap oom.  Don't know which change helped....

jakirkham:
Does this change also work for you, @DrPaulBrewer?

DrPaulBrewer:
I can test later today. Maybe I'll try to formalize a test case.

jakirkham:
That would be very helpful. Thanks, @DrPaulBrewer.

all-spark-notebook/Dockerfile
lbustelo: Oh man... doing too many of these at the same time. It totally does not need to be an ENV. :) I'll inline that.

lbustelo:
@parente I tried a variety of techniques to try to bring all `dev` releases for `0.1.0` without needing to edit the Dockerfile. The `~=` was the only thing that worked. I tried `<=, >` and others and landed on this. Now, I'm not sure if it will install something like `0.1.1`. My hope is that it would not even install `0.1.0` proper, but this is hard to test.

parente:
@lbustelo Elsewhere we used `pip install toree==0.1` which lets the patch release float, but I'm not sure here with the .dev on the end.

The problem is, if there's nothing for us to change in the dockerfile, then there's no way for us to trigger a rebuild. That's why elsewhere we typically edit the dockefiles to move to new major or minor versions, but  rely on unrelated PRs to trigger patch version bumps.

jakirkham:
Another thought might be to use pip install with the hash. I know you are kind of trying to get away from the hash towards a formal version, but I don't believe toree has reach a formal release yet as it is still in incubator status.

In any event, we probably should use the actual repo URL as the GitHub repo is a mirror. ( git://git.apache.org/incubator-toree.git )

DrPaulBrewer:
@lbustelo A docker container built from your pull request is more stable,
but has issues that should block you from pushing it.

(1)  `pip install toree` copies some kernel settings that override
`SPARK_OPTS`, effectively ignoring the `SPARK_OPTS` in the Dockerfile

particularly it sets (1.1) `--master` to `local[2]` and (1.2) ignores any
memory settings

On the plus side, I tried an example that made original all-spark-notebook
fail with heap OOM but the container based on your PR executed without
error,  though configured memory and cores were not used because of (1).

Details to follow.

DrPaulBrewer:
Bossy kernel located.  Apparently setting env vars there overrides earlier setting.  I also encountered this problem in my own fork.

<pre>
paul@home:~/docker-testing/results/lbustelo$ ls -l 
total 44
lrwxrwxrwx 1 paul paul    79 Mar  8 09:06 bossy_kernel_overwrites_env_spark_opts -> ./container-home-jovyan-dotlocal/share/jupyter/kernels/apache_toree/kernel.json
drwxr-xr-x 3 paul paul  4096 Mar  8 08:34 container-home-jovyan-dotlocal
-rw-rw-r-- 1 paul paul 28875 Mar  8 08:58 dockerlogs.good
-rw-rw-r-- 1 paul paul   524 Mar  8 08:58 Notes.txt
paul@home:~/docker-testing/results/lbustelo$ cat bossy_kernel_overwrites_env_spark_opts 
{
  "env": {
    "MAX_INTERPRETER_THREADS": "16",
    "CAPTURE_STANDARD_ERR": "true",
    "CAPTURE_STANDARD_OUT": "true",
    "PYTHONPATH": "/usr/local/spark/python:/usr/local/spark/python/lib/py4j-0.8.2.1-src.zip",
    "SPARK_HOME": "/usr/local/spark",
    "SEND_EMPTY_OUTPUT": "false",
    "SPARK_OPTS": "--master=local[2] --driver-java-options=-Xms1024M --driver-java-options=-Xmx4096M --driver-java-options=-Dlog4j.logLevel=info"
  },
  "display_name": "Apache_Toree",
  "language_info": {
    "name": "scala"
  },
  "codemirror_mode": "scala",
  "argv": [
    "/home/jovyan/.local/share/jupyter/kernels/apache_toree/bin/run.sh",
    "--profile",
    "{connection_file}"
  ]

Notice this is also where `master=local[2]` is getting set.  

My suggestion is to drop the SPARK_OPTS line in the kernel file.

DrPaulBrewer:
Testing details are in two clumps:

[All the testing screenshots, from the notebook and port 4040](https://drive.google.com/folderview?id=0B-XXSCz5WC7NTWw4WFVLS05rUkE&usp=sharing)

[csv file input data and two notebooks, docker commands, docker container logs e.g. heap OOM error details, misc 70Mb .tgz](https://storage.googleapis.com/jupyter-testing-2016mar08/docker-testing-2016Mar08.tgz)

I provide two scala/toree ipynb notebook examples -- one that works in all containers I tested -- and the other with slight additions to caching that generate Heap OOMs in jupyter/all-spark-notebook and stall out my fork drpaulbrewer/iopy-spark-notebook.  The lbustelo PR container runs the problematic example too (good job!).

The container built from `lbustelo's` PR for `all-spark-notebook` functions most reliably except that spark master is set to local[2] and memory is set incorrectly, SPARK_OPTS being overridden by the kernel.json as described in the previous note.  I would be curious if it still reliable with more cores running.

jupyter/all-spark-notebook has memory set incorrectly.  It is set with java options to be a max of 4G but spark only allocates ~500M to cache memory.  I understand that only around 50-66% should get allocated to cache memory but 12% is too low.  master is set ok.  It crashes if the problematic example is run, with heap oom messages.  

My docker image drpaulbrewer/iopy-spark-notebook sets up memory correctly with --driver-memory and the SPARK_OPTS is not overwritten by kernel.json because I deleted the line resetting that env var.   But.... mine, even with the other changes (openjdk8, slightly newer spark), won't run the problematic example without stalling.

lbustelo:
@jakirkham I'm the maintainer of Toree and we are on track to put a `0.1.0` release even while in incubation. The `dev` packages are a way to test things out without having to go through the process of Voting for a release in Apache. 

@DrPaulBrewer I think you are hitting an issue with Apache Toree's installation logic. I agree that we should allow setting these ENV variables outside and not override them on install, but I think that issue should be opened on the Toree side. I would appreciate it if you open a JIRA at https://issues.apache.org/jira/browse/TOREE and link back to this issue. It is a little bit more work, but it is good to show the cross community aspect of Toree to the Apache community.

lbustelo:
I removed the `ENV` step and decided to peg the version of Apache Toree to have a more deterministic environment. I will make sure that I submit PRs for other `dev` releases until we get to a stable `0.1.0`.

parente:
Thanks @lbustelo. :+1: for the explicit versioning until you reach a non-dev release. 

@DrPaulBrewer Thanks so much for diving in and testing. I agree with @lbustelo that the env var problems you noted should probably be opened against Toree itself.

We can merge after travis goes green (assuming it doesn't time out.)

jakirkham:
@lbustelo, pinning to a dev release is fine. Thanks for your help.

parente:
I'll start the build on the VM.

PR #146, opened 2016-03-07
WIP: Makefile: Refresh images using git sha tag too

jakirkham:
When performing refresh, it is possible that user wants the image that matches the git sha tag. This makes sure to pull in that image, but continues to pull in latest as was done before.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,7 @@ push/%:
 refresh/%:
 # skip if error: a stack might not be on dockerhub yet
 	-docker pull $(OWNER)/$(notdir $@):latest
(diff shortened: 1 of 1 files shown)


jakirkham:
Sort of wonder how useful this is as it stands. To get it to work in the intended way, we should tag the image that matches the commit release as latest instead of what is being done now.

parente:
I'm missing the use case and intent.

jakirkham:
Suppose I want to rebuild an older commit that is not master. If I checkout that commit and build, it will pull the latest images, which are too recent. Instead the user would want to pull in older tagged images if possible and have those treated as if they were latest. That was the idea. In practice, maybe it is not so useful.

jakirkham:
Yeah, I think I am going to abandon this.

PR #148, opened 2016-03-07
Clean apt-get list

jakirkham:
Related: https://github.com/jupyter/docker-stacks/issues/62

Simply empties `/var/lib/apt/lists/` of contents. This directory is repopulated on any call to `apt-get update` so it is not a problem to remove it. All it serves as is a cache. Should save a little bit more space for each image (~10MB).

Refs:
* ( http://unix.stackexchange.com/a/217371 )
* ( http://askubuntu.com/questions/179955/var-lib-apt-lists-huge-in-12-04/179992#179992 )

### DIFF
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -7,13 +7,14 @@ MAINTAINER Jupyter Project <jupyter@googlegroups.com>
 USER root
 
 # Util to help with kernel spec later
-RUN apt-get -y update && apt-get -y install jq && apt-get clean
+RUN apt-get -y update && apt-get -y install jq && apt-get clean && rm -rf /var/lib/apt/lists/*
 
 # Spark dependencies
 ENV APACHE_SPARK_VERSION 1.6.0
 RUN apt-get -y update && \
     apt-get install -y --no-install-recommends openjdk-7-jre-headless && \
-    apt-get clean
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
 RUN cd /tmp && \
         wget -q http://d3kbcqa49mib13.cloudfront.net/spark-${APACHE_SPARK_VERSION}-bin-hadoop2.6.tgz && \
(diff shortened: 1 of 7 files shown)


jakirkham:
Will squash once we are happy with it.

jakirkham:
Exceeded build time during the first build. Expect this will continue as it is building all the images, which I think is just outside Travis CI's time limit.

jakirkham:
To give some idea of the effects, the `minimal-notebook` will be reduced by `~10 MB` and the `datascience-notebook` will be reduced by `~50 MB`. These aren't exactly massive savings, but they are easy to get. Images that use `apt-get` in many `RUN` steps will likely benefit more than those that don't.

DrPaulBrewer:
Just remember that because of docker layering, a "RUN apt-get clean" on its
own line still stores all the old files from the previous line in an
internal layer.

I'm sure you know that already, and are being careful, but not everyone is
aware of the consequences of docker's layering.

jakirkham:
I am aware. I don't think I add such a line. Do you see some?

DrPaulBrewer:
No, it was a knee-jerk reaction. I'm trying to concentrate on two other
things at same time.

jakirkham:
Failed again due to reaching max time limit.

@parente, do you have any thoughts on this? It appears to be working fine we are just running into time limit issues.

parente:
I'm fine with the change. Want to squash it before merge?

jakirkham:
Squashed and rebased on master. Don't think there is any point in waiting for the tests as they will just time out again, but I will leave it in your hands.

jakirkham:
Thanks.

jakirkham:
Building on tmux release. Decided to build because we don't want these additional cleaning steps interfering with running CI.

parente:
Sure. Build whenever you think it's time. It's a pretty stress free process. Either it completes, or it doesn't and you re-run.

PR #155, opened 2016-03-14
Rebase datascience-notebook on scipy-notebook

### LINKED ISSUE #139: Refactor datascience-notebook to use scipy-notebook as base image
It looks like there is quite a bit shared in common between the `scipy-notebook` image and the `datascience-notebook` image. In fact, it appears the `datascience-notebook` image is just the `scipy-notebook` plus some more stuff. I wonder if it would make sense to refactor `datascience-notebook` to use `scipy-notebook` as its base image (`FROM`) instead. This would hopefully cut down build time, release time, disk usage, and maintenance burden. Though it could be that I am missing something.

parente:
Fixes #139

### DIFF
--- a/datascience-notebook/Dockerfile
+++ b/datascience-notebook/Dockerfile
@@ -1,7 +1,6 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
-
-FROM jupyter/minimal-notebook
+FROM jupyter/scipy-notebook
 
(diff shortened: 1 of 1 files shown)


jakirkham:
LGTM

jakirkham:
This refactoring has been crucial as it has brought our build time down to roughly 1/3 of what it was. By doing this, even though the cache keeps busting for mysterious reasons, we are in better shape to consider doing releases from Travis CI.

minrk:
Nice!

Issue #153, opened 2016-03-13
Curl errors adding QuantEcon package in datascience container


mkolovic:
```
Pkg.add("QuantEcon")
INFO: Cloning cache of Blosc from git://github.com/stevengj/Blosc.jl.git
INFO: Cloning cache of DSP from git://github.com/JuliaDSP/DSP.jl.git
INFO: Cloning cache of FactCheck from git://github.com/JuliaLang/FactCheck.jl.git
INFO: Cloning cache of Formatting from git://github.com/lindahua/Formatting.jl.git
INFO: Cloning cache of Graphs from git://github.com/JuliaLang/Graphs.jl.git
INFO: Cloning cache of HDF5 from git://github.com/JuliaLang/HDF5.jl.git
INFO: Cloning cache of IterationManagers from git://github.com/spencerlyon2/IterationManagers.jl.git
INFO: Cloning cache of MAT from git://github.com/simonster/MAT.jl.git
INFO: Cloning cache of Polynomials from git://github.com/Keno/Polynomials.jl.git
INFO: Cloning cache of QuantEcon from git://github.com/QuantEcon/QuantEcon.jl.git
INFO: Cloning cache of Zlib from git://github.com/dcjones/Zlib.jl.git
INFO: Installing Blosc v0.1.4
INFO: Installing DSP v0.0.11
INFO: Installing FactCheck v0.4.2
INFO: Installing Formatting v0.1.5
INFO: Installing Graphs v0.6.0
INFO: Installing HDF5 v0.4.9
INFO: Installing IterationManagers v0.0.1
INFO: Installing MAT v0.2.14
INFO: Installing Polynomials v0.0.5
INFO: Installing QuantEcon v0.1.2
INFO: Installing Zlib v0.1.12
INFO: Building Blosc
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (77) error setting certificate verify locations:
  CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
================================[ ERROR: Blosc ]================================

failed process: Process(`curl -f -o c-blosc-1.7.0.tar.gz -L https://github.com/Blosc/c-blosc/archive/v1.7.0.tar.gz`, ProcessExited(77)) [77]
[...]

PR #157, opened 2016-03-14
Add hdf5-tools package

mkolovic:
hdf5-tools package is needed for HDF5 Julia package. See Issue #153 .

### DIFF
--- a/minimal-notebook/Dockerfile
+++ b/minimal-notebook/Dockerfile
@@ -22,6 +22,7 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     bzip2 \
     unzip \
     libsm6 \
+    hdf5-tools \
     pandoc \
     texlive-latex-base \
     texlive-latex-extra \


parente:
Do you want to put it in minimal-notebook or scipy-notebook? The most common use will be in the scipy/datascience containers I would imagine. r-notebook might also need it, but that image is getting very little traction. Spark+HDF5 is doable but less common, I imagine.

jakirkham:
+1 for scipy-notebook

jakirkham:
@mkolovic, any thoughts on moving this to the scipy-notebook? Minimal notebook is a bit big already (texlive, etc.) and some users have pushed to make it smaller. Also, we already don't include thing like the SciPy stack to really get the mileage out of this.

mkolovic:
Sorry, I just threw it into minimal without looking carefully. I believe for my purposes it could even go under Julia dependencies in the datascience stack but if scipy stack is a good place I can move it there.

parente:
datascience-notebook builds from scipy-notebook so you'll get the 2 for the price of 1.

jakirkham:
Literally was just writing that.

jakirkham:
Just checked and these actually should already be in the both of those actually (via `conda`). What release are you using, @mkolovic.

mkolovic:
The datascience image ID is SHA256:703912fd02e3833c595fb87ae330d7511ab076c94cd1851550a5533ddc9ec8c1. I pulled it from Docker Hub about a week ago. The scipy stack has h5py installed via `conda`, is that what you are referring to? That implementation works fine in a Python 3 notebook, both from the scipy stack and the datascience stack. I tested today by firing both up and working through the first few exercises in the h5py quickstart guide. I think the installation via `conda` bypasses the installation of the underlying library. So while Python is fine in terms of HDF5 support in both the scipy stack and datascience stack, Julia is lacking HDF5 support in the datascience stack. Hopefully I'm not too far off.

jakirkham:
> The scipy stack has h5py installed via conda, is that what you are referring to?

Yes, that includes `libhdf5` and the `h5*` tools.

> That implementation works fine in a Python 3 notebook, both from the scipy stack and the datascience stack.

That's good. Should be in Python 2, also.

> Julia is lacking HDF5 support in the datascience stack.

This what we need to solve. We should just install Julia's HDF5 bindings package just as we do with Python's. Will need to figure out how to tell it where to look for HDF5.

> Hopefully I'm not too far off.

You're not. Just trying to get the lay of the land here.

jakirkham:
Added this issue for adding HDF5 support for Julia ( https://github.com/jupyter/docker-stacks/issues/165 ).

jakirkham:
I am going to go ahead and close this. As I think we should start a fresh PR to add this to Julia.

PR #160, opened 2016-03-18
[ci skip] Add badges to stack READMEs

parente:
Pulls and stargs


jakirkham:
Should we add [Image Layers]( https://imagelayers.io/ ) too?

PR #162, opened 2016-03-18
build JupyterHub singleuser images for each notebook stack

minrk:
For every `*-notebook` stack, a `*-singleuser` image is also built

cf jupyter/dockerspawner#82

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,11 +29,12 @@ help:
 	@echo '     tag/<stack-dirname> - tags the latest stack image with the HEAD git SHA'
 
 build/%: DARGS?=
-build/%:
-	docker build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@)
 
 build/%-singleuser: build/%-notebook
 	./internal/build-singleuser $(OWNER)/$*-notebook $(OWNER)/$*-singleuser
+
+build/%:
+	docker build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@)
 
 build-all: $(patsubst %,build/%, $(ALL_IMAGES))
 
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,11 @@ ALL_STACKS:=minimal-kernel \
 	datascience-notebook \
 	pyspark-notebook \
 	all-spark-notebook
+
+ALL_SINGLEUSERS:=$(shell echo $(ALL_STACKS) | sed "s/ /\n/g" | grep notebook | sed s/notebook/singleuser/g)
+
+ALL_IMAGES:=$(ALL_STACKS) $(ALL_SINGLEUSERS)
+
 GIT_MASTER_HEAD_SHA:=$(shell git rev-parse --short=12 --verify HEAD)
 
 help:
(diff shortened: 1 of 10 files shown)


parente:
Nice. Will wait to see how it fares on travis.

Note to self / maintainers: make sure we manually create the repos on Docker Hub and assign the proper build permissions before merging and running `make release`. (The root README here has the details in the "new stack" section).

parente:
We should also mention these both in the top-level README in the tip section and within each *-notebook README. What they're for, how to use them with JupyterHub, pointer to JupyterHub doc about singleuser.

minrk:
Looks like I've missed something in the Makefile

parente:
> docker build  --rm --force-rm -t jupyter/minimal-singleuser:latest ./minimal-singleuser

It's trying to build in a folder named `./minimal-singleuser` but it looks like the bash script reuses the folder `singleuser=dockerspawner/singleuser`

jakirkham:
@parente, I don't believe I have permissions on Docker Hub. Is there some email I missed where I needed to accept something?

parente:
@jakirkham You're all set now. Part of the "stacks" team.

jakirkham:
Just got the notification. Thanks @parente.

rgbkrk:
Looks like binstar went out during the last build. Is this good to go otherwise?

parente:
I think the DH projects still need to be created before we can push.

minrk:
I *think* so, but let's give Travis one more shot. I managed to introduce a BSD-sed vs GNU-sed incompatibility in one of the last builds.

@parente I made the DH projects just now, I think.

Makefile
```diff
@@ -12,6 +12,11 @@ ALL_STACKS:=minimal-kernel \
 	datascience-notebook \
 	pyspark-notebook \
 	all-spark-notebook
+
+ALL_SINGLEUSERS:=$(shell echo $(ALL_STACKS) | sed "s/ /\n/g" | grep notebook | sed s/notebook/singleuser/g)
```
parente: echoing this gives me `minimal-kernelnminimal-singleusernr-singleusernscipy-singleuserndatascience-singleusernpyspark-singleusernall-spark-singleuser`.

Lack of whitespace is going to screw up later steps?

minrk: There isn't a lack of whitespace, they are newlines instead of spaces.

parente:
@minrk See them now. Can you copy/paste the long description from the exist repos and adjust accordingly? Don't want to recreate the docker-hell of unknown images we had before. (The README in this repo has the steps in the Maintainer section.)

One more comment about the Makefile.

IMHO, I think we should wait til Travis makes it all the way through before merging this. There's quite a bit of magic in the makefile now. Either we debug it when travis fails or we debug it after merge and when the manual build fails later.

minrk:
DockerHub descriptions updated.

parente:
Travis is green. Output in the logs looks right to me. @minrk can you confirm?

minrk:
Yes, I think so.

parente:
Running the build /  push now.

jakirkham:
I might do another release after you. :wink: We have some changes for datascience that we would like to ship. ( https://github.com/jupyter/docker-stacks/pull/166 )

parente:
@jakirkham That's fine. We should be doing a separate build for each PR anyway. That way if something goes wrong, we know what change broke things.

parente:
Permission problems. `stacks` doesn't have access on the new repos.  Fixing now.

parente:
It's pushing now. Apparently, even though the filesystem layers already exist in other jupyter/* repositories on DH, they still have to push again to the new repos.

Disk is free, right? :money_with_wings:

jakirkham:
Went back and checked a few times. Seemed to have some issues pushing, but we did get to deploy in the last one. So, it should all be up there.

PR #171, opened 2016-03-25
Add diagram of image relationships

### LINKED ISSUE #129: Diagram of how images relate
Maybe this is somewhere and then a good nudge in the right direction would be great.

I think it would be really nice to have a diagram that shows us how the different images here are related. It seems like we have quite a few and it would really help to understand which images are affected by making a change in one Dockerfile. It could probably be some SVG so it shows up nicely in the browser (even mobile) and we could embed it in the top level Readme. Thoughts?

parente:
Fixes #129


jakirkham:
Nice. Maybe we should do SVG so it scales nicely on different devices.

Was meaning to do this as I got more invovled, but then there were issues to fix. Thanks for doing this. :smile:

parente:
Unfortunately, can't `` an svg from a relative GitHub path. I tried that first. Doesn't work.

jakirkham:
Fair enough. We can play with later when GitHub adds this feature.

jakirkham:
Thanks @parente. :smile:

PR #172, opened 2016-03-25
Move the overview image up on README page

willingc:
I really like this image and want to move it higher on the page so it catches more views :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,12 @@ If this is your first time using Docker or any of the Jupyter projects, do the f
 2. Open the README in one of the folders in this git repository.
 3. Follow the README for that stack.
 
+## A visual overview of stacks
+
+Here's a diagram of the `FROM` relationships between all of the images defined in this project:
+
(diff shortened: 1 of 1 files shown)


jakirkham:
Sure, I'm ok with that. @parente, thoughts?

Also, @willingc feel free to add `[ci skip]` to these sorts of commits. No need to build docker images for doc changes. :smile:

willingc:
@jakirkham Sure. Where do I put [ci skip] in the commit message or somewhere in the PR? Learning something new. Cool.

jakirkham:
Normally, I stick with the commit message as most CIs know how to handle that. Though some will add a merge commit with `master` that includes the PR title. In that case, the PR title will work too. No need to do it now though.

jakirkham:
I'm going to go ahead and merge this. We can always play with the Readme more. I think @parente is probably busy.

parente:
It's OK. My only comment is that this feels more like how stuff works under the hood and that it might scare off some users who just want to get an image and run it.

jakirkham:
Fair enough, what if we add a brief summary before that just gives a feel as to how the stacks are grouped and then segue to the diagram. Do you think that would make the transition a little nicer?

willingc:
@parente and @jakirkham Alternatively, a new visual might be created. What I liked about this one is that it gave a very easy view of all the different stacks that are available. 

Thanks for the CI tip too.

PR #174, opened 2016-03-30
[WIP] Add JupyterHub Docker Swarm example.

jtyberg:
This is very much a work in progress.  I'm putting it out there so folks know it's coming, and because it's taken me quite a bit to get to this point.

I intend on breaking it to make it better.

### DIFF
--- a/examples/jupyterhub-swarm/.gitignore
+++ b/examples/jupyterhub-swarm/.gitignore
@@ -1,0 +1,1 @@
+userlist
--- a/examples/jupyterhub-swarm/Dockerfile
+++ b/examples/jupyterhub-swarm/Dockerfile
@@ -1,0 +1,11 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+FROM jupyter/jupyterhub:master
+
+# Install dockerspawner and its dependencies
+RUN /opt/conda/bin/pip install \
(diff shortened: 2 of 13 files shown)


minrk:
Hi, this is super cool! We have been discussing putting together some reference deployments of JupyterHub, and one of them might look very much like this, though perhaps not with swarm. Would you be interested in helping out with that? The target would be a little different from docker-stacks, which are meant more for use as base images, whereas the reference deployments would be meant more as examples on which to base deployments. As such, we were planning to do it in dedicated repos.

jtyberg:
@minrk First of all, I agree that docker-stacks is not the right place for reference deployments of JupyterHub.  If there are plans to create new repos for such examples, I'm all for it.  Just show me the links.

I'll be happy to help with the reference deployments too.  Fair warning: I'm a strong proponent of using Docker because the Docker community has a lot of momentum, and they're figuring out how to make deploying things easier.   For example, I just updated this example to persist notebook directories using a third-party Docker volume driver that supports volumes backed by NFS.  

That said, if folks want to deploy JupyterHub using Swarm or Mesos or Kubernetes as their cluster scheduler, fine.  If they want to deploy to SoftLayer or AWS or wherever, fine.  Whatever their choices, I think we need to have enough examples that folks are able to deploy JupyterHub on their platform in a couple of hours, max.

PR #183, opened 2016-04-11
minimal-notebook/README: Change NB_USER to NB_UID

jcfr:
Update text to be consistent with list of documented options.


parente:
Thanks @jcfr. Good catch. Unfortunately, the mistake is copy-pasted to all the READMEs. If you'd like to add a commit with all those as well, we'll surely take it. If not, we can fix after this one.

jcfr:
@parente Thanks for the review. Closing this PR. Will submit an other one fixing all READMEs

PR #184, opened 2016-04-11
README: Change NB_USER to NB_UID

jcfr:
Update text to be consistent with list of documented options.


jcfr:
@parente Here is an updated PR to follow up on #183

parente:
Much obliged! 🍰

jcfr:
Merci

PR #189, opened 2016-04-22
Update README.md

wernight:
Fixes https://github.com/jupyter/docker-stacks/issues/190

Fixes inconcistent PEM path (described as being `/home/jovyan/.local/share/jupyter/notebook.pem` and not `/home/jovyan/.ipython/profile_default/security/notebook.pem`).

Merges with README of https://github.com/jupyter/docker-stacks/tree/master/minimal-notebook

### DIFF
--- a/datascience-notebook/README.md
+++ b/datascience-notebook/README.md
@@ -38,7 +38,7 @@ You can sidestep the `start-notebook.sh` script entirely by specifying a command
 You may customize the execution of the Docker container and the Notebook server it contains with the following optional arguments.
 
 * `-e PASSWORD="YOURPASS"` - Configures Jupyter Notebook to require the given password. Should be conbined with `USE_HTTPS` on untrusted networks.
-* `-e USE_HTTPS=yes` - Configures Jupyter Notebook to accept encrypted HTTPS connections. If a `pem` file containing a SSL certificate and key is not found in `/home/jovyan/.ipython/profile_default/security/notebook.pem`, the container will generate a self-signed certificate for you.
+* `-e USE_HTTPS=yes` - Configures Jupyter Notebook to accept encrypted HTTPS connections. If a `pem` file containing a SSL certificate and key is not provided (see below), the container will generate a self-signed certificate for you.
 * `-e NB_UID=1000` - Specify the uid of the `jovyan` user. Useful to mount host volumes with specific file ownership. For this option to take effect, you must run the container with `--user root`. (The `start-notebook.sh` script will `su jovyan` after adjusting the user id.)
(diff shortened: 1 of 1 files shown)


jakirkham:
Please consider adding `[ci skip]` to commit messages that are doc only.

parente:
@wernight Thanks for the PR. The old path actually appears in the READMEs of all the other stacks as well. (We duplicate for reader convenience instead of having end users understand the inheritance chain of images and options.)

Would you mind updating the other READMEs with the same fix? If you're busy, we can merge this and get the others in a follow-on.

wernight:
Sorry I'm kind of busy. :+1:

parente:
NP. I'll send a PR for the others.

jakirkham:
I wonder if we should have some sort of rendering step maybe using jinja to propagate changes like this through the READMEs?

So, there would be a template that we fix and then something we run that changes all of the actual READMEs. Not sure that they are standardized enough for this, but we could at least move in that direction.

parente:
I was thinking about that as I wrote the comment above. Could just do `make docs` and have it render. But I'd hate to do something custom while the rest of jupyter is using sphinx and readthedocs. On the other hand, having everything here in github as a README so that it just shows up is valuable.

jakirkham:
> On the other hand, having everything here in github as a README so that it just shows up is valuable.

Oh, I'm not say that shouldn't be the case. I am thinking a git pre-commit hook that kicks the process off.

> But I'd hate to do something custom while the rest of jupyter is using sphinx and readthedocs.

Though we do have different needs here so I think that is an ok reason to investigate something that works for us. For instance, we want to have READMEs that are on Docker Hub. We like being able to see them when browsing that directory in GitHub or locally. Sphinx and readthedocs aren't really designed for that (unless I am missing something).

That being said, I'd certainly appreciate feedback from other people in the Jupyter ecosystem on this point.

parente:
> Sphinx and readthedocs aren't really designed for that

This is the point I was trying to make. You and I are on the same page. :)

jakirkham:
Ok, so we are agreeing and one of us (me) is not awake. :smile:

parente:
Nah. I just did a bad job of making the point. 🙊

wernight:
If you have some job, please also update Docker Hub. With autobuild it's done automatically (i wonder why it cannot be autobuilt anymore).

jakirkham:
It's proved horribly problematic. Basically the relationships between images aren't always respected. Also, it is hard to tell what changes are included in a release. This, in turn, makes hard to track down bugs.

Though we have considered using Travis to build and release, this was not initially possible due to the build time. We did recently cut that down significantly so we could revisit it.

Though there are a couple of challenges. Travis reliably busts the cache early. We cannot reproduce this in any configuration we have tried. This will mean many extra layers must be pulled down even if they haven't changed. Also, Docker Hub often fails to receive pushed images so we have to retry several times. This is quite slow meaning it could cause time limits to be exceeded on Travis.

PR #207, opened 2016-05-12
Update README links

willingc:
Update files for changed link.

### DIFF
--- a/all-spark-notebook/README.md
+++ b/all-spark-notebook/README.md
@@ -195,7 +195,7 @@ Connection to Spark Cluster on Standalone Mode requires the following set of ste
 
 ## Notebook Options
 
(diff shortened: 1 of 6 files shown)


jakirkham:
Yeah, @parente and I were talking about this before. Maybe you have some thoughts on this, @willingc.

It seems like there is a lot of repetition in the READMEs. One thing that would aid updates like this is if we generate the repeated parts using a common source document. This would allow one to make one change in one place and propagate it. The main downside is that updating the READMEs becomes a little bit more complicated.

jakirkham:
LGTM. Are there more changes planned here? Anything else we need to wait for here?

willingc:
@jakirkham No more changes right now. I'll ask around at the next couple of conferences that I'm at for tips on the best way/simplest way to minimize the repetition. Thanks. On another note, I have been working with Docker on OS X Beta and it is soooooooo much better than boot2docker.

jakirkham:
Sounds good. In it goes. Thanks @willingc.

jakirkham:
Thanks for looking into how to reduce repetition in doc maintenance.

Ooh. I need to try that. Still rocking `docker-machine` ATM. `boot2docker` was painful. Getting rid of the old CLI helped.

PR #208, opened 2016-05-14
Update debian base image for CVEs

parente:
https://github.com/docker-library/official-images/pull/1695


parente:
Have been unable to build and push this so far. Every attempt is met with failure apt-get installing a random package from the large list of packages in minimal-notebook. Rate limit?

rgbkrk:
Does the host need to be updated, DNS cache cleaned out?

parente:
Nope. 10th try in a row was a charm.

rgbkrk:
Ah yes the good old "do it 10 times and then it works" trick.

Issue #205, opened 2016-05-09
images are huge


gimoh:
Current images are quite huge, which is especially surprising regarding the minimal images.  Are all these packages like editors (jed, vim, emacs), pandoc, tex, etc. really necessary for running the notebook?

parente:
> jed, vim, emacs

These are not required, but were oft requested as being necessary for getting work done in the terminal window of the notebook server. (@fperez)

> pandoc, tex

These are required by the notebook server for many of the "Download as" menu items to work, otherwise they return 500 errors.

It's been asked in the past if there could an tiny-notebook stack that includes the bare minimum to get the notebook server to start, but doesn't enable all of its features. I bet there could be.

I'm assuming you're looking for a base image for your own stack?

gimoh:
Well, I only do a minimal customisation, but yeah a more modular stack would be useful.  The problem with size is transfer time and obviously storage cost.

Mind you I haven't checked which exact packages are bumping the size significantly, these are just the ones that looked suspicious ;-)

As for editors, I usually avoid putting them in base image and instead just run another container with an editor (like the web-based [brackets](https://hub.docker.com/r/fabric8/brackets/)) and `--volumes-from=FOO` which allows me to edit files in data volumes.  This works for me but I suppose it doesn't handle all scenarios (files outside of volumes).

minrk:
Docker's linear inheritance makes it *super* bad at doing anything modular. It's possible that we should remove tex from the minimal-notebook image, though it does mean that some notebook functionality (download as .pdf) would stop working. If we do remove it from minimal, we should add it back in ~every descendent image.

gimoh:
Aye, maybe modular wasn't the right word, I meant the split into tiny-notebook and minimal-notebook (or even minimal-notebook (which is really minimal) and full-notebook), like @parente mentioned.

willingc:
tiny-notebook seems a good idea for the spartan, no-frills, no-extras version. We could add a note that some functionality like download as PDF is not available in this version.

fperez:
I'm OK dropping those editors as well from the tiny one too. Do we have the package sizes for each? I thought jed was tiny, emacs is obviously much bigger and can probably be dropped from most (jed is a decent emacs stand-in at very small size). vim without too many plugins should also be pretty small, but we should check.

parente:
I think we need to identify the use case for tiny before we can make a decision on what to have in it vs what to drop.

[...]

PR #209, opened 2016-05-14
Base notebook stack

parente:
Continuing the discussion from #205. Not ready to merge. Includes #208 for most up to date size comparison.

With this definition, tiny-notebook is about ~1/5th the size of minimal-notebook.

```
jupyter/tiny-notebook         latest    1703acbd84d4    Less than a second ago   416.9 MB
jupyter/minimal-notebook      latest    4a0d21eb14ef    15 minutes ago           1.869 GB
```

Decisions for discussion:

* tiny-notebook includes the notebook config and startup scripts that currently enable the various configuration features (e.g., USE_HTTP, PASSWORD, etc.) in minimal-notebook and, by inheritance, all of the other images. I think it's important to include these in tiny-notebook for consistency with the other docker-stacks images. Otherwise, what's the point of starting from a docker-stacks image? A developer can just grab any OS image, add python, and pip/conda install notebook.
* tiny-notebook does install jupyterhub and the correspoding start-singleuser script. (+18 MB, included in the size reported above
* tiny-notebook does not install build-essentials or python-dev at a savings of 200 MB and a cost of being able to build C extensions out of box.

Other notes:

* terminado is now an explicit prereq to notebook 4.2 according to conda so I removed the explicit mention of it from the notebook install line.


parente:
@gimoh, @fperez any comments on this definition of "tiny"?

willingc:
@parente How much space does it add for vim and git? 

On another note, have you tried Alpine yet? While I believe we should have a Debian tiny, it might be worth playing around with this at some point. It may be helpful for Raspberry Pi and education.

parente:
> How much space does it add for vim and git?

Will check.

> On another note, have you tried Alpine yet? While I believe we should have a Debian tiny, it might be worth playing around with this at some point. It may be helpful for Raspberry Pi and education.

I tried alpine at one point. It's tends of MB smaller to start, but once you lob in miniconda and such, the savings over debian really doesn't matter much in the desktop / cloud realm. That said, for new use cases, I agree another micro stack might still make sense.

parente:
@willingc, with git and vim:

```
jupyter/tiny-notebook              latest              cd85f25a2c47        Less than a second ago   505.9 MB
```

Nearly 100 MB increase. I think they should be excluded here since they're not required dependencies for the jupyter components to work.

willingc:
@parente agreed. Thanks for checking.

minrk:
I do feel a bit odd having tiny < minimal. If I had my druthers, I'd rename 'minimal' to 'base' or 'basic' and then have 'minimal' be the truly minimal stack, but I don't know who might be relying on minimal being not-so-minimal.

> tiny-notebook does install jupyterhub and the correspoding start-singleuser script. (+18 MB, included in the size reported above

Installing JupyterHub is not required to get the single-user script. In particular, it pulls in unused dependencies. This could be simplified to just fetch the script directly:

    wget https://raw.githubusercontent.com/jupyterhub/jupyterhub/0.6.1/scripts/jupyterhub-singleuser

parente:
> I do feel a bit odd having tiny < minimal. If I had my druthers, I'd rename 'minimal' to 'base' or 'basic' and then have 'minimal' be the truly minimal stack, but I don't know who might be relying on minimal being not-so-minimal.

I agree the naming is not a great idea, but I'm worried about the ~5000 pulls of the minimal-notebook:

https://gist.github.com/parente/316d5c242aeb484484c8

I'd certainly be peeved if one day I decided to go with the latest minimal-notebook and fund that it changed from "minimal image with a fully functioning notebook to get work done" to "minimal image with a minimally functioning notebook for use as a base image".

Maybe this new image should be called base-notebook, since it is at the top of the hierarchy and is meant to serve as a base Docker image, and minimal-notebook stays as is?

> Installing JupyterHub is not required to get the single-user script.

Yes, except @jtyberg felt yanking one file out of GitHub felt messy in the original PR (https://github.com/jupyter/docker-stacks/pull/185#issuecomment-208983308).

willingc:
@parente @minrk Good points. IMHO 'base' is an improvement over 'tiny'. Probably makes sense to add a hierarchy to the Docker readme: base < minimal < ...

parente:
@willingc I'll definitely update the inheritance diagram in the root README once we settle on the names.

gimoh:
Fantastic, it's looking really good now!  Outstanding, thanks :-)

parente:
Rebased. I think this can go in now.

parente:
Starting the build for this.

parente:
Getting anaconda over capacity and `Fetching package metadata: ....Error: Could not find anaconda.org user r` errors. Will keep retrying.

PR #213, opened 2016-05-27
Discuss: Remove minimal kernel

parente:
The minimal-kernel image sees very little use (less than 600 pulls since it was added). It's purpose was to get the kernel gateway (KG) into a small image that could be the basis for adding kernels and libraries for specific use cases. The work over in the incubator on the dashboard projects use the KG heavily, but we quickly found that adding it to any of the other fully-featured stack images (all-spark-notebook, datascience-notebook, ...) was always easier than trying to build up what was needed from a small image.

Anyone object to killing off this stack?


rgbkrk:
No objection here. It's too bad we can't see when people have pulled images, over time.

parente:
I thought about turning https://gist.github.com/parente/316d5c242aeb484484c8 into something that collects the stats once a day over time like https://gist.github.com/parente/facb555dfbae28e817e0 but haven't gotten around to it.

rgbkrk:
Well done! Where have you been storing these? Just updating the CSV in the gist?

parente:
I have a (mostly) scheduled task to clone the gist, nbconvert execute the notebook, commit and push the updated files back to the gist.

minrk:
Seems fine to me. Thanks, @parente!

parente:
Rebased on master after sending in #209.

jakirkham:
I'd always thought about using this, but never ended up doing so. Sounds reasonable to get rid of it to me too.

rgbkrk:
Thank you for deletion! You even cleaned up the diagrams which is super stellar.

Issue #188, opened 2016-04-22
Jupyter images cannot be run in a secured multi tenant Docker environment.


GrahamDumpleton:
Although you have tried to set up the Docker images to be able to run as a specific non `root` user, this is not sufficient in a Docker environment which provides support for multi tenant hosting.

In such an environment, as well as enforcing that images cannot be run as `root`, they will also run the Docker containers as different user IDs for different real users or customers, or even different applications.

Different user IDs are used to reduce the risk that were someone to break out of the Docker container, even though they wouldn't be `root`, that they then still cannot access the data of, or otherwise interact with, another application.

Even with what you have tried to do with ensuring that a non `root` user is used, it is not done in a way that a hosting environment can actually verify the container isn't running as `root`. This is because you use a named user for the `USER` statement in Docker. This says nothing about the real uid behind it, which is going to be dictated by the local passwd file within the Docker container, something which can't readily be verified by a hosting environment. A secured Docker environment is not therefore going to allow your image to run.

To work in such an environment a number of changes are required to the Jupyter images. I am going to describe the required changes here and explain why they are needed.

I am happy to provide a pull request containing all the changes, but would only do that if there is an indication that you would be willing to make the changes. If for some reason you are going to baulk on making the changes, then some other solution would need to be found. If no solution can be found that you are happy with, then the Jupyter Notebooks will not be able to be run in such environments and so you would miss out on potential users. Given that such environments are going to become more prevalent in enterprise cloud deployments where security is very important, it is a large space you would be missing out on.
## Integer User ID vs Named User

The first problem as mentioned above is the use of:

```
USER jovyan
```

when closing out the `Dockerfile`. This results in the `ContainerConfig.User` setting within the Docker manifest data being `jovyan`.

In a hosting environment which blocks the running of Docker images as `root`, this image would still not be allowed to run. This is because a hosting environment cannot assume that `jovyan` hasn't actually been mapped to use the uid of 0 in the `passwd` file within the container.

To remedy this issue, the `Dockerfile` for each image should instead be closed out with:

```
USER 1000
```

where 1000 is the actual uid for the `jovyan` user. In other words, the integer uid of the UNIX user account should be used in `USER` and not the name of the user. That way the hosting environment can actually trust that your image is actually going to run as uid `1000`.
## Group That User Belongs To

The next issue is the group that the `jovyan` user is added to, as created by:

```
# Create jovyan user with UID=1000 and in the 'users' group
[...]

Issue #10, opened 2015-08-20
Doc / alias pip for python 2 and 3


parente:
See chat starting about here:

https://gitter.im/jupyter/docker-stacks?at=55d5beb41250e6600d08d27d

basnijholt:
Chat went down, but the discusion was that there is no easy way to find out how to switch to Python 2 environment when you start the image with `/bin/bash`.

parente:
I added a section to the bottom of every relevant README about the conda environments present in each image, how to switch between them, and what key commands are available in each (python, ipython, pip, conda, ...)

PR #217, opened 2016-05-30
Various tuneups

parente:
Fixed a handful in one PR to avoid multiple costly rebuilds.

* Fix permissions when host mounting SSL cert/key
* Disable pip cache during docker build (#188)
* Set HOME env var (#188)
* Add pip2, pip3 symlinks for images with python2 (#10)

Based on PR #216 so that the tests run on travis post build. Merge that PR before this one.


base-notebook/Dockerfile
```diff
@@ -49,7 +50,7 @@ USER jovyan
 # Setup jovyan home directory
 RUN mkdir /home/$NB_USER/work && \
     mkdir /home/$NB_USER/.jupyter && \
-    mkdir /home/$NB_USER/.local && \
+    mkdir -p -m 700 /home/$NB_USER/.local/share/jupyter && \
```
minrk: why does this directory need to be created manually?

parente: If a user follows the directions in the readme to host mount a ssl certificate where we have preconfigured the notebook server to look for it, docker creates the intervening directories with root ownership and read only permissions. This effectively prevents the notebook server from writing other files into this directory like the notebook signing key and breaks everything.

We could change the notebook config to look for the pem file in another directory and break backward compatibility with older images. I opted for keeping the file where it has been and making the subdirectories ahead of time.

minrk: Gotcha, thanks.

Issue #210, opened 2016-05-20
unable to plot in jupyter notebook using R kernel


fyears:
I am using default configuration of [`jupyter/datascience-notebook`](https://github.com/jupyter/docker-stacks/tree/master/datascience-notebook).

By accessing `http://docker.local:8888/`, I could start a notebook with R kernel, but I could not generate plots inside the notebook.

The simple official example:

```R
require(stats) # for lowess, rpois, rnorm
plot(cars)
lines(lowess(cars))
```

![capture](https://cloud.githubusercontent.com/assets/1142836/15414326/76ff94ce-1dee-11e6-85ac-7e5413bbfc4d.PNG)

The #54 might relevant, but I am using the most up-to-todate image (20160519) .

parente:
Thanks for the report. Definitely looks like #54 again. libxrender moved from datascience-notebook to minimal-notebook (https://github.com/jupyter/docker-stacks/blob/master/minimal-notebook/Dockerfile#L33) so it should still be available. Must be something else wrong now.

michael-erasmus:
I'm running into the exact same issue. 

Not sure if this helps, but if I run `start-notebook` is see the following output whenever I try plotting anything. It also creates PDF files on disk.

```
dev.new(): using pdf(file="Rplots3.pdf")
```

michael-erasmus:
I also spotted this in the output, right before the Kernel crashed:

```
Warning messages:
1: In doTryCatch(return(expr), name, parentenv, handler) :
  unable to load shared object '/opt/conda/lib64/R/modules//R_X11.so':
  libjpeg.so.8: cannot open shared object file: No such file or directory
2: In png(tf, width, height, "in", pointsize, bg, res, antialias = antialias) :
  unable to load shared object '/opt/conda/lib64/R/library/grDevices/libs//cairo.so':
  libjpeg.so.8: cannot open shared object file: No such file or directory
3: In png(tf, width, height, "in", pointsize, bg, res, antialias = antialias) :
  failed to load cairo DLL
4: In file(con, "rb") :
  cannot open file '/tmp/RtmpjYzEfu/file183dd7bd4f.png': No such file or directory
5: In svg(tf, width, height, pointsize, FALSE, family, bg, antialias) :
  failed to load cairo DLL
6: In file(con, "rb") :
  cannot open file '/tmp/RtmpjYzEfu/file181f35e31f.svg': No such file or directory
[...]

PR #222, opened 2016-06-23
Added shell for accessing odbc to supply data to Jupyter.  Fixed a fe…

SternAndrew:
Added shell for accessing odbc to supply data to Jupyter.  Fixed a few issues by pulling in newer versions of rpy2 and ipywidgets.

1) Added freetds to allow access to database instances using odbc.
2) Create a shell .odbc.ini file that can be filled out to access a database instance.
3) Updated the version of rpy2 since the %R and %%R magic wouldn't work from the Python kernels so that R can be used in cells within a Python sheet.
4) Updated ipywidgets so that the widgets can be used within a Python3 sheet. Without this update it wouldn't show the controls in a Python3 sheet after running the sheet
5) Pinned the conda to jpeg 8 since the ggplot2 graphs wouldn't showup in R sheets.  See https://github.com/jupyter/docker-stacks/issues/210

### DIFF
--- a/datascience-notebook/Dockerfile
+++ b/datascience-notebook/Dockerfile
@@ -94,12 +94,12 @@ RUN echo "[ODBC Data Sources]" >> ~/.odbc.ini && \
     echo "Driver           = /home/jovyan/odbcdriver/lib/libtdsodbc.so" >> ~/.odbc.ini && \
     echo "Trace            = No" >> ~/.odbc.ini && \
     echo "TraceFile        = /home/jovyan/prodms.log" >> ~/.odbc.ini && \
-    echo "Server           = MachineIPAddress >> ~/.odbc.ini && \
-    echo "Host             = MachineHostName >> ~/.odbc.ini && \
+    echo "Server           = MachineIPAddress" >> ~/.odbc.ini && \
+    echo "Host             = MachineHostName" >> ~/.odbc.ini && \
     echo "Port             = MachineDBPort" >> ~/.odbc.ini && \
     echo "Database         = DatabaseInstanceName" >> ~/.odbc.ini && \
     echo "UID              = DatabaseUserName" >> ~/.odbc.ini && \
-    echo "PWD              = DatabasePassword >> ~/.odbc.ini && \
+    echo "PWD              = DatabasePassword" >> ~/.odbc.ini && \
     echo "Protocol         =" >> ~/.odbc.ini && \
     echo "ReadOnly         = No" >> ~/.odbc.ini && \
     echo "RowVersioning    = No" >> ~/.odbc.ini && \
--- a/datascience-notebook/Dockerfile
+++ b/datascience-notebook/Dockerfile
@@ -27,12 +27,14 @@ RUN apt-get update && \
 
 USER jovyan
 
-# R packages including IRKernel which gets installed globally.
+# Fix the version of rpy2 since the 2.7 is broken.
 # Note that there is a bug with rpy2 where the %R and %%R don't work.
-# We need an newer version not yet available on conda to fix this so uninstall and install new version at end.
+# We need an newer version not yet available on the default for conda to fix this.
+RUN conda install -c bioconda rpy2=2.7.8
+
+# R packages including IRKernel which gets installed globally.
 RUN conda config --add channels r && \
     conda install --quiet --yes \
-    'rpy2=2.7*' \
     'r-base=3.2*' \
     'r-irkernel=0.5*' \
     'r-plyr=1.8*' \
@@ -69,11 +71,6 @@ RUN julia -e 'Pkg.add("IJulia")' && \
 RUN echo 'push!(Sys.DL_LOAD_PATH, "/opt/conda/lib")' > /home/$NB_USER/.juliarc.jl && \
 julia -e 'Pkg.add("Gadfly")' && julia -e 'Pkg.add("RDatasets")' && julia -F -e 'Pkg.add("HDF5")'
 
-# Fix the graphing issue with ggplot2
-# from: https://github.com/jupyter/docker-stacks/issues/210
-RUN echo "jpeg 8*" >> /opt/conda/conda-meta/pinned
-RUN conda update --all -y
-
 # Add odbc drivers to the install
 RUN cd /home/jovyan && \
     wget ftp://ftp.freetds.org/pub/freetds/stable/freetds-patched.tar.gz && \
@@ -108,14 +105,4 @@ RUN echo "[ODBC Data Sources]" >> ~/.odbc.ini && \
     echo "FakeOidIndex     = No" >> ~/.odbc.ini && \
     echo "ConnSettings     =" >> ~/.odbc.ini && \
     echo "TDS_Version      = 7.0" >> ~/.odbc.ini 
-
-# Fix the version of rpy2 since the 2.7 is broken.  See note above
-RUN conda uninstall rpy2 && \
-    conda install -c bioconda rpy2=2.7.8
-
-# Fix the ipywidgets since the base image has an older version of ipywidgets.
-# This older version works with the Python2 instance but not the Python3 instance.
-# Installing a newer version allows the widgets to work with both Python versions.
-RUN conda uninstall ipywidgets && \
-    conda install -c conda-forge ipywidgets=5.1.5
 
--- a/datascience-notebook/Dockerfile
+++ b/datascience-notebook/Dockerfile
@@ -18,6 +18,9 @@ RUN apt-get update && \
     gcc && apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
+# Python pyodbc
+RUN pip install pyodbc
+
 # Julia dependencies
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
--- a/datascience-notebook/Dockerfile
+++ b/datascience-notebook/Dockerfile
@@ -27,15 +27,18 @@ RUN apt-get update && \
 
 USER jovyan
 
+# R packages including IRKernel which gets installed globally.
+RUN conda config --add channels r && \
+    conda install --quiet --yes \
+    'r-base=3.2*' 
+
 # Fix the version of rpy2 since the 2.7 is broken.
 # Note that there is a bug with rpy2 where the %R and %%R don't work.
 # We need an newer version not yet available on the default for conda to fix this.
 RUN conda install -c bioconda rpy2=2.7.8
 
 # R packages including IRKernel which gets installed globally.
-RUN conda config --add channels r && \
-    conda install --quiet --yes \
-    'r-base=3.2*' \
+RUN conda install --quiet --yes \
     'r-irkernel=0.5*' \
     'r-plyr=1.8*' \
     'r-devtools=1.9*' \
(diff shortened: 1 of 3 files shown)


parente:
@SternAndrew Thanks for the contribution.

I see there's quite a few fixes going into this PR, done by uninstalling packages and installing new ones here. It would probably be better to fix these issues in the base images so that we're not bloating the image with the old layers plus the new ones.

I'm not clear on how reusable the ODBC support is that you're adding. I see the sample configuration. Are users expected to provide their own? I'm asking to understand if this makes sense as an addition directly to the docker stack image, if it should be a recipe on the wiki, or if it should be a separate image entirely that just starts `FROM jupyter/datascience-notebook`.

SternAndrew:
I expect at some point the base versions of both rpy2 and ipywidgets will get updated in conda and that these patch lines at the end can just be removed.  Better to localize the fixes since we shouldn't need them in the long run.  I feel the same way about the fix for the ggplot2 using the jpeg 8 setting.  If these were long term additions I would agree fully with you but the desire to remove these as quickly as possible and to shrink the image should remain. This increases the desire to get the conda versions updated to fix the problems but until they are fixed there is no reason for new users trying out your docker image to experience these issues.  It took quite some time to figure out how to fix these issues and others shouldn't have to go through the same process.

I added the .odbc.ini file but you are correct that.additional sections should be added by the user.  This just provides a bit of documentation that will help new users setup and get information from odbc sources as not everyone is using sample data or csv files.  The database provides some real usage results that I was able to demo.  The addition of freetds and the .odbc.ini file will help others so that they only need to append to the file I created with their database configuration to access data within their environment.  I expect them to use this as a base and just add some appended lines before having a running instance. By placing it somewhere else it just makes it harder for new users to find the information.  The increase in space is minor but makes it much faster for users to see what Jupyter can really do for them.

I have attempted to also add Ruby to my running base image but the IRuby kernel dies so I didn't include it in the docker file.  Although it would be nice to have a base image with Ruby available if any new users like that language since it is fairly popular.

My general impression is that trying out the docker image is the first experience with Jupyter and that it should just work and have everything they would need.  It helps them quickly create demos and figure out how helpful this project could be as a tool.  I expect that after the demo on the docker image they will create a fuller instance on hardware or create more complete docker images for corporate use.

jakirkham:
To give you some background @SternAndrew, the `jpeg` problem is a bit more complex. We decided at [conda-forge]( http://conda-forge.github.io/ ) that `jpeg` 9 was necessary and trying to build with `jpeg` 8 was problematic. However, `defaults` hasn't caught up yet. This affects a wide range of things including `qt`. As a result, it will be a bit before `defaults` switches to `jpeg` 9. It seems totally sensible to me to set that constrain lower in the stack to ensure things like `matplotlib` don't get messed up (in short I agree with @parente). Now knowing this would you be ok with moving that lower in the stack?

SternAndrew:
I have moved the fixes earlier in the stack.  Thank you for the background it did help.

parente:
>  It seems totally sensible to me to set that constrain lower in the stack to ensure things like matplotlib don't get messed up (in short I agree with @parente). Now knowing this would you be ok with moving that lower in the stack?

@jakirkham did you mean "into the base images" by lower in the stack?

jakirkham:
Yes, sorry, that was a bit vague. Does that still seem reasonable to you or did I misunderstand something?

parente:
> Yes, sorry, that was a bit vague. Does that still seem reasonable to you or did I misunderstand something?

Yes. I think the PR might need to be split up a bit to target fixes in the right places. Otherwise, the datascience image has all the goodness / upgrades / fixes while the rest are left out in the cold.

Fixes aside, we also have decide if the ODBC support is something that belongs only in this image, in the base images, or outside docker-stacks as a recipe / other image. If it does wind up here, it at least needs some doc about what the user has to do starting from the template in order to us it.

SternAndrew:
I have completed breaking up the fixes.  I'm fine if you move the odbc stuff into an earlier base image.

parente:
@SternAndrew I'd like to do a bit more cleanup on your PR before merging and experiment a bit with where the odbc support should go. I'm going to leave a couple notes here. I'll branch from your PR to make the follow-on edits myself in the near future (retaining your original commits of course). If you'd like to carry on with making the changes yourself, don't let me stop you.

datascience-notebook/Dockerfile
```diff
@@ -11,9 +11,16 @@ RUN apt-get update && \
...
     gfortran \
+    unixodbc-dev \
+    libtool-bin \
+    autoconf \
+    automake \
     gcc && apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
+# Python pyodbc
+RUN pip install pyodbc
```
parente: Can we get from conda instead?

datascience-notebook/Dockerfile
```diff
@@ -56,4 +75,40 @@ RUN julia -e 'Pkg.add("IJulia")' && \
 # Show Julia where conda libraries are
 # Add essential packages
 RUN echo 'push!(Sys.DL_LOAD_PATH, "/opt/conda/lib")' > /home/$NB_USER/.juliarc.jl && \
-    julia -e 'Pkg.add("Gadfly")' && julia -e 'Pkg.add("RDatasets")' && julia -F -e 'Pkg.add("HDF5")'
+julia -e 'Pkg.add("Gadfly")' && julia -e 'Pkg.add("RDatasets")' && julia -F -e 'Pkg.add("HDF5")'
+
+# Add odbc drivers to the install
+RUN cd /home/jovyan && \
+    wget ftp://ftp.freetds.org/pub/freetds/stable/freetds-patched.tar.gz && \
```
parente: Checksum for insecure package downloads.

datascience-notebook/Dockerfile
```diff
@@ -56,4 +75,40 @@ RUN julia -e 'Pkg.add("IJulia")' && \
...
+    tar xvzf freetds-patched.tar.gz && \
+    cd /home/jovyan/freetds-1.00.6 && \
+    ./configure --prefix=/home/jovyan/odbcdriver && \
+    make && \
+    make install
+
+# Add ODBC configuration file
+RUN echo "[ODBC Data Sources]" >> ~/.odbc.ini && \
```
parente: Need to update README to describe what the user does to this. A sample that works out of the box when bound to a DB in another docker container would be good too. If we merge this, we have to support this, so we better have a way to ensure it works.

datascience-notebook/Dockerfile
parente: Also, pin version.

datascience-notebook/Dockerfile
```diff
@@ -43,8 +57,13 @@ RUN conda config --add channels r && \
     'r-nycflights13=0.1*' \
     'r-caret=6.0*' \
     'r-rcurl=1.95*' \
+    'r-dbi=*' \
+    'r-scales=*' \
```
parente: We should really pin versions. We've had no end of pain with failed rebuilds when versions are allowed to float arbitrarily.

datascience-notebook/Dockerfile
SternAndrew: Haven't tried that out but if it works it would be preferred over the non-conda.  Do you have a database to test against?  We should install the conda version for both Python2 and Python3.

scipy-notebook/Dockerfile
```diff
@@ -14,9 +14,13 @@ RUN apt-get update && \
 
 USER jovyan
 
+# Fix the ipywidgets since the base image has an older version of ipywidgets.
+# The older version works with the Python2 instance but not the Python3 instance.
+# Installing a newer version allows the widgets to work with both Python versions.
+RUN conda install -c conda-forge ipywidgets=5.1.5
```
parente: No need to do this separate. conda-forge is configured as a system channel in the base-notebook (https://github.com/jupyter/docker-stacks/blob/master/base-notebook/Dockerfile#L64)

I confirmed locally that a regular install without specifying the channel picks up ipywidgets 5.1.5. (But would be nice if @jakirkham can confirm that this is the correct behavior.)

datascience-notebook/Dockerfile
SternAndrew: Agreed.  Not sure how this would be implemented correctly so accepting of any solution you suggest or code.

parente: There's a few examples in the other stack images. base or minimal notebook has one for the miniconda download.

datascience-notebook/Dockerfile
parente: When I run conda install for pyodbc I get v3.0.10. I get the same when running pip install. I'm assuming their equivalent.

Re: database to test against, I think that's something we're going to want to document in the README (see below). Linking to one of the official maria, mysql, postgres, etc. docker containers to do it is probably the simplest way.

datascience-notebook/Dockerfile
SternAndrew: Agreed.  Should pin as of versions 4 days ago when I did my last full build and what I'm currently using.  Here are the versions I pulled down:

jovyan@xxx:~/work$ conda list | grep dbi
r-dbi                     0.3.1                 r3.2.2_2a    r
jovyan@xxx:~/work$ conda list | grep scales
r-scales                  0.3.0                 r3.2.2_0a    r

datascience-notebook/Dockerfile
SternAndrew: Seems like a good idea.  Perhaps the .odbc.ini file could be prepared to use the correct credentials for the maria, mysql image.  Perhaps a base database per-populated and setup as a test odbc image preconnected to test the odbc connection.  That way the automated build might be able to run a system test using this populated data  so future changes can make sure the odbc stuff still is working.

datascience-notebook/Dockerfile
SternAndrew: Added checksum for this tar package.

datascience-notebook/Dockerfile
SternAndrew: Added pinning for dbi and scales.

SternAndrew:
Note that I locked in verison 1.00.9 as the version to use but that my current setup is using freetds-1.00.6.tar.gz.

datascience-notebook/Dockerfile
parente: Prepping the default odbc.ini to work with one of the official database images sounds like a good approach. Would you like to take a crack at documenting that setup in the README? If not, I can at some point.

SternAndrew: I have a few items on my plate at this time.  I think you might get to it before me.

scipy-notebook/Dockerfile
jakirkham: So, I think we should start using a much newer version of `conda`. It appears to be 3.19.1 [currently]( https://github.com/jupyter/docker-stacks/blob/2485724a08c9ce55c296fdd486abdc5244eabe1c/base-notebook/Dockerfile#L63 ), but to get channel priority to work we need 4.1.x. Though this may come with some other surprises. That all being said, I pulled down the `base-notebook` and found it has 4.1.4 installed. So, this pinning is not actually respected. Something we should think about. Though I would recommend using something on the 4.1.x series and we should discuss fixing that pinning in a new PR.

In other words, `conda-forge` should be preferred without this channel argument.

parente: I'm not sure we ever intended to pin conda itself. The initial install has to point to some release. Thereafter we let conda upgrade itself. If the entire package mgmt system is busted I think we'd find or during the build process.

parente:
Hi @SternAndrew. So clearly neither you nor I found the time to get this PR across the finish line and into a state that would make all the additions easy for users to understand and us to maintain. I think it's best to close it out at this point and reopen it if either of us, or someone else, has the time and interest in seeing it all through. Thoughts?

PR #223, opened 2016-06-28
update sympy to 1.0

leosartaj:


parente:
This keeps failing due to network problems on travis. I'll give it a quick shot here and merge if sympy goes in without issue.

leosartaj:
Okay. Other packages here do not depend on Sympy ASFAIK, so this shouldn't break anything.

parente:
Thanks for the submission!

parente:
Build is underway but might take a bit because docker was upgraded on the build machine and apparently that broke docker cache from the last images.

leosartaj:
Thanks! @parente

parente:
Pushed as 9f5bbec91d6a

PR #224, opened 2016-06-28
Support on-demand creation of mounted work directory

okoethibm:
When a notebook container is started (as root/sudo) with a host mount (for a different user) as work directory like
    root> docker run -it --rm -v ~someuser/mynotebooks:/home/jovyan/work -e NB_UID=someuid --user=root jupyter/minimal-notebook
and the host directory ~someuser/mynotebooks does not exist, it is created on-demand by docker run but will not be writable for the notebook user, so all attempts to create a notebook will fail.
This can be avoided by adding the current directory (container work directory) to the chown call.


parente:
Yup. Makes sense. Thanks!

Going to merge this but group it in a single build/push to DockerHub with #225 and #227 to avoid multiple, giant pushes for minor, independent changes (I hope).

PR #225, opened 2016-06-28
Make use of available environment variables

eigengrau:
base-notebook defines environment variables for the Conda install path and the
notebook user. However, in some instances, these locations were still hardcoded.
Let’s use the variables instead.

### DIFF
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -17,7 +17,7 @@ RUN apt-get update && \
     gcc && apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-USER jovyan
+USER $NB_USER
 
 # R packages
 RUN conda config --add channels r && \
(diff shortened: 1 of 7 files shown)


parente:
I think docker used to treat env vars in `USER` statements as literal usernames and not fail in any noticeable way. Can you confirm that ops in the dockerfile are indeed using `jovyan` and not a `$NB_USER` user?

eigengrau:
> Can you confirm that ops in the dockerfile are indeed using jovyan and not a $NB_USER user?
Ran the containers; can confirm that the variables are evaluated.

```
Step 34 : USER $NB_USER
 ---> Using cache
 ---> e59282e100b1
Step 35 : RUN id
 ---> Running in 50b55c952f16
uid=1000(jovyan) gid=100(users) groups=100(users)
```

parente:
Thanks for the quick check. There's a series of PRs that came in last night and we still do pushes to dockerhub somewhat manually. I'll get this in, in order.

parente:
Merging but grouping it in a single build/push to DockerHub with #224 and #227 to avoid multiple, giant pushes for minor, independent changes (I hope).

PR #226, opened 2016-06-28
inherit pyspark from scipy-notebook

minrk:
removes duplicate installation of Python 2 env and Python packages in both Dockerfiles.

There are very few packages in scipy not in pyspark, so removing the duplication seems logical to me.

Does this make sense @parente?

I'm not sure how to update the inheritance diagram.


parente:
> There are very few packages in scipy not in pyspark, so removing the duplication seems logical to me.

Makes sense. They've grown closer over time.

> I'm not sure how to update the inheritance diagram.

If you open the README in github and click the diagram, it takes you to the block diagram editor site. I edit the source there to adjust the diagram, save off the SVG as a PNG into `internal/inherit-diagram.png` here, and then update the link in the README to be the permalink for the updated diagram. (The diagram editor gives you the link under the source box.)

jakirkham:
I've also felt like these two images are pretty similar. Thanks for reorganizing them @minrk. 😄

parente:
@minrk needs a rebase and the diagram update. Otherwise, should be GTG.

minrk:
Cool, I hadn't seen the diagram editor before. Should be rebased and updated now.

parente:
You have to grab the SVG from the diagram site, convert it to a PNG, and commit it as `internal/inherit-diagram.png` as well. Otherwise, the visible diagram is still the old one and only the link leads to the new one.

The conversion is a pain. I thought about linking directly to the SVG on the diagram site, but I believe it's being regenerated on every visit. I'm sure all that extra traffic would not be welcome by the author.

README.md
```diff
@@ -25,7 +25,7 @@ If this is your first time using Docker or any of the Jupyter projects, do the f
 
 Here's a diagram of the `FROM` relationships between all of the images defined in this project:
 
-[![Image inheritance diagram](internal/inherit-diagram.png)](http://interactive.blockdiag.com/?compression=deflate&src=eJyFzLEOwjAMRdGdr7C6d0dURXRjZ0QIOY2LTNM4SsxQEP9Ow5YuXe87tnHSj5bxAZ8dgEQmr6gsHloIEjUia7MslgZ8Ob0P4jXxm5Z5n3sHV4eGXFtZMoz-8KSUmE6Xc1fdmt1f1EcwmKj2omRExnxXhCwm9jyhK9C6ZZd6DnOhypKNRcWlku9p81_cFGFOAeNYuHXLDp2r1_L7A-hzeKU)
+[![Image inheritance diagram](internal/inherit-diagram.png)](http://interactive.blockdiag.com/image?compression=deflate&encoding=base64&src=eJyFzLEOwjAMRdGdr7C6d0dURXRjZ0QIOY2LTNM4SsxQEP9Ow5YuXe87tnHSj5bxAZ8dgEQmr6gsHloIEjUia7MslgZ8Ob0P4jXxm5Z5n3sHV4eGXFtZMoz-8KSUmE6Xc1fdmt1f1EcwmKj2omRExnxXhCwm9jyhK9C6ZZd6DnOhypKNRcWlku9p81_c-BXmFDCOhVq37NC5ei2_P_VBd-Y)
```
parente: Need to link to the svg not the png. (Last time I tried this, github didn't support the svg hence the png conversion) 😕

minrk:
Oops, sorry about that. Updated.

parente:
Yea, try to look at the README.md now in your branch. For me, it doesn't render the SVG like it did the PNG.

I think GitHub plays some games with caching and proxying images and it breaks SVGs.

minrk:
I didn't push the svg because I noticed that, too, but I forgot to turn the link in README back to the png after trying. Should be okay, now.

parente:
Heh. Got it. 

Looks like this went green once and the only change thereafter was the diagram, so merging.

parente:
Thanks @minrk!

parente:
Pushed as `e736784a1a8f`.

jakirkham:
The reason for the PNG is relative links to SVGs are not supported by GitHub. Oddly they do render it in the blob. See this issue ( https://github.com/isaacs/github/issues/316 ) for details. Note that is an unofficial GitHub issue tracker run by other users.

We could try using [RawGit]( https://rawgit.com/ ) if we find this conversion tedious. Though that will include a link that is explicitly tied to a branch.

PR #227, opened 2016-06-28
Only move julia kernelspec

minrk:
rather than all kernelspecs that might be in the user kernelspec dir

cf https://github.com/jupyter/docker-demo-images/pull/94


parente:
Will merge when it goes green and kick off a push to DockerHub,.

parente:
Running build and pushing.

parente:
Pushed as `dd0b2ed864e0`.

PR #236, opened 2016-07-09
Default matplotlib to inline backend

### LINKED ISSUE #235: error importing matplotlib in scipy-notebook
Hi,

I just pulled a fresh copy of jupyter/scipy-notebook from dockerhub. When I run

```
import matplotlib.pyplot as plt
```

I get the following error:
```
ImportErrorTraceback (most recent call last)
<ipython-input-5-eff513f636fd> in <module>()
----> 1 import matplotlib.pyplot as plt

/opt/conda/lib/python3.5/site-packages/matplotlib/pyplot.py in <module>()
    112 
    113 from matplotlib.backends import pylab_setup
--> 114 _backend_mod, new_figure_manager, draw_if_interactive, _show = pylab_setup()
    115 
    116 _IP_REGISTERED = None

/opt/conda/lib/python3.5/site-packages/matplotlib/backends/__init__.py in pylab_setup()
     30     # imports. 0 means only perform absolute imports.
     31     backend_mod = __import__(backend_name,
---> 32                              globals(),locals(),[backend_name],0)
     33 
     34     # Things we pull in from all backends

/opt/conda/lib/python3.5/site-packages/matplotlib/backends/backend_qt4agg.py in <module>()
     16 
     17 
---> 18 from .backend_qt5agg import FigureCanvasQTAggBase as _FigureCanvasQTAggBase
     19 
     20 from .backend_agg import FigureCanvasAgg

/opt/conda/lib/python3.5/site-packages/matplotlib/backends/backend_qt5agg.py in <module>()
     13 
     14 from .backend_agg import FigureCanvasAgg
---> 15 from .backend_qt5 import QtCore
     16 from .backend_qt5 import QtGui
     17 from .backend_qt5 import FigureManagerQT

/opt/conda/lib/python3.5/site-packages/matplotlib/backends/backend_qt5.py in <module>()
     29     figureoptions = None
     30 
---> 31 from .qt_compat import QtCore, QtGui, QtWidgets, _getSaveFileName, __version__
     32 from matplotlib.backends.qt_editor.formsubplottool import UiSubplotTool
     33 

/opt/conda/lib/python3.5/site-packages/matplotlib/backends/qt_compat.py in <module>()
    135     # have been changed in the above if block
    136     if QT_API in [QT_API_PYQT, QT_API_PYQTv2]:  # PyQt4 API
--> 137         from PyQt4 import QtCore, QtGui
    138 
    139         try:

ImportError: libicui18n.so.56: cannot open shared object file: No such file or directory
```
everything is internal to the docker container, so this shouldn't be an issue with my machine. Everything except matplotlib seems to be working. Also, I know very little about Qt and how Python interacts with it. Google doesn't turn up anything either.

Any ideas??

parente:
Fixes #235 and avoids repeat regressions

### DIFF
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -15,6 +15,8 @@ RUN apt-get update && \
 USER $NB_USER
 
 # Install Python 3 packages
+# Remove pyqt and qt pulled in for matplotlib since we're only ever going to
+# use notebook-friendly backends in these images
 RUN conda install --quiet --yes \
     'ipywidgets=5.1*' \
     'pandas=0.18*' \
@@ -32,12 +34,15 @@ RUN conda install --quiet --yes \
     'dill=0.2*' \
     'numba=0.23*' \
     'bokeh=0.11*' \
-    'h5py=2.5*' \
-    && conda clean -tipsy
+    'h5py=2.5*' && \
+    conda remove --quiet --yes --force qt pyqt && \
+    conda clean -tipsy
 # Activate ipywidgets extension in the environment that runs the notebook server
 RUN jupyter nbextension enable --py widgetsnbextension --sys-prefix
 
 # Install Python 2 packages
+# Remove pyqt and qt pulled in for matplotlib since we're only ever going to
+# use notebook-friendly backends in these images
 RUN conda create --quiet --yes -p $CONDA_DIR/envs/python2 python=2.7 \
     'ipython=4.2*' \
     'ipywidgets=5.1*' \
@@ -57,8 +62,9 @@ RUN conda create --quiet --yes -p $CONDA_DIR/envs/python2 python=2.7 \
     'numba=0.23*' \
     'bokeh=0.11*' \
     'h5py=2.5*' \
-    'pyzmq' \
-    && conda clean -tipsy
+    'pyzmq' && \
+    conda remove -n python2 --quiet --yes --force qt pyqt && \
+    conda clean -tipsy
 # Add shortcuts to distinguish pip for python2 and python3 envs
 RUN ln -s $CONDA_DIR/envs/python2/bin/pip $CONDA_DIR/bin/pip2 && \
     ln -s $CONDA_DIR/bin/pip $CONDA_DIR/bin/pip3
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -63,6 +63,10 @@ RUN conda create --quiet --yes -p $CONDA_DIR/envs/python2 python=2.7 \
 RUN ln -s $CONDA_DIR/envs/python2/bin/pip $CONDA_DIR/bin/pip2 && \
     ln -s $CONDA_DIR/bin/pip $CONDA_DIR/bin/pip3
 
+# Configure ipython kernel to use matplotlib inline backend by default
+RUN mkdir -p $HOME/.ipython/profile_default && \
+    echo "c.InteractiveShellApp.matplotlib = 'inline'" > $HOME/.ipython/profile_default/ipython_kernel_config.py
+
 USER root
 
 # Install Python 2 kernel spec globally to avoid permission problems when NB_UID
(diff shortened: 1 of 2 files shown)


minrk:
This should happen on import, not trigger import of matplotlib on startup of every kernel, which slows kernel startup significantly. There's an [open issue](https://github.com/ipython/ipykernel/pull/82) in IPython to do this by default. I use [this import hook](https://github.com/minrk/profile_default/blob/master/startup/mplimporthook.py) in my startup files to do it in the meantime.

parente:
@minrk Do you think your import hook is good enough to include here for now ("This is horrible, and probably nobody should use it." comment aside)? I think anything,  import hook or slowed kernel startup, beats the heck out of `*.so` import errors that keep coming up every month or so.

minrk:
Sure, let's give the import hook a try. I'm not sure it does the right thing in every case, but it does work for my typical imports of matplotlib.

parente:
Import hook works in Python 3. In Python 2, this code:

```
        try:
            # remove myself from the import hooks
            sys.meta_path.remove(self)
        except ValueError:
            pass
```

causes an IndexError when doing `import matplotlib.pyplot as plt`. The sys.meta_path is being modified during iteration over `sys.meta_path` I expect.

Any thoughts on how else we might unregister the finder in Python 2? I'm not coming up with much.

minrk:
The `self._called` makes it a no-op, so it's probably enough. The other option is to assign sys.meta_path to a new copy of the list, rather than removing it in-place:

```python
sys.meta_path = [ loader for loader in sys.meta_path if loader is not self ]
```

which should work on py2 and 3

parente:
Should be GTG. Green on Travis and tested the matplotlib behavior manually with a local build.

jakirkham:
Interesting idea with the import hook. I'll have to keep this in mind.

Something also to consider is there is a `notebook` or `nbagg` backend as well. It is a little newer, but is pretty nice as it allows a bit more interactivity with the plots. Basically the stuff that you would get with a backend like `pyqt4` except that it is all in Javascript and embedded in the notebook. It's also easy to convert an interactive plot to an inline one after the fact. Just something else to consider.

jakirkham:
Should add that we could go a step further and remove `qt` and `pyqt` from the image completely. While this isn't necessary and normally I wouldn't suggest it, but they are bulky packages and removal would help cutdown on image size.

To do so, we would need to do `conda remove --force qt pyqt`. The `--force` is required as they are dependencies of existing packages (i.e. `matplotlib`). So it tells `conda` we don't care and would rather remove them anyways.

parente:
@jakirkham I'll give pyqt removal a shot right now.

parente:
> they are bulky packages and removal would help cutdown on image size.

Removing qt and pyqt cuts about 120 MB from the image size. Both the inline and notebook backends for matplotlib continue to work as expected. When I try to switch to the qt backend, I get a reasonable exception informing me that I don't have qt.

jakirkham:
Awesome! Thanks for trying @parente. This seems totally reasonable to me.

scipy-notebook/Dockerfile
```diff
@@ -57,8 +62,9 @@ RUN conda create --quiet --yes -p $CONDA_DIR/envs/python2 python=2.7 \
     'numba=0.23*' \
     'bokeh=0.11*' \
     'h5py=2.5*' \
-    'pyzmq' \
-    && conda clean -tipsy
+    'pyzmq' && \
+    conda remove --force qt pyqt && \
```
jakirkham: I think this need the prefix too ( `-p $CONDA_DIR/envs/python2` ). Also, I forgot to add `--quiet --yes`.

scipy-notebook/Dockerfile
```diff
@@ -32,12 +34,15 @@ RUN conda install --quiet --yes \
     'dill=0.2*' \
     'numba=0.23*' \
     'bokeh=0.11*' \
-    'h5py=2.5*' \
-    && conda clean -tipsy
+    'h5py=2.5*' && \
+    conda remove --force qt pyqt && \
```
jakirkham: Forgot to add `--quiet --yes`.

parente: Nah. It's my bad. I copy/pasted the command into the Python 2 environment setup, but forgot that I'll need to target that environment.

minrk:
Thanks!

PR #238, opened 2016-07-13
Update R version to 3.3* in r-notebook and datascience-notebook

poplav:


parente:
Thanks @poplav . I'm curious to see how many of the packages work with R 3.3. :)

poplav:
Just realized there is the [datascience-notebook](https://github.com/jupyter/docker-stacks/blob/master/datascience-notebook/Dockerfile) as well [hand over face].  I'll update there as well.

parente:
@poplav No worries. I clearly don't have all the places nailed down in my mind either!

poplav:
I am seeing https://github.com/jupyter/docker-stacks/issues/210.  I need to verify it is not a result of these changes, so hold off on merging please.

poplav:
It looks like `2485724a08c9` reintroduced #210 making libjpeg v9.  Noting this is then not attributed to the recent R upgrades.

parente:
Thanks for tracking that down. I'm going to do one last check on the all-spark-notebook vs datascience-notebook for the tag you mention. I'm wondering if the R problem only exists in one and not the other because of the libs that are touched by other conda actions.

poplav:
I did see that the fix mentioned in #210:
```
echo "jpeg 8*" >> /opt/conda/conda-meta/pinned
conda update --all -y
```
does work after the image is built.  In my case I could run this in the dev_image step of declarative widgets, but that would not be optimal/time consuming

parente:
Nah. Don't fix it in your build. I'm looking into how to fix it in the main images.

parente:
Pushed `97a5071c5775`.

Issue #248, opened 2016-07-27
missing SQLAlchemy for python2 in scipy-notebook 


midnightradio:
While using datascience-notebook, I met a problem with importing a package sqlalchemy. And I found that there was no sqlalchemy installed for python2 which I was using with the notebook. I address the problem caused by scipy-notebook image installing conda packages for both python2 and python3 is based on base-notebook which only installs python3 and other pakcages including sqlalchemy. 

Though I think base-notebook should contain python2 with sqlalchemy instead of supporting only python3, to make a workaround solution much easier, I added the package to the parameters for conda install command for scipy-notebook. I built, tested and it works.

parente:
SQLAlchemy is not installed directly in base-notebook. It's a dependency of jupyterhub which is installed there.

That said, I don't have a problem with adding it explicitly to both the Python2 and Python3 environments the scipy-notebook image. Would you mind updating your PR to include a statement to install it for both Python2 and 3 in that image?

jakirkham:
Should we just added in base though if that is where it is getting added to Python 3?

parente:
Base doesn't even have the Python 2 environment iirc. It's only in base by virtue of being an implicit dependency of jhub.

midnightradio:
I was curious about which package in base-notebook leads sqlachemy to be installed implicitly and hesitate to put explicit installation in scipy-notebook skipping base and minimal. Now it seems reasonable to do so, right?

One other thing I had to think about after PR was it is still useless without any drivers for a database but we cannot provide all drivers before one got chosen by a user. However, putting necessaries of MySQL drivers along with sqlal hemy would help many users with supporting the most popular uses but also with providing a starting point letting certain users to know where should they modify to meet their own needs.

With those consideration, PR updated.

parente:
Fixed by #249. Thanks again! 🎈

PR #249, opened 2016-07-27
add sqlalchemy to conda install command

midnightradio:
workaround for the issue #248

### DIFF
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -9,7 +9,6 @@ USER root
 # libav-tools for matplotlib anim
 RUN apt-get update && \
     apt-get install -y --no-install-recommends libav-tools && \
-    apt-get install -y --no-install-recommends libmysqlclient-dev && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


jtyberg:
@midnightradio Can you add sqlalchemy to the python3 conda install as well?  We should explicitly install it in both environments, rather than rely on it being pulled in.

midnightradio:
Not only that, I added some mysql drivers which would not make the package useless at least for MySQL users.

parente:
Hmm. I was fine with adding the sqlalchemy package, but adding the mysql driver / client goes too far I think. If we add mysql, now where do we draw the line when other folks ask for postgres, maria, Oracle, ... The image will be bloated with maybe 1-2 drivers a few people use, and many that most people never touch.

How about we simply add the SQLAlchemy lib to the Python 2 and 3 environments explicitly, and then add to the [recipes wiki page](https://github.com/jupyter/docker-stacks/wiki/Docker-recipes) about how to create new Docker images deriving from the stacks to install specific database engines?

midnightradio:
Maybe I was personally hoping that I could use the image anywhere with no changes and went too far. I agree with your words. In build testing before updating my PR.

parente:
Not a problem and thanks for understanding. It's *really* hard to put a strict definition on what should and shouldn't go in the images here.

jtyberg:
New Docker images are built with tag `1d374670daaa`.

Issue #150, opened 2016-03-08
Travis: Fix COPY command to not bust cache


jakirkham:
This [`COPY`]( https://github.com/jupyter/docker-stacks/blob/630ceec2576a309ace2ebee1d766601a064bb0f3/minimal-notebook/Dockerfile#L93 ) command busts the cache every time on Travis CI. It would be nice to figure out why this is occurring on Travis CI and fix it. This is causing tests to run ~45min in length. Even simple changes that don't invalidate the builds at all like documentation changes suffer from this long build time.

parente:
Here's a brain teaser. I went to do a new build. I noticed the build VM disk was nearly full. I did the following:

```
# get the latest code
git fetch origin
git reset --hard origin/master

# nuke all the existing images
service docker stop
rm -rf /var/lib/docker
service docker start

# build!
make release-all
```

Unlike on travis, the only step that broke cache was the actual change in the git diff from the last build. That's pretty surprising considering I was starting from a brand new docker store in `/var/lib/docker` and a brand new docker daemon process.

jakirkham:
So, you mean the VM worked the way Travis should work. ;)

Yeah, I tried computing the SHA256 hash for the file being copied in this statement on Travis, the VM, and my Mac. The result was the same in all cases. Still confused.

parente:
> So, you mean the VM worked the way Travis should work. ;)

Yes. Yes it did. :hurtrealbad:

jakirkham:
Just as another interesting tidbit. The VM and Travis are both using the exact same version of docker.

jakirkham:
Honestly, I think we should raise this issue with Travis. It looks like we have enough evidence at this point to suggest something weird is happening on Travis. Maybe they can let us inspect one of their VMs to figure out what is going wrong. As your last release went out, I am rebuilding master so we can use this as an example for them to look at.

jakirkham:
Here is the corresponding issue on Travis' issue tracker. ( https://github.com/travis-ci/travis-ci/issues/5764 )

parente:
I'm going to close this issue out because we were unable to determine what causes the docker cache invalidation on a VM graciously given to us by travis. If we start a new, strong push to deploy from a CI system, we can try to debug it again.

Issue #131, opened 2016-03-03
Add retry loop for docker push


jakirkham:
When we are getting 500 errors from trying to push to DockerHub, we want to be able to retry automatically some number of times. Possibly this number should be fixed based on the number of repeats needed in practice.

Related: https://github.com/docker/hub-feedback/issues/533
Related: https://github.com/jupyter/docker-stacks/issues/7

parente:
This should be pretty easy to add to the `make release-all` target.

jakirkham:
Do you know how many times you have to repeat the push normally?

parente:
Max I saw was 5 I think.

parente:
Lately, I've seen more failures fetching packages from debian and anaconda repos than pushing to DHub. If we're going to add a loop, it should go around the whole `make release-all`.

PR #250, opened 2016-07-29
Build improvements

parente:
* Don't pull images onto travis. Caching doesn't work. See #150 
* Add retry option #131


jakirkham:
Sad, but true. 😢

parente:
I think this is good to go. No build needed.

parente:
Tnx!

PR #253, opened 2016-08-03
Update conda to 4.1.11 in base-notebook

### LINKED ISSUE #252: Conda version is old
Its seems that the base notebook conda version is old
https://github.com/jupyter/docker-stacks/blob/master/base-notebook/Dockerfile#L60

I'm running 4.1.9 on my machine. This is leading to problems with `Warning: somethings is wrong with binstar_client (No module named binstar_client.utils)` when resolving some dependancies on subsequence conda installs. This should be resolved by a more modern conda https://github.com/conda/conda/issues/3141

Should we update the base notebook to use a more modern conda version?

parente:
Fixes #252

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -56,11 +56,11 @@ RUN mkdir /home/$NB_USER/work && \
 # Install conda as jovyan
 RUN cd /tmp && \
     mkdir -p $CONDA_DIR && \
-    wget --quiet https://repo.continuum.io/miniconda/Miniconda3-3.19.0-Linux-x86_64.sh && \
-    echo "9ea57c0fdf481acf89d816184f969b04bc44dea27b258c4e86b1e3a25ff26aa0 *Miniconda3-3.19.0-Linux-x86_64.sh" | sha256sum -c - && \
-    /bin/bash Miniconda3-3.19.0-Linux-x86_64.sh -f -b -p $CONDA_DIR && \
-    rm Miniconda3-3.19.0-Linux-x86_64.sh && \
-    $CONDA_DIR/bin/conda install --quiet --yes conda==3.19.1 && \
+    wget --quiet https://repo.continuum.io/miniconda/Miniconda3-4.1.11-Linux-x86_64.sh && \
+    echo "efd6a9362fc6b4085f599a881d20e57de628da8c1a898c08ec82874f3bad41bf *Miniconda3-4.1.11-Linux-x86_64.sh" | sha256sum -c - && \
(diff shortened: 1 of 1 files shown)


jakirkham:
Thanks @parente.

jakirkham:
Building a release from this merged commit.

base-notebook/Dockerfile
```diff
@@ -56,11 +56,11 @@ RUN mkdir /home/$NB_USER/work && \
 # Install conda as jovyan
 RUN cd /tmp && \
     mkdir -p $CONDA_DIR && \
-    wget --quiet https://repo.continuum.io/miniconda/Miniconda3-3.19.0-Linux-x86_64.sh && \
-    echo "9ea57c0fdf481acf89d816184f969b04bc44dea27b258c4e86b1e3a25ff26aa0 *Miniconda3-3.19.0-Linux-x86_64.sh" | sha256sum -c - && \
-    /bin/bash Miniconda3-3.19.0-Linux-x86_64.sh -f -b -p $CONDA_DIR && \
-    rm Miniconda3-3.19.0-Linux-x86_64.sh && \
-    $CONDA_DIR/bin/conda install --quiet --yes conda==3.19.1 && \
+    wget --quiet https://repo.continuum.io/miniconda/Miniconda3-4.1.11-Linux-x86_64.sh && \
+    echo "efd6a9362fc6b4085f599a881d20e57de628da8c1a898c08ec82874f3bad41bf *Miniconda3-4.1.11-Linux-x86_64.sh" | sha256sum -c - && \
```
damianavila: Not sure if you want that, but you could use `latest` instead of a version number... maybe you want to pin that specific version, but if it is not the case, using `latest` helps...

parente: Good to know. We do try to pin all first order dependencies in the stacks so that upgrades are explicit in PRs and rebuilds.

jakirkham: Though I think I have seen `conda` getting upgraded later anyways. 😒 Would need to double check where I saw this happening though.

parente: Yes. I think I saw the same too, though I thought I remember hearing newer versions of conda don't self-upgrade implicitly anymore?

jakirkham: Hmm...interesting. Maybe so. Haven't heard that before though, but maybe I just missed it.

damianavila: You have a config option to avoid auto-upgrade (`auto_update_conda`) or you can use this thing: http://www.damian.oquanta.info/posts/how-to-pin-conda.html

jakirkham: Adding in PR ( https://github.com/jupyter/docker-stacks/pull/257 ).

PR #255, opened 2016-08-09
Fixed errors in READMEs

abtrout:
This PR fixes a few broken hyperlinks and inconsistencies in README files. In particular:

* each README file contained hyperlinks for start scripts, which pointed to the `minimal-notebook` instead of `base-notebook`
* the [base-notebook README](https://github.com/jupyter/docker-stacks/blob/master/base-notebook/README.md#jupyterhub) includes an example which references `minimal-notebook` rather than `base-notebook`

It looks like these were left over after some recent reorganization.


parente:
Thanks! :cake:

PR #257, opened 2016-08-09
Disable `conda` auto-update

jakirkham:
This configures `conda` so that it should not auto-update. Instead users will need to explicitly update it.


jakirkham:
Based on local testing this seems to work ok. Tried downgrading `conda` to `4.0.11` (a version that still supports this config) and then updating `setuptools` to `25.1.6` in the `root` environment. In the process, `conda` did not try to update itself and stayed `4.0.11`.

parente:
:shipit:

PR #258, opened 2016-08-09
Add a generic start.sh script to images

parente:
Refactors `start-notebook.sh` to make it easy to specify alternative container commands while retaining the special features of the docker-stacks (e.g., `GRANT_SUDO`). For example, here's how to get Jupyter Lab running in a docker-stacks image:

```dockerfile
FROM jupyter/scipy-notebook

RUN pip install jupyterlab && \
    jupyter serverextension enable --py jupyterlab
CMD ["start.sh", "jupyter", "lab"]
```

The PR contains updates to all the READMEs with additional examples (e.g., IPython console, kernel gateway).

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -78,13 +78,15 @@ RUN pip --no-cache-dir install 'jupyterhub==0.5'
 
 USER root
 
-# Configure container startup as root
 EXPOSE 8888
 WORKDIR /home/$NB_USER/work
+
+# Configure container startup
 ENTRYPOINT ["tini", "--"]
-CMD ["start-notebook.sh"]
+CMD ["start-jupyter.sh", "notebook"]
 
 # Add local files as late as possible to avoid cache busting
+COPY start.sh /usr/local/bin/
 COPY start-notebook.sh /usr/local/bin/
 COPY start-singleuser.sh /usr/local/bin/
 COPY jupyter_notebook_config.py /home/$NB_USER/.jupyter/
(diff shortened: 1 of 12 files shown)


parente:
@jakirkham, @jtyberg The PR is big due to the fact that we still don't generic READMEs from a single source (nearing the top of my must-fix list). I went over it a few times. If you get a chance to skim it, that would be a good sanity check.

base-notebook/Dockerfile
```diff
@@ -78,13 +78,15 @@ RUN pip --no-cache-dir install 'jupyterhub==0.5'
...
 USER root
 
-# Configure container startup as root
 EXPOSE 8888
 WORKDIR /home/$NB_USER/work
+
+# Configure container startup
 ENTRYPOINT ["tini", "--"]
-CMD ["start-notebook.sh"]
+CMD ["start-jupyter.sh", "notebook"]
```
jtyberg: should this be `CMD ["start.sh", "notebook"]` ?

parente: Doh. Should be no change at all since start-notebook still exists for backward compat. I'm surprised the Travis build / test worked with that.

parente: Test runner was not using exit code 1 when every test retry failed. The test was indeed failing with the bad start command, but the travis job was not. Fixed both the start command and the make test target.

jtyberg:
LGTM

parente:
Pushed to docker hub with tag `69a5c1ab7ff7`. Also added the Jupyter Lab example to the recipes wiki page.

PR #259, opened 2016-08-10
Upgrade hdf5 and h5py conda packages

tylere:
This upgrade allows for the HDF5 package license to be displayed from within the container.

cat /opt/conda/pkgs/hdf5-1.8.17-2/info/LICENSE.txt


parente:
🍰

parente:
Pushed to docker hub with tag `7001472997fa`

jakirkham:
This is also true of `h5py`, but we had to correct the missing license in that package first. It was deployed yesterday. So I believe it should be included in any recent build of docker-stacks.

Issue #256, opened 2016-08-09
tini license is not available from within the Docker image


tylere:
The **tini** library has an MIT license, which includes the requirement:

> The above copyright notice and this permission notice shall be included in
> all copies or substantial portions of the Software.

However, the **base_notebook** container image only copies the **tini** binary file (via wget), and does not include a copy of license in the container image.

parente:
Good catch. Do you know if the many official docker images that include tini pull the license text in or take some alternative route? I'd like to mimic whatever the official images do here.

parente:
Or https://github.com/krallin/tini/issues/46 ...

tylere:
I don't know of any docker images that use tini and pull the license text. Jenkins appears to [do something similar to docker-stacks](https://github.com/jenkinsci/docker/blob/master/Dockerfile).

parente:
We're all copy/pasta-ing [the recipe from the tini readme](https://github.com/krallin/tini#using-tini) (unfortunately). 

I think @jakirkham has the right idea in krallin/tini#46, or at least we'll settle on a best practice for including it that everyone can adopt.

jakirkham:
Sorry, meant to xref that back here. Including the license in the binary seems like the simple way to ensure everyone is compliant. Though if we want to put it in somewhere like `/usr/share/tini` that could work. For some other projects I have used the rpm, which doesn't appear to include the license either.

tylere:
Here are a few options the could be used to include it, without modifying the binary file.

1. Along the lines of the [Using tini](https://github.com/krallin/tini#using-tini) section of the tini docs, you [could](url) append an additional ADD command to the Dockerfile that copies the license file. The license file URL would need to be updated every time there is a new release. I.e.
ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
ADD https://github.com/krallin/tini/blob/2c2bb517b0a5e283ea6dd3e561f00ae006780949/LICENSE /LICENSE

2. Following the example shown in [tini-images](https://github.com/krallin/tini-images) container images, you could [use git clone](https://github.com/krallin/tini-images/blob/master/src/build-app) to pull the source code.

jakirkham:
Just for creative flair, we could create a conda package for `tini` and `conda` has a standard way of including licenses. Honestly, it might be fun to create `tini` package for `conda` just to see that it can be done.

parente:
[...]

PR #260, opened 2016-08-13
Update `tini` to `v0.10.0`

jakirkham:
Closes https://github.com/jupyter/docker-stacks/issues/256

Includes a copy of `tini` that provides its license.

cc @tylere


parente:
On DockerHub as `8ee3ea7a590a`.

Issue #203, opened 2016-05-06
Generate READMEs from common source


parente:
There's a lot of duplicate text across the READMEs for reader convenience. @jakirkham and I briefly discussed improving the authoring experience a bit by generating these READMEs from a common source using a simple templating approach. Even this has a cost, however, as new contributors looking to update READMEs will have to send PRs against the source, not the READMEs themselves.

parente:
We also briefly discussed the pros/cons of doing a full RTD site like all the other Jupyter projects, but both agreed we didn't want to lose the value of the simple README per stack that is immediately visible and accessible on GitHub and linked from Docker Hub.

willingc:
I agree that the simple README per stack is most important. I'm going to do some research when I'm at Write The Docs conference in a couple of weeks to see if others have ideas. We may be able to use jinja to help generate the individual READMEs. For now, I would put a note in a CONTRIBUTORS.md reminding folks to search and replace across all stacks when making a change to an individual README. Worst case: we should be aware of it during review ;-)

parente:
> We may be able to use jinja 

This was my first thought too, though it's a bit heavy. I wonder if there's a markdown way of doing simple includes?

minrk:
There's not an *official* markdown way to do includes (official markdown is a not-at-all-extensible shorthand for HTML), but you might find a markdown implementation that supports it. However, it may be simplest to do a simple template version, such that we can do a `make readmes` which fills in the common blocks.

Given how this repo is used, though, the generated results should be committed to the repo.

parente:
See comment in #261.

PR #261, opened 2016-08-15
[WIP] Generate READMEs from jinja templates

parente:
The docs/generate.py script renders the jinja templated markdown files 
in docs/ into the README.md files in all of the other folders.

Things left to do:
- [ ] `make docs`
- [ ] CONTRIBUTING.md describing what folks need to do to add to or update the READMEs
- [ ] prefix generated READMEs with a note that warns they are **generated**
- [ ] run doc gen on travis to make sure it works

For issue #203.

### DIFF
--- a/all-spark-notebook/README.md
+++ b/all-spark-notebook/README.md
@@ -18,12 +18,11 @@
 * A [start.sh](../base-notebook/start.sh) script useful for running alternative commands in the container (e.g. `ipython`, `jupyter kernelgateway`, `jupyter lab`)
 * Options for HTTPS, password auth, and passwordless `sudo`
 
-
 ## Basic Use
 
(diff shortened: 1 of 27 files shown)


parente:
This change does make inconsistencies across READMEs less likely and requires edits in fewer places. It it also raises the bar for folks wanting to make PRs as they can no longer contribute directly to the READMEs but rather need to find the proper include file to edit.

I'm on the fence.

jakirkham:
Yep, I hear ya. Feeling similarly. Though, for me personally, jinja has become an accepted fact of life for many problems that I work on. So am ok with that way of solving this problem.

parente:
We haven't made major changes to the docs in a while that would warrant needing to generate them from a common source. I think we can close this as YAGNI.

PR #274, opened 2016-09-06
Adds crayon to the r notebook image

poplav:
Resolves https://github.com/IRkernel/IRkernel/issues/399 by adding a dependency that was introduced.


parente:
Included in `749017d878d6`

PR #279, opened 2016-09-10
allow grant sudo to use pip

fdavis:
If I run a notebook with sudo privileges I cannot use pip without specifying the full path.

`!sudo pip install package` returns `sudo: pip: command not found`


parente:
Do you need to run pip using sudo? The jovyan user in the container has read-write access to `/opt/conda`.

fdavis:
Oh wow, you definitely don't. Well played!

Thanks

Issue #269, opened 2016-08-26
Regression: changing NB_UID


parente:
The start.sh script changes the owner of `.` which is the `/home/jovyan/work directory` but misses changing `/home/jovyan/` so notebook (and the user) have no access to write files anywhere in the home dir.

I think we purposely took out the chown at one point to avoid changing ownership over files when the home directory is host mounted. But clearly that breaks when the home dir is NOT mounted as seen below.

```
pparente@VM4-002 ~ ❯❯❯ docker run -it --rm -p 8888:8888 -e NB_UID=$(id -u) --user root -v /home/pparente/test:/home/jovyan/work jupyter/scipy-notebook
Traceback (most recent call last):
  File "/opt/conda/lib/python3.5/site-packages/traitlets/traitlets.py", line 501, in get
    value = obj._trait_values[self.name]
KeyError: 'runtime_dir'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/bin/jupyter-notebook", line 6, in <module>
    sys.exit(notebook.notebookapp.main())
  File "/opt/conda/lib/python3.5/site-packages/jupyter_core/application.py", line 267, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/opt/conda/lib/python3.5/site-packages/traitlets/config/application.py", line 595, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-7>", line 2, in initialize
  File "/opt/conda/lib/python3.5/site-packages/traitlets/config/application.py", line 74, in catch_config_error
    return method(app, *args, **kwargs)
  File "/opt/conda/lib/python3.5/site-packages/notebook/notebookapp.py", line 1069, in initialize
    self.init_configurables()
  File "/opt/conda/lib/python3.5/site-packages/notebook/notebookapp.py", line 842, in init_configurables
    connection_dir=self.runtime_dir,
  File "/opt/conda/lib/python3.5/site-packages/traitlets/traitlets.py", line 529, in __get__
    return self.get(obj, cls)
  File "/opt/conda/lib/python3.5/site-packages/traitlets/traitlets.py", line 508, in get
    value = self._validate(obj, dynamic_default())
  File "/opt/conda/lib/python3.5/site-packages/jupyter_core/application.py", line 99, in _runtime_dir_default
    ensure_dir_exists(rd, mode=0o700)
[...]

PR #281, opened 2016-09-12
Workaround for the linux filesystem issue

amatsus:
This is a workaround for #269. OverlayFS - the linux union filesystem had a bug that look up lower-layer perms before upper-layer perms and it was fixed on https://github.com/torvalds/linux/commit/38b78a5f18584db6fa7441e0f4531b283b0e6725 (kernel 4.6 and later). Same bug of AUFS was also fixed on https://github.com/sfjro/aufs4-linux/commit/625634d1dc3e6a0d03fbd918c074912797f5903b. But these patch isn't attached yet on most linux distros, e.g. Ubuntu Server 16.04, RHEL/CentOS 7.2 and Docker for Mac.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -4,12 +4,14 @@
 
 set -e
 
+mkdir -p -m 700 /home/$NB_USER/.local/share/jupyter
+
 # Handle special flags if we're root
 if [ $UID == 0 ] ; then
     # Change UID of NB_USER to NB_UID if it does not match
     if [ "$NB_UID" != $(id -u $NB_USER) ] ; then
         usermod -u $NB_UID $NB_USER
-        chown -R $NB_UID $CONDA_DIR .
+        chown -R $NB_USER. $CONDA_DIR /home/$NB_USER/.local .
     fi
 
     # Enable sudo if requested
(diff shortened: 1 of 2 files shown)


fdavis:
Cool, thanks!

parente:
I'd like to give this a quick test in a couple host-mounted volume scenarios before merging.

base-notebook/start.sh
```diff
@@ -4,12 +4,14 @@
 
 set -e
 
+mkdir -p -m 700 /home/$NB_USER/.local/share/jupyter
```
parente: We used to create this directory ahead of time to avoid issues with permissions at runtime. I can't remember the specifics of those problems now, though I vaguely recall they are workarounds. I'm going to try removing the mkdir from the Dockerfile, avoid changing anything in start.sh, and see what happens in the various images.

parente:
There's another spot down in the datascience-notebook where ~/.local/share/jupyter gets created as part of the Julia kernel installation. The directory has to be removed as part of the command else this workaround fails for containers spawned from that image.

parente:
Sorry for letting this linger. I'll get back on fixing the datascience-notebook issue soon. 

@amatsus can you grant maintainers permission to push to your PR branch (a new github checkbox on the right side of the PR screen for you). That way I can add a commit to fix the other image right here.

parente:
Continued in #294.

amatsus:
Sorry for my late reply. You mean this image? checkboxofprscreen It has been that way all along.

parente:
@amatsus Dang. Maybe I missed it but I didn't see any indicator of it from my side.

PR #289, opened 2016-09-26
Build matplotlib font cache during docker build.

danielballan:
Matplotlib builds a font cache the first time it's imported. By importing it during the docker build, we can avoid this:

screen shot 2016-09-26 at 9 06 14 am

### DIFF
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -76,7 +76,7 @@ RUN ln -s $CONDA_DIR/envs/python2/bin/pip $CONDA_DIR/bin/pip2 && \
 
 # Import matplotlib the first time to build the font cache.
 ENV XDG_CACHE_HOME /home/$NB_USER/.cache/
-RUN $CONDA_DIR/envs/$ENV_NAME/bin/python -c "import matplotlib.pyplot"
+RUN $CONDA_DIR/envs/python2/bin/python -c "import matplotlib.pyplot"
 
 # Configure ipython kernel to use matplotlib inline backend by default
 RUN mkdir -p $HOME/.ipython/profile_default/startup
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -76,7 +76,7 @@ RUN ln -s $CONDA_DIR/envs/python2/bin/pip $CONDA_DIR/bin/pip2 && \
 
 # Import matplotlib the first time to build the font cache.
 ENV XDG_CACHE_HOME /home/$NB_USER/.cache/
-RUN $CONDA_DIR/envs/python2/bin/python -c "import matplotlib.pyplot"
+RUN MPLBACKEND=Agg $CONDA_DIR/envs/python2/bin/python -c "import matplotlib.pyplot"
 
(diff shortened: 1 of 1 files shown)


parente:
Thanks, Dan. Do we need to do this against the Python 2 environment in the image too, or is the cache shared?

minrk:
Thanks, Dan! @parente the cache is shared, so this should only needs to run once.

scipy-notebook/Dockerfile
```diff
@@ -74,6 +74,10 @@ RUN conda create --quiet --yes -p $CONDA_DIR/envs/python2 python=2.7 \
 RUN ln -s $CONDA_DIR/envs/python2/bin/pip $CONDA_DIR/bin/pip2 && \
     ln -s $CONDA_DIR/bin/pip $CONDA_DIR/bin/pip3
 
+# Import matplotlib the first time to build the font cache.
+ENV XDG_CACHE_HOME /home/$NB_USER/.cache/
+RUN $CONDA_DIR/envs/python2/bin/python -c "import matplotlib.pyplot"
```
parente: Travis is failing because it's trying to use the qt backend by default. If you move these lines below line 83 where the import hook makes matplotlib inline the default and use ipython to run the import instead of python, it might work.

minrk: I think setting MPL_BACKEND=Agg for this command should do it. I don't think the inline backend will work below, because IPython startup doesn't get loaded for `python -c`.

parente: Didn't know you could control the backend via env var. Cool.

`python -c` definitely won't work, and that's why I was thinking running `ipython -c` might do the trick instead.

danielballan: This qt4 issue is a transient problem created by anaconda. They released qt5 (the new default package for 'qt') but retained their matplotlib configuration that expects qt4 specifically. Seems like this was fixed upstream yesterday, so simply re-kicking the build might be enough. But `MPL_BACKEND=Agg` is safer anyway.

danielballan:
What is harfbuzz? Both of the last Travis builds have died on this. I assumed it was an anaconda.org hiccup the first time, but it's repeatable.

```
Error: HTTPError: 404 Client Error: NOT FOUND for url: https://conda.anaconda.org/r/linux-64/harfbuzz-0.9.35-6.tar.bz2: https://conda.anaconda.org/r/linux-64/harfbuzz-0.9.35-6.tar.bz2
```

parente:
Looks like a dependency of one of the packages we're installing. The package, version, and build number show on the anaconda site:

https://anaconda.org/r/harfbuzz/files

I'm not sure why the download link is busted, even when I try it through the UI.

parente:
🍒  harfbuzz is back! 🍒 

Restarting the build.

danielballan:
Hahaha A+ usage of cherries for celebration.

parente:
We're failing on the matplotlib import again. Can you try the `ipython -c` approach instead?

danielballan:
`MPL_BACKEND` -> `MPLBACKEND` did the tricky. Ready to merge.

parente:
Included in build tag `28515ed64e5e`

PR #290, opened 2016-09-28
use cdn for apt

minrk:
httpredir fails semi-regularly when installing lots of packages,
and will be deprecated in stretch.

Based on info from http://deb.debian.org


parente:
Included in build tag `28515ed64e5e`

PR #296, opened 2016-10-20
Don't mix mkl and open_blas packages

### LINKED ISSUE #295: Error from numexpr and mkl when doing math with pandas Series objects
I am using scipy-notebook

I get an error when running the following code:
```python
import pandas as pd
N = 10001 # no problems if N=10000
x = pd.Series([1.0] * N)
x ** 2  # error
```
Not all math operations cause a crash. I tried `x / 2` which crashes, while `x * 2` and `x + 2` do not crash.

When it crashes, I get the following error message:
```
pythonpython: : symbol lookup errorsymbol lookup error: : /opt/conda/lib/python3.5/site-packages/numexpr/../../../libmkl_vml_mc3.so/opt/conda/lib/python3.5/site-packages/numexpr/../../../libmkl_vml_mc3.so: : undefined symbol: mkl_serv_getenvundefined symbol: mkl_serv_getenv
```

I'm pretty sure this is related to https://github.com/conda/conda/issues/2031
As suggested in that issue, I fixed it by doing:
```
conda install mkl-rt
```

What is the right way to fix this so I don't get these errors out of the box?

parente:
* Bump numexpr to 2.6 so it comes from conda-forge
* Install nomkl

Fixes #295

### DIFF
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -18,9 +18,10 @@ USER $NB_USER
 # Remove pyqt and qt pulled in for matplotlib since we're only ever going to
 # use notebook-friendly backends in these images
 RUN conda install --quiet --yes \
+    'nomkl' \
     'ipywidgets=5.2*' \
     'pandas=0.19*' \
-    'numexpr=2.5*' \
+    'numexpr=2.6*' \
     'matplotlib=1.5*' \
     'scipy=0.17*' \
     'seaborn=0.7*' \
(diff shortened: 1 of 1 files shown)


parente:
In tagged build `6939ea1b1df3`

jakirkham:
Sorry to be late to the party. A better conda-forge style fix would be to do something like this `conda install -yq blas=*=openblas`. You could also pin the version too (by replacing `*`), but this is a metapackage. So I don't know that it has much value.

jakirkham:
Added PR ( https://github.com/jupyter/docker-stacks/pull/602 ) to use the `blas` package. Would be good to know if there is something else I'm missing by suggesting this switch.

PR #300, opened 2016-11-14
Add badges with image size and layers

thewtex:
This uses the MicroBadger badge service. It allows prospective image users to
quickly get an idea of the size of an image from glancing at the README.


parente:
Thanks for the badges and for showing me how to do alt text in markdown. Did not know that!

thewtex:
Thanks for the review and merge!

PR #302, opened 2016-11-17
Update Julia version to 0.5

alanhdu:
Julia 0.3 is pretty old and most Julia packages no longer support it. These commits move us to 0.5, which was released in September.

Unfortunately, none of the Debian repositories seem to support 0.5, so I have a hack to use Ubuntu's stable Julia PPA instead. I've done some moderate sanity checks (playing around with Gadfly, reading and writing HDF5 files), but it's pretty kludgey.

One option would be to stick with Julia 0.4.7, which is in the Sid repositories for Debian. I think that the changes in Julia 0.5 are major enough that we should go with 0.5... OTOH I don't actually use Julia that much so `¯\_(ツ)_/¯`.


parente:
Thanks for the PR. I don't use Julia either, but others have asked for the upgrade in the past and no one has really had the time or expertise to do it.

Your changes look clean to me. Let's get them in. Anyone that wants to remain on 0.3 can do so by sticking with the last build tag before this change.

parente:
Pushed to Docker Hub with tag `c411f52fcc93`.

minrk:
Thanks @parente!

PR #307, opened 2016-11-29
Add vincent, beautifulsoup, and xlrd to scipy-notebook

slarson:
Also adds support to install from pip.

### DIFF
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -8,7 +8,7 @@ USER root
 
 # libav-tools for matplotlib anim
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends libav-tools && \
+    apt-get install -y --no-install-recommends libav-tools python-pip python-dev build-essential && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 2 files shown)


scipy-notebook/Dockerfile
```diff
@@ -8,7 +8,7 @@ USER root
 
 # libav-tools for matplotlib anim
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends libav-tools && \
+    apt-get install -y --no-install-recommends libav-tools python-pip python-dev build-essential && \
```
parente: Is there a reason to add OS packages for pip and python when the new installs are coming from conda?

scipy-notebook/Dockerfile
```diff
@@ -37,7 +37,10 @@ RUN conda install --quiet --yes \
     'bokeh=0.11*' \
     'sqlalchemy=1.0*' \
     'hdf5=1.8.17' \
-    'h5py=2.6*' && \
+    'h5py=2.6*' \
+    'vincent=0.4.4*' \
+    'beautifulsoup4=4.5.1*' \
```
parente: One more small request that I missed before: would you mind removing the patch part of the version number so that it's allowed to vary like in the other package specs:

```
'vincent=0.4.*' \
'beautifulsoup4=4.5.*' \
```

Same with the packages in the Python2 environment below.

slarson: fixed!

parente:
Thanks for the PR @slarson. 🍰 

I'll report back with the build tag when it finishes pushing to Docker Hub.

parente:
In tag `bde52ed89463`

slarson:
Nice!  Is that the same as latest now?

parente:
Yes. The latest tag always updates automatically to point to the last successful build.

PR #310, opened 2016-12-05
JupyterHub 0.7

minrk:
and get it from conda (conda-forge)

single-user protocol should still be compatible, so this shouldn't change supported Hub versions


parente:
`18e5563b7486`

PR #312, opened 2016-12-08
TensorFlow 0.11.0

skumagai:
The tensorflow-notebook image currently ships TensorFlow version 0.9.0. However, the official TensorFlow website no longer carries documentations for this version. Documentations are available for versions 0.10.0, 0.11.0, and 0.12.0.

Given 0.12.0 is still in RC0 and not available in conda-forge yet, I opted for 0.11.0. I am new to TensorFlow, and I don't know how soon the release version of 0.12.0 would come out. It may be better not  updating the docker image just now. However, it would be great to have a version with  still accessible official documentations.


parente:
Thanks for the PR. 🍰

PR #294, opened 2016-10-12
Fix #269, NB_UID regression

parente:
Also update irkernel to fix base_display problem on rebuild

PR #315, opened 2016-12-20
Update Dockerfile

wizardbc:
DO NOT remove all the files in `$HOME/.local/`


parente:
@wizardbc if I recall, the complete wipe of .local was being done to avoid permission when using the  `NB_UID` env var. The creation the `jupyter` subdir at build time causes problems when writing to it later at runtime under a different user. #294 and #269 have the history.

Did you hit a problem with the full wipe of .local?

wizardbc:
@parente In jupyter/docker-demo-images, `update-dockerfile-include` merges `all-spark-notebook` and `datascience-notebook`.

Removing `.local` directory part removes apache spark scalar kernel.

Maybe you can add some copy code in `all-spark-notebook`.

parente:
Ah, I see. I wonder if we should just install the Toree kernel spec to the global directory instead of to the user home.

PR #320, opened 2016-12-31
Install Toree to conda env instead of user home

parente:
Alternative fix for #315 to allow docker-demo-image to combine the current
datascience-notebook image which clears ~/.local directory with
all-spark-notebook.


parente:
@wizardbc This should fix the issue you hit.

parente:
Included in tag `b17be6205231` on Docker Hub.

PR #321, opened 2016-12-31
[WIP] Test travis caching again

parente:


parente:
The `docker build --cache-from` option in docker 1.13 might make it possible to build quickly and release from Travis https://github.com/docker/docker/pull/26839.

jakirkham:
Did you talk to Travis already about when 1.13 might be available on their images?

parente:
I have not. 

I did start playing with the `--cache-from` locally and found it did not behave the way I assumed it would when rebuilding. I need to dig some more.

parente:
Ah. https://github.com/docker/docker/issues/20131

parente:
Maybe https://github.com/tonistiigi/buildcache before 1.13 lands on travis.

parente:
Using the combination of docker save of all intermediate layers, the travis cache, and docker load suggested on https://giorgos.sealabs.net/docker-cache-on-travis-and-docker-112.html seems to work.

https://travis-ci.org/jupyter/docker-stacks/builds/217608413

If we don't hit the max cache size that travis allows, we can start using this to do builds on merge to master (no more manual step on a VM) and later avoid having to pull the images back to travis.

Going to encode the steps for all layers in the Makefile next.

parente:
Saving all of the images into separate cache tarballs pushes the build over the 50 minute travis limit. Going to try saving the layers into a single tarball, but it might still be living on the edge.

parente:
No such luck. Images are just to big. Even if we forgo the caching, trying to push fresh images to Docker Hub on every build will certainly run over the timeout limit.

Issue #303, opened 2016-11-18
Permission denied on container directory when mounting SSL cert + key


clintworrell:
When running the following:
docker run -d -e USE_HTTPS=yes -v /home/testing/test.pem:/home/jovyan/.local/share/jupyter/notebook.pem -p 8080-9999:8888 jupyter/datascience-notebook

The container doesn't start and the logs show the following:
Traceback (most recent call last):
  File "/opt/conda/lib/python3.5/site-packages/traitlets/traitlets.py", line 526, in get
    value = obj._trait_values[self.name]
KeyError: 'runtime_dir'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/bin/jupyter-notebook", line 6, in <module>
    sys.exit(notebook.notebookapp.main())
  File "/opt/conda/lib/python3.5/site-packages/jupyter_core/application.py", line 267, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/opt/conda/lib/python3.5/site-packages/traitlets/config/application.py", line 652, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-7>", line 2, in initialize
  File "/opt/conda/lib/python3.5/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/opt/conda/lib/python3.5/site-packages/notebook/notebookapp.py", line 1069, in initialize
    self.init_configurables()
  File "/opt/conda/lib/python3.5/site-packages/notebook/notebookapp.py", line 842, in init_configurables
    connection_dir=self.runtime_dir,
  File "/opt/conda/lib/python3.5/site-packages/traitlets/traitlets.py", line 554, in __get__
    return self.get(obj, cls)
  File "/opt/conda/lib/python3.5/site-packages/traitlets/traitlets.py", line 533, in get
    value = self._validate(obj, dynamic_default())
  File "/opt/conda/lib/python3.5/site-packages/jupyter_core/application.py", line 99, in _runtime_dir_default
    ensure_dir_exists(rd, mode=0o700)
  File "/opt/conda/lib/python3.5/site-packages/ipython_genutils/path.py", line 167, in ensure_dir_exists
    os.makedirs(path, mode=mode)
  File "/opt/conda/lib/python3.5/os.py", line 241, in makedirs
    mkdir(name, mode)
[...]

PR #322, opened 2017-01-01
debian, miniconda, notebook version, option updates

parente:
* Upgrade to latest debian base image
* Upgrade to Notebook 4.3
* Upgrade to Miniconda 4.2.12
* Remove USE_HTTPS env var in favor of command line options for key and cert
* Add GEN_CERT env var for generating a self-signed certificate
* Remove PASSWORD env var in favor of the new Notebook 4.3 default token auth
  or the more secure a hashed password command line option

### DIFF
--- a/all-spark-notebook/README.md
+++ b/all-spark-notebook/README.md
@@ -5,7 +5,7 @@
 
 ## What it Gives You
 
-* Jupyter Notebook 4.2.x
+* Jupyter Notebook 4.3.x
 * Conda Python 3.x and Python 2.7.x environments
 * Conda R 3.3.x environment
 * Scala 2.10.x
@@ -17,16 +17,17 @@
 * [tini](https://github.com/krallin/tini) as the container entrypoint and [start-notebook.sh](../base-notebook/start-notebook.sh) as the default command
 * A [start-singleuser.sh](../base-notebook/start-singleuser.sh) script useful for running a single-user instance of the Notebook server, as required by JupyterHub
 * A [start.sh](../base-notebook/start.sh) script useful for running alternative commands in the container (e.g. `ipython`, `jupyter kernelgateway`, `jupyter lab`)
-* Options for HTTPS, password auth, and passwordless `sudo`
-
+* Options for a self-signed HTTPS certificate and passwordless `sudo`
 
 ## Basic Use
 
-The following command starts a container with the Notebook server listening for HTTP connections on port 8888 without authentication configured.
+The following command starts a container with the Notebook server listening for HTTP connections on port 8888 with a randomly generated authentication token configured.
 
 ```
-docker run -d -p 8888:8888 jupyter/all-spark-notebook
+docker run -it --rm -p 8888:8888 jupyter/all-spark-notebook
 ```
+
+Take note of the authentication token included in the notebook startup log messages. Include it in the URL you visit to access the Notebook server or enter it in the Notebook login form.
 
 ## Using Spark Local Mode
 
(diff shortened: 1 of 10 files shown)


parente:
Also fixes #303.

parente:
@minrk or @jakirkham I'd appreciate a review on this one. It removes a couple of env vars we've had since the beginning in favor of better security and fewer magic options. I think moving in this direction is the right thing to do. Users relying on latest might be surprised by the change of options, but this is why we tag every image build IMHO.

minrk:
👍 on all counts, thanks @parente!

parente:
In image tag `6018daa19365` on DHub.

pyspark-notebook/README.md
```diff
@@ -112,32 +114,55 @@ For example, to set the base URL of the notebook server, run the following:
...
 * `-v /some/host/folder/for/work:/home/jovyan/work` - Host mounts the default working directory on the host to preserve work even when the container is destroyed and recreated (e.g., during an upgrade).
-* `-v /some/host/folder/for/server.pem:/home/jovyan/.local/share/jupyter/notebook.pem` - Mounts a SSL certificate plus key for `USE_HTTPS`. Useful if you have a real certificate for the domain under which you are running the Notebook server.
-* `-p 4040:4040` - Opens the default port for the [Spark Monitoring and Instrumentation UI](http://spark.apache.org/docs/latest/monitoring.html). Note every new Spark context that is created is put onto an incrementing port (ie. 4040, 4041, 4042, etc.) by default, and it might be necessary to open multiple ports using a command like `docker run -d -p 8888:8888 -p 4040:4040 -p 4041:4041 jupyter/pyspark-notebook`. You can also control the port Spark uses for its web UI with the `spark.ui.port` config option.
```
jakirkham: I'm not really a Spark user. So this may be an ignorant question. Should we still be noting the port options for the Monitoring and Instrumentation UI or is this not really needed any more for some reason?

all-spark-notebook/README.md
```diff
@@ -207,33 +208,54 @@ For example, to set the base URL of the notebook server, run the following:
...
 * `-v /some/host/folder/for/work:/home/jovyan/work` - Host mounts the default working directory on the host to preserve work even when the container is destroyed and recreated (e.g., during an upgrade).
-* `-v /some/host/folder/for/server.pem:/home/jovyan/.local/share/jupyter/notebook.pem` - Mounts a SSL certificate plus key for `USE_HTTPS`. Useful if you have a real certificate for the domain under which you are running the Notebook server.
-`-p 4040:4040` - Opens the default port for the [Spark Monitoring and Instrumentation UI](http://spark.apache.org/docs/latest/monitoring.html). Note every new Spark context that is created is put onto an incrementing port (ie. 4040, 4041, 4042, etc.) by default, and it might be necessary to open multiple ports using a command like `docker run -d -p 8888:8888 -p 4040:4040 -p 4041:4041 jupyter/pyspark-notebook`. You can also control the port Spark uses for its web UI with the `spark.ui.port` config option.
```
jakirkham: Same question as below.

jakirkham:
Sorry for the delay.

This seems like a pretty nice set of changes. Had a question about a change to the Spark documentation.

FWICT it seems like the security options should be much easier for new users to work with. So that's a good thing.

All in all a good addition.

pyspark-notebook/README.md
parente: Good catch. I was too heavy handed on the cut across all the READMEs. I'll put it back in.

PR #326, opened 2017-01-04
Add issue template

parente:


parente:
@willingc, @minrk I modeled this after the template you setup in Jupyter Hub. Mind taking a look based on your experience about how well that template is panning out?

minrk:
Thanks!

PR #350, opened 2017-03-13
Bump all the package versions

basnijholt:
This PR changes the versions of packages such that we can easily switch to Python 3.6 later.

Once Spark 2.1.1 is released and `spylon-kernel` updated to `py36` we can change `conda` to 4.3* and profit 😄

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -59,10 +59,10 @@ RUN mkdir /home/$NB_USER/work && \
 # Install conda as jovyan
 RUN cd /tmp && \
     mkdir -p $CONDA_DIR && \
-    wget --quiet https://repo.continuum.io/miniconda/Miniconda3-4.3.11-Linux-x86_64.sh && \
-    echo "b9fe70ce7b6fa8df05abfb56995959b897d0365299f5046063bc236843474fb8 *Miniconda3-4.3.11-Linux-x86_64.sh" | sha256sum -c - && \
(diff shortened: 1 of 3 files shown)


basnijholt:
This is failing because of https://github.com/conda-forge/jupyterhub-feedstock/issues/8

minrk:
I think you can retitle this to "bump all the things" now.

basnijholt:
Bumping `matplotlib` to 2.0 and `statsmodels` to 0.8 was not really necessary for Python 3.6, but did it anyways while I am at it :)

basnijholt:
Damn, this is not yet going to work because https://github.com/maxpoint/spylon-kernel only supports Python 3.5.

parente:
@basnijholt thanks for digging into this. If it turns out to be too much of a bear to do all the updates at once, feel free to chunk it up.

> Damn, this is not yet going to work because https://github.com/maxpoint/spylon-kernel only supports Python 3.5.

And that because Spark only supports python 3.5 until 2.1.1 comes out.

basnijholt:
So Python 3.6 is not going to be possible yet.

For now I have removed the commit that changed the `conda` version.

Once Spark 2.1.1 comes out it should be easy to change to Python 3.6, since all Python packages that are in the Dockerfiles now have corresponding `py36` builds (except `spylon-kernel` of course).

parente:
Thanks again. Builds might take a while to reach DH because of all the changes.

parente:
Tagged on DH as `07292b8ebdd5`.

PR #352, opened 2017-03-15
Bump Py4J to 0.10.4

rudolfkral:
Spark 2.1.0 requires Py4J 0.10.4, not 0.10.3.


parente:
Good catch. Thanks.

parente:
Tagged as `6383877f1242` on DH.

PR #356, opened 2017-03-19
Add a tiny alpine based notebook

yuvipanda:
When doing live demos of JupyterHub  container based deployments,
showmanship is important. And one big aspect of it is 'time taken
from nothing at all to running single-user container'. Image pull
times are a big contributor to this first time demo experience -
and all the current docker-stacks images are too big to give a
'pop'.

This image is optimized for size and demo-worthiness only. Not
to be used for anything else, ideally. Part of the demo can be
configuring the spawner to use a different dockerstack image -
this provides a nice way to explain why it is slow without it
taking away from the initial demo's speed.

### DIFF
--- a/tiniest-notebook/Dockerfile
+++ b/tiniest-notebook/Dockerfile
@@ -1,0 +1,11 @@
+FROM alpine:3.5
+
+# We install py-zmq from alpine, since installing it from pip requires we have
+# gcc and build tools installed. manylinux1 wheels don't work on / for alpine yet
+RUN apk add --no-cache python3 py-zmq
+
+RUN pip3 install --no-cache-dir notebook ipykernel jupyterhub
+
+CMD ["/usr/bin/jupyter", "notebook", "--ip=0.0.0.0", "--port=8888"]
+
+EXPOSE 8888
--- a/tiniest-notebook/README.md
+++ b/tiniest-notebook/README.md
@@ -1,0 +1,26 @@
+# Tiniest notebook stack #
+
+This is the tiniest possible docker image that can run a Jupyter Notebook. It
+is primarily meant for demo purposes where speed of pulling is important.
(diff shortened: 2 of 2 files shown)


yuvipanda:
I'm currently using https://github.com/yuvipanda/jupyterhub-simplest-k8s/blob/master/singleuser/Dockerfile for demos, but that's in my personal repo and hence not ideal. Also is too big ;)

yuvipanda:
Am also happy to move this to a different repo - there needs to be somewhere for the kubespawner specific hub to live as well.

parente:
Hey @yuvipanda. This is a good thing to have. Let's chat about whether we want to maintain it in docker-stacks or not. Do you see Jupyter users adopting it for demos or just the core team? If you think many will benefit from it, then there's a few more things that should be done before we merge it:

1. Update the Makefile in the root of docker-stacks to rebuild this image in the same manner as the other images.
2. Add it to the image inheritance diagram in the README so users know that it sits outside the rest of the image hierarchy. (The source+editor link of the image is in the alt-tag of the image in the README.)
3. Pin the alpine:3.5 image to an image sha256 if it tends to rev without a tag version change. (The updates to debian:jessie used to cause grief.)
4. Create the new repo on DockerHub to host it. (I can do this part. Steps are in the README here.)

yuvipanda:
Thanks @parente. The more I think about it, the more I feel it doesn't actually belong here. It shouldn't be used for anything other than demo purposes, and is outside the stack hierarchy here. So I think this should live somewhere else (TBD), and I'll close this PR for now.

parente:
There is jupyter/docker-demo-image. It currently contains the kitchen-sink image running under tmpnb.org with its own Makefile for building it. Maybe it would fit there?

Issue #242, opened 2016-07-18
jupyterhub - systemuser mode


steverweber:
can the systemuser.sh also be included in the base?

```
https://github.com/jupyterhub/dockerspawner/blob/master/systemuser/systemuser.sh
```

parente:
@jtyberg @minrk Does systemuser.sh make sense here?

steverweber:
humm there is more to it....
https://github.com/jupyterhub/dockerspawner/blob/master/singleuser/Dockerfile

parente:
@steverweber Well, the singleuser config is already included and all of the READMEs have a section on how to use it. (https://github.com/jupyter/docker-stacks/tree/master/base-notebook#jupyterhub)

You're asking about systemuser instead, right?

steverweber:
/systemuser/

jtyberg:
Well, I suppose it's the same thing as including singleuser.sh in the image.   Since jupyterhub/dockerspawner has both spawners, including both scripts seems consistent.

steverweber:
@jtyberg I assume the same... however the Dockfiles are a little different from one-another so perhaps a little extra overview is needed.

https://github.com/jupyterhub/dockerspawner/blob/master/systemuser/Dockerfile
https://github.com/jupyterhub/dockerspawner/blob/master/singleuser/Dockerfile

jtyberg:
`docker-stacks` doesn't use the Dockerfiles over in jupyterhub/dockerspawner.

All we do is copy the [start-singleuser.sh](https://github.com/jupyter/docker-stacks/blob/master/base-notebook/start-singleuser.sh) bash script into the container image so it's available.   

https://github.com/jupyter/docker-stacks/blob/master/base-notebook/Dockerfile#L89

So if someone comes along and wants to run the image in a JupyterHub setup, they just run the  script to run the notebook in a mode compatible with JHub.  We could do the same for `start-systemuser.sh`, I would think.

parente:
@steverweber Would you like to submit the start-systemuser.sh script addition as a PR? The line to mimic has moved a bit since @jtyberg's comment. It's now here:

https://github.com/jupyter/docker-stacks/blob/master/base-notebook/Dockerfile#L94

natbusa:
All the systemuser.sh currently does (on top of singleuser.sh) is to remove the default user jovyan and `useradd -u $USER_ID -s $SHELL $USER`.

[...]

PR #359, opened 2017-03-26
Enable jupyterlab as a jupyter serverextension

natbusa:
It's required when working with jupyterhub script start-singleuser.sh and jupyterhub-singleuser

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -74,6 +74,8 @@ RUN conda install --quiet --yes \
     'jupyterlab=0.17.*' \
     && conda clean -tipsy
 
+RUN jupyter serverextension enable --py jupyterlab --sys-prefix
+
 USER root
 
 EXPOSE 8888


parente:
The conda recipe for the package should be enabling the server extension automatically upon install due to https://github.com/conda-forge/jupyterlab-feedstock/blob/master/recipe/post-link.sh, assuming it's being grabbed from conda-forge. If it's not, we should fix that.

natbusa:
Agreed.

jakirkham:
Are we activating the environment that contains `jupyterlab`? If not, that could be the cause.

parente:
It's installing into the root environment. Does it need to be activated at install time?

natbusa:
I was following up on the systemuser setup:
https://github.com/jupyterhub/dockerspawner/tree/master/systemuser

here I see something I do not fully understand:
the jovyan user is deleted (at build time), a new user is created by the startup script (run time). 
also the lab is reinstalled -apparently- again, this time under root.

related to this thread:
https://github.com/jupyter/docker-stacks/issues/242

jakirkham:
> Does it need to be activated at install time?

No, only at run time. We can probably add it to the `start-notebook.sh` script or add an entrypoint script that handles it.

natbusa:
I retested with clean docker images and it's all fine. 

jupyter lab is indeed installed and serverextensions are enabled.
Thank you for pointing me to the conda post-link.sh hooks.

PR #374, opened 2017-04-30
changes for accomomdating lmodern

naveenkumarmarri:


parente:
Thanks again @naveenkumarmarri

PR #375, opened 2017-05-02
reduce files in home directory a bit

minrk:
- add a few missing `--system` flags to conda config
- install default notebook config to /etc/jupyter instead of ~/.jupyter
- add a few missing `conda clean`s

cc @yuvipanda

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -53,7 +53,6 @@ USER $NB_USER
 
 # Setup jovyan home directory
 RUN mkdir /home/$NB_USER/work && \
-    mkdir /home/$NB_USER/.jupyter && \
     echo "cacert=/etc/ssl/certs/ca-certificates.crt" > /home/$NB_USER/.curlrc
 
 # Install conda as jovyan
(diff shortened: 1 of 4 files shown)


base-notebook/Dockerfile
```diff
@@ -87,8 +86,7 @@ CMD ["start-notebook.sh"]
 COPY start.sh /usr/local/bin/
 COPY start-notebook.sh /usr/local/bin/
 COPY start-singleuser.sh /usr/local/bin/
-COPY jupyter_notebook_config.py /home/$NB_USER/.jupyter/
-RUN chown -R $NB_USER:users /home/$NB_USER/.jupyter
+COPY jupyter_notebook_config.py /etc/jupyter/
```
parente: Should this be chown'ed to the `jovyan` user so that root isn't required to edit it?

yuvipanda:
Nice! 

Putting it in non-home locations but still having it be user editable seems important. I usually just isolate everything into /srv/<something> for this reason (and have that be owned by jovyan), but that might be too drastic a change for docker stacks

minrk:
@yuvipanda the biggest thing I think you'll lose by mounting $HOME as a volume is ~all of Julia in the datascience-notebook, which is installed in home (as is Julia's convention). I think we can move this somewhere else if we carefully set some options (maybe `JULIA_PKGDIR` is enough)

parente:
Small bump. Is this ready to merge or are you also planning to tackle the Julia problem as part of it?

minrk:
@parente this should be good to go. I'll tackle Julia in another PR

PR #380, opened 2017-05-11
Add fonts-liberation requirement, fix #372

### LINKED ISSUE #372: Seaborn inclusion results in font-family not found when plotting

`jupyter/all-spark-notebook`

`docker run -it --rm -p 8889:8888 jupyter/all-spark-notebook:latest`

1. Visit http://localhost:8889
2. Start an Python 3 Notebook
3. Execute in cell:

```python
%matplotlib inline
import pandas as pd
import seaborn as sns

pd.Series([1,2,3]).plot(kind='bar')
```

A plot, with no emitted warnings.

**What actually happens?**

The following gets emitted as a warning:

```
/opt/conda/lib/python3.5/site-packages/matplotlib/font_manager.py:1297: UserWarning: findfont: Font family ['sans-serif'] not found. Falling back to DejaVu Sans
  (prop.get_family(), self.defaultFamily[fontext]))
```

I believe the problem is caused by `seaborn`'s stylesheets for `matplotlib`. Plotting without seaborn does not result in this warning. Looking in the matplotlib fonts data directory, I see now sans-serif fonts. However, I don't know the rationale for the specific Debian jesse `FROM`, and I don't know much (anything) about debian font management. This is a minor issue. However, I suspect a debian sys-admin or someone familiar with seaborn could fix more "correctly" than I could.

jbn:
I added the `fonts-liberation` requirement to `base-notebook`. Since it's so early in the image layers, it requires a fairly massive rebuild, which may not be desirable. On the other hand, it should solve the font issue everywhere, and it doesn't add terribly much to the image size. 

See: [Commonly Used Fonts on Debian](https://wiki.debian.org/Fonts#Commonly_Used_Fonts)

Also, FWIW, I considered using,

```sh
fc-list -f "%{file}\n" | grep .ttf | \
while read f ; \
    do cp $f /opt/conda/lib/python3.5/site-packages/matplotlib/mpl-data/fonts/ttf; \
done
```

This copies all available `ttf`s to the `matplotlib` font directory. But, font-liberation seems like a cleaner solution.

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -21,6 +21,7 @@ RUN REPO=http://cdn-fastly.deb.debian.org \
     ca-certificates \
     sudo \
     locales \
+    fonts-liberation \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 


parente:
@jbn Thank you for your first PR here! I favor using an existing package over a custom script, so this LGTM. I'll merge and rebuild the images. 🍰

jbn:
Thanks, @parente! Expect plenty more. I'm using / evangelizing these stacks everywhere. Like, friends have gotten annoyed, even. Got two derivative works coming soon; and, just released [nbmerge](https://github.com/jbn/nbmerge) a few days ago. Jupyter, FTW!

parente:
Thanks for the nbmerge pointer. Happy to have any help here or elsewhere in Jupyter that you're willing to provide!

parente:
Tagged in `cfddc5a3163f` on Docker Hub.

Issue #331, opened 2017-01-26
Consider not writing in user home


akhmerov:
All the images (starting from base notebook) write part of the configuration inside the user home folder, and assume that `/home/$NB_USER/work` will be mounted. This has a practical limitation that it is not easy to setup a hub such that useful folders like `.ssh`, or e.g. `.bash_history` persist across container restarts.

I might be missing underlying assumptions and use cases, but I suggest to assume that home itself would be mounted, and instead modify the global configuration options.

parente:
You make a good point. We've been slowly moving config files out of home and either into the sys-prefix or root prefix over time. It looks like the remaining files of significance in the user's home directory are:

```
.local/share/jupyter/jupyter_notebook_config.py
.ipython/profile_default/startup/mplimporthook.py
.conda/environments.txt
```

I know the first can move into the conda prefix /opt/conda/etc/jupyter. I think the second can as well. I have seen any negative consequences of simply deleting or ignoring the third, but I also haven't experimented or researched it much beyond running `conda env list`.

akhmerov:
Are you referring to a specific derivative dockerfile? In `base-notebook` I only see `.bash_logout  .bashrc  .continuum  .curlrc  .jupyter  .profile  work` in home, and `.continuum` is empty.

akhmerov:
I see now, it's in `scipy-notebook`. I might be confusing something, but I don't understand the role of this file:

```bash
docker run -it --rm jupyter/scipy-notebook bash
jovyan@9d2aacb3af29:~/work$ rm -rf ~/.conda
jovyan@9d2aacb3af29:~/work$ conda env list
# conda environments:
#
python2                  /opt/conda/envs/python2
root                  *  /opt/conda
```

parente:
@akhmerov I don't know the role of the file either. It could be harmless to remove it. https://github.com/conda/conda/issues/945 seems to have a bit of info about it.

akhmerov:
We're using an image without that file for a long time already. It appears it is indeed harmless to remove.

yuvipanda:
This also means that you can't use this in KubeSpawner with Persistent storage, and I don't want to encourage the default to attach persistent storage somewhere other than $HOME. https://github.com/jupyterhub/kubespawner/issues/31 is related.

jakirkham:
Raised an issue about `.conda` ( https://github.com/conda/conda/issues/5388 ) and `.continuum` ( https://github.com/conda/conda/issues/5389 ).

[...]

PR #386, opened 2017-05-17
install Julia packages in /opt/julia

minrk:
This gets close to #331, but we're not quite there. I believe the only files left in $HOME are `.curlrc` and `.ipython/profile_default/startup/mplimporthook`

@parente do you recall what the curlrc config solves and/or if it can be put in `/etc/` somewhere?

### DIFF
--- a/datascience-notebook/Dockerfile
+++ b/datascience-notebook/Dockerfile
@@ -15,13 +15,21 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/*
 
 # Julia dependencies
+# install Julia packages in /opt/julia instead of $HOME
+ENV JULIA_PKGDIR=/opt/julia
+
 RUN echo "deb http://ppa.launchpad.net/staticfloat/juliareleases/ubuntu trusty main" > /etc/apt/sources.list.d/julia.list && \
     apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 3D3D3ACC && \
     apt-get update && \
     apt-get install -y --no-install-recommends \
     julia \
     libnettle4 && apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+    rm -rf /var/lib/apt/lists/* && \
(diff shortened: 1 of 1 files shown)


parente:
The `.curlrc` fixes a problem with using curl to access sites over HTTPS: https://github.com/jupyter/docker-stacks/blob/bb222f49222eceb726ee3b9004905f1cd9121aed/base-notebook/Dockerfile#L57. Without it, every certificate was rejected as untrusted long ago. Maybe newer images don't have this problem or maybe there's a better way to configure it. https://ec.haxx.se/cmdline-configfile.html seems to suggest curl only looks in the home directory on linux.

jakirkham:
Is this just the system `curl`? We built `curl` at conda-forge as well, which uses certificates that we pulled from the certifi package. It generally works ok.

ref: http://github.com/conda-forge/curl-feedstock
ref: http://github.com/conda-forge/ca-certificates-feedstock

parente:
It does indeed predate most of conda forge. It's worth trying to switch to a curl package installed in the root conda env so that we it's available on the default path.

minrk:
I just ran in a base-notebook container:

```bash
rm ~/.bashrc
apt update && apt install curl
curl https://github.com
```

and it works fine. Does that mean we don't need this file anymore?

jakirkham:
Maybe it is worth digging through `git blame` to see who added the file and why. Just in case there is something we are missing. Then again I don't know much about it.

parente:
I'm pretty sure I added the .curlrc. I'll go poke through the history.

parente:
Nope, it was in be4ca700173b8f794934022f1daaa55f0b37b3c9 for issue #153, cert error when Julia downloaded the QuantEcon package.

parente:
Let's get this in. If curl breaks, we can always build again with a fix.

jakirkham:
Should add we have [`julia`]( https://github.com/conda-forge/julia-feedstock ) in `conda-forge` now.

PR #387, opened 2017-05-19
Change GID of NB_USER to NB_GID if NB_GID is passed as a parameter

stolho:
Hi,

we are using jupyter in multi-tenant environment. That means that we are mounting folder with notebooks and data. These folders on host system have different permissions based on uid and gid. The problem is, that currently it allows you to specify only the uid of the user, sending NB_UID as a parameter: `-e NB_UID=1000`

It will be nice to have also the possibility to specify the gid using: `-e NB_GID=100`.

I've also added `-o` parameter to be able to specify non-unique groups.

Best,
Stanislav

### DIFF
--- a/all-spark-notebook/README.md
+++ b/all-spark-notebook/README.md
@@ -208,7 +208,7 @@ Connection to Spark Cluster on Standalone Mode requires the following set of ste
 
 ## Notebook Options
 
-The Docker container executes a [`start-notebook.sh` script](../base-notebook/start-notebook.sh) script by default. The `start-notebook.sh` script handles the `NB_UID` and `GRANT_SUDO` features documented in the next section, and then executes the `jupyter notebook`.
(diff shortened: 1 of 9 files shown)


parente:
I vaguely recall that we used to support change the group ID, but removed that for some reason. I'll do some poking around in the history to see what actually happened and why.

parente:
I was thinking of *Group That User Belongs To* in #188 which recommends adding the NB_USER to GID=0 for reasons explained that text.

FRosner:
I looked at #188 before but afaik the solution was that the guys are having their own fork? I understand that maybe changing the group ID does not satisfy their requirements but it would be consistent with the current configuration options to change the UID.

Is there any drawback in supporting a flexible GID? If you don't specify the parameter then it will just use the default one like normal :)

parente:
I'm cool with this change. Would you mind updating the READMEs to describe this new env var alongside the NB_UID that already exists?

stolho:
I've updated the README files with NB_GID. Hope this is ok :)

parente:
The doc updates look great! Let's merge and build. 🍰

PR #388, opened 2017-05-20
finish removing files from $HOME

### LINKED ISSUE #331: Consider not writing in user home

minrk:
follow-up to #386

- mplimporthook should be obsolete with matplotlib 2.0
- curlrc *seems* to be obsolete now (updates to debian base image?)
- set default WORKDIR to $HOME instead of $HOME/work (leave work there for
compatibility)

There are still a couple of artefacts in $HOME, but nothing that would suffer from being wiped out due to $HOME being mounted as a fresh volume.

cc @yuvipanda

closes #331


base-notebook/Dockerfile
```diff
@@ -52,9 +52,8 @@ RUN useradd -m -s /bin/bash -N -u $NB_UID $NB_USER && \
 
 USER $NB_USER
 
-# Setup jovyan home directory
-RUN mkdir /home/$NB_USER/work && \
-    echo "cacert=/etc/ssl/certs/ca-certificates.crt" > /home/$NB_USER/.curlrc
+# Setup work directory for backward-compatibility
+RUN mkdir /home/$NB_USER/work
```
jakirkham: Is there anything else in docker-stacks using this at this point or is this of concern for some known downstream use cases?

minrk: > Is there anything else in docker-stacks using this

No, this directory is deliberately created and left empty, so that it might be mounted as a volume. Nothing is done with it in docker-stacks.

> is this of concern for some known downstream use cases?


juptyerhub-deploy-docker [uses this](https://github.com/jupyterhub/jupyterhub-deploy-docker/blob/master/jupyterhub_config.py#L34) by default, but since it mounts it as a volume, it probably wouldn't be affected. We can possibly get away with removing the directory without harming anything I *know of*, at least.

akhmerov:
Would it make sense to make `$NB_USER` configurable at this stage?

base-notebook/Dockerfile
parente: `~/work` is configured as the default `notebook-dir` both so that it could be mounted without disturbing all of the files in the home directory and to avoid cluttering the notebook dashboard page.

datascience-notebook/Dockerfile
```diff
@@ -15,13 +15,21 @@ RUN apt-get update && \
...
     apt-get install -y --no-install-recommends \
     julia \
     libnettle4 && apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+    rm -rf /var/lib/apt/lists/* && \
+    # Show Julia where conda libraries are \
+    echo "push!(Libdl.DL_LOAD_PATH, \"$CONDA_DIR/lib\")" >> /usr/etc/julia/juliarc.jl && \
+    # Create JULIA_PKGDIR \
+    mkdir $JULIA_PKGDIR && \
+    chown -R $NB_USER:users $JULIA_PKGDIR
```
parente: Also need to add a chown of this directory to the start-notebook.sh script as well in order to support `NB_USER` and (as of #387) `NB_GID` at docker run time. Look for the similar chown for the conda root directory.

minrk: 👍  done. #387 removed the chown of `.`, which means that $HOME won't have the right permissions in this case, will it? Should I put that back, as well?

yuvipanda:
Is this mergeable now?

datascience-notebook/Dockerfile
parente: @minrk I missed your comment because GitHub folded the diff after your update to the PR. Yes, I think the `.` was removed by mistake and I missed it in code review. I'll send a follow-on with it.

parente: That `.` used to resolve to `$HOME/work`. Long ago, we chmod'ed all of `$HOME` instead until someone pointed out that when `$HOME` is a host-mounted user home directory, the logic in the startup script could potentially wind up breaking ownership on the host-mounted files if `$NB_USER` is set poorly. I don't really see around this, however, if we want ownership set correctly from the container perspective.

PR #391, opened 2017-06-04
Update Tensorflow example

saulshanabrook:
`initialize_all_variables` is deprecated, although it still works, and `global_variables_initializer` is the replacement.


parente:
Thanks for the update @saulshanabrook . 🍰

PR #392, opened 2017-06-04
Fix chown of $NB_USER home

parente:
Follow-up from missed comment in #388.


akhmerov:
@parente do you mean my comment about customizing `$NB_USER`? If no, then disregard the following.

If so, then I don't believe this will be sufficient: if the container end-user passes a nonexistent `$NB_USER` as a variable to `start.sh`, the script will fail because the custom `$NB_USER` was not yet created.

parente:
Sorry I meant the comment from @minrk.

PR #395, opened 2017-06-08
Force ownership change on /home/jovyan/work

Stibbons:
In case Jupyter is started by Marathon with a persistant volume, `/home/jovyan/work` is owned by [root](url). We still want to use uid 1000, partly because we don't want to change ownership on [/opt/conda](url) and so on).

So, just test ownership of the work directory, if it is wrong for this folder, just chown it.

Note: we cannot controle which user the Marathon Volume will be created with, it is mandatorily owned by `root` user.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -16,6 +16,9 @@ if [ $UID == 0 ] ; then
                 chown -R $NB_UID "$d"
             fi
         done
+    elif [[ -d /home/$NB_USER/work && $(stat -c %U /home/$NB_USER/work) != "$NB_USER" ]]; then
+      echo "Change ownership of /home/$NB_USER/work from $(stat -c %U /home/$NB_USER/work) to $NB_USER"
+      chown -R $NB_UID /home/$NB_USER
     fi
 
     # Change GID of NB_USER to NB_GID if NB_GID is passed as a parameter


parente:
I think we had this feature some time ago, and removed it due to multiple reports from people mounting a host volume as `/home/jovyan/work`, not realizing the `NB_UID` needed to be set, and finding their host directory ownership changed to 1000.

Stibbons:
For me, `/home/jovyan/work` and subfolder should at least get read and write rights.

Can I modify it so an environment variable can gate the `chmod` (for example `ENSURE_CHMOD_HOME` ?

parente:
Well, it looks like we recently re-introduced the chmod'ing if the `NB_UID` is changed from the default anyway: https://github.com/jupyter/docker-stacks/blob/master/base-notebook/start.sh#L15. So maybe we should run with it, but better document that `/home/jovyan` is chown'ed to the `NB_UID` when the container starts with `-u root`.

Would you mind submitting changes to the READMEs saying something to that effect?

Stibbons:
Yes, I saw that, but I don't like to chmod all `/opt/conda`each start. The chmod of `/home/jovyan/work`may already take some times it the user has previously heavilly used its workdir.

parente:
What if we check the top-level directory only and make the assumption that if it doesn't match the `NB_UID`, then a recursive chmod is in order?

Stibbons:
That's my patch :)

base-notebook/start.sh
```diff
@@ -16,6 +16,9 @@ if [ $UID == 0 ] ; then
                 chown -R $NB_UID "$d"
             fi
         done
+    elif [[ -d /home/$NB_USER/work && $(stat -c %U /home/$NB_USER/work) != "$NB_USER" ]]; then
```
parente: I meant, should we check if `/home/$NB_USER` has the right ownership here if the code inside the block is going to change that folder, and not `/home/$NB_USER/work`?

Stibbons: In this case, I would do both. I don't know how you use this docker image, but me, I just mount a Marathon volume with `/home/$NB_USER/work` (we are migrating to kubernetes, I guess it is the same). So actually we should chmod the mount point

yuvipanda: In kubernetes, you can set fsGroup in Security Context to 1000 and solve this problem :) Look at z2jh.jupyter.org for a nice k8s setup!

Stibbons: I'll test it when we have our kube ready. For the moment, on our Marathon install, it does not seem possible to set the user id

ezreal92170:
I'm using marathon and jupyter/docker-stacks with local persistant volume. This PR would help me too.

minrk:
#435 should remove the need for ownership modifications, as long as the container user is a member of the user-writable group (10000)

PR #396, opened 2017-06-08
Fix sudoers.d inclusion with debian img (fix #394)

### LINKED ISSUE #394: Using  "GRANT_SUDO=yes" doesn't work
When trying to use `-e GRANT_SUDO=yes` parameter during startup it won't work.
This is because the last line of `/etc/sudoers` is commented, preventing the new file `/etc/sudoers.d/notebook` to be included in the configuration.

Uncommenting this last line would be enough to solve this issue.

```
# cat /etc/sudoers

...
...
%sudo	ALL=(ALL:ALL) ALL

# See sudoers(5) for more information on "#include" directives:

#includedir /etc/sudoers.d
^^^^^^^^^^^^^^^^^^^^^^^^^^--- issue here
```

Stibbons:

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -25,6 +25,8 @@ if [ $UID == 0 ] ; then
 
     # Enable sudo if requested
     if [ ! -z "$GRANT_SUDO" ]; then
+        # Ensure includedir sudoers.d is not commented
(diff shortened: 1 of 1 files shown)


parente:
Thanks for the fix! 🍰

PR #397, opened 2017-06-08
Add logs to start.sh

Stibbons:
Ease user understanding about what's going out on startup


parente:
Restarting the build.

PR #403, opened 2017-06-21
Revert "Fix sudoers.d inclusion on debian img (fix #394)"

### LINKED ISSUE #394: Using  "GRANT_SUDO=yes" doesn't work

### LINKED ISSUE #400: Sudo does not work in the latest docker hub images. (Relation : #394)
ex)
https://hub.docker.com/r/jupyter/minimal-notebook/tags/
latest is df7a34bebed0

/etc/sudoers is 
```
includedir /etc/sudoers.d
```

parente:
This reverts commit 4b3b6697e78bd286910ccabd7cfe423592e1a841.

\#includedir is the correct syntax, not a comment
https://www.sudo.ws/man/1.8.13/sudoers.man.html#Including_other_files_from_within_sudoers

Fixes #400

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -28,8 +28,6 @@ if [ $UID == 0 ] ; then
 
     # Enable sudo if requested
     if [ ! -z "$GRANT_SUDO" ]; then
-        # Ensure includedir sudoers.d is not commented
-        sed -i -E 's|^#(includedir /etc/sudoers.d.*)|\1|g' /etc/sudoers
         echo "$NB_USER ALL=(ALL) NOPASSWD:ALL" > /etc/sudoers.d/notebook
     fi
 


Stibbons:
really ? that's sucks:
[Man](https://www.sudo.ws/man/1.8.13/sudoers.man.html#Including_other_files_from_within_sudoers)
> The pound sign (‘#’) is used to indicate a comment (unless it is part of a #include directive

Sorry for the bad patch. I agree with this PR

parente:
Not a problem. I reviewed it and didn't even catch the fact that it had always been that way. I wonder what caused the sudo not to work in your original ticket now too.

PR #420, opened 2017-07-21
start.sh: fix check for if we are root

minrk:
UID env is not set when launching as root, so root condition is never triggered. Use `id -u` check instead.


parente:
Wonder when this stopped working. LGTM at any rate.

PR #424, opened 2017-08-03
update home folder and username on startup

### LINKED ISSUE #414: consider allowing to change the username
I'm bringing this up more as a discussion point rather than feature request (it is easy to work around this limitation): right now the username is fixed in the images as `NB_USER=jovyan`. In some circumstances it'd be nicer to allow changing the username on startup: sometimes a different generic name is more suitable (e.g. `student`), or one could even pass the username from the authenticator.

I believe changing this would amount to:
* Checking in `start.sh` whether `$NB_USER=jovyan`.
* If not, renaming `jovyan` into `$NB_USER`, changing the location of the home folder, and only after that updating permissions.

akhmerov:
closes #414 

Some remarks:
* This now assumes that the base image does not rename the notebook user, and keeps it `jovyan`. Is that OK?
* Like the previously existing code, I assume that the home is located in `/home/$NB_USER`. This isn't given, but out of scope of the MR
* I do not copy anything from `/home/jovyan` to the new location, assuming that the users know what they want to put in the new location.
* I am not a linux administration expert and I may be forgetting something obvious.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -6,6 +6,9 @@ set -e
 
 # Handle special flags if we're root
 if [ $(id -u) == 0 ] ; then
+    # Handle username change. Since this is cheap, do this unconditionally
+    usermod -d /home/$NB_USER -l $NB_USER jovyan
+
     # Change UID of NB_USER to NB_UID if it does not match
     if [ "$NB_UID" != $(id -u $NB_USER) ] ; then
         echo "Set user UID to: $NB_UID"


parente:
Thanks for the PR @akhmerov. The PR looks good on the face of it. I hope to find a few minutes soon to test it before merging.

parente:
Tested here. Seems to do the right thing. We'll need to update all of the READMEs to indicate that changing the user name is a viable option. But let's merge this first. Happy to take another PR if you have time. If not, we'll get to it. 🍰

akhmerov:
Sorry, I probably won't have time to do this in a while.

PR #425, opened 2017-08-07
warn about unused config

minrk:
if some env vars are set that aren't used, complain about it so behaviour is a bit less surprising.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -27,7 +27,8 @@ if [ $(id -u) == 0 ] ; then
     fi
 
     # Enable sudo if requested
-    if [ "$GRANT_SUDO" == "1" ]; then
+    if [ "$GRANT_SUDO" == "1" || "$GRANT_SUDO" == 'yes' ]; then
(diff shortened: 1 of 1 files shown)


base-notebook/start.sh
```diff
@@ -27,14 +27,23 @@ if [ $(id -u) == 0 ] ; then
     fi
 
     # Enable sudo if requested
-    if [ ! -z "$GRANT_SUDO" ]; then
+    if [ "$GRANT_SUDO" == "1" ]; then
```
parente: All of the READMEs give `-e GRANT_SUDO='yes'` as an example. Should we keep that or update all the documentation? I'd imagine supporting `1` and only `1` going forward might break a few users.

minrk: Good catch, I'll make sure it accepts 'yes'. I mainly wanted to catch 'GRANT_SUDO=0' from accidentally enabling it

parente:
LGTM. Thanks @minrk!

PR #427, opened 2017-08-09
update base image to debian:stretch

minrk:
- Mostly minor s/jessie/stretch/ in a few places
- Switched Julia from ppa to official binaries (ppa doesn't work for stretch due to libssl upgrade, but official binaries are probably better anyway for non-ubuntu)
- removed no-longer-needed backports for jdk-8

<del>The only thing missing is mesos. There doesn't appear to be a .deb for stretch, yet, and I couldn't find any portable mesos binaries.</del>

mesos is installed for jessie, pulling libssl-1.0.0 from jessie-backports.

mesos on stretch: https://github.com/mesosphere/mesos-deb-packaging/issues/109

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Starting with [git commit SHA 9bd33dcc8688](https://github.com/jupyter/docker-st
 
 **When there's a security fix in the Debian base image, do the following in place of the last command:**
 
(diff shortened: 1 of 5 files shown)


minrk:
closed in favor of #428

PR #428, opened 2017-08-09
proposal: base image on ubuntu:16.04 (LTS)

minrk:
alternative to #427

Especially since we use at least one PPA, we might have an easier time tracking Ubuntu.

I haven't been able to find a really good way to choose between debian and ubuntu. Docker best practices reference debian as a good default starting point, but package support seems better on ubuntu.


parente:
I believe we went with debian at first because of the smaller image size. These days, the ubuntu image is not that much larger. I'm OK making the switch, especially since many people have asked for it.

parente:
To put some numbers on sizes:

```
REPOSITORY              TAG                 IMAGE ID            CREATED             SIZE
debian                  stretch             a20fd0d59cf1        2 weeks ago         100MB
ubuntu                  16.04               14f60031763d        2 weeks ago         120MB
```

minrk:
Makes sense. Given our image sizes, 20MB isn't much.

PR #431, opened 2017-08-10
make jupyterhub env->cli args optional

minrk:
DockerSpawner will stop passing these args, so make them optional before we start passing `--arg=<empty>`

### DIFF
--- a/base-notebook/start-singleuser.sh
+++ b/base-notebook/start-singleuser.sh
@@ -4,19 +4,35 @@
 
 set -e
 
-notebook_arg=""
-if [ -n "${NOTEBOOK_DIR:+x}" ]
-then
-    notebook_arg="--notebook-dir=${NOTEBOOK_DIR}"
+# set default ip to 0.0.0.0
+if [[ "$NOTEBOOK_ARGS $@" != *"--ip="* ]]; then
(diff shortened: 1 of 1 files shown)


base-notebook/start-singleuser.sh
```diff
@@ -4,19 +4,35 @@
...
-  --hub-prefix=$JPY_HUB_PREFIX \
-  --hub-api-url=$JPY_HUB_API_URL \
-  ${notebook_arg} \
-  $@
+# handle some deprecated environment variables
+# from DockerSpawner < 0.8.
+# These won't be passed from DockerSpawner 0.9,
+# so avoid specifying --arg=empty-string
+if [ ! -z "$NOTEBOOK_DIR" ]; then
+  NOTEBOOK_ARGS="--notebook-dir=$NOTEBOOK_DIR $NOTEBOOK_ARGS"
```
parente: Do `$NOTEBOOK_DIR` and any of the other env vars need to be quoted to handle spaces and other special chars when passed as command line parameters to jupyterhub-singleuser below?

minrk: Good call. That's the only one that's a path, so I think the rest should be okay.

PR #433, opened 2017-08-10
Remove Python 2

### LINKED ISSUE #432: What to do with Python 2
Right now, `scipy-notebook` sets up a full scientific Python stack for Python 2, and there are currently no images that have a reasonably complete stack without a big legacy install next to it. That seems pretty wasteful.

We should consider when/how to push Python 2 out of the default images. Options include:

- remove python2 entirely
- add python2 *sibling* to scipy-notebook which has a full Python 2 environment, but only the notebook server in the default Python 3 env
- add python2 *descendant* of scipy-notebook that adds a Python2 scipy-stack (this image would be the same same as scipy-notebook is today)

The first option would be my preference. Derivative images can always `conda create -n python2 python=2 ipykernel`.

The result of these would be that descendant images (datascience, pyspark, tensorflow) would all lose Python 2. I think that's a good thing, given that these are supposed to be "opinionated stacks", and my opinion is that nobody should be using Python 2 with Jupyter :)

minrk:
Only Python 3 from now on

Implementation of first proposal in #432 

closes #432

### DIFF
--- a/all-spark-notebook/README.md
+++ b/all-spark-notebook/README.md
@@ -6,7 +6,7 @@
 ## What it Gives You
 
 * Jupyter Notebook 4.3.x
-* Conda Python 3.x and Python 2.7.x environments
+* Conda Python 3.x environment
 * Conda R 3.3.x environment
 * Scala 2.11.x
 * pyspark, pandas, matplotlib, scipy, seaborn, scikit-learn pre-installed for Python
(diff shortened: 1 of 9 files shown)


parente:
Merge whenever you're ready @minrk.

doctapp:
Why remove Python 2.x? Please make this a Docker build option

ericdill:
For all of those wondering why py2 was removed: https://github.com/jupyter/roadmap/blob/master/accepted/migration-to-python-3-only.md

PR #435, opened 2017-08-11
use groups for managing write-access to files

### LINKED ISSUE #434: start.sh performance issue
Hi there! I am currently working at my master theses, which is about integrated practicing environments for computer science classes. For the practical part of my thesis the target is to create a scalable, secure and and easy to use Jupyterhub setup for my university. I am currently working at this [solution](https://github.com/wakonp/jupyterhub-fhjoanneum), which covers the [SwarmSpawner](https://github.com/cassinyio/SwarmSpawner), the [LDAPAuthenticator](https://github.com/jupyterhub/ldapauthenticator), [nbgrader](https://github.com/jupyter/nbgrader) and a persitant [nfs storage](https://github.com/ErezHorev/dockerized_nfs_server). I can report that all works fine and I hope that I get the documentation ready in the next few days.

Coming to my question:
I am using the `NB_UID` and `NB_GID` change steps, when executing the `start.sh` as root. This is  awesome, because in my setup every students also gets a NFS User, where the UID and GID must match. I can report that, the following part of the script creates perform issues when spawning the notebook server:

```
# Careful: $HOME might resolve to /root depending on how the
# container is started. Use the $NB_USER home path explicitly.
for d in "$CONDA_DIR" "$JULIA_PKGDIR" "/home/$NB_USER"; do
    if [[ ! -z "$d" && -d "$d" ]]; then
        echo "Set ownership to uid $NB_UID: $d"
        chown -R $NB_UID "$d"
    fi
done
```
The main problem is, that the `$CONDA_DIR aka /opt/conda` gets most likely bigger with each kernel you install and the recursive `chown` while starting the server is not the best approach I think. In my setup I used the `scipy-notebook` as my base image and added the `IHaskell`kernel from [here](https://github.com/gibiansky/IHaskell) and the `Javascript` kernel from [here](https://github.com/notablemind/jupyter-nodejs). Dockerimage Size 9.57GB [Dockerfile](https://github.com/wakonp/jupyterhub-fhjoanneum/blob/master/fhj-notebooks/teacher-notebook/Dockerfile.teachernotebook) - [DockerHub](https://hub.docker.com/r/walki12/teachernotebook/).

I had do modify the `start.sh` script and remove the `chown` loop to make it work. To overcome a permission issue at server start, I installed all additional kernels with `UMASK 0002` to assure that the `NB_Group`(users) gets full-access to all the installed kernel files. 

I think that the `start.sh` should be modified and maybe it would be nice if it is possible to change the username too. My solution can to it:

```
#!/bin/bash

set -e

if [ $(id -u) == 0 ] ; then
    # Change UID of NB_USER to NB_UID if it does not match
    if [ "$NB_UID" != $(id -u $NB_USER) ] ; then
        echo "Set user UID to: $NB_UID"
        usermod -u $NB_UID -l $JPY_USER $NB_USER
        NB_USER=$JPY_USER
        # Careful: $HOME might resolve to /root depending on how the
        # container is started. Use the $NB_USER home path explicitly.
        #for d in "$CONDA_DIR" "$JULIA_PKGDIR" "/home/$NB_USER"; do
        #    if [[ ! -z "$d" && -d "$d" ]]; then
        #        echo "Set ownership to uid $NB_UID: $d"
        #        chown -R $NB_UID "$d"
        #    fi
        #done
    fi

    # Change GID of NB_USER to NB_GID if NB_GID is passed as a parameter
    if [ "$NB_GID" ] ; then
        echo "Change GID to $NB_GID"
        groupmod -g $NB_GID -o $(id -g -n $NB_USER)
    fi

    # Enable sudo if requested
    if [ ! -z "$GRANT_SUDO" ]; then
    fi

    # Exec the command as NB_USER
    echo "Execute the command as $NB_USER"
    exec su $NB_USER -c "env PATH=$PATH $*"
else
    # Exec the command
    echo "Execute the command"
    exec $*
fi
```
see [here](https://github.com/wakonp/jupyterhub-fhjoanneum/blob/master/fhj-notebooks/teacher-notebook/start.sh). The `$JPY_USER` will get added by the SwarmSpawner when executing the `start-singleuser.sh` script.

I am not quite sure if this could help you.

minrk:
- any files the user should be able to write should have group `user-permissions` with `g+rwX`
- remove `chown` from start.sh because it is no longer needed
- add `fix-permissions` script for setting the user-writable permissions on a path
- users group is GID 100
- containers can set group with `--group-add users` if they want to run with a different uid/gid
  (without -u root -e NB_UID -e NB_GID, which make this unnecessary), for instance:
  

       docker run -u 501:501 --group-add users -it jupyter/base-notebook

Questions:

- is there a reason we need a dedicated high-gid group for this? Or can we leave it as the default `users` group 100?

closes #434
closes #395

I believe this resolves issues brought up in #188 without needing to consider the possible implications of running with gid=0

### DIFF
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -25,13 +25,19 @@ RUN conda install --quiet --yes \
     'r-irkernel=0.7*' \
     'r-ggplot2=2.2*' \
     'r-sparklyr=0.5*' \
-    'r-rcurl=1.95*' && conda clean -tipsy
+    'r-rcurl=1.95*' && \
+    conda clean -tipsy && \
+    fix-permissions $CONDA_DIR
 
 # Apache Toree kernel
-RUN pip --no-cache-dir install https://dist.apache.org/repos/dist/dev/incubator/toree/0.2.0/snapshots/dev1/toree-pip/toree-0.2.0.dev1.tar.gz
-RUN jupyter toree install --sys-prefix
+RUN pip install --no-cache-dir \
+    https://dist.apache.org/repos/dist/dev/incubator/toree/0.2.0/snapshots/dev1/toree-pip/toree-0.2.0.dev1.tar.gz \
+    && \
+    jupyter toree install --sys-prefix && \
+    fix-permissions $CONDA_DIR
 
 # Spylon-kernel
 RUN conda install --quiet --yes 'spylon-kernel=0.4*' && \
-    conda clean -tipsy
-RUN python -m spylon_kernel install --sys-prefix
+    conda clean -tipsy && \
+    python -m spylon_kernel install --sys-prefix && \
+    fix-permissions $CONDA_DIR
--- a/base-notebook/README.md
+++ b/base-notebook/README.md
@@ -58,7 +58,7 @@ You may customize the execution of the Docker container and the command it is ru
 * `-e NB_GID=100` - Specify the gid of the `jovyan` user. Useful to mount host volumes with specific file ownership. For this option to take effect, you must run the container with `--user root`. (The `start-notebook.sh` script will `su jovyan` after adjusting the group id.)
 * `-e GRANT_SUDO=yes` - Gives the `jovyan` user passwordless `sudo` capability. Useful for installing OS packages. For this option to take effect, you must run the container with `--user root`. (The `start-notebook.sh` script will `su jovyan` after adding `jovyan` to sudoers.) **You should only enable `sudo` if you trust the user or if the container is running on an isolated host.**
 * `-v /some/host/folder/for/work:/home/jovyan/work` - Mounts a host machine directory as folder in the container. Useful when you want to preserve notebooks and other work even after the container is destroyed. **You must grant the within-container notebook user or group (`NB_UID` or `NB_GID`) write access to the host directory (e.g., `sudo chown 1000 /some/host/folder/for/work`).**
-* `--group-add user-writable` - use this argument if you are also specifying
+* `--group-add users` - use this argument if you are also specifying
   a specific user id to launch the container (`-u 5000`), rather than launching the container as root and relying on NB_UID and NB_GID to set the user and group.
 
 
--- a/base-notebook/README.md
+++ b/base-notebook/README.md
@@ -58,6 +58,9 @@ You may customize the execution of the Docker container and the command it is ru
 * `-e NB_GID=100` - Specify the gid of the `jovyan` user. Useful to mount host volumes with specific file ownership. For this option to take effect, you must run the container with `--user root`. (The `start-notebook.sh` script will `su jovyan` after adjusting the group id.)
 * `-e GRANT_SUDO=yes` - Gives the `jovyan` user passwordless `sudo` capability. Useful for installing OS packages. For this option to take effect, you must run the container with `--user root`. (The `start-notebook.sh` script will `su jovyan` after adding `jovyan` to sudoers.) **You should only enable `sudo` if you trust the user or if the container is running on an isolated host.**
 * `-v /some/host/folder/for/work:/home/jovyan/work` - Mounts a host machine directory as folder in the container. Useful when you want to preserve notebooks and other work even after the container is destroyed. **You must grant the within-container notebook user or group (`NB_UID` or `NB_GID`) write access to the host directory (e.g., `sudo chown 1000 /some/host/folder/for/work`).**
+* `--group-add user-writable` - use this argument if you are also specifying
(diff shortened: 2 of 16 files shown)


wakonp:
Exactly want I wanted 👍

akhmerov:
This won't fix permissions on mounted containers, right?

minrk:
> This won't fix permissions on mounted containers, right?

Correct, and rightly so - mounted volumes *shouldn't* have their permissions touched, right?

> Since home is often mounted, shouldn't fix_permissions still be run on startup on home?

I don't think so. Since home is often mounted, it should already be mounted with the right permissions. It shouldn't be modified after mounting.

parente:
> Correct, and rightly so - mounted volumes shouldn't have their permissions touched, right?

They should not. Whenever we've accidentally done that, we've had bugs opened here with users saying their host permissions were changed. I think it's up to the user to match the host permission to whatever UID / GID they decide to use in the container.

akhmerov:
Thanks for the clarifications.

base-notebook/README.md
```diff
@@ -58,7 +58,11 @@ You may customize the execution of the Docker container and the command it is ru
 * `-e NB_GID=100` - Specify the gid of the `jovyan` user. Useful to mount host volumes with specific file ownership. For this option to take effect, you must run the container with `--user root`. (The `start-notebook.sh` script will `su jovyan` after adjusting the group id.)
 * `-e GRANT_SUDO=yes` - Gives the `jovyan` user passwordless `sudo` capability. Useful for installing OS packages. For this option to take effect, you must run the container with `--user root`. (The `start-notebook.sh` script will `su jovyan` after adding `jovyan` to sudoers.) **You should only enable `sudo` if you trust the user or if the container is running on an isolated host.**
 * `-v /some/host/folder/for/work:/home/jovyan/work` - Host mounts the default working directory on the host to preserve work even when the container is destroyed and recreated (e.g., during an upgrade).
-
+* `--group-add user-writable` - use this argument if you are also specifying
+  a specific user id to launch the container (`-u 5000`), rather than launching the container as root and relying on NB_UID and NB_GID to set the user and group.
+  
+  It does not
```
parente: Leftover? Same with line below?

all-spark-notebook/Dockerfile
```diff
@@ -25,13 +25,19 @@ RUN conda install --quiet --yes \
     'r-irkernel=0.7*' \
     'r-ggplot2=2.2*' \
     'r-sparklyr=0.5*' \
-    'r-rcurl=1.95*' && conda clean -tipsy
+    'r-rcurl=1.95*' && \
+    conda clean -tipsy && \
+    fix-permissions $CONDA_DIR
```
parente: Is it possible to set the group sticky bit (+s) on the root conda directory so that group permissions are preserved across all installs without running this script each time?

minrk: Setting the sticky bit gets us *almost* there, and I did set it. Good call. New files have the right owners, but not the right *permissions*. For that, I think we need [acls](https://unix.stackexchange.com/a/1315) or the ability to set umask for subsequent RUN commands, but AUFS doesn't support ACLs and I don't see a way to set umask, so I think we're stuck. I'd love to be wrong, though.

parente:
Giving this a shot locally, mostly to better understand the changes and to see if we really need the `NB_USER` and such overrides going forward. I did immediately see that `docker run` on v 17.xCE does not recognize `-g`. Is that flag a typo in the description of the PR or am I missing something?

base-notebook/README.md
```diff
@@ -111,11 +113,6 @@ conda install some-package
 ```python
 # Spawn user containers from this image
 c.DockerSpawner.container_image = 'jupyter/base-notebook'
-
-# Have the Spawner override the Docker run command
-c.DockerSpawner.extra_create_kwargs.update({
-	'command': '/usr/local/bin/start-singleuser.sh'
-})
```
parente: Is start-singleuser no longer needed? If so, we should pull it from the other READMEs as well.

minrk: Yes. `start.sh` already dispatches to `start-singleuser.sh` based on environment variables. Users don't need to know the difference.

parente:
I've tested this locally and it works great. We should consider adding more doc about when / how it's useful vs the NB_USER, NB_UID, etc. env vars. 

Will merge this after the current round of images finishes building. (There's no way to communicate from one build to the next on docker cloud which git SHA to pull and build, so merging again in the meantime means we'll get a mix of code changes across the images for the same tag).

minrk:
The `-g` was just a typo. It should have been `-u uid[:gid]`.

Issue #440, opened 2017-08-17
Removal of Python 2 should be more documented, with a workaround


guillaumerosinosky:
Hi

The removal of Python 2 in jupyter/scipy-notebook provoked lots of errors and astonishment in my scripts. I speak of the following pull requests : https://github.com/jupyter/docker-stacks/pull/432 and https://github.com/jupyter/docker-stacks/pull/433 .

As I probably won't be alone to have this problem, perhaps you could precise in the README.md a workaround (or your maybe planned derived image ?). For instance, I use jupyter/scipy-notebook:b4dd11e16ae4 instead.

Thank you for all the good work.

parente:
To close this out: need an entry on the recipes page about how to add Python 2 back in via a child Dockerfile.

parente:
Documented steps to add a Python 2.x environment here:

https://github.com/jupyter/docker-stacks/wiki/Docker-recipes#add-a-python-2x-environment

doesnotexist:
The verbatim in the documentation does not work. It fails to find qt or pyqt to remove and it can't find kernda in the path.  With some slight modifications to the dockerfile I got it to work.  Here's what works for me: 

```# Choose your desired base image
FROM jupyter/scipy-notebook:latest

# Create a Python 2.x environment using conda including at least the ipython kernel
# and the kernda utility.
RUN conda create --quiet --yes -p $CONDA_DIR/envs/python2 python=2.7 ipython kernda ipykernel && \
    conda clean -tipsy

USER root

# Create a global kernelspec in the image and modify it so that it properly activates
# the python2 conda environment.
RUN $CONDA_DIR/envs/python2/bin/python -m ipykernel install && \
    $CONDA_DIR/envs/python2/bin/kernda -o -y /usr/local/share/jupyter/kernels/python2/kernel.json

USER $NB_USER
```

rancas:
@doesnotexist I've built a docker using your indications and I am now able to run notebooks with python2.
Unfortunately, I've no scipy and pandas installed for Python 2.7 notebooks.
Can this be fixed or do I need to install 2.7 all modules during docker build?

doesnotexist:
I've don't really know much about conda, so I could be wrong, but I imagine you could just tack on any packages that are conda installable to the end of the conda create line, and it would install those packages for that version of python when its creating the env.

parente:
> Can this be fixed or do I need to install 2.7 modules during docker build?

[...]

PR #441, opened 2017-08-17
[ci skip] Note in README about Python 2.x

parente:
Addresses #440

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ Starting with [git commit SHA 9bd33dcc8688](https://github.com/jupyter/docker-st
 
 ## Other Tips and Known Issues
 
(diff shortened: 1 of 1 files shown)


parente:
@guillaumerosinosky Sorry for the surprise. We try to make it clear that `latest` is an ever changing set of images in the README, but I understand that in the case of removing Python 2, a note on the google group at least probably would have been a good idea.

This PR adds a note about the last tag containing Python 2.7. I think the recipes wiki page would be the right place for a longer Dockerfile example for how to add Python 2.x back into your own images. Do you agree?

minrk:
> I think the recipes wiki page would be the right place for a longer Dockerfile example for how to add Python 2.x

Yes, that makes sense.

Issue #389, opened 2017-05-25
Keras stack 


hakandilek:
Hi There,

I would like to bring up a discussion about a keras stack image based on the existing tensorflow stack.

Stibbons:
+1. Keras should be installed inside the tensorflow image

parente:
Happy to take a PR here. https://github.com/conda-forge/keras-feedstock exists so this should be a an addition to conda install in that Dockerfile.

parente:
`786611348de1` on docker hub/cloud.

PR #448, opened 2017-08-25
Keras stack introduced

hakandilek:
For #389


parente:
@hakandilek thanks for this. To cut down on the cost of maintaining all the different stack here, do you think Keras could be added to the exist tensorflow-notebook stack instead? We could rename it for clarity if that's a sticking point.

hakandilek:
@parente well, keras is just one extra pip-package on top of tensorflow. For me it does not make a difference. The question is if someone would like to have that stack without the keras package.

parente:
Since it's just one additional install, it's much lower cost in terms of maintenance and ease for users to add it to the tensorflow image I think. Would you mind making that change to the PR?

jakirkham:
Would add that `keras` is available from `conda-forge` as well.

parente:
Thanks @jakirkham. Let's definitely add this to the conda-install in the tensorflow image. We can rename that image jupyter/deep-notebook or some such if tensorflow is too specific.

hakandilek:
Rolling it back due to #456.

PR #455, opened 2017-09-05
Fix working directory problems with custom NB_USER

### LINKED ISSUE #442: Permission denied error when starting the container with custom user and UID/GID

`jupyter/base-notebook`

`docker run -it --user=root -e NB_USER=user -e NB_UID=50000 -e NB_GID=50000 jupyter/base-notebook /usr/local/bin/start-singleuser.sh`

Run the docker run command

I do not see any errors and Jupyter server is started.

**What actually happens?**

observed `PermissionError: [Errno 13] Permission denied: '/home/user'`. I believe this is due to the directory of the new user has not been created. The user and the home directory have been changed by [this command](https://github.com/jupyter/docker-stacks/blob/master/base-notebook/start.sh#L10) but the new home directory does not yet exist at that point. Hence, it produced an error when starting `jupyterhub-singleuser` conda command.

Possible solution may be to add `mv /home/jovyan /home/$NB_USER` command after the change of user and its home directory .

parente:
Improve start.sh logging along the way. Fixes #442.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -6,18 +6,23 @@ set -e
 
 # Handle special flags if we're root
 if [ $(id -u) == 0 ] ; then
+    # Switch to the root of the container in case we start adjusting paths
+    # that impact the default working directory
+    cd /
+
     # Handle username change. Since this is cheap, do this unconditionally
-    usermod -d /home/$NB_USER -l $NB_USER jovyan
+    echo "Set username to: $NB_USER"
+    usermod -m -d /home/$NB_USER -l $NB_USER jovyan
 
     # Change UID of NB_USER to NB_UID if it does not match
     if [ "$NB_UID" != $(id -u $NB_USER) ] ; then
-        echo "Set user UID to: $NB_UID"
+        echo "Set $NB_USER UID to: $NB_UID"
         usermod -u $NB_UID $NB_USER
     fi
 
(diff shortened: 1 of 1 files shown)


base-notebook/start.sh
```diff
@@ -6,18 +6,23 @@ set -e
...
 # Handle special flags if we're root
 if [ $(id -u) == 0 ] ; then
+    # Switch to the root of the container in case we start adjusting paths
+    # that impact the default working directory
+    cd /
+
     # Handle username change. Since this is cheap, do this unconditionally
-    usermod -d /home/$NB_USER -l $NB_USER jovyan
+    echo "Set username to: $NB_USER"
+    usermod -m -d /home/$NB_USER -l $NB_USER jovyan
```
minrk: This should only happen if `/home/$NB_USER` doesn't exist

base-notebook/start.sh
```diff
@@ -27,8 +32,11 @@ if [ $(id -u) == 0 ] ; then
         echo "$NB_USER ALL=(ALL) NOPASSWD:ALL" > /etc/sudoers.d/notebook
     fi
 
+    # Switch to the user's home directory after adjusting any paths
+    cd /home/$NB_USER
```
minrk: This needs to be aware of what $PWD was before. We don't want the workdir to be set unconditionally to $HOME

Issue #406, opened 2017-06-25
More integration tests


parente:
The makefile currently tests for notebook server HTTP port liveliness in a container after building an image. We should also consider checking:

* Various docker run flags we support
* Kernel startup using the API
* Execution of a basic notebook, maybe using nbconvert

parente:
Foundation for this went in as PR #408. Leaving open because I think it's valuable to have additional kernel-specific integration tests using nbconvert, the execution preprocessor, and some notebook-based tests.

parente:
Folks have contributed tests over time and will hopefully continue to do so. I'm going to close this out.

PR #458, opened 2017-09-07
homedir fixes

minrk:
- don't relocate home if home exists (e.g. mounted volume)
- respect workdir, only cd to new home if we were in /home/jovyan/

#455 broke SystemUserSpawner, where workdir is already set and /home/$NB_USER is a mounted volume that should not be modified.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -6,13 +6,26 @@ set -e
 
 # Handle special flags if we're root
 if [ $(id -u) == 0 ] ; then
-    # Switch to the root of the container in case we start adjusting paths
-    # that impact the default working directory
-    cd /
 
     # Handle username change. Since this is cheap, do this unconditionally
     echo "Set username to: $NB_USER"
-    usermod -m -d /home/$NB_USER -l $NB_USER jovyan
+    usermod -d /home/$NB_USER -l $NB_USER jovyan
+
+    # handle home and working directory if the username changed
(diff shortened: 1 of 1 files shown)


parente:
Whoopsie. That was a merge misclick in #455.

base-notebook/start.sh
```diff
@@ -6,13 +6,26 @@ set -e
...
-    # Switch to the root of the container in case we start adjusting paths
-    # that impact the default working directory
-    cd /
 
     # Handle username change. Since this is cheap, do this unconditionally
     echo "Set username to: $NB_USER"
-    usermod -m -d /home/$NB_USER -l $NB_USER jovyan
+    usermod -d /home/$NB_USER -l $NB_USER jovyan
+
+    # handle home and workding directory if the username changed
```
parente: working

parente:
The number of options we support and the scenarios we need to keep in mind when making changes no longer fits in my head between contributions here. I'll spend some time on #406 to alleviate that pain.

Thanks for the save @minrk

base-notebook/start.sh
minrk: Thanks! fixed.

parente:
https://cloud.docker.com/app/jupyter/repository/docker/jupyter/all-spark-notebook/timeline failed to build (out of disk). I've restarted it.

Issue #460, opened 2017-09-09
Cannot start at `/home/jovyan/work` (scipy-notebook)


mandaiy:
Hi! Thanks for using Jupyter's docker-stacks images.

If you are requesting a library upgrade or addition in one of the existing images, please state the desired library name and version here and disregard the remaining sections.

If you are reporting an issue with one of the existing images, please answer the questions below to help us troubleshoot the problem. Please be as thorough as possible.

**What docker image you are using?**
jupyter/scipy-notebook:latest

**What complete docker command do you run to launch the container (omitting sensitive values)?**

Example: `docker container run --rm -it -p 8888:8888  jupyter/scipy-notebook:latest`

**What steps do you take once the container is running to reproduce the issue?**

Example:

1. Visit http://localhost:8889

**What do you expect to happen?**
`/home/jovyan/work` should be displayed

**What actually happens?**

The files and directories of `/home/jovyan/fecets` are displayed.

I checked that everything was fine when I used jupyter/minimal-notebook.
I guess the WORKDIR should be changed after running `jupiter nbextension` in scipy-notebook/Dockerfile.

parente:
Recently introduced regression. Will be fixed by #462.

parente:
Thanks again @mandaiy for the bug report and the initial PR to fix. 🍰

PR #461, opened 2017-09-09
Change starting directory of scipy-notebook

mandaiy:
Solves #460


parente:
@mandaiy Thanks for the fix. Turns out there's more that needs to be done to fix permissions and whatnot that I forgot when merging the original facets PR. I'll open a new PR based on your commit with the additional fixes.

Issue #454, opened 2017-09-04
Update to Julia 0.6


appleparan:
If you are requesting a library upgrade or addition in one of the existing images, please state the desired library name and version here and disregard the remaining sections.

* Please update Julia kernel to v0.6. It's been a 2 months after Julia v0.6 have been released. I think it's time to update Julia kernel to v0.6 because it introduce new type system and some breaking changes ([changelog](https://github.com/JuliaLang/julia/blob/master/NEWS.md#julia-v060-release-notes)).

naveenkumarmarri:
@parente I can try this out
Can you point me which files to change?

parente:
Julia is installed from https://launchpad.net/~staticfloat/+archive/ubuntu/juliareleases which currently has 0.5.2 as the latest available release. It's installed here in the datascience-notbook/Dockerfile: https://github.com/jupyter/docker-stacks/blob/master/datascience-notebook/Dockerfile#L22

I'm open to changing how we install Julia if that necessary. Or maybe that upstream channel could use help in packaging 0.6. I'm not a Julia user (Julian?) so contributions on how to move this forward are most welcome.

jakirkham:
Related note, we have started building Julia in conda-forge. Haven't really added any related packages yet. Though could still try using the standard Julia mechanisms for installation with it.

parente:
Somehow, I knew you'd be popping in, @jakirkham, I just didn't know how soon. 💯 

https://github.com/conda-forge/julia-feedstock

@naveenkumarmarri If you want to give it a shot, try removing the install of julia using apt and run `conda install julia` instead. Then see if the various `julia -e 'Pkg.init()'` commands further down in the Dockerfile succeed or not.

If you start on the PR, and give maintainers access to contribute to it, we can iterate on it together.

naveenkumarmarri:
@parente sure

appleparan:
https://github.com/staticfloat/julia-docker/tree/master/julia 

you can check these files for updating julia, it needs for docker images.

naveenkumarmarri:
@parente I checked `conda search julia --channel conda-forge`
The version is 0.5.2
I couldn't find any package with 0.6 version
Should we install from source code?

parente:
PR for Julia 0.6 opened for the conda-forge feedstock 2 hours ago.

https://github.com/conda-forge/julia-feedstock/pull/7

tkelman:
[...]

PR #471, opened 2017-09-25
Update julia to v0.6.0 from upstream binaries

jpfairbanks:
Thanks for providing these awesome dockerfiles, my libraries require julia-v0.6+. So I updated the dockerfile.

download and install julia from julialang.org install to /usr/local/bin/


jpfairbanks:
addresses #454

parente:
PR #510 pushed Julia 0.6 across the finish line. Thanks for the submission though.

PR #473, opened 2017-09-26
Upgrade to notebook 5.1

gnestor:


takluyver:
I don't see any reason not to do this...

parente:
Docker Cloud builds choked here: https://cloud.docker.com/app/jupyter/repository/registry-1.docker.io/jupyter/scipy-notebook/builds/531023ea-8288-453f-8d37-a472e290d7ab

I've restarted that build. If it succeeds, down stream images should build as well.

Issue #472, opened 2017-09-26
ppc64le dockerhub images


vaibhavsood:
Hi, I'm looking to enable docker containers for packages like jupyter for multiple architectures on dockerhub, starting off with a focus on ppc64le. Wanted to check on the work that will be involved to do that

I have locally been able to successfully test ppc64le changes to the dockerfiles in this repo - changes for enabling additional arch would be minimum.

Would like to know how a multi-arch image for this can be initiated? The official images have this method for enabling multi-arch : https://github.com/docker-library/official-images#multiple-architectures, however this repo isnt an official image yet. Would the first step be to make this an official image or are there any other ways the dockerhub jupyter image can be enabled for ppc64le?

Hi! Thanks for using Jupyter's docker-stacks images.

If you are requesting a library upgrade or addition in one of the existing images, please state the desired library name and version here and disregard the remaining sections.

If you are reporting an issue with one of the existing images, please answer the questions below to help us troubleshoot the problem. Please be as thorough as possible.

**What docker image you are using?**

Example: `jupyter/scipy-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

Example: `docker run -it --rm -p 8889:8888 jupyter/all-spark-notebook:latest`

**What steps do you take once the container is running to reproduce the issue?**

Example:

1. Visit http://localhost:8889
2. Start an R noteobok
3. ...

**What do you expect to happen?**

Example: ggplot output appears in my notebook.

**What actually happens?**

Example: No output is visible in the notebook and the notebook server log contains messages about ...

parente:
There was a contribution of a ppc64 Dockerfile a while back: https://github.com/jupyter/docker-stacks/tree/master/base-notebook. We haven't kept it up to date or invested in automatically building the images for that platform since then.

vaibhavsood:
Ok thanks! 

Want to check if you would be willing to make the docker images official https://docs.docker.com/docker-hub/official_repos/#how-do-i-create-a-new-official-repository , that would offload the build process to docker. They also build for multi-arches so that could be used to auto build additional arches like ppc64le.

Alternatively, i can help to bring the ppc64le dockerfile upto date (i work with the original contributor of the ppc64le dockerfile @seth-priya ) and can also help with setting up a CI build for autobuilding ppc64le images locally, would that be okay?

My interest here is to publish ppc64le docker images (ideally official images), want to know your thoughts, thanks!

parente:
[...]

PR #482, opened 2017-10-13
Upgrade to notebook 5.2.0

gnestor:


parente:
Thanks @gnestor. Looks like the conda package for Linux isn't quite baked yet according to the build failure.

gnestor:
Oh, do we need to get the conda package published before we can update docker-stacks?

parente:
Yep. Once it’s available we can retrigger the build here and see how it goes.

jakirkham:
Toggling for CI.

gnestor:
Looks like we're good to go!

base-notebook/Dockerfile.ppc64le.patch
```diff
@@ -81,14 +81,14 @@
...
 -    'jupyterhub=0.7.*' \
 -    'jupyterlab=0.18.*' \
 -    && conda clean -tipsy
 +# Install Jupyter notebook and Hub
 +RUN yes | pip install --upgrade pip
 +RUN yes | pip install --quiet --no-cache-dir \
-+    'notebook==5.1.*' \
++    'notebook==5.2.*' \
 +    'jupyterhub==0.7.*' \
 +    'jupyterlab==0.18.*'
```
jakirkham: Should we be switching to `pip` here? Also `pip` is a `conda` package. So it might be better to upgrade `pip` via `conda` if it is needed.

parente: Are there ppc64 versions of these packages?

jakirkham: Good point! I guess not since this is a recent `notebook` release.

jakirkham: Though all the other packages had come from Miniconda for ppc64le before. Maybe we should just drop the `pip` upgrade line. The others (excepting `jupyterhub`) require the `notebook`.

parente: We currently don't have a way to build the ppc64 versions ourselves so I'm a bit hesitant to change anything with the setup beside the minor notebook version number in this PR. Issue #472 is open to address ppc64 going forward, and @seth-priya, @vaibhavsood, or anyone can follow-up with other improvements.

PR #491, opened 2017-11-02
Fix wrong permissions when changing NB_UID or NB_GID.

dsevilla:
When changing UID or GID of the jovyan user, I noticed that a usermod and/or groupmod commands are issued, but then the own home directory and `/opt/conda` still belong to the previous UID of jovyan, so permissions should be granted again in the case of a new UID or GID is requested via container environment variables.

This change takes some time. It could be made in the background, but then it may make some tests fail that execute just after starting the notebook server.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -39,6 +39,14 @@ if [ $(id -u) == 0 ] ; then
         groupmod -g $NB_GID -o $(id -g -n $NB_USER)
     fi
 
+    # Grant permissions on home and in /opt/conda to $NB_USER
+    if [ "$NB_UID" != $(id -u $NB_USER) -o "$NB_GID" ] ; then
+	echo "Granting permissions to $NB_USER on /home/$NB_USER and /opt/conda"
+        CURRENT_GID=${NB_GID:-$(id -g $NB_USER)}
+	chown -R $NB_UID:$CURRENT_GID /home/$NB_USER
+	chown -R $NB_UID:$CURRENT_GID /opt/conda
+    fi
+
     # Enable sudo if requested
     if [[ "$GRANT_SUDO" == "1" || "$GRANT_SUDO" == 'yes' ]]; then
         echo "Granting $NB_USER sudo access"


base-notebook/start.sh
```diff
@@ -39,6 +39,14 @@ if [ $(id -u) == 0 ] ; then
         groupmod -g $NB_GID -o $(id -g -n $NB_USER)
     fi
 
+    # Grant permissions on home and in /opt/conda to $NB_USER
+    if [ "$NB_UID" != $(id -u $NB_USER) -o "$NB_GID" ] ; then
+	echo "Granting permissions to $NB_USER on /home/$NB_USER and /opt/conda"
+        CURRENT_GID=${NB_GID:-$(id -g $NB_USER)}
+	chown -R $NB_UID:$CURRENT_GID /home/$NB_USER
```
parente: We used to perform a chown on the home directory, but ran into problems when the home directory is mounted from the host. #435 and #458 have some of the history. The fix-permission script was introduced to avoid having to chown because the user is in a group that has default read/write access to /opt/conda and the home dir. However, if the user's GID is changed, then I'm sure that falls apart.

I think we should avoid reintroducing the chown if possible, but I'm not sure how just yet. I think having the ability to change UID, GID, username, sudo permission, etc. might be causing more problems than they're worth and maybe some should be deprecated.

dsevilla: OK, you're right there seems to be a bit too many options for these. The fact is that in all the combinations I tried I didn't try one that only changed the NB_UID and not the NB_GID, allowing then write access to these directories. This has worked for the moment.

However, if not all the options, I think that Jupyter being a tool that most of the time needs another volume mounted where your notebooks reside, it would be nice to seamlessly support at least this case, when the user of the machine is not exactly UID 1000, or at least add it as a common case explaining that you can use NB_UID, but *not* NB_GID if you want to have access both to the externally mounted volumes of use with UID NB_UID *and* to the /opt/conda directory to be able to install packages.

For the moment, then, I'll close the pull reqest. Really thanks for your help,
diego.

Issue #489, opened 2017-10-24
Unable to write R/lib path on jupyter/all-spark-notebook (bbfb8aad625b)


blacknred0:
I am not sure if this is related to [Issue 443](https://github.com/jupyter/docker-stacks/issues/447), but it could be.  I was trying to install a package and wasn't able through jupyter.  

I ended up solving the issue by 

1. Starting `docker run -it --rm -p 8888:8888 -p 4040:4040 -v $(pwd):/home/jovyan -e GRANT_SUDO=yes --user root jupyter/all-spark-notebook start.sh bash`
2. Then `sudo chmod -R 777 /usr/local/spark-2.2.0-bin-hadoop2.7/R/lib/`
3. Then `jupyter-notebook` and was able to install with no issues.

**What docker image you are using?**

`jupyter/all-spark-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

`docker run -it --rm -p 8888:8888 -p 4040:4040 -v $(pwd):/home/jovyan -e GRANT_SUDO=yes --user root jupyter/all-spark-notebook`

**What steps do you take once the container is running to reproduce the issue?**

Example:

1. Visit http://localhost:8889
2. Start an R notebook
3. Run this command to install a package
```
install.packages("Hmisc", dependencies = TRUE, repos = "http://cran.rstudio.com/")
```

**What do you expect to happen?**

Installation of the package should take place successfully.

**What actually happens?**

The package is unable to install due to `/usr/local/spark-2.2.0-bin-hadoop2.7/R/lib` not being able to write to the path.

```
Loading required package: Hmisc
Warning message in library(package, lib.loc = lib.loc, character.only = TRUE, logical.return = TRUE, :
“there is no package called ‘Hmisc’”Installing package into ‘/usr/local/spark-2.2.0-bin-hadoop2.7/R/lib’
(as ‘lib’ is unspecified)
Warning message in install.packages("Hmisc", dependencies = TRUE, repos = "http://cran.rstudio.com/"):
“'lib = "/usr/local/spark-2.2.0-bin-hadoop2.7/R/lib"' is not writable”
Error in install.packages("Hmisc", dependencies = TRUE, repos = "http://cran.rstudio.com/"): unable to install packages
Traceback:

1. install.packages("Hmisc", dependencies = TRUE, repos = "http://cran.rstudio.com/")   # at line 3 of file <text>
2. stop("unable to install packages")
```

[...]

PR #496, opened 2017-11-12
Patch R package install user issues

blacknred0:
-The following patch addresses issue #489
(https://github.com/jupyter/docker-stacks/issues/489)


parente:
Thanks @blacknred0. 🍰

PR #499, opened 2017-11-14
prepend R channel

minrk:
instead of appending, which causes broken R due to mismatched icu version.

gives R channel lower priority than conda-forge.

This is probably due to some package having an improper pin somewhere.


minrk:
Merging since R doesn't work at all right now without this.

parente:
Thanks @monrk. Good catch. I wonder if we need that channel at all anymore. Entirely possible conda-forge has everything we install on its own now.

Issue #466, opened 2017-09-13
Missing timezone data OS package


szaydel:
**What docker image you are using?**

`jupyter/datascience-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

`docker run -e GRANT_SUDO=yes --user=root -d -p 8123:8888 -v /var/lib/docker/volumes/jupyter/r/$NAME:/home/jovyan/work --name=jupyter-$NAME jupyter/datascience-notebook start-notebook.sh --NotebookApp.token='' $NAME` 
...
where `$NAME == $1` from command line. This is a snippet from a shell script.

**What steps do you take once the container is running to reproduce the issue?**

1. Visit http://localhost:8123
2. Start an R noteobok
3. require(tidyquant)

The problem I am observing is inability to process CSV files due to missing timezone data. It seems like the timezone data package is not by default installed with the base OS image.

I addressed this issue via shell as such: `sudo apt-get update ; sudo apt install tzdata` . In my mind this package should have really been there from the start.

anatoliykmetyuk:
Any news about this one? It is pretty strange that a standard feature of a standard library does not work out of the box.

Am not able to do `sudo apt-get update ; sudo apt install tzdata`, because "passwordless" sudo requires password, which I don't know.

anatoliykmetyuk:
For anyone else struggling.

I have used a custom `Dockerfile` with the following contents:

```
FROM jupyter/r-notebook

USER root
RUN apt-get update
RUN apt install tzdata

USER $NB_USER
```

Then build and run this image as specified in the [official tutorial](https://docs.docker.com/get-started/part2/).

jakirkham:
Care to do a PR @anatoliykmetyuk?

Issue #502, opened 2017-11-18
R missing timezone data and Olson database


fasib:
**What docker image you are using?**

`jupyter/r-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

`docker run -it --rm -p 8889:8888 jupyter/r-notebook:latest`

**What steps do you take once the container is running to reproduce the issue?**

Start an R noteobok

Run:
```
Sys.timezone()
```
and
```
library(tidyverse)
default_locale()
```

**What do you expect to happen?**

Something like:
```
'Twilight Zone'
```
and
```
[library loading output]
<locale>
Numbers:  123,456.78
Formats:  %AD / %AT
Timezone: UTC
Encoding: UTF-8
<date_names>
Days:   Sunday (Sun), Monday (Mon), Tuesday (Tue), Wednesday (Wed), Thursday
        (Thu), Friday (Fri), Saturday (Sat)
Months: January (Jan), February (Feb), March (Mar), April (Apr), May (May),
        June (Jun), July (Jul), August (Aug), September (Sep), October
        (Oct), November (Nov), December (Dec)
AM/PM:  AM/PM
```

**What actually happens?**

```
Warning message in normalizePath("/etc/localtime"):
“path[1]="/etc/localtime": No such file or directory”

'NA'
```
and
```
[library loading output]
Warning message in OlsonNames():
“no Olson database found”

Error: Unknown TZ UTC
Traceback:

1. default_locale()
2. locale()
3. check_tz(tz)
4. stop("Unknown TZ ", x, call. = FALSE)
```

**Fix i'm using right now:**
- Added `tzdata` to the pre-requisites.

~~Should I just create a pull request with this change?~~ created a PR

PR #503, opened 2017-11-18
(fix) Add tzdata as R pre-requisites. #502

fasib:


parente:
Also fixes #466. Thanks for the PR @fasib .

PR #508, opened 2017-11-28
pytest integration tests

parente:
* Port server liveliness test from make to test/test_notebook.py
* Add placeholders for base-notebook tests
* Update travis config to install test dependencies on the host

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,6 @@ sudo: required
 services:
     - docker
 install:
-    - pip install pytest docker
+    - pip install docker pytest requests
 script:
     - make build-test-all
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,6 @@ sudo: required
 services:
     - docker
 install:
-    - pip install docker pytest requests
+    - make test-env
 script:
     - make build-test-all
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
-language: generic
-
+language: python
+python:
+    - 3.6
 sudo: required
 services:
     - docker
-
+install:
+    - pip install pytest docker
 script:
     - make build-test-all
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,6 @@ sudo: required
 services:
     - docker
 install:
-    - make test-env
+    - make test-reqs
 script:
     - make build-test-all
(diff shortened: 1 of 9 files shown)


parente:
Probably good enough for review now. It doesn't add tests of specific kernels and containers yet, but it's a reasonable starting point I think.

base-notebook/test/test_container_options.py
```diff
@@ -0,0 +1,114 @@
...
+    c = container.run(
+        user='1010:1010',
+        group_add=['users'],
+        command=['start.sh', 'id']
+    )
+    rv = c.wait(timeout=5)
+    assert rv == 0
+    assert 'uid=1010 gid=1010 groups=1010,100(users)' in c.logs(stdout=True).decode('utf-8')
+
+@pytest.mark.dev
```
parente: Need to remove this.

parente: And from the test above.

base-notebook/test/test_container_options.py
```diff
@@ -0,0 +1,114 @@
...
+    echo "test content" > test.txt
+    cat test.txt
+    ''')
+    script.chmod(0o755)
+
+    c = container.run(
+        volumes={
+            home.strpath: {'bind': '/home/jovyan', 'mode': 'rw'},
+        },
+        command=['start.sh', '/home/jovyan/test.sh']
```
parente: This test script should change to trying to write to a subdirectory that is not chown'ed to `jovyan:users`on the host and then to one that is. It should then assert the first fails and the second succeeds. Right now, it's not testing much

.travis.yml
```diff
@@ -1,8 +1,10 @@
...
-
+language: python
+python:
+    - 3.6
 sudo: required
 services:
     - docker
-
+install:
+    - pip install docker pytest requests
```
ericdill: should this be moved in to the `make test` command? Otherwise it seems to make it slightly more challenging to run the tests

parente:
Latest test failing because of a jupyterlab / jupyter-widgets version conflict. Updating jupyterlab and to 0.30.* and trying again.

parente:
Merging to get the JLab compatibility fix in. Looks like master rebuilds are broken at the moment because of it.

minrk:
Hooray! Thanks, @parente!

PR #509, opened 2017-11-29
 Fix for granting SUDO to jovyan user and run bash commands.

tlinnet:
Adding extra information to the main README file, for easier access to sub images.

Fix for granting SUDO to jovyan user and run bash commands.

Running:
docker run -ti --rm -p 8888:8888 -e GRANT_SUDO=yes --user root jupyter/base-notebook:033056e6d164 start.sh

Will immediately exit the container.

With the fix, the image stays in bash after granting SUDO to jovyan user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,13 +14,14 @@ If you're familiar with Docker, have it configured, and know exactly what you'd 
 # -ti: pseudo-TTY+STDIN open.
 # -rm: remove the container on exit.
 # -p: publish port to the host
-
 docker run -ti --rm -p 8888:8888 jupyter/<your desired stack>:<git-sha-tag>
+
+# And mount current directory into 'work' directory in image
+docker run -ti --rm -p 8888:8888 -v "$PWD":/home/jovyan/work jupyter/<your desired stack>:<git-sha-tag>
 
 # Bagkground mode:
 # -d: detach, run container in background.
 # -P: Publish all exposed ports to random ports
-
 docker run -d -P jupyter/<your desired stack>:<git-sha-tag>
 ```
 
--- a/README.md
+++ b/README.md
@@ -34,6 +34,16 @@ If this is your first time using Docker or any of the Jupyter projects, do the f
 3. Follow the README for that stack.
 
 ## Visual Overview
+[Click here for commented build history, with reference to SHA values.](https://github.com/jupyter/docker-stacks/wiki/Docker-build-history)
+
+* base-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/base-notebook), [SHA list](https://hub.docker.com/r/jupyter/base-notebook/tags/)
+* minimal-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/minimal-notebook), [SHA list](https://hub.docker.com/r/jupyter/minimal-notebook/tags/)
+* scipy-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/scipy-notebook), [SHA list](https://hub.docker.com/r/jupyter/scipy-notebook/tags/)
+* r-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/r-notebook), [SHA list](https://hub.docker.com/r/jupyter/r-notebook/tags/)
+* tensorflow-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/tensorflow-notebook), [SHA list](https://hub.docker.com/r/jupyter/tensorflow-notebook/tags/)
+* datascience-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/datascience-notebook), [SHA list](https://hub.docker.com/r/jupyter/datascience-notebook/tags/)
+* pyspark-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/pyspark-notebook), [SHA list](https://hub.docker.com/r/jupyter/pyspark-notebook/tags/)
+* all-spark-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/all-spark-notebook), [SHA list](https://hub.docker.com/r/jupyter/all-spark-notebook/tags/)
 
 Here's a diagram of the `FROM` relationships between all of the images defined in this project:
 
--- a/README.md
+++ b/README.md
@@ -10,6 +10,17 @@ Opinionated stacks of ready-to-run Jupyter applications in Docker.
 If you're familiar with Docker, have it configured, and know exactly what you'd like to run, this one-liner should work in most cases:
 
 ```
+# Foreground mode:
+# -ti: pseudo-TTY+STDIN open.
+# -rm: remove the container on exit.
+# -p: publish port to the host
+
+docker run -ti --rm -p 8888:8888 jupyter/<your desired stack>:<git-sha-tag>
+
+# Bagkground mode:
+# -d: detach, run container in background.
+# -P: Publish all exposed ports to random ports
+
 docker run -d -P jupyter/<your desired stack>:<git-sha-tag>
 ```
 
--- a/README.md
+++ b/README.md
@@ -7,19 +7,23 @@ Opinionated stacks of ready-to-run Jupyter applications in Docker.
 
 ## Quick Start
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -10,6 +10,18 @@ Opinionated stacks of ready-to-run Jupyter applications in Docker.
 If you're familiar with Docker, have it configured, and know exactly what you'd like to run, this one-liner should work in most cases:
```
parente: Probably need to update this lead-in sentence since now there's a few one-liners. (And thanks for that!)

README.md
```diff
@@ -10,6 +10,18 @@ Opinionated stacks of ready-to-run Jupyter applications in Docker.
...
 
 ```
+# Foreground mode:
+# -ti: pseudo-TTY+STDIN open.
+# -rm: remove the container on exit.
+# -p: publish port to the host
+docker run -ti --rm -p 8888:8888 jupyter/<your desired stack>:<git-sha-tag>
+
+# And mount current directory into 'work' directory in image
```
parente: Maybe expand this comment to say why you'd pick this form (retain your work across containers) vs the first (throwaway instance).

parente: Also, something probably needs to be said about permissions on `$PWD` so that the default 1000:100 (jovyan:user) in the container can write to it.

README.md
```diff
@@ -22,6 +34,16 @@ If this is your first time using Docker or any of the Jupyter projects, do the f
 3. Follow the README for that stack.
 
 ## Visual Overview
+[Click here for commented build history, with reference to SHA values.](https://github.com/jupyter/docker-stacks/wiki/Docker-build-history)
+
+* base-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/base-notebook), [SHA list](https://hub.docker.com/r/jupyter/base-notebook/tags/)
```
parente: 💯

README.md
```diff
@@ -22,6 +34,16 @@ If this is your first time using Docker or any of the Jupyter projects, do the f
...
+* scipy-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/scipy-notebook), [SHA list](https://hub.docker.com/r/jupyter/scipy-notebook/tags/)
+* r-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/r-notebook), [SHA list](https://hub.docker.com/r/jupyter/r-notebook/tags/)
+* tensorflow-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/tensorflow-notebook), [SHA list](https://hub.docker.com/r/jupyter/tensorflow-notebook/tags/)
+* datascience-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/datascience-notebook), [SHA list](https://hub.docker.com/r/jupyter/datascience-notebook/tags/)
+* pyspark-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/pyspark-notebook), [SHA list](https://hub.docker.com/r/jupyter/pyspark-notebook/tags/)
+* all-spark-notebook: [README](https://github.com/jupyter/docker-stacks/tree/master/all-spark-notebook), [SHA list](https://hub.docker.com/r/jupyter/all-spark-notebook/tags/)
 
 Here's a diagram of the `FROM` relationships between all of the images defined in this project:
```
parente: A nit: Maybe keep this sentence + the image first in the section since the title is "A Visual Overview", then follow with the links for more details?

base-notebook/start.sh
```diff
@@ -47,7 +47,12 @@ if [ $(id -u) == 0 ] ; then
 
     # Exec the command as NB_USER
     echo "Execute the command: $*"
-    exec su $NB_USER -c "env PATH=$PATH $*"
+    if [ $# -eq 0 ]; then
```
parente: I think `start.sh <required command>` was the original intent of the script. If we want to make running bash the default, that's cool. But is the separate sudo form necessary?

parente:
@tlinnet Thanks for the doc contributions. I've started feeling like docker-stacks has grown to the point where a ReadTheDocs site is warranted with more content. The content you've provided is all a great step in that direction.

tlinnet:
Hi @parente.

I am using a Dockerfile which pulls from the scipy image. Then i fiddle with packages and setup.

Here, the easiest would be to lunch the image with sudo for jovyan and install packages. And when it is working, transfer the commands to the Dockerfile.

Currently, the setup would be to run image as root, install packages, su to jovyan and test. Exit back to root, change... 

This is annoying. :)

base-notebook/start.sh
tlinnet: One could do.

... bash start.sh bash 

This change aims for 
... start.sh
Would arrive at bash.
Right now, it will exit image.

README.md
tlinnet: Hi @parente.

When you mount a volume, my test showed you don't need to change permissions.
Permission change is only for internal folders in image, for example after a COPY command. Is this right?

README.md
tlinnet: My problem was that I not easily could find the SHA values 'easily'.

This change reflect that. The order could be rearranged. :)

README.md
parente: It appears to depend on either the version of docker or the OS. The latest Docker for Mac seems to allow `jovyan` read-write access to the volume when mounted over `/home/jovyan`. The version of Docker running on Travis CI for Linux seems to respect the ownership and permissions of the host directory when mounted as volume, and disallows write access by `jovyan` by default (ref: integration test failure https://travis-ci.org/jupyter/docker-stacks/builds/310366105).

base-notebook/start.sh
parente: I'm OK with that behavior change, but we should also make `bash` the default when `GRANT_SUDO` is not set for consistency.

parente:
@tlinnet I'm just getting back to looking at PRs. I'd like to make a few minor edits to the project README (ordering, additional command doc, typo fixes ) before merging. Can you check or confirm that you've checked the "Allow edits from maintainers?" box in your PR? If you're busy, I can merge what's here and just send my updates as a separate PR later.

tlinnet:
Hi @parente . The "Allow edits from maintainers" is showing ticked and green. 
Best Troels

parente:
Pushed on commit touching up some of the wording in the main README, then rebased this PR master to pick up the integration tests that merged into master ahead of this PR. Watching Travis for green ...

parente:
UID and GID tests failed. Looking at whether the expected behavior is now different.

base-notebook/start.sh
```diff
@@ -45,9 +45,23 @@ if [ $(id -u) == 0 ] ; then
...
+            sudo -H -u $NB_USER bash -c 'env; PATH=$PATH; bash'
+        else
+            exec su $NB_USER -c "env PATH=$PATH $*"
+        fi
+    else
+        if [ $# -eq 0 ]; then
+            echo "No arguments supplied"
+            bash -c 'env; PATH=$PATH; bash'
+        else
+            env PATH=$PATH $*
```
parente: The test failure is due to the user never becoming the appropriate `NB_USER` in the case where the container is run as root to configure an alternative `NB_UID` or `NB_GID` without `GRANT_SUDO` set. The original `start.sh` logic used the `exec su $NB_USER` command regardless of whether `GRANT_SUDO` was set or not to get the correct behavior here.

I think using the `exec su` form on line 62 should fix that problem, and will make that change after verifying the tests pass locally and refactoring the duplicate script lines. I'll need to toy with line 60 some to see how that behaves in the various sudo/no-sudo, root-on-start/no-root-on-start, etc. configurations. Some additional tests will probably help here.

Sidenote: yay for tests!

parente:
@tlinnet I think I've cleaned up all of the relevant test failures while retaining the desired behavior of running bash when no other command is specified under all configurations we support (e.g., `--user root`, `-e GRANT_SUDO=yes`, neither). I made one semantic change to improve consistency in `start.sh`, namely using `exec sudo` in all cases when the initial `uid == 0`. This change means multi-part commands like `start.sh 'foo && bar'` now need to run as `start.sh bash -c 'foo && bar'`.

I'm going to merge what's here to avoid it getting stale. If I've botched anything, please comment and we can continue in a follow-on PR.

PR #519, opened 2017-12-25
Add option to use PORT environment variable

riywo:
This commit will add an ability to configure listening port number at runtime just providing `PORT` environment variable.

It can be workaround by passing `start-notebook.sh --NotebookApp.port=8080` as a runtime command, but passing `PORT` environment variable is much popular way to configure listening port by 12-factor apps.

### DIFF
--- a/base-notebook/jupyter_notebook_config.py
+++ b/base-notebook/jupyter_notebook_config.py
@@ -9,7 +9,7 @@ import stat
 
 c = get_config()
 c.NotebookApp.ip = '*'
-c.NotebookApp.port = 8888
+c.NotebookApp.port = int(os.getenv('PORT', 8888))
 c.NotebookApp.open_browser = False
 
 # Generate a self-signed certificate


parente:
Hi @riywo. We used to support a `PORT` environment variable, but found it conflicted with env vars for some use cases (https://github.com/jupyter/docker-stacks/pull/33) and ultimately decided to remove it to reduce the number of feature flags we need to maintain across the scripts and images here.

riywo:
Hi @parente ,

Thank you for explaining the history. I didn't know that, sorry.

So, without using environment variables, `start-notebook.sh --NotebookApp.port=8080` is the option to configure the listening port. Could you add this to README doc or somewhere else?

parente:
@riywo No problem. It seems reasonable to add that info to the recipes page or maybe the README in the root of the project. Out of curiosity, though, do you have a use case for adjusting the port used by the notebook server within the container instead of the Docker mapping to an alternative external port (e.g., `docker run -it -p 5555:8888 jupyter/base-notebook`)?

riywo:
@parente Thank you for adding the README. The use case is sometime you can't use port mapping. For example, AWS Fargate uses host type network, so container listening port must be exposed to outside. In such case, you want to configure container port so that the workaround is needed.

I'm closing this issue. Thank you for your advice.

parente:
@riywo Thanks for the explanation. Would you like to add an entry to the https://github.com/jupyter/docker-stacks/wiki/Docker-recipes page about how to adjust the internal port?

PR #525, opened 2018-01-04
Append /opt/conda/bin to the sudo secure_path

parente:
Fixes #524. Adds a test to prevent regression.

Issue #524, opened 2018-01-03
sudo: jupyter: command not found


sergesvr:
Hi! Thanks for using Jupyter's docker-stacks images.

If you are requesting a library upgrade or addition in one of the existing images, please state the desired library name and version here and disregard the remaining sections.

If you are reporting an issue with one of the existing images, please answer the questions below to help us troubleshoot the problem. Please be as thorough as possible.

**What docker image you are using?**

`jupyter/datascience-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

`docker run -dit --name datascience --user root -e GRANT_SUDO=yes -e NB_UID=1000 -e NB_GID=1000 --restart always -p 80:8888 -v /vagrant:/home/jovyan/work start-notebook.sh --NotebookApp.notebook_dir=work --NotebookApp.token=''`

**What steps do you take once the container is running to reproduce the issue?**

docker logs datascience

Set username to: jovyan
usermod: no changes
Set jovyan GID to: 1000
Granting jovyan sudo access
Executing the command: jupyter notebook --NotebookApp.notebook_dir=work --NotebookApp.token=
sudo: jupyter: command not found
Set username to: jovyan
usermod: no changes
Granting jovyan sudo access

**What do you expect to happen?**

see the jupyter notebook initialization logs

**What actually happens?**

the server does not start as the sudo user does not find jupyter

parente:
Regression due to #509. I need to add more integration tests clearly.

In the meantime, the workaround is to pin to the prior build tag, 0b3ec811c968.

parente:
Gah, right, forgot about `secure_path` in `sudoers`. We can either add `/opt/conda/bin` permanently as part of the build process or switch to full paths like `/opt/conda/bin/jupyter-notebook` if we don't want to trust things that might get installed in the container.

nbonnotte:
I have the same issue with `docker run -ti --user=root -e NB_UID=1000 -e NB_GID=1000 jupyter/datascience-notebook`

I think `sed -ri "s#Defaults\s+secure_path=\"([^\"]+)\"#Defaults secure_path=\"\1:$CONDA_DIR/bin\"#" /etc/sudoers` should have been added outside the  GRANT_SUDO condition

Or the documentation should say that the GRANT_SUDO condition is necessary in that case, but I don't think that's the idea

Bidek56:
Has this been resolved? I am still getting "sudo: jupyter: command not found" when running the latest 

[...]

PR #529, opened 2018-01-09
Add /bin to sudo secure_path

nbonnotte:
see issue #524

cf. PR #525

### DIFF
--- a/base-notebook/test/test_container_options.py
+++ b/base-notebook/test/test_container_options.py
@@ -85,6 +85,18 @@ def test_sudo_path(container):
     assert c.logs(stdout=True).decode('utf-8').rstrip().endswith('/opt/conda/bin/jupyter')
 
 
+def test_sudo_path_without_grant(container):
+    """Container should include /opt/conda/bin in the sudo secure_path."""
+    c = container.run(
+        tty=True,
+        user='root',
+        command=['start.sh', 'which', 'jupyter']
+    )
+    rv = c.wait(timeout=10)
+    assert rv == 0
+    assert c.logs(stdout=True).decode('utf-8').rstrip().endswith('/opt/conda/bin/jupyter')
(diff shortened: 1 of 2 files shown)


parente:
@nbonnotte Thanks for the PR. It looks like the tests failed waiting for a notebook server to startup in many cases. I'll retry them, but it's possible the change in start.sh broke something that needs local debugging.

base-notebook/test/test_container_options.py
```diff
@@ -77,7 +77,6 @@ def test_sudo_path(container):
     c = container.run(
         tty=True,
         user='root',
-        environment=['GRANT_SUDO=yes'],
```
parente: It'd be nice to keep the `GRANT_SUDO` option in this test to check that `jovyan` can run the `sudo which jupyter` command, and write another test (or parameterize this one) to check the case where the container launches as root without granting the user sudo.

parente:
Replicating the test that is hanging:

```
~/p/docker-stacks ❯❯❯ docker run -it --rm --user root jupyter/base-notebook start.sh sudo which jupyter
Set username to: jovyan
usermod: no changes
Executing the command: sudo which jupyter
[sudo] password for jovyan:
```

The issue is that `start.sh` does `exec sudo -E -H -u $NB_USER PATH=$PATH $cmd` to switch to the `$NB_USER` which is `jovyan` by default. But then as `jovyan`, it's trying to execute the `sudo which jupyter` command given in the test or on the command line above. 

When running the container as `--user root` with the `start.sh` script, the sudo test is implicit. I think we just need the new test to check `which jupyter` instead of `sudo which jupyter`. That or the test should not use `start.sh` and just run `sudo which jupyter` directly.

nbonnotte:
@parente The tests are now passing

Issue #517, opened 2017-12-17
[Discussion] How to support additional stacks?


parente:
We’ve been collecting PRs that add new image definitions to this project (see: #512, #486 #444). On one hand, I'd love to showcase the diversity of languages supported by Jupyter. On the other, I’m not sure we have the maintainer bandwidth for PHP, Ruby, Cling, etc. How might we proceed? Some ideas:

* Create a Jupyter "docker-stacks starter kit" project that people can use to set up their own custom image build process based on what we use here. Have a central place to link to such community images.
* Go ahead and accept additional image PRs here, but ask the contributors if they're willing to help maintain the images. Put maintainer info in each stack README.
* Go ahead and accept additional image PRs here, noting which we're most adept at maintaining and which are maintained by best effort.
* Create separate repos under github.com/jupyter for each image. Maintain each project as best we can.

minrk:
I like option one the most. Building a generation tool and a wiki-style index of community images seems a lot more sustainable than growing this repo forever. Part of what's great about docker hub, etc. is that there's no reason for all of these images to be in one place, as long as they are discoverable. The only reason to do that is to consolidate maintenance effort, and we are already at (or past) our limit with the images we have, so we can't very responsibly add more. At least not without a commitment from each PR author that they will support their proposed image indefinitely.

Maybe we can learn from [docker official repositories](https://docs.docker.com/docker-hub/official_repos/) for how to draw these lines.

parente:
I'm in favor of the first option as well. I'm thinking that a ReadTheDocs site for this repo could serve as place to post a list of community maintained images, not to mention serve as motivation for me to finally consolidate the duplicate text spread across the READMEs.

taqtiqa-mark:
Frankly Eight containers seems like overkill.  
The main advantage of this effort is complex integrations are done 'once'.

Maybe the maint. burden can be eased and value increased at the same time.
A defensible approach could be to delineate three/four containers:

Base: In memory work loads, single machine (but still multiple container use cases)
Middle: Mix of big data and in memory work. Several machines (<10).
Upper: Big Data Cloud environment use cases.  Multiple locations many multiples of machines
Blue-sky: Experimental components and configurations

To help determine what goes in it helps to have additional defined (arbitrary) constraints.
Some constraints could be:
- Container image sizes (compressed): 
    - Base <1GB, Middle <2.5 Upper  <3.5 GB, Blue-sky unlimited
- Runtime memory use (at startup or for hello world type apps):
[...]

Issue #516, opened 2017-12-13
[Discussion] Status CUDA Support for tensorflow image


pascalwhoop:
Hi all,
As far as I understood from [this post](https://github.com/jupyter/docker-stacks/issues/264#issuecomment-241606872), it's a legal issue if you can roll out NVIDIA CUDA with the images or not. 
1. is that true?
2. could we extend the docs with a link to an external post (to be written?) that shows how to port this Dockerimage to one that supplies CUDA support? I am currently working on this, but I am not sure yet if I will succeed. Have others used it together? What were your experiences?

parente:
@pascalwhoop I'm not familiar with the licensing and distributions requirements for CUDA. Maybe @jakirkham can chime in.

> could we extend the docs with a link to an external post (to be written?) that shows how to port this Dockerimage to one that supplies CUDA support?

The recipes wiki page (https://github.com/jupyter/docker-stacks/wiki/Docker-recipes) sounds like the right place to put a link.

jamestwebber:
Adding a 👍 here to register interest in this feature, and happy to help debug if I can. A GPU-enabled TF image would be super useful.

pascalwhoop:
Okay I made some progress today. I built a Dockerfile based on the nvidia cuda image and then added everything else myself. I had issues with the [tensorflow image](gcr.io/tensorflow/tensorflow:latest-gpu) hosted on gcr.io, because my Python3 kernels actually used python 2 underneath (which is a big nono obviously)

The image contains:

- jupyter of course
- python 2+3, all packages are in 3 though
- TensorFlow
- OpenAI Gym
- Roboschool
- bullet (to avoid MujoCo proprietary stuff)
- ffmpeg
- X server + VNC ability

It's big (5GB) but I guess it has a lot packing in it. 

The Python libs are the same as the datascience notebook, I left all R stuff out though. I think that can be a separate notebook

The repo with my built Dockerfile is [this one](https://github.com/pascalwhoop/tf_openailab_gpu_docker) and I would love some feedback. If it pleases the community we can think about the best place to put it. I am unsure if it is a "Jupyter" or a "TensorFlow" or even a "AI research" Image... so under which organization to place it remains to be seen.

I am currently building the image on a google cloud VM and I will push it to the docker hub so people can check it out and see if it works for them. Building it yourself requires you to DL the cudnn.so which you can only get with an nvidia account (yuck) but we should digress and swallow the proprietary pill for now...

parente:
[...]

PR #533, opened 2018-01-14
Add PyTorch notebook

narumiruna:
Add pytorch=0.3.0 and torchvision=0.2.0 based on scipy-notebook.

I pushed the images to my docker hub:
```
$ docker run -it --rm -p 8888:8888 narumi/pytorch-notebook
```

### DIFF
--- a/pytorch-notebook/.dockerignore
+++ b/pytorch-notebook/.dockerignore
@@ -1,0 +1,2 @@
+# Documentation
+README.md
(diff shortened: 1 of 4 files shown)


parente:
@narumiruna Thanks for working on the PyTorch Dockerfile. We're currently discussing ways to support additional stacks in #517. Please feel free to chime in with your thoughts there. We're leaning toward having new stacks live in separate repositories and listing them in a central documentation site here.

There's a separate conversation in #516 about CUDA for Tensorflow which includes some open questions about whether or not the CUDA license allow redistribution in Docker images. You may want to check that out as well.

narumiruna:
@parente Since I'm not familiar with CUDA license, I removed the CUDA support.

PR #538, opened 2018-01-22
ReadTheDocs site, User Guide sections

parente:
For #517

I'll setup a temporary RTD site out of my fork/branch for easier
consumption when the docs are far enough along. At the moment, I
don't think they're worth reviewing yet, except perhaps the general
outline of the doc sections.

PR #539, opened 2018-01-28
jupyterhub-lab support will be enabled if $JUPYTER_ENABLE_LAB is set.

### LINKED ISSUE #531: Consider adding jupyterhub-lab to images with JupyterHub and Lab
We currently install JupyterHub and JupyterLab into all the notebook images. It'll be great to install https://github.com/jupyterhub/jupyterlab-hub too in them, since without it Lab doesn't have clean links to JupyterHub control panel.

clkao:
Closes #531.

### DIFF
--- a/base-notebook/README.md
+++ b/base-notebook/README.md
@@ -29,6 +29,12 @@ Take note of the authentication token included in the notebook startup log messa
 
 The Docker container executes a [`start-notebook.sh` script](./start-notebook.sh) script by default. The `start-notebook.sh` script handles the `NB_UID`, `NB_GID` and `GRANT_SUDO` features documented in the next section, and then executes the `jupyter notebook`.
 
+You can launch [JupyterLab](https://github.com/jupyterlab/jupyterlab) by setting `JUPYTER_ENABLE_LAB`:
+
+```
+docker run -it --rm -e JUPYTER_ENABLE_LAB=1 --rm -p 8888:8888 jupyter/base-notebook
+```
+
 You can pass [Jupyter command line options](https://jupyter.readthedocs.io/en/latest/projects/jupyter-command.html) through the `start-notebook.sh` script when launching the container. For example, to secure the Notebook server with a custom password hashed using `IPython.lib.passwd()` instead of the default token, run the following:
 
 ```
--- a/base-notebook/start-notebook.sh
+++ b/base-notebook/start-notebook.sh
@@ -8,5 +8,9 @@ if [[ ! -z "${JUPYTERHUB_API_TOKEN}" ]]; then
   # launched by JupyterHub, use single-user entrypoint
   exec /usr/local/bin/start-singleuser.sh $*
 else
-  . /usr/local/bin/start.sh jupyter notebook $*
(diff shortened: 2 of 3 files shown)


parente:
@clkao thanks for looking into this. Is that environment variable set by JupyterHub automatically? If so, great. If not, we prefer documenting how users may change the start command instead of introducing a custom env var that we have to support long term.

clkao:
@parente Yes I should add this to the Notebook Options section in README.  It seem I should be adding to all of them?  Just want to make sure sure that they are not rendered by templates somewhere else.

parente:
@clkao If there's documentation in JupyterHub or the jupyterhub-lab extension that describes the environment variables and when it's set, we should point to that in the READMEs. That involves updating all of the READMEs at the moment since #538 is still a work in progress.

clkao:
@parente I've cleaned up this a bit to use `JUPYTER_ENABLE_LAB`, which is more general for enabling lab via env for both standalone notebook and hub. Let me know if the base-notebook README changes are fine and I'll copy to other READMEs.

base-notebook/README.md
```diff
@@ -29,6 +29,12 @@ Take note of the authentication token included in the notebook startup log messa
 
 The Docker container executes a [`start-notebook.sh` script](./start-notebook.sh) script by default. The `start-notebook.sh` script handles the `NB_UID`, `NB_GID` and `GRANT_SUDO` features documented in the next section, and then executes the `jupyter notebook`.
 
+You can launch [JupyterLab](https://github.com/jupyterlab/jupyterlab) by setting `JUPYTER_ENABLE_LAB`:
+
+```
+docker run -it --rm -e JUPYTER_ENABLE_LAB=1 --rm -p 8888:8888 jupyter/base-notebook
```
parente: Later in the READMEs we document the following way to start JupyterLab instead of Notebook in these images with:

```
docker run -it --rm -p 8888:8888 jupyter/base-notebook start.sh jupyter lab
```

I'm thinking we should extend the start-singleuser script to accept an alternative command like start.sh does like so:

```
# defaults to the original behavior of running jupyterhub-singleuser
docker run -it --rm -p 8888:8888 jupyter/base-notebook start-singleuser.sh 
# overrides the command to run jupyter labhub instead
docker run -it --rm -p 8888:8888 jupyter/base-notebook start-singleuser.sh jupyter labhub
```

That way we can avoid introducing a env var that we have to maintain as a special feature of the stack images. What do you think?

clkao: That works too. I still think env var is somewhat less intrusive though.  Image you have a jupyterhub with multiple images that some have lab installed, and some might be older ones.  Setting the env var now suggests using jupyterlab whenever it is available.

If we require the spawner to specify singleuser_cmd, people using formspawner/profilesspawner will need to carefully select the appropriate start command for individual image.

parente: Thanks for the clarification about the use case. I think the env var makes sense, now that I understand some images might have the env var set while others may not.

parente:
@clkao LGTM. I think the addition to the base-notebook README is fine for now in hopes that we'll have a consolidated ReadTheDocs site soon.

PR #540, opened 2018-01-28
optional GPU builds

### LINKED ISSUE #516: [Discussion] Status CUDA Support for tensorflow image

clkao:
This PR enables GPU builds by adding `GPU=1` to make, which switches to nvidia/cuda:8.0-cudnn7-devel-ubuntu16.04 for the base image, and builds notebook images as the `latest-gpu` tags to be consumed by downstream images. #540

The tensorflow-notebook also installs tensorflow-gpu accordingly.

Closes #516.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,12 @@ arch_patch/%: ## apply hardware architecture specific patches to the Dockerfile
 	fi
 
 build/%: DARGS?=
+ifeq ($(GPU),1)
+build/%: TAG_POSTFIX?=-gpu
+build/%: BARGS?=--build-arg BASE=nvidia/cuda:8.0-cudnn7-devel-ubuntu16.04 --build-arg TAG_POSTFIX=$(TAG_POSTFIX)
+endif
 build/%: ## build the latest image for a stack
-	docker build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@)
+	docker build $(BARGS) $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest$(TAG_POSTFIX) ./$(notdir $@)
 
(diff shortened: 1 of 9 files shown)


parente:
@clkao Thank you for the PR. However, there's an unresolved issue in #516 about whether we can legally build and distribute images that include CUDA or not. The EULA for CUDA is rather lengthy and is not a standard open source license. I don't think we can accept PRs that add CUDA until we have an OK from Nvidia to maintain docker images that include it or legal counsel on the matter.

clkao:
@parente I am aware of the license issue. Do you think it's ok if we have the build tooling in the repository, without actually publishing the ones based on nvidia/cuda to our docker hub?

I think the images with gpu support are popular demands, and the build process should probably be shared to reduce potential duplicated efforts. Merely referring to the CUDA image should leave the EULA binding to whoever is actually building it.

parente:
> Do you think it's ok if we have the build tooling in the repository, without actually publishing the ones based on nvidia/cuda to our docker hub?

As a matter of principal, we do not wish to deal with the distribution of non-open source software. This sentiment includes both the binary Docker images and the toolchain we use to build them.

We will be happy to revisit this issue if the CUDA license changes. Until then, users who wish to include CUDA in their Docker images will need to accept the license agreement and make their own builds.

That said, we appreciate the time you took to open this PR and raise the issue, @clkao. Please don't hesitate to do so again!

PR #541, opened 2018-01-31
Support provisioned storage with incorrect permissions

cam72cam:
I ran into an issue when trying to get this to work with a NFS server which I did not have direct control over (EFS).  As part of the PersistentVolumeClaim, there is no easy way to set the UID and GID of the created directory.on the networked FS.

My only concern with this chown is that some user out there might be running jupyterhub in an odd configuration where $NB_USER is not supposed to have these exact permissions on the storage.  I think this is quite unlikely, but it is worth mentioning. 

I chronicled my experiences with working around this issue and setting up z2jh on EFS in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/421 with @yuvipanda.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -20,7 +20,10 @@ if [ $(id -u) == 0 ] ; then
     
     # Handle case where provisioned storage does not have the correct permissions by default
     # Ex: default NFS/EFS (no auto-uid/gid)
-    chown $NB_UID:$NB_GID /home/$NB_USER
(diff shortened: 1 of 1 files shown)


minrk:
Changing permissions is known to cause problems for mounted volumes. Perhaps more appropriate is to do this if and only if the chosen user+group combo cannot write in their home directory. Or more simply, gate it behind an off-by-default flag, e.g  `CHOWN_HOME`.

cam72cam:
@minrk I took your suggestion and added a commit to gate it behind $CHOWN_HOME

minrk:
Thanks!

PR #544, opened 2018-02-03
Fixing PYTHONPATH in base-notebook/start.sh

Bidek56:
In order to fix the PYTHONPATH problem in pyspark-notebook, I have added PYTHONPATH to sudo command in start.sh


parente:
Thanks for the PR. The test failures are unrelated and being discussed in #545. I think this is fine to merge in the meantime.

Issue #549, opened 2018-02-15
Use integer $NB_UID and not $NB_USER for USER in Dockerfile.


GrahamDumpleton:
All the ``docker-stacks`` images currently use:

```
USER $NB_USER
```

to switch from ``root`` to non privileged ``jovyan`` user.

This is not ideal when it comes to platforms for deploying containers which enforce secure methods for deploying images in containers. This is because the platform may restrict users from deploying images that will run as ``root``. Because a user name is used with ``USER`` via ``$NB_USER``, the platform cannot verify that isn't ``root``, as the image could be faking it, by actually having a ``/etc/passwd`` file which maps ``jovyan`` to UNIX ID ``0``.

To allow for more options in using ``docker-stacks`` images with more secure platforms for deploying images in containers, the ``Dockerfile`` for any ``docker-stacks`` image should instead use:

```
USER $NB_UID
```

The consequence of this is that when you run ``docker inspect`` on an image, instead of seeing:

```
        "ContainerConfig": {
            "User": "jovyan",
```

you will see:

```
        "ContainerConfig": {
            "User": "1000",
```

Because the ``User`` in the image manifest now has a integer UNIX ID, the platform can trust that the image will not in fact run as ``root``.

Examples of what this will allow are as follows.

On Kubernetes and systems based on it, such as OpenShift, it is possible to create a service account in the platform with a restrictive Security Context Constraint (SCC) which allows a user to run images as non root user specified by the image:

```
NAME               PRIV      CAPS      SELINUX     RUNASUSER          FSGROUP     SUPGROUP    PRIORITY   READONLYROOTFS   VOLUMES
nonroot            false     []        MustRunAs   MustRunAsNonRoot   RunAsAny    RunAsAny    <none>     false            [configMap downwardAPI emptyDir persistentVolumeClaim projected secret]

```

This can only work where the ``User`` in the image manifest is an integer ID. When it isn't the SCC falls back to the default, which in a environment with full security means that the container is assigned a user ID, which then fails as the images aren't designed to run as an arbitrary user ID where group would be GID of 0.

The ability to allow a user to run with ``nonroot`` SCC is safer than granting them the ``anyuid`` SCC in Kubernetes, which allows them to run images as ``root`` as well.

The use of an integer ID for ``USER`` in ``Dockerfile``, also allows the image to be used as a Source-to-Image (S2I) builder in OpenShift as it will also block use of images which do not use an integer UNIX UID for ``USER``.

---

[...]

PR #550, opened 2018-02-15
Use NB_UID for USER statement in Dockerfile so deployment platform can verify that image doesn't run as root.

GrahamDumpleton:
For full details on the requirement for this patch see https://github.com/jupyter/docker-stacks/issues/549


parente:
Thanks for the explanation and PR, @GrahamDumpleton. I suspect the build failure is unrelated (a Julia compilation issue) but I restarted it for good measure.

PR #551, opened 2018-02-15
Templates for deploying Jupyter Notebooks to OpenShift.

GrahamDumpleton:
This is an examples directory which includes templates which can be used to deploy the ``docker-stacks`` images to OpenShift.

There are still some separate tweaks that could be made to the ``docker-stacks`` images to make things run better in Kubernetes/OpenShift environments with more strict security regime in place, but I am creating separate issues and PRs for them. These templates are not dependent on those other changes and will still work.


sherl0cks:
@GrahamDumpleton small thing, but just noticed that the delete command in the `README` does not delete the config map. perhaps because it does not have `app=${APPLICATION_NAME}` label?

GrahamDumpleton:
Okay, will fix that shortly after done with next PR working on for docker-stacks.

GrahamDumpleton:
Fixed label for config map so is deleted when using command given in README. Also added notes in README for how to change password.

Should be good to go again.

sherl0cks:
works for me. thanks @GrahamDumpleton

minrk:
Thanks!

parente:
@GrahamDumpleton Thanks. A small heads up: we're slowly moving toward a model where we maintain a core set of images here, and support community members in creating their own stack and examples in other repositories. #517 has some information,

GrahamDumpleton:
Thanks for the heads up, I have a whole separate project going on about creating customised notebook images where can optionally select kernels you want used at time of build, along with what language packages you also want installed. Wouldn't expect that to be picked up by main project, but may be of side interest to some people.

Issue #552, opened 2018-02-15
Lack of passwd/group file entries when run as arbitrarily assigned user ID.


GrahamDumpleton:
**What docker image you are using?**

All are affected by this issue.

**What complete docker command do you run to launch the container (omitting sensitive values)?**

Two commands to consider.

```
$ docker run -it -u 1000000 --group-add 100 jupyter/minimal-notebook:latest bash
```

and:

```
$ docker run -it -u 1000000 --group-add 100 --group-add 1000000 jupyter/minimal-notebook:latest bash
```

The examples are where images are run as arbitrarily assigned user ID. This scenario can arise in container deployment platforms which enforce a security model where images can't run as ``root`` or even the user the image specifies. This is the case with Kubernetes (and OpenShift), where Role Base Access Control (RBAC) and Security Context Constraints (SCC) are enabled to enforce a secure multitenant deployment environment.

Note that the images will not actualy work out of the box this way because the group they run as will be GID 0. To compensate for this, you need to add GID 100 as supplemental group for the container. A platform might also itself add the assigned user ID as a supplemental group as well.

**What steps do you take once the container is running to reproduce the issue?**

Once shell is running the container, run the commands:

```
id
whoami
groups
```

In the first case the result will be:

```
I have no name!@3dffcb336b17:~$ id
uid=1000000 gid=0(root) groups=0(root),100(users)
I have no name!@3dffcb336b17:~$ whoami
whoami: cannot find name for user ID 1000000
I have no name!@3dffcb336b17:~$ groups
root users
```

In the second it will be:

```
I have no name!@bf6a70260f15:~$ id
uid=1000000 gid=0(root) groups=0(root),100(users),1000000
I have no name!@bf6a70260f15:~$ whoami
whoami: cannot find name for user ID 1000000
I have no name!@bf6a70260f15:~$ groups
root users groups: cannot find name for group ID 1000000
1000000
```

**What do you expect to happen?**

No errors.

**What actually happens?**

Errors.

The reason this is a problem is some Python packages out there aren't tolerant of the user they are running as not having passwd and group file entries. An example of typical exception that might occur is:

```
Traceback (most recent call last):
  File "/opt/app-root/bin/jupyterhub", line 3, in <module>
    from jupyterhub.app import main
  File "/opt/app-root/lib/python3.6/site-packages/jupyterhub/app.py", line 118, in <module>
    class NewToken(Application):
[...]

PR #553, opened 2018-02-15
Add passwd/group file entries when running images as assigned user ID not in passwd file.

GrahamDumpleton:
Details of the reason for this change in https://github.com/jupyter/docker-stacks/issues/552

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -47,9 +47,12 @@ ENV PATH=$CONDA_DIR/bin:$PATH \
 ADD fix-permissions /usr/local/bin/fix-permissions
 # Create jovyan user with UID=1000 and in the 'users' group
 # and make sure these dirs are writable by the `users` group.
+# Make /etc/passwd and /etc/group writable so can add entries
+# for uid/gid when container run as arbitrarily assigned uid.
 RUN useradd -m -s /bin/bash -N -u $NB_UID $NB_USER && \
     mkdir -p $CONDA_DIR && \
     chown $NB_USER:$NB_GID $CONDA_DIR && \
+    chmod g+w /etc/passwd /etc/group && \
     fix-permissions $HOME && \
     fix-permissions $CONDA_DIR
 
@@ -90,13 +93,15 @@ EXPOSE 8888
 WORKDIR $HOME
 
 # Configure container startup
-ENTRYPOINT ["tini", "--"]
+ENTRYPOINT ["container-entrypoint"]
 CMD ["start-notebook.sh"]
 
 # Add local files as late as possible to avoid cache busting
 COPY start.sh /usr/local/bin/
 COPY start-notebook.sh /usr/local/bin/
 COPY start-singleuser.sh /usr/local/bin/
+COPY container-entrypoint /usr/local/bin/
+COPY fix-passwd-entry /usr/local/bin/
 COPY jupyter_notebook_config.py /etc/jupyter/
 RUN fix-permissions /etc/jupyter/
 
--- a/base-notebook/container-entrypoint
+++ b/base-notebook/container-entrypoint
@@ -1,0 +1,15 @@
+#!/bin/bash
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+# Add entries to /etc/passwd and /etc/group if being run as user ID
+# other than which the image has specified by the USER statement.
+
+fix-passwd-entry
+
+# Run 'tini' as a mini supervisor. This will manage the actual
+# application process, passing on signals received by the container, and
+# also reaping zombie processes. Must use 'exec' so that it inherits
+# process ID 1 of the container.
+
+exec tini -- "$@"
--- a/base-notebook/fix-passwd-entry
+++ b/base-notebook/fix-passwd-entry
@@ -1,0 +1,22 @@
+#!/bin/bash
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+# By default the image would run as user ID 1000. If the user is run as
+# an arbitrarily assigned user ID by the container platform it will be
+# much higher and the lack of entries in /etc/passwd and /etc/group can
+# cause failure of any Python code which doesn't tolerate having no
+# entries for uid/gid. Assume that if running with larger uid that we
+# need to add entries into the respective files for that user. Have the
+# check be greater than 100000 to allow room in case custom images
+# derived from this image were adding additional explict users/groups of
+# their own.
+
+if [ `id -u` -ge 100000 ]; then
+    cat /etc/passwd | sed -e "s/^jovyan:/nayvoj:/" > /tmp/passwd
+    echo "jovyan:x:`id -u`:`id -g`:,,,:/home/jovyan:/bin/bash" >> /tmp/passwd
+    cat /tmp/passwd > /etc/passwd
+    rm /tmp/passwd
+
+    echo "jovyan:x:`id -u`:" >> /etc/group
+fi
--- a/base-notebook/fix-passwd-entry
+++ b/base-notebook/fix-passwd-entry
@@ -12,11 +12,17 @@
 # derived from this image were adding additional explict users/groups of
 # their own.
 
-if [ `id -u` -ge 100000 ]; then
+NB_UID=`id -u`
+NB_GID=`id -g`
+
+if [ $NB_UID -ge 100000 ]; then
     cat /etc/passwd | sed -e "s/^jovyan:/nayvoj:/" > /tmp/passwd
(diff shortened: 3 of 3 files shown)


GrahamDumpleton:
Forgot to modify ``Dockerfile.ppc64le``. Further update to come.

GrahamDumpleton:
Should be noted that ``Dockerfile.ppc64le`` doesn't seem to have been tracking ``Dockerfile`` for other changes. So looks like it is out of date and missing things.

base-notebook/Dockerfile
```diff
@@ -47,9 +47,12 @@ ENV PATH=$CONDA_DIR/bin:$PATH \
 ADD fix-permissions /usr/local/bin/fix-permissions
 # Create jovyan user with UID=1000 and in the 'users' group
 # and make sure these dirs are writable by the `users` group.
+# Make /etc/passwd and /etc/group writable so can add entries
+# for uid/gid when container run as arbitrarily assigned uid.
 RUN useradd -m -s /bin/bash -N -u $NB_UID $NB_USER && \
     mkdir -p $CONDA_DIR && \
     chown $NB_USER:$NB_GID $CONDA_DIR && \
+    chmod g+w /etc/passwd /etc/group && \
```
minrk: This makes it trivial for any user to become root in their container:

```bash
cp /etc/passwd /tmp/passwd
rootpass=`openssl passwd -1 root`
cat /tmp/passwd | sed "s/root:x/root:${rootpass}/" > /etc/passwd
```

Now the user can `su root` with the password 'root'

I think we need another solution that doesn't grant universal `su root` access.

This does only affect putting users into the root group, so not the default case, but it does mean that putting users into the root group is equivalent to letting the users run as root. Is that the intended effect?

GrahamDumpleton: I will check, but I don't believe that will work in secure environments like Kubernetes/OpenShift where you are blocked from running setuid executables and various related system calls behind the ability to change real user/group.

GrahamDumpleton: Yep, this is restricted in Kubernetes/OpenShift.

```
jovyan@image:~$ su root
Password:
initgroups: Operation not permitted
```

This is because secure container environments will drop capabilities.

```
                "securityContext": {
                    "capabilities": {
                        "drop": [
                            "KILL",
                            "MKNOD",
                            "SETGID",
                            "SETUID"
                        ]
                    },
                    "privileged": false,
                    "runAsUser": 1000030000,
                    "seLinuxOptions": {
                        "level": "s0:c6,c0"
                    }
                },
```

parente:
Re: Dockerfile.ppc64le, I think it's safe to leave that be for now. It's a good example of items we're not able to effectively maintain in this repo that should probably reside in another repository.

base-notebook/container-entrypoint
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+# Add entries to /etc/passwd and /etc/group if being run as user ID
+# other than which the image has specified by the USER statement.
+
+fix-passwd-entry
```
parente: Is it possible for this logic to live in the `start*` scripts after tini runs instead of introducing additional scripts? I understand that Python and other programs may balk when the UID is not found in `/etc/users` and `/etc/group`, but does tini have that problem?

Simply looking to keep the number of separate workflow scripts we need to maintain low.

GrahamDumpleton: The reason for adding it in an ``ENTRYPOINT`` script and doing it in there is that it gets applied even if someone overrides the ``CMD`` to do something else. For example, in JupyterHub, the ``CMD`` for the notebook images is overridden to execute ``jupyterhub-singleuser`` directly, bypassing ``start.sh``.

GrahamDumpleton:
Note, I will make one more change to this. I will add a check that doesn't do anything if ``id -g`` is not ``0``. This is because need to have group gid of ``0`` to update files. This only occurs when image run with UID not in passwd file to begin with. So further constrain this to narrow use case where required.

GrahamDumpleton:
Have made change to check running as GID of 0 since relying on group being that to write passwd and group files. So should all be good to go.

Summarising what we have.

* Add ``container-entrypoint`` script from which can make passwd/group changes because need to be done even if someone overrides ``CMD`` such as in JupyterHub.
* Add separate ``fix-passwd-entry`` script instead of having it on ``container-entrypoint`` so can be executed by someone in custom container entrypoint script of derived image if necessary.
* Make passwd/group files writable by root group. In normal use cases they still aren't able to be updated. Only time would be able to be updated is specific case where UID not in passwd file is used and so falls back to using GID of 0 as a result, such as is case in multitenant Kubernetes/OpenShift.
* Although could technically do whenever GID is 0, only do it if UID greater than 100000. Kubernetes/OpenShift actually use user IDs 1000000000 and above, which is well above that value. Not touching things when less than 100000 avoids potential that people with derived images already would be affected.
* Add the new entry in passwd file for arbitrary user ID being used which isn't in passwd file.
* Map original 1000 user ID of jovyan to different name (nayvoj - jovyan backwards), so those user IDs still show a name.
* In specific case where supplemental group ID same as user ID is added, add an entry in group for that as well.

As far as is known there is no risk with making passwd/group files writable to root and doesn't provide any ability for use to get elevated privileges.

FWIW, the only other way this whole thing could be solved is by by fiddling around with shared libraries and forced preloading of them into processes. This is talked about in:

* http://blog.dscpl.com.au/2015/12/unknown-user-when-running-docker.html

That other methods has lots of problems, including that nss_wrapper isn't packaged in stable for Debian/Ubuntu and that ``LD_PRELOAD`` and other environment variables will not be set if someone uses ``docker exec`` to enter container, or if application strips environment variables when running sub processes etc. So right now the writable passwd/group files and doing it from container entrypoint has been found to be best option.

base-notebook/container-entrypoint
minrk: This will depend on the Spawner. It is true of KubeSpawner, but isn't true of DockerSpawner in general (deployments *may* override CMD, but it is not overridden by default).

Doing this as an entrypoint makes sense to me. I think the same might be said for much of what's in start.sh, though. Effectively, putting this and only this in entrypoint is saying that the passwd entry is more important than and should be harder to opt-out of than anything in start.sh. That may be right, but I think we should think carefully about this in the future.

base-notebook/fix-passwd-entry
```diff
@@ -0,0 +1,28 @@
...
+if [ $NB_GID -eq 0 -a $NB_UID -ge 100000 ]; then
+    cat /etc/passwd | sed -e "s/^jovyan:/nayvoj:/" > /tmp/passwd
+    echo "jovyan:x:$NB_UID:$NB_GID:,,,:/home/jovyan:/bin/bash" >> /tmp/passwd
+    cat /tmp/passwd > /etc/passwd
+    rm /tmp/passwd
+
+    id -G | grep -q -w $NB_UID; STATUS=$?
+    if [ $STATUS -eq 0 ]; then
+        echo "jovyan:x:$NB_UID:" >> /etc/group
+    fi
```
minrk: We have similar logic to this in start.sh, and it currently enables:

- setting user id
- setting user name
- granting sudo access

The start.sh logic requires launching the image as the root *user* rather than the root group. Effectively, we have a fork of two different behaviors with the same goal - one dedicated for openshift/some kubernetes deployments (this one) and one for most others (in start.sh). They accomplish the same thing in different ways and in different code paths. Perhaps we should reconcile them. It would probably be nice if we could accomplish the existing root-requiring logic in start.sh via root *group* rather than root *user*. Would you like to take a stab at that?

GrahamDumpleton: The problem still gets back to fact that JupyterHub only runs ``jupyterhub-singleuser`` and not ``start-singleuser.sh``. Because of my use case for JupyterHub that isn't a complete stumbling block though as it is already necessary to configure ``KubeSpawner`` to set supplemental group of ``users`` on the container since ``docker-stacks`` images aren't setup to use group ``root`` for everything, but rely on groups ``users`` when user ID is not 1000. Since already doing that, no drama to also override the startup command to be ``start-singleuser.sh``. Since it internally uses ``start.sh`` to run ``jupyterhub-singleuser`` then integrating it with ``start.sh`` is probably okay.

I can't see that this will change the need to make ``passwd`` and ``group`` files writable to ``root`` group. The only way around that is messy and requires compiled C programs that are ``setuid`` so can become effective user of ``root``, but not real user as kernel setuid() call blocked, such that can update files in that C program. Am not confident that isn't blocked by some configurations for running containers.

Even if make ``passwd`` and ``group`` writable, my recollection is that commands like ``usermod`` and ``groupmod`` don't work as that change is not enough. When I tested that it may have been on CentOS though so possible that Debian is different. So chance that have to still make changes to ``passwd`` and ``group`` manually. I will do some more testing and see what can be done using ``start.sh``.

PR #556, opened 2018-02-17
improve secure_path modification by storing in sudoers.d

avinson:
changes to /etc/sudoers can be overwritten by later changes or package installs. it is preferable to place this config change into /etc/sudoers.d


parente:
Thanks @avinson

PR #557, opened 2018-02-19
Sample builder scripts for build custom notebook images using Source-to-Image.

GrahamDumpleton:
This example directory contains sample builder scripts for being able to build custom notebook images using the Source-to-Image (S2I) tool. Details on the S2I tool can be found at:

* https://github.com/openshift/source-to-image

The S2I tool makes it easy to combine notebooks, data files etc., from a Git repository or local directory, with a base image. In this case the examples are using the ``docker-stacks`` images. As well as combine notebooks and data files with an image, the build scripts can also automatically install additional Python packages required by the notebooks during the image build.

The S2I tool and sample build scripts here provide similar capabilities to ``mybinder.org`` but without the complexity of having to setup any special infrastructure.  Instead you can run builds on your own local machine.

Templates are also provided to allow use of S2I support integrated into OpenShift.

This examples directory may be of interest in discussion under https://github.com/jupyter/docker-stacks/issues/517

### DIFF
--- a/examples/source-to-image/README.md
+++ b/examples/source-to-image/README.md
@@ -1,0 +1,159 @@
+This example provides scripts for building custom Jupyter Notebook images containing notebooks, data files, and with Python packages required by the notebooks already installed. The scripts provided work with the Source-to-Image tool and you can create the images from the command line on your own computer. It provides similar capabilities to ``mybinder.org`` but without the complexity of having to setup any special infrastructure. Templates are also provided to enable running builds in OpenShift, as well as deploying the resulting image to make it available.
+
+Source-to-Image Project
+-----------------------
+
+Source-to-Image (S2I) is an open source project which provides a tool for creating container images. It works by taking a base image, injecting additional source code or files into a running container created from the base image, and running a builder script in the container to process the source code or files to prepare the new image.
+
+Details on the S2I tool, and executable binaries for Linux, macOS and Windows, can be found on GitHub at:
+
+* https://github.com/openshift/source-to-image
+
+The tool is standalone, and can be used on any system which provides a docker daemon for running containers. To provide an end-to-end capability to build and deploy applications in containers, support for S2I is also integrated into container platforms such as OpenShift.
+
+Getting Started with S2I
+------------------------
+
+As an example of how S2I can be used to create a custom image with a bundled set of notebooks, run:
+
+```
+s2i build \
+  --scripts-url https://raw.githubusercontent.com/jupyter/docker-stacks/source-to-image/examples/source-to-image \
+  --context-dir docs/source/examples/Notebook \
+  https://github.com/jupyter/notebook \
+  jupyter/minimal-notebook:latest \
+  notebook-examples
+```
+
+This example command will pull down the Git repository ``https://github.com/jupyter/notebook`` and build the image ``notebook-examples`` using the files contained in the ``docs/source/examples/Notebook`` directory of that Git repository. The base image which the files will be combined with is ``jupyter/minimal-notebook:latest``, but you can specify any of the Jupyter Project ``docker-stacks`` images as the base image.
+
+The resulting image from running the command can be seen by running ``docker images``.
+
+```
+REPOSITORY         TAG     IMAGE ID      CREATED        SIZE
+notebook-examples  latest  f5899ed1241d  2 minutes ago  2.59GB
+```
+
+You can now run the image.
+
+```
+$ docker run --rm -p 8888:8888 notebook-examples
+Executing the command: jupyter notebook
+[I 01:14:50.532 NotebookApp] Writing notebook server cookie secret to /home/jovyan/.local/share/jupyter/runtime/notebook_cookie_secret
+[W 01:14:50.724 NotebookApp] WARNING: The notebook server is listening on all IP addresses and not using encryption. This is not recommended.
+[I 01:14:50.747 NotebookApp] JupyterLab beta preview extension loaded from /opt/conda/lib/python3.6/site-packages/jupyterlab
+[I 01:14:50.747 NotebookApp] JupyterLab application directory is /opt/conda/share/jupyter/lab
+[I 01:14:50.754 NotebookApp] Serving notebooks from local directory: /home/jovyan
+[I 01:14:50.754 NotebookApp] 0 active kernels
+[I 01:14:50.754 NotebookApp] The Jupyter Notebook is running at:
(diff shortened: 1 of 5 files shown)


examples/source-to-image/assemble
```diff
@@ -0,0 +1,42 @@
...
+rm -rf /tmp/src
+
+# Install any Python modules. If we find an 'environment.yml' file we
+# assume we should use 'conda' to install packages. If 'requirements.txt'
+# use 'pip' instead. Ensure we are in the 'work' directory so relative
+# directory paths in these files resolve okay if installing packages
+# from subdirectories.
+
+if [ -f /home/$NB_USER/work/environment.yml ]; then
+    (cd /home/$NB_USER/work && conda env update --name root --file environment.yml)
```
minrk: without spawning a shell, this can `conda env update -f /path/to/environment.yml`. Same below for pip.

examples/source-to-image/README.md
```diff
@@ -0,0 +1,159 @@
+This example provides scripts for building custom Jupyter Notebook images containing notebooks, data files, and with Python packages required by the notebooks already installed. The scripts provided work with the Source-to-Image tool and you can create the images from the command line on your own computer. It provides similar capabilities to ``mybinder.org`` but without the complexity of having to setup any special infrastructure. Templates are also provided to enable running builds in OpenShift, as well as deploying the resulting image to make it available.
```
minrk: I don't think this jab at mybinder.org is useful or accurate. What it's more comparable to is one *piece* of Binder, repo2docker, which also requires nothing but python and docker.

examples/source-to-image/assemble
```diff
@@ -0,0 +1,42 @@
...
+# an S2I build was used to create a new base image with pre-installed
+# Python packages, with the new image then subsequently being used as a
+# S2I builder base image.
+
+rm -f /home/$NB_USER/work/environment.yml
+rm -f /home/$NB_USER/work/requirements.txt
+
+# Copy injected files to correct place in 'work' directory.
+
+cp -Rf /tmp/src/. /home/$NB_USER/work
```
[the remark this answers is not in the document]
GrahamDumpleton: Okay, I thought it must have been some existing convention that the work directory existed and that people were expected to always do stuff in it for some reason. I could see that as reasonable as mounting onto ``$HOME`` actually causes problems where as part of the image build things were created/cached in ``$HOME`` that would be required later. Mounting on ``$HOME`` would hide any such cached files and potentially break things or cause things to be regenerated. As an example, mounting on ``$HOME`` would hide ``~/.cache/matplotlib`` generated during the build of ``scipy-notebook``, meaning the font cache index (???) has to be regenerated first time ``matplotlib`` is imported in a notebook.

Having only ``$HOME`` to work with is a bit problematic in more general case of S2I because S2I builders may want to generate special configuration and startup scripts. You cannot put these in the directory which is likely going to be what people would mount a volume on as then it all stops working when you deploy the image. Because S2I doesn't run as ``root``, you don't really have any ability to create directories anywhere else really you can put stuff. The more typical scenario in S2I builders is that you have ``/opt/app-root`` under which the S2I builder can put stuff, not a user home directory, and just ``/opt/app-root/src`` is ``$HOME`` for the user account, and is what is used by the application. Thus S2I builders can happily write extra stuff under ``/opt/app-root`` and know that things will not be broken if single subdirectory of ``/opt/app-root/src`` underneath is used as mount point for volume.

If am to ignore the ``work`` directory, and possibly even start removing it since no need for it these examples, then means I have to generate some configuration files elsewhere, perhaps under ``/etc/jupyter`` since that is writable to ``users`` group.

examples/source-to-image/assemble
GrahamDumpleton: So changing sub directory in sub shell may not be required, never actually checked whether does the right thing and to be safe would always ensure in the same directory as source code and that file. If don't do things in ``work`` directory but ``$HOME`` then becomes moot as ``$HOME`` will be the current working directory anyway.

The ``--name root`` option is used because I have found that if don't have that and ``environment.yml`` has ``name`` attribute for environment, then it creates a separate environment under ``/opt/conda/envs``. I assumed this would not be want you want, because that environment isn't going to be activated and things would run in the root environment still, meaning installed packages would not be found. Thus to be safe was overriding things so always installed into the root environment.

GrahamDumpleton:
As far as examples go I have no issues providing support for them. I will also accept if you decide you don't want the examples after all and also remove the other example I provided. I will maintain them in a separate repository.

GrahamDumpleton:
If I move both examples to separate repositories, can I still at least have an examples subdirectory for each with a ``README`` file in them which some details and links off to the separate repositories or docs site?

examples/source-to-image/assemble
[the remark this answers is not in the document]
minrk: We used to put required configuration, kernelspecs, etc. in $HOME and that caused us to call 'work' the place we wouldn't touch where users could put their files and could be safely mounted as a volume.  We've since fixed that, and by design no files in $HOME may be required for any of these images to work. In JupyterHub, the current convention is to mount $HOME as a volume, because then there's no confusion about users losing data because they forgot to put it in their 'persistent' directory.

The only data in $HOME in any of these images is the matplotlib font cache, which is a small performance optimization for the first import of matplotlib. If $HOME is mounted as a volume, the cache will be created as normal for any user and re-used thereafter. So it saves an every-time cost for the no-persistence case and persistent storage saves the same cost when there is persistent storage, after the first launch for that user.

The exception is Binder, where we put the repo data is in $HOME specifically because it's an ephemeral container, and we don't need to persist user data. The same makes sense if s2i is being used for ephemeral containers, but doesn't if the users should be saving their work in $HOME. If s2i should be writing persistent data for the home-as-volume case, *nothing* should be written to $HOME. Writing to /opt or /etc makes sense. Whatever you like.

examples/source-to-image/assemble
minrk: Yes, `--name` is required. Sorry, I didn't mean that my line was complete, just that you could replace cd in a sub-shell with an absolute path in the file argument. The full line:

    conda env update -n root -f /path/to/environment.yml

will work no matter what the cwd.

minrk:
I'm very happy to have the new examples here. Thanks for supporting them!

GrahamDumpleton:
Have tweaked things now to do things directly in home directory and ensure any configuration introduced by OpenShift is not mounted in home directory so the home directory can be mounted over by normal persistent volume.

minrk:
Looks great, thanks!

PR #562, opened 2018-02-22
Update tidyverse to 1.2 in datascience-notebook

betatim:
This moves the tidyverse version forwards to 1.2.

We'd like to use this version for a jupyterhub deployment for http://openhumans.org/. Using this docker image and contributing to its upkeep is much easier than making a derivative.

cc @gedankenstuecke for questions about R version compatibility etc

### DIFF
--- a/datascience-notebook/Dockerfile
+++ b/datascience-notebook/Dockerfile
@@ -46,7 +46,7 @@ RUN conda config --system --append channels r && \
     'r-irkernel=0.7*' \
     'r-plyr=1.8*' \
     'r-devtools=1.12*' \
-    'r-tidyverse=1.0*' \
+    'r-tidyverse=1.2*' \
     'r-shiny=0.14*' \
     'r-rmarkdown=1.2*' \
     'r-forecast=7.3*' \


gedankenstuecke:
Looks good to me, though looking at the Dockerfile it seems that many of the packages (including R itself which is 3.3 instead of 3.5) are a bit dated by now. So maybe it would be nice to bump all of them to the latest versions?

I don't know whether there are any conflicts on other users' ends that would prohibit bumping most of them.

parente:
Build failed with:

```
PackageNotFoundError: Packages missing in current channels:
            
  - r-tidyverse 1.2*
```

Looks like https://github.com/conda-forge/r-tidyverse-feedstock/blob/master/recipe/meta.yaml needs a bump first.

betatim:
Ok, I will go and chase up some conda-forge packages. Should I start down that path with the aim of updating several of the R packages used here and is there someone who cares about them that I should talk to?

parente:
Thanks @betatim. Feel free to update any of the R package versions in the Dockerfiles here. If anyone is dependent on the current versions, they can stick with a tag that does not include your updates.

betatim:
Jupp, that is fine with me. I don't really use R a lot so I'd just be setting things to "the latest version", so if someone with more R experience can take over that would be great.

PR #567, opened 2018-03-02
Add instructions for enabling Jupyter Lab interface with OpenShift deployment.

GrahamDumpleton:
This is a change to README for OpenShift example which shows how the Jupyter Lab interface can be enabled after the notebook has been deployed. There is no template parameter to set it at this point, but could consider adding it later so can optionally enable it at time of first deployment.


Bidek56:
It failed because of max time: "The job exceeded the maximum time limit for jobs, and has been terminated."

parente:
The build failure is related to the max build time allowed on Travis. I've opened #577 so we can address that problem separately.

PR #570, opened 2018-03-06
Update to Toree 0.2.0 RC3

lresende:


parente:
I think the build failure is related to the max build time allowed on Travis. I've opened #577 so we can address that problem separately.

PR #571, opened 2018-03-06
jupyter_lab

tarekbadrshalaan:
just image to run jupyter and jupyter lab


parente:
Hi @tarekbadrshalaan. Thanks for sharing this Dockerfile. At the moment, we're tapped out on how many images we can maintain here and (slowly) working on documentation about how users can maintain images in their own repositories. Please feel free to join the discussion on #517.

Issue #532, opened 2018-01-13
R package hexbin for r-notebook image


terhorst:
The R package `hexbin` (https://cran.r-project.org/web/packages/hexbin) is required to use `geom_hex` in ggplot, however it does not come with the r-notebook image.

parente:
@terhorst Thanks for reporting this issue. If you'd like to add `r-hexbin` to the list of packages conda installed in r-notebook, I'll be happy to review and merge that PR.

PR #573, opened 2018-03-09
add r-htmltools pkg

terhorst:
This is another conda package which is required for some tidyverse functions (specifically, `stringr::str_view`) but not included in the `r-tidyverse` metapackage. (cf #532)


parente:
The build failure is related to the max build time allowed on Travis. I've opened #577 so we can address that problem separately.

Issue #575, opened 2018-03-09
When using a volume, the work dir belongs to root


ewjoachim:
**What docker image you are using?**

`jupyter/scipy-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

`docker run -it --rm -p 127.0.0.1:8888:8888 --user=root -e NB_UID=$UID -e NB_GID=$GID -e CHOWN_HOME=1 -v "$pwd/notebooks:/home/jovyan/work" --name notebook jupyter/scipy-notebook`

**What steps do you take once the container is running to reproduce the issue?**

Example:

1. Visit http://localhost:8888
2. Try creating a notebook in work/

**What do you expect to happen?**

Example: Notebook is created

**What actually happens?**

![screenshot from 2018-03-09 18-51-03](https://user-images.githubusercontent.com/1457576/37221601-d9ba4a8a-23ca-11e8-8a82-24876588e6f8.png)

![screenshot from 2018-03-09 18-51-47](https://user-images.githubusercontent.com/1457576/37221635-f13dd212-23ca-11e8-8b71-2557150e346b.png)

Output of start.sh before the notebook starts:
```
Set username to: jovyan
usermod: no changes
Changing ownership of /home/jovyan to 1000:1000
Set jovyan GID to: 1000
Executing the command: jupyter notebook
```

Given that [these lines](https://github.com/jupyter/docker-stacks/blob/master/base-notebook/start.sh#L24-L25) are executed, I don't see how work could still belong to root, but it is so...?

ewjoachim:
Actually, chown is not -R, that's why it doesn't work. I've made a PR :)

ewjoachim:
Should be closed now. Let's reopen if a new problem arises :)

cam72cam:
This needs to be re-opened

ewjoachim:
What happened ? Do the steps outlined above lead to the same bug ? After it was merged, I think it worked for me...

PR #576, opened 2018-03-09
Chown of home dir needs to be recursive

ewjoachim:
Refs #575


ewjoachim:
Hm, the CI was not passing, I hope it's not because of my work...?

parente:
@ewjoachim no worries. The issue is because of #577. Thanks for the PR.

ewjoachim:
ok, nice. Thank you :)

BTW, you (as in all the contritors) did an awesome job with these images. Thank you and congrats !

PR #579, opened 2018-03-12
Add pyarrow installation

gglanzani:
This way we can profit from pandas UDF

Please advise on the best place to install pyarrow. It's now installed in the inner layer (of this image), together with java.

### DIFF
--- a/pyspark-notebook/Dockerfile
+++ b/pyspark-notebook/Dockerfile
@@ -14,6 +14,12 @@ RUN apt-get -y update && \
     apt-get install --no-install-recommends -y openjdk-8-jre-headless ca-certificates-java && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
+
+# Install pyarrow
+RUN conda install --quiet -y 'pyarrow' && \
+    fix-permissions $CONDA_DIR && \
+    fix-permissions /home/$NB_USER
+
 RUN cd /tmp && \
         wget -q http://apache.claz.org/spark/spark-${APACHE_SPARK_VERSION}/spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz && \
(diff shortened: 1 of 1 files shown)


pyspark-notebook/Dockerfile
```diff
@@ -13,7 +13,8 @@ ENV HADOOP_VERSION 2.7
 RUN apt-get -y update && \
     apt-get install --no-install-recommends -y openjdk-8-jre-headless ca-certificates-java && \
     apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+    rm -rf /var/lib/apt/lists/* && \
+    conda install -y pyarrow
```
parente: I think it's reasonable to install pyarrow in the pyspark image. If there's already a section that installs packages using `conda`, please add `pyarrow` there. If there, isn't it's best to start one since we need to run some cleanup and a script to fix permissions as part of the same line. https://github.com/jupyter/docker-stacks/blob/master/scipy-notebook/Dockerfile#L20 can serve as a guide.

gglanzani: @parente Good point, I've updated the PR.

pyspark-notebook/Dockerfile
```diff
@@ -14,6 +14,15 @@ RUN apt-get -y update && \
     apt-get install --no-install-recommends -y openjdk-8-jre-headless ca-certificates-java && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
+
+# Install pyarrow
+RUN conda install --quiet -y 'pyarrow' && \
+    rm -rf $CONDA_DIR/share/jupyter/lab/staging && \
+    rm -rf /home/$NB_USER/.cache/yarn && \
+    rm -rf /home/$NB_USER/.node-gyp && \
```
parente: It's safe to remove the three `rm -rf` lines above. In the scipy image, we install some nodejs dependencies for jupyterlab and need to clean up working directories. pyarrow should be a pure Python package.

After this change, it LGTM.

gglanzani:
Hi @parente,

thanks for the review. The changes have been pushed.

PR #580, opened 2018-03-13
Add jupyterlab_bokeh extension

gglanzani:
Without the extension it is not possible to use all Bokeh functionalities in JupyterLab


scipy-notebook/Dockerfile
```diff
@@ -48,6 +48,7 @@ RUN conda install --quiet --yes \
     jupyter nbextension enable --py widgetsnbextension --sys-prefix && \
     # Also activate ipywidgets extension for JupyterLab
     jupyter labextension install @jupyter-widgets/jupyterlab-manager@^0.33.1 && \
+    jupyter labextension install jupyterlab_bokeh && \
```
parente: Can you please pin this to `@^0.4.0` so that when the extension updates to be compatible with a a newer jupyterlab, builds don't break here until we're ready to update all the jupyterlab components together?

gglanzani: @parente Pinned

parente:
LGTM now. Thanks!

Issue #577, opened 2018-03-09
Shorten Travis CI build process


parente:
We're running up against the max build time allowed. Ideas for shortening so the tests can run to completion again:

* Be smarter about what images have changed and what to build based on the git diff
* Remove some images
* Remove some libraries from the images
* Study where the build time is spent and focus on that piece

taqtiqa-mark:
AFAICT you'll forever be snookered by this issue without breaking the build steps out of the make files, to take advantage of the parallel build functionality.

The extra benefit of doing this is that same parallel structure can be used in Shippable - this would give you some redundnacy in the build infrastructure.

parente:
Dealing with the partial dependency tree has made me shy away from exploring parallelization options and favoring tricks to shorten the build where I can. 

For example, when someone makes a change to base-notebook, we need to build/test minimal-notebook next before any other image. After minimal is built, we can parallelize r-notebook and scipy-notebook. Once scipy is done, we can then parallelize datascience-notebook, tensorflow-notebook, and pyspark-notebook. Finally, once  pyspark is built, we can build all-spark-notebook.

Do you have experience setting up such a flow in Travis (or elsewhere)? Have an estimate of what it would take to do so here? I'd love to hear about it if you do!

taqtiqa-mark:
@parente I'm currently building a port of the jupyter docker containers to rkt [over here](https://github.com/taqtiqa-admin/rkt-stacks).  The [.travis.yml](https://github.com/taqtiqa-admin/rkt-stacks/blob/master/.travis.yml) isn't the most elegant but you'll see the idea in matrix.

Travis runs 5 jobs at the same time.
The first 'job' builds the base and minimal notebooks.
The rest of the jobs build only one container.

Apologies for the red state I'm in the middle of switch from `rkt fetsh + rkt image export` to `skopeo + docker2aci`  that the halve the current image sizes back closer to what you have with docker.

This is all temporary since - time permitting - I'd like to build the oci containers with Buildah and Alpine Linux.  At the moment I have the `base-notebook` built this way, just need to push the code - let me know if your interested in seeing the wet paint.

parente:
[...]

PR #581, opened 2018-03-13
Disable datascience-notebook precompilation

parente:
I'm thinking the Julia build might be the long pole here. Pushing this just to test how it changes the build time.

Ref: #577

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,8 @@ ALL_STACKS:=base-notebook \
 	minimal-notebook \
 	r-notebook \
 	scipy-notebook \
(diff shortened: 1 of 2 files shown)


parente:
`Ran for 26 min 46 sec`. I suspect the datascience-notebook image with Python, R, and Juilia accounts for the majority of the time removed from the build.

parente:
Re-enabling the builds and disabling just the Julia `using` commands to avoid pre-compilation brings the full build in at around 45 minutes.

It's the install of the Julia hdf5 package that seems to make up nearly half the build time.

PR #596, opened 2018-03-26
Run additional scripts in /usr/local/bin/start-notebook.d/

clkao:
Actual use cases can be:
- impose more restrictive ulimit
- cleanup user trash directory
- whitelist env variables (eg kubernetes service endpoint info removal)

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -11,6 +11,22 @@ else
     cmd=$*
 fi
 
+for f in /usr/local/bin/start-notebook.d/*; do
+  case "$f" in
+    *.sh)
+      echo "$0: running $f"; . "$f"
+      ;;
+    *)
+      if [ -x $f ]; then
+        echo "$0: running $f"
+        $f
+      else
+        echo "$0: ignoring $f"
+      fi
+      ;;
+  esac
+  echo
+done
 # Handle special flags if we're root
 if [ $(id -u) == 0 ] ; then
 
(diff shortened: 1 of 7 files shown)


base-notebook/start.sh
```diff
@@ -111,6 +111,13 @@ else
     if [[ "$GRANT_SUDO" == "1" || "$GRANT_SUDO" == 'yes' ]]; then
         echo 'Container must be run as root to grant sudo permissions'
     fi
+    for f in /usr/local/bin/start-notebook.d/*; do
+      case "$f" in
+        *.sh)     echo "$0: running $f"; . "$f" ;;
+        *)        echo "$0: ignoring $f" ;;
```
minrk: Shouldm't we run all executables in this directory, regardless of extension? Or is it explicitly required to be 'sourced' scripts?

clkao: Right, for my use case calling `ulimit` it needs to be dot-sourced. I don't mind also running everything else if people think that's useful.

base-notebook/start.sh
```diff
@@ -111,6 +111,13 @@ else
...
+      esac
+      echo
+    done
```
parente: Do we want this logic to execute if the user runs the container with `--user root` too? As written, it only runs in the `else` block of the surrounding check for the container user ID.

clkao: good point. I think this should be moved to the top out of the `if` doing uid check.  With that we can gradually refactor this giant if block into smaller scripts.

minrk:
Nice!

PR #602, opened 2018-04-02
Use conda-forge's blas package w/openblas

jakirkham:
As conda-forge does not use the `nomkl` package, but instead uses the `blas` package to select BLAS versions, install the `blas` package and select the `openblas` implementation.

xref: https://github.com/jupyter/docker-stacks/pull/296

### DIFF
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -18,7 +18,7 @@ USER $NB_UID
 # Remove pyqt and qt pulled in for matplotlib since we're only ever going to
 # use notebook-friendly backends in these images
 RUN conda install --quiet --yes \
-    'nomkl' \
+    'blas=*=openblas' \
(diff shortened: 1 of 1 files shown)


parente:
A heads up. All of the PRs that merged in the last day failed to build on Docker Cloud. I’ll check if it’s infrastructure problems or something with the build definitions sometime today.

jakirkham:
Thanks for the heads up.

Had seen some failures on Docker Hub yesterday that appeared to be infrastructure related (building worked fine locally). Maybe related?

Given CI now appears to pass, guessing this was infrastructure related. Is that correct? Do we know the cause?

parente:
A forced manual rebuild worked just fine. All of the builds were reporting successful, and then had errors after the success message that apparently caused either the image push or tagging to fail. Really odd.

Issue #609, opened 2018-04-13
check if username jovyan exists before usermod call in start.sh


datwiz:
When restarting a stopped container, the start.sh script fails if the `NB_USER` env var has been set due to the user `jovyan` not existing.  The username was renamed to `NB_USER` when the container was started the first time.

**What docker image you are using?**
`jupyter/datascience-notebook`
Code impacted is defined in `jupyter/base-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**
Using a docker-compose file to start and stop the notebook container
```
version: '3'
services:
  datascience-notebook:
    user: root
    image: 'jupyter/datascience-notebook:8a1b90cbcba5'
    command: ['/usr/local/bin/start-notebook.sh', '--NotebookApp.token=']
    # command: ['/usr/local/bin/start.sh', '/bin/bash']
    environment:
    - NB_USER=${USER}
    - CHOWN_HOME=yes
    - GRANT_SUDO=yes
    volumes:
    - ${HOME}/work:/home/${USER}/work
    ports:
    - "8888:8888"
```

**What steps do you take once the container is running to reproduce the issue?**
1.  Start a notebook container, specifying an alternate username
2. Stop the container, without removing it
3. Restart the container

```
> ../src/datascience/datascience > docker-compose up -d
Creating network "datascience_default" with the default driver
Creating datascience_datascience-notebook_1 ... done
> ../src/datascience/datascience > docker-compose stop
Stopping datascience_datascience-notebook_1 ... done
> ../src/datascience/datascience > docker-compose start
Starting datascience_datascience-notebook_1 ... done
Attaching to datascience_datascience-notebook_1
datascience-notebook_1  | /usr/local/bin/start-notebook.sh: ignoring /usr/local/bin/start-notebook.d/*
datascience-notebook_1  |
datascience-notebook_1  | Set username to: XXXXXXX
datascience-notebook_1  | usermod: user 'jovyan' does not exist
datascience_datascience-notebook_1 exited with code 6
> ../src/datascience/datascience > docker-compose ps
               Name                             Command               State    Ports
------------------------------------------------------------------------------------
datascience_datascience-notebook_1   tini -- /usr/local/bin/sta ...   Exit 6
>
```

**What do you expect to happen?**
Container restarts and notebook is available for use.

**What actually happens?**
Container fails in the start.sh script when trying to perform a usermod command on a non-existent user.

parente:
Should be fixed by #610. Thanks again.

PR #610, opened 2018-04-13
In start.sh ensure usename jovyan exists before attempting usermod

datwiz:
Added a conditional check in start.sh to ensure that the jovyan user exists before attempting a usermod.
Fix for Issue #609


consideRatio:
Ah! Found my way to this PR that contained the commit, so now it is explained with the referenced issue!

PR #617, opened 2018-04-23
Remove facets directory

mimoralea:
The `cd` command switches to the user's home directory (away from the directory we want to delete)
Then we remove the facets directory using the full path


mimoralea:
I think the build issue is unrelated. I fixed that locally by adding a `--force` to `npm cache clean`

parente:
Thanks for submitting this fix. I agree the build issue is unrelated. I can merge, but upstream builds will likely fail until we address whatever new npm behavior we're seeing with respect to cache during the jupyterlab extension installs.

Issue #627, opened 2018-05-05
jupyter/pyspark-notebook missing conda clean


jayqi:
The Dockerfile for the pyspark-notebook is missing a `conda clean` after installing `pyarrow`.

https://github.com/jupyter/docker-stacks/blob/1fbaef522f17d413535b86ad335b0f58f1042789/pyspark-notebook/Dockerfile#L44-L47

parente:
Fixed by #628.

PR #628, opened 2018-05-05
Added missing conda clean to pyspark-notebook Dockerfile

jayqi:
Added conda clean to pyspark-notebook Dockerfile. Addresses #627.

Building locally, this reduces the image size from 5.16GB -> 5.08GB (~80 MB).


parente:
Good catch. Thanks for the PR. 🍰

PR #629, opened 2018-05-06
update higher version of tensorflow and keras

dexterchan:
current tensorflow 1.3 is outdated.... 
need to update numpy, scipy etc
to update tensorflow to 1.5 and keras to 2.1

### DIFF
--- a/tensorflow-notebook/Dockerfile
+++ b/tensorflow-notebook/Dockerfile
@@ -3,11 +3,56 @@
 FROM jupyter/scipy-notebook
 
 LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
+USER root
+
+ENV DEBIAN_FRONTEND noninteractive
+
+ENV ORACLE_INSTANTCLIENT_MAJOR 11.2
+ENV ORACLE_INSTANTCLIENT_VERSION 11.2.0.4.0
+ENV ORACLE /usr/local/oracle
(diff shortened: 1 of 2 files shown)


tensorflow-notebook/Dockerfile
```diff
@@ -4,10 +4,28 @@ FROM jupyter/scipy-notebook
...
-    'keras=2.0*' && \
+    'numpy' \
+    'matplotlib' \
+    'pandas' \
+    'scipy' \
+    'seaborn' \
+    'scikit-learn' \
+    'scikit-image' \
+    'tensorflow' \
+    'keras' \
```
parente: I think we can get away with just updating the version numbers here without adding everything else. If there happens to be a version of numpy that also needs to be revised in the parent image, we should change that there.

tensorflow-notebook/Dockerfile.WithOracle
```diff
@@ -0,0 +1,56 @@
...
+    'keras' \
+    'plotly' \
+    'quandl' \
+    'tweepy' \
+    'textblob' \
+    'PyYAML' && \
+    conda clean -tipsy && \
+    fix-permissions $CONDA_DIR 
+#RUN pip install gym
+#RUN apt-get install python-opengl
```
parente: Did you mean to submit this file? I don't think we'll want to main this Dockerfile here in the Jupyter repo.

parente:
@dexterchan thanks for sharing what you did. I'd like to start a new PR that updates just the tensorflow and keras package versions, and updates whatever conflicts there might be in the scipy-notebook images.

PR #633, opened 2018-05-07
Don't chown -R a user's home directory

cam72cam:
Only the top level permissions are set wrong due to kubernetes not being able to do this automatically.

This hits significant problems when the chown takes more than 30s and the pod fails to start

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -40,7 +40,7 @@ if [ $(id -u) == 0 ] ; then
     # Ex: default NFS/EFS (no auto-uid/gid)
     if [[ "$CHOWN_HOME" == "1" || "$CHOWN_HOME" == 'yes' ]]; then
         echo "Changing ownership of /home/$NB_USER to $NB_UID:$NB_GID"
-        chown -R $NB_UID:$NB_GID /home/$NB_USER
+        chown $NB_UID:$NB_GID /home/$NB_USER
     fi
 
     # handle home and working directory if the username changed
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -40,11 +40,11 @@ if [ $(id -u) == 0 ] ; then
     # Ex: default NFS/EFS (no auto-uid/gid)
     if [[ "$CHOWN_HOME" == "1" || "$CHOWN_HOME" == 'yes' ]]; then
         echo "Changing ownership of /home/$NB_USER to $NB_UID:$NB_GID"
-        chown $NB_UID:$NB_GID /home/$NB_USER
+        chown $CHOWN_HOME_OPTS $NB_UID:$NB_GID /home/$NB_USER
     fi
     if [ ! -z "$CHOWN_EXTRA" ]; then
         for extra_dir in $(echo $CHOWN_EXTRA | tr ',' ' '); do
-            chown $NB_UID:$NB_GID $extra_dir
+            chown $CHOWN_EXTRA_OPTS $NB_UID:$NB_GID $extra_dir
         done
     fi
 
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -42,6 +42,11 @@ if [ $(id -u) == 0 ] ; then
         echo "Changing ownership of /home/$NB_USER to $NB_UID:$NB_GID"
         chown $NB_UID:$NB_GID /home/$NB_USER
     fi
+    if [ ! -z "$CHOWN_EXTRA" ]; then
+        for extra_dir in $(echo $CHOWN_EXTRA | tr ',' ' '); do
+            chown $NB_UID:$NB_GID $extra_dir
+        done
+    fi
 
     # handle home and working directory if the username changed
     if [[ "$NB_USER" != "jovyan" ]]; then


cam72cam:
@ewjoachim This reverts your work in here: https://github.com/jupyter/docker-stacks/pull/576

I don't think doing a -R is a good solution as we are already running into some major slowness issues on our cluster.  Perhaps adding a list of jupyter subpaths to chown is a better approach for https://github.com/jupyter/docker-stacks/issues/575?

base-notebook/start.sh
```diff
@@ -40,7 +40,7 @@ if [ $(id -u) == 0 ] ; then
     # Ex: default NFS/EFS (no auto-uid/gid)
     if [[ "$CHOWN_HOME" == "1" || "$CHOWN_HOME" == 'yes' ]]; then
         echo "Changing ownership of /home/$NB_USER to $NB_UID:$NB_GID"
-        chown -R $NB_UID:$NB_GID /home/$NB_USER
+        chown $NB_UID:$NB_GID /home/$NB_USER
```
parente: The env var `CHOWN_HOME` setting is designed to me "chown the $NB_USER home directory recursively." If you think there's a need for a flag to indicate a non-recursive chown should be performed instead, I think we should be adding a env var instead of changing the meaning of the existing one.

cam72cam: Please take a look at the git history.  I originally wrote this code due to work discussed in the zero to jupyterhub project.

The discussion about the chown can be found here: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/421#issuecomment-362019068

Doing a chown -R in a user's home directory is a terrible idea by default and will cause significant slowdowns as soon as the environment is seriously used.  We have <10 users and are already having issues due to this change.

If someone really want's to do a chown -R, a flag should be added for that.  However I think a better solution should be found, like specifying paths inside the user's home directory to be chowned.

parente: Sorry, clarifying: I thought the original intent of `CHOWN_HOME` was meant to be recursive and assumed (clearly, incorrectly) that the later PR from #576 was meant to correct the missing `-R` based on the conversation in the original issue #541.

I agree we either need two flags (`CHOWN_HOME`, `CHOWN_HOME_RECURSIVE`) or something more elegant.

cam72cam: Breaking out the two flags would work, but I hope we can find something more elegant 

I wonder what @ewjoachim thinks about adding something like $CHOWN_HOME_SUBDIRS and have that be a list.  He has had the clearest use case for that flag so far.

On the other hand more options for the user is usually a good thing so we could do all 3 options, though I still worry that CHOWN_HOME_RECURSIVE can easily get users into trouble

Sorry if I took the offensive a bit too hard in my initial reply

parente: No worries. My initial response this morning was a too terse from tapping it out on a phone and lead to the initial confusion anyway.

`CHOWN_HOME_SUBDIRS` sounds good to me. And maybe if we support `.` as a value, it can mean the same thing as `CHOWN_HOME_RECURSIVE` without that additional option.

ewjoachim: I'm clearly 👍 on anything that fixes a problem I don't have while not creating problems for me 😄or if it breaks but there's a better workaround, I'll gladly take it too ! I've only made one contribution to the repo, so I'm far from being an expert on anything.

About the `CHOWN_HOME` being recursive or not, I just assumed it was supposed to be like this. If there's a better way to do what I did in #576 for #575, I'm totally ok to change my workflow ! Just let me know here or in the original ticket.

cam72cam: @ewjoachim @parente I've pushed a commit to add CHOWN_EXTRA.  It should allow you to chown additional directories and get rid of the need to do a recursive chown

base-notebook/start.sh
```diff
@@ -40,7 +40,12 @@ if [ $(id -u) == 0 ] ; then
...
+    fi
+    if [ ! -z "$CHOWN_EXTRA" ]; then
+        for extra_dir in $(echo $CHOWN_EXTRA | tr ',' ' '); do
+            chown $NB_UID:$NB_GID $extra_dir
```
parente: Should `$CHOWN_EXTRA` be recursive? I know that risks running into the same problem, but if the user has a deeply nested set of directories that all need modification, listing them all is going to be a bear. Or should we expect users in that situation to modify the uid/gid ahead of the container launch, assuming they know what those values will be? 

@ewjoachim any thoughts on if this will satisfy your original use case?

ewjoachim: My original use case is: 
- I want to run docker on an existing `ipynb` in the current dir. I want to be able to edit the `ipynb` from both inside and outside the docker.

With your solution, what is the command I have to run ?

cam72cam: Is it a single ipynb or do you expect to have a bunch of them in that directory?  I am considering adding a CHOWN_EXTRA_RECURSIVE setting (with a warning that it should only be done on known small directories)

cam72cam: I added CHOWN_HOME_OPTS and CHOWN_EXTRA_OPTS just in case someone decides to add flags like -R to either chown.  It worries me for my use case, but that should not stop other people from having the opportunity

parente:
@cam72cam, thanks for the work on this PR. I think it's good to go now. And, @ewjoachim, thanks for chiming in with your feedback too.

PR #635, opened 2018-05-08
send2trash not work

mnagaku:
https://github.com/jupyter/notebook/issues/3130
Now, files not delete on https://hub.docker.com/r/jupyter/base-notebook/ and derived container images.


parente:
Thanks for the PR. I agree that we want to disable the trash mechanism in the container by default, given the background in the notebook issue. If a user has the home directory mounted from the host instead (and assuming the issue not present in that case), passing `--FileContentsManager.delete_to_trash=True` on the command line should override the new default.

Issue #626, opened 2018-05-04
Architecture: Make 'spark' container: spark, hadoop and cluster manager agnostic


taqtiqa-mark:
Hi! Thanks for all the work that has gone into creating Jupyter's docker-stacks images.

I'm wondering if you are open to breaking out spark into a separate container, rather than the current monolithic container. By "open" I mean willing to work in that direction, i.e. accept pull requests etc.
That means I don't propose this issue resolve all the implementation issues/details.

Some advantages for this are:
- Cluster manager agnostic : Currently tied to mesos, and some environments don't have this on "approved"/licensed technology lists
- Hadoop agnostic: Currently tied to specific hapdoop versions, while spark itself is more flexible.
- Container sizes reduced considerably
- Retire pyspark and maintain only the all-spark container
- Support multiple spark client library versions in one container.  That is configure client library versions at container runtime - use a launch script parameter, e.g. `spark=230`, `spark=210`, etc to set env variables, linked files etc.

Some disadvantages for this are:
- Identify a robust spark container that new users are directed to as an easy default, e.g. [BDE/docker-spark](https://github.com/big-data-europe/docker-spark)
- Spark container needs to be launched, but users can be directed to [such resources](https://grzegorzgajda.gitbooks.io/spark-examples/content/basics/docker.html)
- Example R and Py files showing connection parameters for different cluster managers

Won't go into more detail until there is interest in taking this approach.

parente:
Hi @taqtiqa-mark 

> I'm wondering if you are open to breaking out spark into a separate container, rather than the current monolithic container.

I think I need a bit more information to understand whether we should update all-spark-notebook or manage a set of Spark related images outside this repo. I've got a couple questions about your advantages and disadvantage that should help.

> Identify a robust spark container that new users are directed to as an easy default, e.g. BDE/docker-spark

Would the new image(s) you're proposing derive from `jupyter/base-notebook` or another docker-spark image like `BDE/docker-spark`? I'd like for all the images in this repository to stay within the inheritance tree of `jupyter/base-notebook` for consistency. That said, I completely understand if you're proposing one or more new images that start from a Spark-specific base and add Jupyter instead of the other way around. I just think those images deserve their own repository for clarity.

> Support multiple spark client library versions in one container. That is configure client library versions at container runtime - use a launch script parameter, e.g. spark=230, spark=210, etc to set env variables, linked files etc.
> Container sizes reduced considerably

[...]

Issue #611, opened 2018-04-15
Following Debian upstream in adopting ffmeg over libav-tools


taqtiqa-mark:
Hi! Thanks for all the effort making Jupyter's docker-stacks images.

libav-tools (jupyter/scipy-notebook)

I'm porting the Jupyter stacks to Apline Linux containers.  As part of that, while building  `jupyter/scipy-notebook` I discovered libav is not available on Alpine Linux, but ffmeg is.  The libav team was not very responsive/proactive about the patch to try make it work, and things seem to have stalled.  I don't see any evidence to expect this to change.

On further digging I discovered that, unfortunately, there has been some ping-pong in Debian around libav vs ffmeg. 
 Having dropped ffmeg at one point they seem now to have now dropped libav and picked ffmpeg back up. 
Previously this affected this project per issue #90, where ffmpeg was not available in a ubuntu distribution.  
Some detail behind the change is set out in a [LWN article](https://lwn.net/Articles/650816/).  It seems the reasons (security) for switching back to ffmeg are not trivial.  The [announcement is here](https://lwn.net/Articles/650495/)

Any chance you'd consider following upstream and using ffmpeg in place of libav-tools?  

If not that is fine, I will just need to include a notice in the README.md.

parente:
> Previously this affected this project per issue #90, where ffmpeg was not available in a ubuntu distribution.

I believe the images here were based on Debian Jessie at the time. Now they derive from Ubuntu 16.04, so switching to ffmpeg shouldn't be all that difficult. If you'd like to submit a PR with the change, along with a notebook showing that matplotlib animation still works properly, I think we can merge it.

jakirkham:
There’s also `ffmpeg` in `conda-forge`, which could be used to this end.

taqtiqa-mark:
Unfortunately, I'm going to have to leave this to someone else.  The Alpine build is taking all my spare cycles.

I'll use ffmpeg with the hope/expectation that at some point the switch from libav-tools will be made.

Happy for you to close this or leave it open.  As you wish.

parente:
No problem, @taqtiqa-mark. Please leave the issue open.

jakirkham:
If it's any help, I use `ffmpeg` from `conda-forge` to generate videos either with `matplotlib` or `moviepy` (primarily on our Linux-based cluster). Don't know what bar you had in mind for testing, but maybe that provides some assurance that this does work.

Bidek56:
[...]

PR #641, opened 2018-05-24
Upgrade Ubuntu to latest LTS (Bionic 18.04)

BradenAtSiemens:
For your consideration, an upgrade of Ubuntu to its latest LTS (Bionic 18.04).


parente:
> Hmm... it seems like hkp://p80.pool.sks-keyservers.net:80 occationally cannot be contacted, though using that key server is not part of this change...

Did you experience build failures because of this? /cc @taqtiqa-mark

scipy-notebook/Dockerfile
```diff
@@ -6,9 +6,9 @@ LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 
 USER root
 
-# libav-tools for matplotlib anim
+# ffmpeg for matplotlib anim
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends libav-tools && \
+    apt-get install -y --no-install-recommends ffmpeg && \
```
parente: Noting this resolves https://github.com/jupyter/docker-stacks/issues/626

parente:
I'm 👍 on this since 18.04 is a LTS release. 

I don't have a way to test the older mesos client compatibility, but we're discussing how to make the Spark images more cluster agnostic in #626 anyway.

I'd like to leave this open for a few days to see if anyone has reservations. Then we can ship it.

scipy-notebook/Dockerfile
taqtiqa-mark: Think this resolves #611 rather than #626

taqtiqa-mark:
Yes, if the consensus/decision is to make the `all-spark-notebook` generic then the current Mesos container would be the result of a  [buildah script](https://github.com/taqtiqa-mark/bob) or a Docker [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds) such as:

````
FROM jupyter/all-spark-notebook
FROM mesosphere/mesos
````

scipy-notebook/Dockerfile
parente: Indeed. Wrong #

BradenAtSiemens:
Hmm... looking more carefully at the build log, for some reason in the notebooks that build off of scipy-notebook, `blas` gets downgraded from 1.1 from the `conda-forge` repo to 1.0 from the `defaults` repo, which then seems to trigger a downgrade of `numpy`, `scipy`, and `scikit-learn`...

If I pin the install of `blas` to `1.1*` (currently it is just `*`), it seems to avoid this issue, so I'll add a commit for that.

parente:
Nice catch. I don't see any other concerning downgrades in the latest build log. Let's get this merged.

Thanks for the PR and the testing! 🍰

jakirkham:
@BradenAtSiemens, have also been running into some annoyances with `blas` downgrades recently and know of others struggling with this issue as well. Think we may want to use `conda`'s MatchSpec feature to pin `blas` to `conda-forge`. Believe this is a consequence of changes in `defaults` that are affecting us currently, which this proposed change would workaround. Will get a PR going.

cc @kalefranz @msarahan (for awareness)

jakirkham:
Have placed this change in PR ( https://github.com/jupyter/docker-stacks/pull/649 ).

Bidek56:
Mesos build is not available for 18.04, why would you hardcode the Mesos repo to xenial? Do you know if it works on 18.04?

Bidek56:
According to [MESOS-8921](https://issues.apache.org/jira/browse/MESOS-8921l) Mesos does NOT work with 18.04.

taqtiqa-mark:
Maybe this is a natural point at which to start to disentangle the Spark Cluster Manager (SCM) and spark from the containers?

taqtiqa-mark:
Maybe 18.04 is the natural point at which to remove the dependency on the version of Spark and its Cluster Manager?

PR #652, opened 2018-06-04
upgrade tini to latest version 0.18.0

gurvindersingh:
Also added the option to send signal to all child group processes as mentioned in docs

https://github.com/krallin/tini#process-group-killing


minrk:
Thanks!

Issue #560, opened 2018-02-20
Revisit root permissions and entrypoint for user management


minrk:
In start.sh, we support setting up the user in a variety of ways that currently require the container to be started as root. #552, #553, #559 introduce a mechanism to handle a subset of those cases when the user is in the root *group*, as is typical with openshift. #553 initially did so in an ENTRYPOINT instead of CMD in order to benefit even when CMD is overridden, e.g. directly to `jupyterhub-singleuser` in kubespawner. That logic now resides in start.sh with everything else (#559), but perhaps much of start.sh ought to move to an ENTRYPOINT instead of the CMD.

Things to consider:

1. how much can/should we accomplish with the root *group* instead of the root *user*? I'd love it if we could stop requiring folks to launch the image as the root user, since a slip-up there can be a major permission issue
2. how much of the permissions/user setup belongs in an ENTRYPOINT rather than start.sh? Pro of ENTRYPOINT is that overriding CMD preserves permissions/user-management logic. Con of ENTRYPOINT is that opting out of user-management behavior is a bit harder if desired. So if we move things to the entrypoint, we should make sure that default behavior is always safe and almost always desirable.

Current capabilities in start.sh that require root permissions:

- update $HOME to /home/$NB_USER if username is set specifically
- rename `jovyan` user to `$NB_USER` if set
- set UID/GID of NB_USER
- grant $NB_USER sudo permissions
- chown $HOME (typically just the empty-directory volume case)

parente:
I've stalled on creating the ReadTheDocs site, partially because I'm at the section of the doc where I have to document all the various options the images support. It's daunting.

So I was planning to open an issue this morning to discuss what options we currently support (e.g., `NB_UID`), why they were originally needed, and if they're still needed given the usage patterns that have emerged.

Then I saw this new ticket about root permissions (thanks @minrk !) and wonder if we should tackle the two topics together:

* Can we shed any of the options we currently support?
* How much do we really need to do in entrypoint / cmd that requires root?

aneagoe:
I'm currently using https://github.com/GrahamDumpleton/openshift3-jupyter-stacks to be able to deploy on OpenShift. However, since the dockerfile there specifies changes to /home/$NB_USER, it results in an additional ~3GB layer to the base image, just because it's changing permissions. I can imagine if this enhancement is resolved, then these images would run natively on OpenShift. 
[...]

PR #653, opened 2018-06-06
Remove writable group file due to issues with being able to give su access when shouldn't be allowed.

GrahamDumpleton:
This change is to revert making the ``/etc/group`` file writable as part solution to avoid issues as discussed in https://github.com/jupyter/docker-stacks/issues/560.

Although it means in environment where container run as group ID not in ``/etc/groups`` you will get a warning:

```
groups: cannot find name for group ID 100000
```

in certain situations when creating an interactive shell in the container, there is no known Python packages or applications which will fail if running with a group ID not in ``/etc/group``.

With the ``/etc/group`` file once again not writable, as part of https://github.com/jupyter/docker-stacks/issues/560 can then work out whether restricting ``su`` to members of ``wheel`` group can be done and lock down ``su`` for normal case.

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -50,7 +50,7 @@ ADD fix-permissions /usr/local/bin/fix-permissions
 RUN useradd -m -s /bin/bash -N -u $NB_UID $NB_USER && \
     mkdir -p $CONDA_DIR && \
     chown $NB_USER:$NB_GID $CONDA_DIR && \
-    chmod g+w /etc/passwd /etc/group && \
+    chmod g+w /etc/passwd && \
     fix-permissions $HOME && \
     fix-permissions $CONDA_DIR
 
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -94,8 +94,7 @@ else
         # User is not attempting to override user/group via environment
         # variables, but they could still have overridden the uid/gid that
         # container runs as. Check that the user has an entry in the passwd
-        # file and if not add an entry. Also add a group file entry if the
-        # uid has its own distinct group but there is no entry.
+        # file and if not add an entry.
         whoami &> /dev/null || STATUS=$? && true
         if [[ "$STATUS" != "0" ]]; then
             if [[ -w /etc/passwd ]]; then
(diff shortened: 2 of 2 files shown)


GrahamDumpleton:
Actually, thinking about it some more, not sure removing this makes any difference.

For someone to be able to add an entry to the ``/etc/group`` file they would already need to be ``root`` or running with group ``root``. If they are already running as group ``root``, since there is no ``wheel`` group, it will fallback to allowing you to run ``su`` anyway, even if ``pam_wheel`` is enabled. So remove write access to ``/etc/group`` gains you nothing.

The only thing one can do is document that if running with a ``uid`` not in the ``/etc/passwd`` file, and thus inherit group ``root``, you should always drop capabilities from running ``su``.

GrahamDumpleton:
Withdrawing this as indeed makes no difference and how is currently done can't be improved. Can only document that should drop capabilities as necessary.

GrahamDumpleton:
May yet reopen this. Although there is no ``wheel`` group on Debian/Ubuntu, you can add one and ``su`` will then honour it and will not fallback to allowing anyone in group ``root`` to run ``su``. So, if use:

```
RUN groupadd wheel && \
    echo "auth required pam_wheel.so use_uid" >> /etc/pam.d/su
```

even if when run with uid not in ``/etc/passwd`` and so run as gid of ``root``, ``su`` will fail.

```
$ docker run -it --rm -u 1000000 -p 8888:8888 base-notebook bash
I have no name!@b41018af87fc:~$ id
uid=1000000 gid=0(root) groups=0(root)
I have no name!@b41018af87fc:~$ start.sh bash
/usr/local/bin/start.sh: ignoring /usr/local/bin/start-notebook.d/*

Adding passwd file entry for 1000000
Container must be run with group "users" to update files
Executing the command: bash
jovyan@b41018af87fc:~$ cp /etc/passwd /tmp/passwd
jovyan@b41018af87fc:~$ rootpass=`openssl passwd -1 root`
jovyan@b41018af87fc:~$ cat /tmp/passwd | sed "s%root:x%root:${rootpass}%" > /etc/passwd
jovyan@b41018af87fc:~$ su root
Password:
su: Permission denied
jovyan@b41018af87fc:~$ id
uid=1000000(jovyan) gid=0(root) groups=0(root)
jovyan@b41018af87fc:~$ exit
exit
I have no name!@b41018af87fc:~$ exit
exit
```

GrahamDumpleton:
Note that this PR was superseded by https://github.com/jupyter/docker-stacks/pull/654 which contains the addition of ``wheel`` group and enabling of ``pam_wheel`` as explained.

Issue #642, opened 2018-05-25
Future issue with jupyter notebook 5.5 regarding hostname in docker ?


tomjorquera:
By toying with upgrading jupyter to version 5.5 inside a container, I stumbled upon the following issue: 
https://github.com/jupyter/notebook/issues/3605

Apparently, jupyter-notebook behavior regarding the url has been changed in 5.5, and it will now try to display the host name instead of `localhost`. Inside docker, it means that the url that now appears will be something like `https://8bc26195e104:8888/?... ` (which does not work when copy-pasted into the browser, host must be changed manually to `localhost`)

Feels like it's a somewhat significant issue for the project.

Maybe a workaround would be to propose an option to jupyter-nb to disable name resolution ?

parente:
I agree this can be a confusion. In fact, the `localhost` link printed by Jupyter only ever directly worked if a user was running the docker container on a local machine. When running a container on a remote server, like a VM in the cloud, Jupyter Notebook still prints the URL as `localhost` because it is blissfully unaware that it is running in a container on another machine, as it should be.

[Docker does allow a user to specify the fully qualified domain name](https://docs.docker.com/engine/reference/run/#env-environment-variables) of the container with `docker run --hostname my.host.name` command. I'm pretty sure this will impact what the notebook server will print when it starts. I'm not sure how useful it is in practice.

parente:
I took a crack at updating the quick start steps to describe what hostname should be used to access the docker image when a container launches. Additional PRs agains the docs are welcome. 

I'm inclined to close this issue, as it's already being tracked upstream in jupyter/notebook and I don't have a good solution at the moment other than to continue to refine the instructions as people find specific fault with them.

minrk:
The ip and port output from the notebook should generally be ignored when run in containers, since it's ~never right and never can be by default. Only in the very specific case of local docker where the forwarded and exposed ports match can it ever be correct. I'm not sure what the best solution is other than *allowing* override (https://github.com/jupyter/notebook/pull/3668) which doesn't change the fact that the default.

parente:
Closing this issue as the docs have been updated and we don't have any ideas at the moment on how to make the docker behavior clearer. If anyone has a suggested improvement, we'll be happy to take a PR.

mshytikov:
Faced the same issue and it took me a while to figure out the way. 

As was mentioned above _allowing override_ solution did work for me. 

In my case to keep _'copy-paste url'_ experience as it was before,
[...]

PR #655, opened 2018-06-07
bump some packages

minrk:
- notebook 5.5
- matplotlib 2.2
- pandas 0.23
- scipy 1.1
- scikit-image 0.14
- statsmodels 0.9


tomjorquera:
Careful, notebook 5.5 currently causes issue #642 !

parente:
Maybe we should update the quickstart blurb in the README (and duplicated in the index.rst of the docs) to say visit `localhost:8888` explicitly and ignore what gets printed. I'm thinking it might help first-time users who aren't yet versed in the nuances of internal vs external container ports and hostnames.

parente:
Thanks for the updates @minrk.

PR #656, opened 2018-06-08
Updated recipe for pip and added conda example.

aaronjswank:
Updated the recipe for pip to include requirements.txt file.
Also included example for using conda to install packages.

### DIFF
--- a/docs/using/recipes.md
+++ b/docs/using/recipes.md
@@ -2,7 +2,7 @@
 
 Users sometimes share interesting ways of using the Jupyter Docker Stacks. We encourage users to [contribute these recipes](../contributing/recipes.html) to the documentation in case they prove useful to other members of the community by submitting a pull request to `docs/using/recipes.md`. The sections below capture this knowledge.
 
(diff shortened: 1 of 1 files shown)


docs/using/recipes.md
```diff
@@ -19,6 +19,33 @@ Then build a new image.
...
+
+```dockerfile
+# Start from a core stack version
+FROM jupyter/datascience-notebook:9f9e5ca8fe5a
+# Install from requirements.txt file
+COPY requirements.txt /tmp/
+RUN pip install --requirement /tmp/requirements.txt && \
+    fix-permissions $CONDA_DIR && \
+    fix-permissions /home/$NB_USER
+COPY . /tmp/
```
parente: It might be good to add a comment about what this extra copy is doing in both cases. I'm guessing its to add additional project files?

aaronjswank: Yes, the extra copy (COPY . /tip/ ) is not needed.  It was there for adding additional project files in my setup.

parente:
Thanks again!

PR #657, opened 2018-06-13
Add Azure data sources support

random82:
* Support for Azure Storage over "wasb://"
* Support for MSSQL
* Support for Azure CosmosDB
* Support via Hadoop plugins
* Installed using maven

### DIFF
--- a/pyspark-notebook/Dockerfile
+++ b/pyspark-notebook/Dockerfile
@@ -42,6 +42,90 @@ ENV PYTHONPATH $SPARK_HOME/python:$SPARK_HOME/python/lib/py4j-0.10.6-src.zip
 ENV MESOS_NATIVE_LIBRARY /usr/local/lib/libmesos.so
 ENV SPARK_OPTS --driver-java-options=-Xms1024M --driver-java-options=-Xmx4096M --driver-java-options=-Dlog4j.logLevel=info
 
+
+
+# Azure Storage provider for Spark 
+ENV HADOOP_SEMVER $HADOOP_VERSION.0
+
+RUN apt-get update -y && \
+    apt-get install -y --no-install-recommends software-properties-common && \
+    apt-add-repository universe && \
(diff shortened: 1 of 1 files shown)


parente:
@random82 Thanks for considering a contribution to the project. At the moment, we are attempting to slim down the spark related images to be more cluster and provider agnostic (#626). I believe your additions would be better served as a [recipe in the documentation](http://jupyter-docker-stacks.readthedocs.io/en/latest/using/recipes.html) (see: [recipe contributor guide](http://jupyter-docker-stacks.readthedocs.io/en/latest/contributing/recipes.html))  or as a community maintained stack (http://jupyter-docker-stacks.readthedocs.io/en/latest/contributing/stacks.html).

Feel free to open a new PR or to repurpose this PR if you'd like to submit a recipe.

random82:
Hi @parente 
The reason why I suggested Azure support is that Azure Blobs (along with AWS S3 and OpenStack Swift) is generally supported by Hadoop stacks and other Spark stacks I was working with so far. 

When I pulled your pyspark image I was expecting that these data sources would be supported out of the box.

I looked at the recipes here: http://jupyter-docker-stacks.readthedocs.io/en/latest/using/recipes.html#using-pyspark-with-aws-s3 but similar setup for WASB will take installing plugins. I'm happy to contribute similar recipe for WASB but without plugins installed it will take cutting new image by a user.

parente:
> I'm happy to contribute similar recipe for WASB but without plugins installed it will take cutting new image by a user.

Having users build (and optionally maintain community) images is the direction we'd like to head. We're basically at the limit of what we can reasonable maintain ourselves in this repo, and there innumerable customizations users would like to contribute to the existing images. Past experience has taught us that kitchen-sink images become unmaintainable over time, so we're trying to slim things down and encourage the growth of an ecosystem (recipes, other repositories linked from the docs) around the core images here.

PR #658, opened 2018-06-13
Bumping up Spark and Conda to the current version

Bidek56:
Bumped Ubuntu and Conda to the current version (4.5.5)
Bumped Spark to 2.3.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,8 @@ Please see the [Contributor Guide on ReadTheDocs](http://jupyter-docker-stacks.r
 
 ## Alternatives
 
-* [jupyter/repo2docker](https://github.com/jupyter/repo2docker) - Turn git repositories into Jupyter enabled Docker Images
+* [jupyter/repo2docker](https://github.com/jupyter/repo2docker) - Turn git repositories into Jupyter-enabled Docker Images
+* [openshift/source-to-image](https://github.com/openshift/source-to-image) - A tool for building/building artifacts from source and injecting into docker images
 * [jupyter-on-openshift/jupyter-notebooks](https://github.com/jupyter-on-openshift/jupyter-notebooks) - OpenShift compatible S2I builder for basic notebook images
 
 ## Resources


base-notebook/Dockerfile
```diff
@@ -63,16 +63,16 @@ RUN mkdir /home/$NB_USER/work && \
...
     wget --quiet https://repo.continuum.io/miniconda/Miniconda3-${MINICONDA_VERSION}-Linux-x86_64.sh && \
-    echo "0c28787e3126238df24c5d4858bd0744 *Miniconda3-${MINICONDA_VERSION}-Linux-x86_64.sh" | md5sum -c - && \
+    echo "a946ea1d0c4a642ddf0c3a26a18bb16d *Miniconda3-${MINICONDA_VERSION}-Linux-x86_64.sh" | md5sum -c - && \
     /bin/bash Miniconda3-${MINICONDA_VERSION}-Linux-x86_64.sh -f -b -p $CONDA_DIR && \
     rm Miniconda3-${MINICONDA_VERSION}-Linux-x86_64.sh && \
     $CONDA_DIR/bin/conda config --system --prepend channels conda-forge && \
     $CONDA_DIR/bin/conda config --system --set auto_update_conda false && \
     $CONDA_DIR/bin/conda config --system --set show_channel_urls true && \
-    $CONDA_DIR/bin/conda install --quiet --yes conda=4.5.4 && \
+    $CONDA_DIR/bin/conda install --quiet --yes conda=${MINICONDA_VERSION} && \
```
parente: I think this will be a no-op given the miniconda installer is using the same version, which is fine. I wonder if there's a value we should put here so we don't have to update it constantly in the long term. Maybe we can add a bit of code to lop off the patch part of `MINICONDA_VERSION` and add a * so that it updates to the latest patch? Or maybe there's a better conda-way to accomplish the same thing that @jakirkham knows?

Not a blocking issue. Just wanted to call it out for ourselves.

minrk: Miniconda releases aren't made for every conda version, so we may well in general be installing e.g. miniconda 4.5.4 and conda 4.5.6. In general, this should not be set using the same variable, which implies the two are linked, when they aren't.

parente:
Interesting. I tried to push one extra commit here to separate the `MINICONDA_VERSION` from the install part of the command and somehow managed to close the PR without the ability to reopen it. Yet when I compare this branch to master locally, there's a delta.

I'll either figure out how to reopen or push a new PR with @Bidek56's changes intact.

PR #665, opened 2018-06-24
Fix #492: NB_GID adds a supplemental group, leaves primary untouched

### LINKED ISSUE #492: Missing /opt/conda/bin path for root user
`jupyter/scipy-notebook`

```
sudo docker run --user root -d -p 18767:8888 --name x-scipy\
                -e GRANT_SUDO=yes\
                -e http_proxy="http://172.17.0.1:3128"\
                -e https_proxy="http://172.17.0.1:3128" --restart unless-stopped\
                -e NB_UID=1011 -e NB_GID=1011 -v /home/x/notebooks:/home/jovyan/work\
                jupyter/scipy-notebook start-notebook.sh\
                --NotebookApp.password='sha1:xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
```

1. Connect to the docker container with `sudo docker exec -itu jovyan x-scipy` or open a terminal in the web interface.
2. Try to use pip (without sudo) -> `pip` works works but fails to install packages due to permissions.
3. Run pip with sudo.
4. Log in as root with `sudo su` and try pip -> `pip` command not found.
5. Full path works -> `sudo /opt/conda/bin/pip`

Run pip with root permissions so it is able to install packages.

**What actually happens?**

`pip` command not found

**Possible solutions**

1. Run pip with sudo (or logged as root) using full path -> `sudo /opt/conda/bin/pip`
2. Add /opt/conda/bin to root `$PATH` env variable.

parente:
This PR changes the behavior of setting `NB_GID` such that it adds `NB_USER` to a supplementary group with the value of `NB_GID` instead of changing the value of the default `users` group (GID 100), causing important directories to become unwritable (e.g., `/opt/conda`)

The only use case for `NB_GID` in the docs, namely supporting host-mounted volumes with an alternative group owner, should continue to work after this change. I'm not sure if there are other use cases that will regress due to this change. I'll leave the PR open for a while for folks to comment.

Fixes #492


parente:
Need to update test assertions

rkdarst:
I just saw this... for my use case, it is also about external volumes which live in a system with its own uid/gid namespace.  But I would prefer `NB_GID` to be the primary gid, not a supplementary one.  In my case, I need files to be created with `NB_GID` so I wanted it primary gid.

In the previous version, I overrode `cmd` in a very hackish way: rename old `users` group to the name I wanted, and add a duplicate `users` group, then when `start.sh` did its stuff, it changed the first one (primary gid with the new name) to `NB_GID`!

I'm posting here to let people know I am working on this now and ask for any thoughts.  If I get something useful, I will make a PR.

... but I might just switch to non-root execution since this may be causing other problems.  Not sure yet.

PR #667, opened 2018-07-09
Fix typo - Juptyer

jamiekt:
Found a typo at http://jupyter-docker-stacks.readthedocs.io/en/latest/using/recipes.html#juptyerhub, `juptyer` should be `jupyter`


parente:
Thanks!

Issue #639, opened 2018-05-22
Issue with setting credentials: server error 500


stefanproell:
Hello,
I am trying to launch a Jupyter container, but after providing the token and the password, there occurs a server error. The file cannot be found

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/jovyan/.jupyter/jupyter_notebook_config.json'
```

Affected image

```
jupyter/datascience-notebook:4ebeb1f2d154
jupyter/minimal-notebook:4ebeb1f2d154
```
## These are the steps:

1. Launch a container with `docker run -p 8888:8888 --name data_science --rm jupyter/minimal-notebook:4ebeb1f2d154`
2. Wait until the token link is shown in the log
3. Open the page in the browser
4. I am directed to the Jupyter directory tree and click on a folder. I am taken to the login page
5. Enter token and a new password
6. Error 500 page is shown

## Error message

The log shows the following error:

```
docker run -p 8888:8888 --name data_science --rm jupyter/minimal-notebook:4ebeb1f2d154
/usr/local/bin/start-notebook.sh: ignoring /usr/local/bin/start-notebook.d/*

Container must be run with group root to update passwd file
Executing the command: jupyter notebook
[I 09:43:46.845 NotebookApp] Writing notebook server cookie secret to /home/jovyan/.local/share/jupyter/runtime/notebook_cookie_secret
[W 09:43:47.052 NotebookApp] WARNING: The notebook server is listening on all IP addresses and not using encryption. This is not recommended.
[I 09:43:47.085 NotebookApp] JupyterLab beta preview extension loaded from /opt/conda/lib/python3.6/site-packages/jupyterlab
[I 09:43:47.085 NotebookApp] JupyterLab application directory is /opt/conda/share/jupyter/lab
[I 09:43:47.092 NotebookApp] Serving notebooks from local directory: /home/jovyan
[I 09:43:47.092 NotebookApp] 0 active kernels
[I 09:43:47.092 NotebookApp] The Jupyter Notebook is running at:
[I 09:43:47.092 NotebookApp] http://[all ip addresses on your system]:8888/?token=cc7c860bdf3de857f55f012dfab7e86ec3a435d769a7a24b
[I 09:43:47.092 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[C 09:43:47.092 NotebookApp] 
    
    Copy/paste this URL into your browser when you connect for the first time,
    to login with a token:
[...]

PR #670, opened 2018-07-11
Adds config file initialization to the base image

stefanproell:
Adds the initial creation of the configuration file jupyter_notebook_config.json to the image by running the generate-config command provided by jupyter notebook. This file was missing as reported in issue #639 and caused an error when trying to set a new password via the Web interface.

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -92,6 +92,9 @@ RUN conda install --quiet --yes \
     fix-permissions $CONDA_DIR && \
     fix-permissions /home/$NB_USER
 
+# Initialize configuration file
+RUN jupyter notebook --generate-config
+
 USER root
 
 EXPOSE 8888


stefanproell:
Build failed because of timeout

stefanproell:
Unrelated to the changes I made, the download of Hadoop in a different Dockerfile is too slow. The domain is suspicious too. Why is Apache Spark in the pyspark-notebook downloaded from claz.org (which seems to be a classified ad site) and not from the official repository? Will write an issue for that. 

This is the link that causes trouble:
`http://apache.claz.org/spark/spark-${APACHE_SPARK_VERSION}/spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz`

stefanproell:
Seems that http://apache.claz.org is in the list of mirrors of Apache:
https://www.apache.org/mirrors/

For the sake of transparency, I would prefer a domain that does not serve ads and is widely known to be reliable. I will change that domain and create a new PR.

PR #671, opened 2018-07-11
Adds config file initialization and updates Apache Mirror due to timeout issue

stefanproell:
As described in issue #639, the config file is not created within the base image. As a result, setting the password in the Web interface after launching the container fails because the file cannot be found. This fix creates a default config file by running the appropriate command in a dedicated step.

When building this image, the build failed due to a timeout of the download of the Spark libraries. The Apache Mirror that was used for downloading was too slow and took more than 18min, which caused the build process to terminate after hitting the 10min timeout. I replaced it with a different official Apache Mirror that is already used in the recipes. I encourage the use of only one mirror in the whole project or fetch the closest mirror in the docker file like this:

```
curl 'https://www.apache.org/dyn/closer.cgi' |
  grep -o '<strong>[^<]*</strong>' |
  sed 's/<[^>]*>//g' |
  head -1
```

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -79,14 +79,20 @@ RUN cd /tmp && \
     fix-permissions $CONDA_DIR && \
     fix-permissions /home/$NB_USER
 
-# Install Jupyter Notebook and Hub
+# Install Jupyter Notebook, Lab, and Hub
+# Generate a notebook server config
+# Cleanup temporary files
+# Correct permissions
+# Do all this in a single RUN command to avoid duplicating all of the
+# files across image layers when the permissions change
 RUN conda install --quiet --yes \
-    'notebook=5.5.*' \
+    'notebook=5.6.*' \
     'jupyterhub=0.8.*' \
     'jupyterlab=0.32.*' && \
     conda clean -tipsy && \
     jupyter labextension install @jupyterlab/hub-extension@^0.8.1 && \
     npm cache clean --force && \
+    jupyter notebook --generate-config && \
     rm -rf $CONDA_DIR/share/jupyter/lab/staging && \
     rm -rf /home/$NB_USER/.cache/yarn && \
     fix-permissions $CONDA_DIR && \
--- a/pyspark-notebook/Dockerfile
+++ b/pyspark-notebook/Dockerfile
@@ -16,7 +16,7 @@ RUN apt-get -y update && \
     rm -rf /var/lib/apt/lists/*
 
 RUN cd /tmp && \
-        wget -q http://apache.claz.org/spark/spark-${APACHE_SPARK_VERSION}/spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz && \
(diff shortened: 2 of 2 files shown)


base-notebook/Dockerfile
```diff
@@ -92,6 +92,9 @@ RUN conda install --quiet --yes \
     fix-permissions $CONDA_DIR && \
     fix-permissions /home/$NB_USER
 
+# Initialize configuration file
+RUN jupyter notebook --generate-config
```
parente: I think we might need to do this as part of the prior command so that the `fix-permissions` script applies to the config file and parent directories in the user home. Otherwise, changes to the uid and gid will not allow proper access to the file.

parente:
Thanks for the PR @stefanproell. I've just the one comment about where the config gets generated in the Dockerfile. I'm OK with switching mirrors. If we keep hitting performance problems, we can take the next step of finding the closest.

base-notebook/Dockerfile
```diff
@@ -88,8 +88,12 @@ RUN conda install --quiet --yes \
...
     npm cache clean --force && \
     rm -rf $CONDA_DIR/share/jupyter/lab/staging && \
-    rm -rf /home/$NB_USER/.cache/yarn && \
-    fix-permissions $CONDA_DIR && \
+    rm -rf /home/$NB_USER/.cache/yarn
+
+# Initialize configuration file
+RUN jupyter notebook --generate-config
+
+RUN fix-permissions $CONDA_DIR && \
```
parente: Thanks for making the change. Theres's one new issue now: The `fix-permissions` needs to be done in the same `RUN` command. If it's done separately, all of the files that have their permissions changed (i.e., everything in the home and conda directories) are duplicated in full across the docker image layers, doubling the size of the original layer in the final image.

Something like the following should do the trick:

```
# Install Jupyter Notebook, Lab, and Hub
# Generate a notebook server config
# Cleanup temporary files
# Correct permissions
# Do all this in a single RUN command to avoid duplicating all of the
# files across image layers when the permissions change
    'notebook=5.6.*' \
    'jupyterhub=0.8.*' \
    'jupyterlab=0.32.*' && \
    conda clean -tipsy && \
    jupyter labextension install @jupyterlab/hub-extension@^0.8.1 && \
    npm cache clean --force && \
    jupyter notebook --generate-config && \
    rm -rf $CONDA_DIR/share/jupyter/lab/staging && \
    rm -rf /home/$NB_USER/.cache/yarn && \
    fix-permissions $CONDA_DIR && \
    fix-permissions /home/$NB_USER
```

(Note: Notebook 5.6 is out with a fix for a critical vulnerability in its bundled mathjax, so it'd be great to bump that version # while we're in here too.)

If you want to make this change, that's great. If not, I can add a commit that does it to your pull request if you check the "Allow edits from maintainers" checkbox in the GitHub pull request screen.

stefanproell:
Hello,
I squashed the changes together in one commit (1584606a40a9980364e7074b8a1a7b1aa5791905) and extended the RUN command by the line for creating the config file. Also I bumped the Notebook Version to 5.6 and changed the mirror as discussed.

parente:
@stefanproell Thank you very much for working this PR to completion. Let's get it merged! 🍰

PR #672, opened 2018-07-12
Fixed README, Jupyter Lab launch sample command

tackman:
Fixed JupyterLab launch sample command options. Option name and image version was wrong


parente:
Thanks for the PR @tackman. FWIW, `e5c5a7d3e52d` is a valid tag on Docker Hub, just one that's a bit older since we don't update the examples in the documentation for every merge we make to master. But I'm glad you bumped it when you had the chance, so thanks again. 🍰

Issue #478, opened 2017-10-04
Use conda-forge's tini


jakirkham:
We package `tini` at `conda-forge` now. Maybe it would make sense to have it used here. Though we would probably want to pin it so that it doesn't get updated by `conda` while the container is running. Thoughts?

parente:
We could switch and pin. We'd have to juggle the build order a bit since tini is currently installed before conda.

minrk:
Sounds good to me.

jakirkham:
FWIW this is how we install and pin `tini` at `conda-forge`. This blocks `conda` from updating `tini` once it is installed.

ref: https://github.com/conda-forge/docker-images/blob/565e3e9add4a7b4462ee94d2838e2b8158152268/linux-anvil/Dockerfile#L73-L74

jakirkham:
Added PR ( https://github.com/jupyter/docker-stacks/pull/677 ), which seemed to do the trick for me locally. Feedback welcome.

PR #677, opened 2018-07-20
Install tini from conda-forge

jakirkham:
Fixes https://github.com/jupyter/docker-stacks/issues/478

Switches to getting `tini` from `conda-forge`.


base-notebook/Dockerfile
```diff
@@ -79,6 +73,13 @@ RUN cd /tmp && \
     fix-permissions $CONDA_DIR && \
     fix-permissions /home/$NB_USER
 
+# Install Tini
+RUN conda install --quiet --yes 'tini=0.18.0' && \
+    conda list tini | grep tini | tr -s ' ' | cut -d ' ' -f 1,2 >> $CONDA_DIR/conda-meta/pinned && \
+    conda clean -tipsy && \
+    fix-permissions $CONDA_DIR && \
+    fix-permissions /home/$NB_USER
```
parente: Do we need to fix permissions on the user home directory? Or just `$CONDA_DIR`?

jakirkham: Not totally sure. `conda` does create things like `~/.conda`. I don't think it's contents would have changed here, but it seemed ok to keep this to be safe.

jakirkham:
Thanks for the review :)

PR #685, opened 2018-07-28
point README badges to dockerhub

delgadom:
Right now URLs simply point to the badge image source


parente:
The badge currently points to https://microbadger.com/images/jupyter/base-notebook which includes some metadata about the images (e.g., size, tags, date created, etc.). Is there an equivalent page on Docker Hub?

delgadom:
The first two badges (pulls, stars) just point to the badge graphic src. I changed them to point to DockerHub. The third (size/layers) currently points to the MicroBadger profile for the base-notebook image - I left this badge alone. You could point them all to MicroBadger if you'd like, but it seems more useful to direct people to the dockerhub repo. And no, unfortunately, I don't think DockerHub provides statistics like MicroBadger does. But it seems intuitive to me to have the pulls & starts badges point to DockerHub and the image stats, which are provided by MicroBadger, point to that site.

parente:
@delgadom Thanks for the update. I misread the pull-request diff as to what changed. Merging! 🍰

delgadom:
Thanks. Probably could have used a better description, even simple as it was.

PR #687, opened 2018-08-01
Update group handling: set primary gid, leave suplemental with group users

rkdarst:
In my deployment, I need the uid and gid of files to match up with the outside university systems.  As such, the current behavior of leaving the primary gid at 100 and changing the supplemental gid to NB_GID isn't very useful.

It's a philosophical question which is better, primary gid=100 or NB_GID.  Perhaps some things will break for others.  But since this was recently changed, maybe things are still in a flux.  This patch is my contribution to get the discussion started.  If you are running kubespawner somehow connected to existing storage, this will probably be useful.  If you are using cloud-type storage, this may be less useful but then in that case, gids may not matter at all.

Let me know what you think and I can update.  I'm using this now and it makes things much easier... also much easier than the hacks I used before da11c3157, #665.

---

- If NB_GID is set, it is likely that files created should have NB_GID
  as their gid (if it doesn't matter, then this change is irrelevant.
  If it does matter, the liklihood that 100 matches with what the user
  wants is low).
- When NB_GID is set, create a new group with that gid.  Then, set
  this as the primary gid of the user.  Set the user's supplemental
  gids to NB_GID,100.


parente:
Change to supplemental was made in #665 for #492 so at least @iyanmv might be interested.

rkdarst:
Yep, I saw #492.  The group order doesn't seem to matter in that case.  This also restores previous behavior of changing primary gid to the new value, which might be good(?).

With this PR, pip/conda install still works (I spent a lot of time trying to make everything work, which ended up caused by the state of things both pre-665 and post-665).

base-notebook/start.sh
```diff
@@ -70,11 +70,12 @@ if [ $(id -u) == 0 ] ; then
...
 
-    # Add NB_USER to NB_GID if it's not the default group
+    # Set NB_USER primary gid to NB_GID (after making the group).  Set
+    # supplementary gids to NB_GID and 100.
     if [ "$NB_GID" != $(id -g $NB_USER) ] ; then
         echo "Add $NB_USER to group: $NB_GID"
-        groupadd -g $NB_GID -o $NB_USER
-        usermod -a -G $NB_GID $NB_USER
+        groupadd -g $NB_GID -o ${NB_GROUP:-${NB_USER}}
+        usermod -g $NB_GID -a -G $NB_GID,100 $NB_USER
```
parente: I agree this change does not regress the fix for #492. Would you mind updating the documentation in https://github.com/jupyter/docker-stacks/blob/master/docs/using/common.md#docker-options to describe the effect setting `NB_GID` has now and introduce `NB_GROUP` as another optional env var that the user can set?

rkdarst:
Updated - thanks for the feedback!

... next time I'll [ci-skip] the docs updates...

rkdarst:
Thanks!

iyanmv:
Hi! @parente thanks for mentioning. I didn't see the PR before, but today I came because I was facing some problems with the last image and I saw this last commit.

I always set NB_UID and NB_GID to match OS user and groups. Till this commit new files where created respecting these ids but now new files use gid=100 instead of NB_GID.

Anyone else facing the same issue?

iyanmv:
Forget it, I cannot replicate the issue in my laptop. Everything works as expected. I am building an image on top of `jupyter/scipy-notebook` and something must be failing there.

rkdarst:
@iyanmv, this PR should fix the problem of files being created with gid=100.  It would have begun only about a month ago.  So things might be in a little bit of a flux now with old versions and caching.

PR #688, opened 2018-08-02
Add a posibliity for a pre-start hook

rkdarst:
I had to add some hooks to the end of `start.sh` in order to hack some custom image configuration (in this case, it was some stuff that had to be done after the users got set up).  It could also be used for debugging (which would have been quite useful for me a month ago).

This is a little bit of a hackish solution, but it seems that the problem is so minimal now that it's not worth making a more heavyweight solution.  But if this is deemed useful, I can improve some more.

Still possibly needs tests.

---
- The start.sh script does different operations on users, and
  sometimes one may need to do operations *after* that is done (as
  opposed to /usr/local/bin/start-notebook.d/ which is run before).
  This provides that possibility.  (One use case is hooks which must
  be run after the users are set up.)
- To be technically perfect, one might want to use something similar
  to /usr/local/bin/pre-start-notebook.d/.  But let's start with the
  simplest thing possible for now - the expected use of this command
  is only if people use NB_USER, etc in a way that requires small
  follow up commands to be run.
- If notebook is running as root, this allows users to run arbitrary
  commands pre-sudo.  Consider security model.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -89,6 +89,7 @@ if [ $(id -u) == 0 ] ; then
 
     # Exec the command as NB_USER with the PATH and the rest of
     # the environment preserved
+    eval "$NB_PRE_START_HOOK"
     echo "Executing the command: $cmd"
     exec sudo -E -H -u $NB_USER PATH=$PATH XDG_CACHE_HOME=/home/$NB_USER/.cache PYTHONPATH=$PYTHONPATH $cmd
 else
@@ -133,5 +134,6 @@ else
 
     # Execute the command
     echo "Executing the command: $cmd"
+    eval "$NB_PRE_START_HOOK"
     exec $cmd
 fi
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -11,22 +11,33 @@ else
     cmd=$*
 fi
 
-for f in /usr/local/bin/start-notebook.d/*; do
-  case "$f" in
-    *.sh)
-      echo "$0: running $f"; . "$f"
-      ;;
-    *)
-      if [ -x $f ]; then
-        echo "$0: running $f"
-        $f
-      else
-        echo "$0: ignoring $f"
-      fi
-      ;;
-  esac
-  echo
-done
+run-hooks () {
+    # Source scripts or run executable files in a directory
+    if [[ ! -d "$1" ]] ; then
+	return
+    fi
+    echo "$0: running hooks in $1"
+    for f in "$1"/*; do
+	case "$f" in
+	    *.sh)
+		echo "$0: running $f"
+		source "$f"
+		;;
+	    *)
+		if [[ -x "$f" ]] ; then
+		    echo "$0: running $f"
+		    "$f"
+		else
+		    echo "$0: ignoring $f"
+		fi
+		;;
+	esac
+    echo "$0: done running hooks in $1"
+    done
+}
+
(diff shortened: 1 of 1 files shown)


parente:
I can see the use for this extension point. Before we merge, I'd like to see on a set of phases when scripts can be run (before user setup, after user setup, when else?) and a single way to include those scripts (.d folders named after the phases would be nice, I think) so that the start.sh doesn't become even more laden with disparate features and we can write clear documentation about it.

I'm willing to work with you on it @rkdarst. To kick things off, how about:

* `/usr/local/bin/before-start.d/` - custom scripts executed at the start of start.sh
* `/usr/local/bin/before-notebook.d/` - custom scripts executed after start.sh performs all of its logic other than running `$cmd`
* `/usr/local/bin/start-notebook.d/` - symlink to `before-start.d` for backward compatibility

rkdarst:
Sure, sounds like it's better to take some time to do it well now.

`before-start` and `before-notebook` seem like good phrases to use and I can't think of anything better yet.

One extra thing to consider: hooks could be run before and after the root setup runs, and also before and after the user setup runs.  My original problem was that I had to run something as the user's uid, er, but this could not be done using existing hook because the user didn't exist then.  (this is a managed deployment with kubernetes and NFS with root_squash so root can't do it).  It is difficult to make a hook *after* the sudo, because sudo only starts one command.  And *maybe* it is good to separate scripts that always run as the user to those that always run as root...?

My complex idea implementing the above is below (names should be improved).  If you ask me, this solves more problems than I actually have right now and is a bit too much like over-engineering.  But still, such an extensive suite of tools could have saved me many hours of hacking around the different limitations.

- `run-hooks /user/local/bin/before-root-setup.d`
- `run-hooks /user/local/bin/after-root-setup.d`
- `sudo -u $NB_USER "$0" "$@"` - start.sh recursively invokes itself as non-root.  If the container is run as non-root, it starts directly at next step.  The necessary support in the script is basically there - but logic of the non-root part would need a bit of updating.
- `run-hooks /user/local/bin/before-user-setup.d`  (but as of now, not much happens in between before-user-setup and after-user-setup)
- `run-hooks /user/local/bin/after-user-setup.d`
- exec the notebook.

parente:
@rkdarst Thanks for the follow-on. Bit swamped at work at the moment, but I will get back to you and give your latest approach the time it deserves.

parente:
@rkdarst I think the implementation you've got in this PR is a good place to start. I'd like to avoid adding other hooks under the YAGNI principal until someone surfaces a need for them. That said, your design is a good place to start if and when we need the hooks in the future.

It looks like all of the tests failed on Travis. I'll try pulling the changes here and seeing if something in the changes prevents start.sh from working properly.

rkdarst:
Thanks for the review!

I updated the PR and rebased onto new master.  It is still the minimum change to get a hook at the end before starting, but a good place anyway.

Thinking about it, I tihnk we can just have the sudo command run `$0 $cmd` and then there can be different hooks before and after the root part.  Backwards compatibility is the most annoying part.  But that's for another time.

parente:
@rkdarst I'm going to this merge this PR. If people raise issues with the switch sourcing `.sh` files instead of invoking them, we can have a discussion in a follow-on issue or PR.

We should also update the docs to point out the new .d directory. I'll open an issue about that for the future.

Thanks again for the contribution.

rkdarst:
Thanks!  For sure there's still some to be done but this is minimally
good to move forward.

Issue #663, opened 2018-06-21
About switch to Ubuntu bionic. Mesos Spark support is now broken


asyd:
**What docker image you are using?**

Example: `jupyter/pyspark-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

Example: `docker run -p 8888:8888 jupyter/pyspark-notebook:latest`

**What steps do you take once the container is running to reproduce the issue?**

Example:

1. Visit http://localhost:8889
2. Start an python notebook
3. Run a notebook that use a Mesos to run spark executors

**What do you expect to happen?**

Well, I expect that work:)

**What actually happens?**

spark-submit failed, with

```
Py4JJavaError: An error occurred while calling None.org.apache.spark.api.java.JavaSparkContext.
: java.lang.UnsatisfiedLinkError: /usr/lib/libmesos-1.2.3.so: libevent_openssl-2.0.so.5: cannot open shared object file: No such file or directory
```

After digging a bit, I find the following issue:

```
docker exec -it <contaienr name> bash
jovyan@cdcc15e1ffe0:~$ ldd /usr/local/lib/libmesos.so  | grep 'not found'
        libevent_openssl-2.0.so.5 => not found
        libevent_pthreads-2.0.so.5 => not found
        libevent-2.0.so.5 => not found

jovyan@cdcc15e1ffe0:~$ dpkg -l mesos

ii  mesos                                          1.2.3-2.0.1                  amd64                        Cluster resource manager with efficient resource isolation

```

The issue is related to the ubuntu bump version. libevent-* packages are now in version` 2.1.8-stable-4build1`, so that explain the link failed (mesos expect a `2.0.x` version). Since there is no packages available for mesos on mesosphere, I guess a custom build of mesos is required. 

Regarding your feedbacks I may try to help :)

Thanks you

Bidek56:
It's a know issue because of [MESOS-8921](https://issues.apache.org/jira/browse/MESOS-8921)
I have pointed it out in the [PR641](https://github.com/jupyter/docker-stacks/pull/641) but @parente merged the PR anyway, so it will be broken till MESOS fixes it and we do another PR.
I would suggest pulling an older image from Docker Hub which contains Ubuntu 16.04, before PR641

asyd:
Ok, thanks for the feedback.

Bidek56:
[...]

PR #693, opened 2018-08-05
[WIP] Deprecate pyspark-notebook, r-notebook, Mesos support

parente:
* Removes the pyspark-notebook and r-notebook image definitions based on usage stats https://nbviewer.jupyter.org/gist/parente/316d5c242aeb484484c8 (#517, #577)
* Removes the Mesos client libraries from the all-spark-notebook as a step toward making the image Hadoop and cluster manager agnostic (#626, #663)
* Updates documentation

It's not meant to be the one and only PR for #626, only a step in that direction.


doctapp:
Why remove based on stats? We extensively use pyspark-notebook by modding it and only pulling it's base image (scipy-notebook)... Also, how will Mesos be supported if deprecating it?
Thanks

parente:
Thanks for the input @doctapp. 

> Why remove based on stats? 

Mainly to cut down on build time and maintenance cost here.

> How will Mesos be supported if deprecating it?

It wouldn't be supported here, since we (the current maintainers) have no way to test it against a Mesos cluster. We're thinking of making all-spark-notebook a cluster manager agnostic image, and rely on someone with Mesos experience and interest in the community to build a descendant image or one that could be used in conjunction with it. 

The issues linked in the description have some discussion on these topics.

doctapp:
Understood, thanks.

docs/using/specifics.md
```diff
@@ -58,111 +58,15 @@ val rdd = sc.parallelize(0 to 999)
...
-
-```scala
-// should print the value of --master in the kernel spec
-println(sc.master)
-
-// do something to prove it works
-val rdd = sc.parallelize(0 to 99999999)
-rdd.sum()
-```
-
```
taqtiqa-mark: Maybe this content in `docs/using/specifics.md` can be moved to `examples/mesos.md`?

parente: That's an option. Another is to move them out entirely. One reason for deprecating these cluster-specific notes is that we don't have a way to maintain them.

taqtiqa-mark: Ack  No problem with deleting.  Just seemed a waste to dispose of all the effort someone put into learning.  Anyway likely better suited to the Docker-recipes wiki page referenced in a comment to #626.

dungdm93:
If pyspark-notebook, r-notebook are deprecated, should we simplify `all-spark-notebook` name to `spark-notebook`?

romainx:
Hello,

Just to share some thoughts about this. It makes sense since.

- `r-notebook` could be replaced without loss of features by `datascience-notebook`
- `pyspark-notebook` could be replaced without loss of features by `all-spark-notebook`

If this decision is taken, we could

- Put a deprecation notice in the documentation for these two images -> I can take care of this point if you want.
- Maybe move them to community images in order to give to users a chance to continue using them.

Best.

parente:
I'm going to close this PR as irrelevant.

* We removed the mesos clients in another PR.
* We haven't hit issues exceeding CI time limits due to the number of images in quite some time.
* We're not running into issues hosting all of the existing images on Docker Hub.
* It's not terribly difficult to maintain the r-notebook and pyspark-notebook alongside the others.
* We already closed lll of the issues linked from the description.

If conditions change, we can reconsider. Right now, there's little benefit to doing this work and only disruption for users of the two images.

Issue #707, opened 2018-08-26
NB_USER is not applied in docker run


marcraminv:
**What docker image you are using?**

`jupyter/scipy-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**
 
```
docker run \
--rm \
-it \
--name jupyter \
-p 8888:8888 \
-e NB_USER=marc \
jupyter/scipy-notebook:latest
```

 **What steps do you take once the container is running to reproduce the issue?**

1. docker run to power the container:
```
docker run \
--rm \
-it \
--name jupyter \
-p 8888:8888 \
-e NB_USER=marc \
jupyter/scipy-notebook:latest
```
The console output:
![1](https://user-images.githubusercontent.com/13021787/44628012-b3868380-a938-11e8-899a-6c13453a9c6c.png)

2. `docker exec -it jupyter /bin/bash`
console output:
![2](https://user-images.githubusercontent.com/13021787/44628013-b8e3ce00-a938-11e8-992f-2f06353099eb.png)

3. `env`

console ouput:
![3](https://user-images.githubusercontent.com/13021787/44628016-c39e6300-a938-11e8-89ec-29b501d73ea0.png)

4. docker run with `root` user:
```
docker run \
--rm \
-it \
--name jupyter \
-p 8888:8888 \
-e NB_USER=marc \
--user root \
jupyter/scipy-notebook:latest
```
console output:
![4](https://user-images.githubusercontent.com/13021787/44628025-e761a900-a938-11e8-8ce5-c0dd093d3f25.png)

5. `docker exec -it jupyter /bin/bash`

console output:
![5](https://user-images.githubusercontent.com/13021787/44628026-edf02080-a938-11e8-8024-aaa511964247.png)

**What do you expect to happen?**

The user `marc` should be changed by joyan as documentation says.
![6](https://user-images.githubusercontent.com/13021787/44628100-1cbac680-a93a-11e8-8611-f295b7be81b0.png)

**What actually happens?**

The start script does not apply **NB_USER**. I don't know if I understood something wrong reading the documentation, but I thought that something was wrong with start script and permissions set up.

parente:
The documentation is missing the key sentence that appears next to the `NB_UID` and other options that modify the default user:

> For this option to take effect, you must run the container with --user root.

Sorry for the trouble. Happy to take a PR to update that bullet in the docs if you'd like to contribute the fix. If not, we'll get to it!

marcraminv:
[...]

PR #723, opened 2018-10-05
Update docs about NB_USER set up

marcraminv:
Try to fix this misunderstood situation #707 . thanks to @edesz and @parente for the comments


parente:
@marcraminv Thanks for the doc clarification! 🍰

Issue #727, opened 2018-10-16
FROM container as an arg


graemes:
One small change I've made at my site is to change the Dockerfiles to use an ARG for the FROM clause (see below).  This makes it simpler for me to change the base container to something other than the hardcoded entry i.e. <cough>cuda</cough> or my own registry in my build scripts.  Would there be any benefit to submitting a PR with the change made throughout the tree?

ARG BUILD_CONTAINER=jupytper/base-notebook
FROM $BUILD_CONTAINER

parente:
Seems reasonable. If you submit a PR, we can watch how the build process goes on Travis.

ericdill:
@parente closed by #731

PR #731, opened 2018-10-19
Convert FROM specification to ARG (#727)

graemes:
This changes the FROM specification in the Dockerfile to use an ARG.  This allows the base container to be overridden in scripts so that local registries may be used (for example).


parente:
Appears everything went right on Travis. Let's give it a whirl. Thanks @graemes. 🍰

PR #733, opened 2018-10-23
Fix the common options link in the docs

bskarda:


bskarda:
Looks like like the travis build failed because of a 503 from  https://conda-forge.github.io/feedstocks. If an admin can kick off another travis run it should succeed.

parente:
Thanks @bskarda. 🍰

PR #739, opened 2018-10-26
Add note about customizing the startup with before-notebook.d

### LINKED ISSUE #715: Update docs for .d folders
https://github.com/jupyter/docker-stacks/pull/688 introduced `/usr/local/bin/before-notebook.d/`.

We should update the docs to note the current behavior.

ericdill:
Closes #715

Following the suggestion from @parente in #715. Minimal documentation to start since the use case seems pretty self-explanatory. I'm happy to add more prose if necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,4 @@ dockerspawner
 dockerspawner.tar.gz
 *.orig
 .ipynb_checkpoints/
+.vscode/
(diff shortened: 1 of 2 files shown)


parente:
I worked some minor updates into this PR about the other hooks, `start-notebook.d` to get it over the finish line. Merging now. Further improvements welcome in follow-on PRs.

ericdill:
Thanks for pushing this across the finish line @parente

masip85:
I need to make an assumption after reading start.sh. , that  /usr/local/bin/before-notebook.d exists because its added somwhow. So i GUESS my scripts must go in hooks folder, but after trying and failing I guess my point need more accurate info to success. 
I have tried an script under hooks folder and also under hooks/before-notebook.d folder.

romainx:
Hello @masip85 here is an example that uses the hook.

https://github.com/jupyter/docker-stacks/blob/fad26c25b8b2e8b029f582c0bdae4cba5db95dc6/pyspark-notebook/Dockerfile#L48-L51

So you need to create the directory `/usr/local/bin/before-notebook.d`.

Best

Issue #749, opened 2018-11-06
datascience-notebook: for julia 1.0, JULIA_PKGDIR is replaced with JULIA_DEPOT_PATH


xuwaters:
## datascience-notebook

for julia 1.0, `JULIA_PKGDIR` environment variable is not working now
packages are still installed under `$HOME/.julia` folder
which causes julia packages could not be found when mounting external volume to image home

so the solution is add another environment variable
```
ENV JULIA_DEPOT_PATH=/opt/julia
```

for more information, please refer to:

- https://github.com/JuliaLang/Pkg.jl/issues/216
- https://docs.julialang.org/en/v1/stdlib/Pkg/index.html#Glossary-1

PR #750, opened 2018-11-06
fix: #749; datascience-notebook: for julia 1.0, JULIA_PKGDIR is repla…

xuwaters:
fix: #749; datascience-notebook: for julia 1.0, JULIA_PKGDIR is replaced with JULIA_DEPOT_PATH


parente:
Thanks, @xuwaters. 🍰

Issue #675, opened 2018-07-17
Adding a spellchecker


avitase:
could you add one of the spellchecker plugins?

parente:
This issue is old, but I'd like to leave it open as a good starting point for someone who wants to contribute here.

mathbunnyru:
There is a spellchecker for jupyterlab and the documentation is quite good about installing it, so I'm closing this issue.
https://github.com/jupyterlab-contrib/spellchecker

PR #755, opened 2018-11-08
Adding recipe for nbextension spellchecker for issue 675

jussmi:
Adding recipe for including nbextensions (specifically spellchecker) per https://github.com/jupyter/docker-stacks/issues/675

The nbextension plugin does not cover jupyterlab, there is an option for that here https://github.com/ijmbarr/jupyterlab_spellchecker I haven't included.

Tested on Windows using minimal-notebook base image


ericdill:
✅ I can confirm that this Dockerfile produces a working docker image with the spellchecker enabled.

parente:
Thanks @jussmi and @ericdill

PR #756, opened 2018-11-08
Bumped Spark to 2.4.0 and miniconda to 4.5.11

Bidek56:
Upgrading Spark to 2.4.0

PR #757, opened 2018-11-08
Bumped Spark to 2.4.0

borregore:


pyspark-notebook/Dockerfile
```diff
@@ -8,7 +8,7 @@ LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 USER root
 
 # Spark dependencies
-ENV APACHE_SPARK_VERSION 2.3.1
+ENV APACHE_SPARK_VERSION 2.4.0
```
parente: Thanks. If py4j 0.10.8.1 is required, line 42 `ENV PYTHONPATH $SPARK_HOME/python:$SPARK_HOME/python/lib/py4j-0.10.7-src.zip` needs to be updated as well.

borregore: Looks like there is no problem in the time to build.

borregore: though I don't know the exact way to test Spark.

Bidek56: You can update the path but if the .zip does NOT exists, Spark won't run. Please pull this PR before it gets merged and it breaks Spark for everyone.

parente: It's OK @Bidek56. We'll make sure the right tarballs are in place before it merges.

@borregore the way I usually check is to go to the Spark website, download the release manually, unpack it, and see what version it includes. Then I make the Dockerfile match.

Bidek56: Or test the container that you build to make sure it works? :)
I am not sure that py4J 0.10.8.1 will work with Spark 2.4.0
PR #756 has Spark 2.4 and MINICONDA_VERSION bump, it maybe a better solution.

parente: @borregore is a new contributor so I'd like to take the opportunity to work through the pull request submission and code review cycle. I appreciate the PR you also submitted, @Bidek56, but would like to see this one through since @borregore opened the original issue too.

I've pushed a commit to this PR to switch the py4j zip back to 0.10.7. Once the build passes, we can merge.

borregore: Why are the tests not passing if a previous test with the same content did pass?

Bidek56: Did you check the logs? 
The build process is very fragile b/c it pulls a lot of packages from different sources.

  Could not install requirement https://dist.apache.org/repos/dist/dev/incubator/toree/0.3.0-incubating-rc1/toree-pip/toree-0.3.0.tar.gz because of error 404 Client Error: Not Found for url: https://dist.apache.org/repos/dist/dev/incubator/toree/0.3.0-incubating-rc1/toree-pip/toree-0.3.0.tar.gz
Could not install requirement https://dist.apache.org/repos/dist/dev/incubator/toree/0.3.0-incubating-rc1/toree-pip/toree-0.3.0.tar.gz because of HTTP error 404 Client Error: Not Found for url:

Bidek56: It looks like toree needs to use a mirror not apache.org

parente:
The build is failing due to the Toree 0.3.0 RC1 being removed from the apache servers. #762 fixes this problem.

parente:
Let's get this in and deal with the fallout in another PR if Toree 0.3.0 final is not Spark 2.4.0 compatible.  Thanks for hanging in there @borregore and @Bidek56 while I've been off doing other Jupyter things.

leobenkel:
I am looking for the jupyter container for spark 2.3.1, which one would it be https://hub.docker.com/r/jupyter/all-spark-notebook/tags?page=3 ?

leobenkel:
I think `50d1eb9ec2d8` is the latest one before spark 2.4.0

PR #759, opened 2018-11-11
fixing inline script problem

dungdm93:
In some special use-cases, I'd like to run `start.sh` with inline script (for example in GitLab-CI), then I got bellow error:
screenshot from 2018-11-11 10-46-43
After some workaround, I figure out that problem lead from [`cmd=$*`](https://github.com/jupyter/docker-stacks/blob/50d1eb9ec2d8ad559bf34e51ac4d680266af4bc8/base-notebook/start.sh#L11) expand all parameter into a single string. Here explain why:
```bash
$ cat ./script.sh
#!/bin/bash

foo=$@ # or even $*, "$@", "$*"
bar=( "$@" )

echo "--- original ----------------"
for param in "$@";        do echo "[$param]"; done
echo "--- foo ---------------------"
for param in "$foo";      do echo "[$param]"; done
echo "--- bar ---------------------"
for param in "${bar[@]}"; do echo "[$param]"; done

$ ./script.sh echo "Hello world
> I'm superman"
--- original ----------------
[echo]
[Hello world
I'm superman]
--- foo ---------------------
[echo Hello world
I'm superman]
--- bar ---------------------
[echo]
[Hello world
I'm superman]
```
You see, `$@` is an array, when assign to `foo`, it become a string.


parente:
@dungdm93 I hope to give this a a review in the next few days.

dungdm93:
I also update docs to a bit more clearer for you :)

base-notebook/start.sh
```diff
@@ -6,32 +6,32 @@ set -e
...
     # Source scripts or run executable files in a directory
     if [[ ! -d "$1" ]] ; then
-	return
+        return
     fi
     echo "$0: running hooks in $1"
-    for f in "$1"/*; do
-	case "$f" in
-	    *.sh)
-		echo "$0: running $f"
```
dungdm93: Change indent from tabs to spaces

PR #773, opened 2018-12-07
Update sparklyr ->0.9.3 for Spark 2.4

zgfh:


parente:
Looks like https://github.com/conda-forge/r-sparklyr-feedstock/blob/master/recipe/meta.yaml needs to be updated for 0.9.3 first.

parente:
Noarch will work if it’s now available. Looks like it appeared 4 days ago.

parente:
Here goes the build again ...

Issue #776, opened 2018-12-12
apt-get clean redundant in these ubuntu images


consideRatio:
Apparently apt-get clean is a redundant command if using an official ubuntu image as we are doing, so at worst we have additional time:

> In addition, when you clean up the apt cache by removing /var/lib/apt/lists it reduces the image size, since the apt cache is not stored in a layer. Since the RUN statement starts with apt-get update, the package cache is always refreshed prior to apt-get install.
> 
> Official Debian and Ubuntu images automatically run apt-get clean, so explicit invocation is not required.

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

Feel free to close this without further action!

parente:
Fixed in #777.

PR #777, opened 2018-12-17
Remove apt-get clean command. Unecessary on official Ubuntu images.

tryantwit:
References issue #776 

Removing apt-get clean commands from Dockerfiles.

Ran build against base-book with and without apt-get clean line and both came out to 738MB build.


parente:
@tryantwit Thanks for submitting this PR and checking the results. Let's :shipit: 

🍰

PR #781, opened 2018-12-28
umask flag

parente:
Squash, merge conflict fix, and removal of default umask setting in PR #702.


parente:
@iyanmv I believe this PR completes the final change we discussed in #702, removing the default umask value from the Dockerfile. I decided to squash the other PR since it had gone stale, but preserved your commit messages in the squash commit.

iyanmv:
Ok, perfect @parente! To be honest, I went back to study Physics and I also lost track of this PR. Thanks for putting it back ;)

PR #784, opened 2018-12-29
Fix bug where container user is "nayvoj"

parente:
`whoami &> /dev/null || STATUS=$? && true` causes STATUS to be set to an empty string when the container starts with option `--add-group="root"` resulting in both `nayvoj` and `jovyan` having `uid=1000` and `gid=100`. The first match in `/etc/passwd` wins and so the container user ends up being `nayvoj` accidentally.

Avoid this by checking that `whoami`'s `STATUS` is neither blank nor 0 before adding a new `/etc/passwd` entry

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -110,7 +110,7 @@ else
         # container runs as. Check that the user has an entry in the passwd
         # file and if not add an entry.
         whoami &> /dev/null || STATUS=$? && true
-        if [[ "$STATUS" != "0" ]]; then
+        if [[ -n "$STATUS" && "$STATUS" != "0" ]]; then
             if [[ -w /etc/passwd ]]; then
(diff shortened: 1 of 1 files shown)


parente:
/cc @GrahamDumpleton if you'd like to give a quick review. I stumbled upon this minor bug by running the following while reviewing another PR.:

```
docker run -it --group-add root --rm jupyter/base-notebook start.sh
```

GrahamDumpleton:
Perhaps use:

```
STATUS=0 && whoami &> /dev/null || STATUS=$? && true
```

instead just to ensure ``STATUS`` is set with a value to begin with. That eliminates problems in very unlikely scenario that ``STATUS`` was set as environment variable to container.

GrahamDumpleton:
LGTM

parente:
Thanks for the look-see @GrahamDumpleton

PR #787, opened 2018-12-29
[WIP] Have start.sh run through both user and root parts

rkdarst:
I'm using docker-stacks in a jupyterhub deployment.  I run as uid=0, and set NB_UID, NB_USER, etc.  I do extensive local customization at run time.  I even have a switch to swap between running as uid=0 with NB_UID and NB_UID directly.

One problem is that sometimes the hooks may run as root, and sometimes they may run as the user.  This could be handled with conditionals, but I'd rather more built-in flexibility.

The bigger issue is that sometimes, as running with uid=0, I want to be able to run hooks as the user.  (example: filesystems are mounted from NFS with root_squash, so root *can't* access and do the necessary things at all).  I've hacked this by having the root hooks run sudo themselves, but I wish there was a better way.  There are many other hacks I am ending up doing and I just feel there has to be a better way.

This PR adds one hack to replace many (but is not ready to use yet).  If `start.sh` is running as root, instead of starting `jupyter` at the end using sudo, re-execute the same `start.sh` script using sudo.  This way, the user-specific setup gets run (in its current form, this does nothing, but in the future).

As an initial useful feature, there are now six forms of hooks: `start-notebook.d`, `start-notebook-root.d`, `start-notebook-user.d`, and `before-notebook-{,root,user}.d`.  If this was accepted, most of my current hacks could be replaced.  As another example of what could be done, instead of doing system setup in `jupyter_notebook_config.py` where it doesn't belong, umask can easily be set here regardless of if you run as user or root (this would have helped me a lot six months ago).  Similar for generating a certificate.

But there are problems:
* This is a hack.  A much better way would be to split `start.sh` into a user and root part, maybe `start-singleuser.sh` could have some of this logic.
* A larger rearrangement would make more use of this.
* Is it worth making such a massive change for specialized use?  I think probably so, because it makes hooks more useful and thus ensure that special use cases are made easier, with fewer upstream changes, later.

I would say do *not* accept this PR yet, but it serves as a basis for discussion to see if the concept is useful.  If useful, discuss the right implementation and do that instead.

=====

Before, start.sh would run *either* its root or its user part.  If you
were running as uid=0 but wanted to run a user hook before starting,
it can't easily be done.  Also scripts in start-notebook.d and
before-notebook.d could be run as either the user or root, depending
on how things are run.

This is a simple, but unelegant solution, which re-invokes start.sh as
the user, if it was originally invoked as root.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -2,6 +2,21 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+# This script can be executed as either a user or as root.  If it's
+# executed as root, then it does the basic setup and re-executes it as
+# a user, so that user-specific setup can also be done if any (this
+# would mainly be in the form of user hooks).
+
+# Default hook directories:
+# - start-notebook is before this script runs
+# - before-notebook is right before the Jupyter commmand is executed
+# /usr/local/bin/start-notebook.d/     - BOTH user and root, backwards compat
+# /usr/local/bin/start-notebook-root.d/
+# /usr/local/bin/start-notebook-user.d/
+# /usr/local/bin/before-notebook.d/    - BOTH user and root, backwards compat
+# /usr/local/bin/before-notebook-root.d/
+# /usr/local/bin/before-notebook-user.d/
+
 set -e
 
 # Exec the specified command or fall back on bash
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -40,6 +40,7 @@ run-hooks /usr/local/bin/start-notebook.d
 
 # Handle special flags if we're root
 if [ $(id -u) == 0 ] ; then
+    run-hooks /usr/local/bin/start-notebook-root.d
 
     # Only attempt to change the jovyan username if it exists
     if id jovyan &> /dev/null ; then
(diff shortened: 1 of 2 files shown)


parente:
@rkdarst Thanks for the write up. My initial reaction is that it makes sense to support additional "hook points" and that we need to refactor `start.sh` to make those additions (and even what currently exists) more maintainable. We might want to consider adding more tests to avoid regressions before starting such a refactor.

rkdarst:
Thank for the review.  I'm glad I'm not completely off here.

So, I've been using this on my own deployment and it seems to work.  I'm also a bit surprised that the tests are passing here.

Realistically, I probably won't have time to make bigger changes - and I am probably not the best person to do it anyway since I don't have a good dev environment set up and am not sure how the tests work.  And I don't want to start a branch if I can't finish it quickly

I think if I did do something, I would split `start.sh` into a user and root portion, root portion calls user portion, and then `start-notebook.sh` calls the right one depending on the current uid.

rkdarst:
This has been inactive for a while, so I thought I would update you.  I have been using the original change for a year.  The ability to have both user and root hooks has been really useful to avoid hacks.  But I guess this isn't equally useful for everyone, it's mainly needed in large-scale, customizable deployments, not for day to day local users.

I've rebased the PR to the current master, and also added some documentation.  I can think of one thing that should be done first: if running as root, the before-notebook.d and start-notebook.d scripts should be only run once, as root.  As the PR is now, the hooks will run as both root and the user (before, they could run as root *or* the user).

What do you think?  Should this go on?

rkdarst:
I think to do proper tests (beyond what's already done), we would have to add a way to mount sample scripts into the container during the test.

maresb:
FWIW, I personally think this makes a lot of sense and would be useful.

After looking at the implementation, it's not clear to me what is the purpose of `bash "$BASH_SOURCE"` and how it works.

rkdarst:
> After looking at the implementation, it's not clear to me what is the purpose of `bash "$BASH_SOURCE"` and how it works.

It serves the purpose of `bash $0` - re-executing the same file.  But since this file is sourced from `start-notebook.sh`, not executed, so `$0` was probably `start-notebook.sh` and not `start.sh`.  But `$BASH_SOURCE` probably worked (I say probably, because I haven't gone and re-verified this now - but this is probably correct).

parente:
I'm going to close this PR since it's still marked WIP and hasn't been touched in a while, and I think it's a subset of the work proposed in #1055 and #1052.

PR #786, opened 2018-12-29
Updated R packages in r-notebook to match all-spark-notebook update in PR #782

romainx:
Updated R packages in `r-notebook` to match `all-spark-notebook` update in PR #782

```bash
# First update to be 
r-base=3.4.1 -> 3.5.1
r-sparklyr=0.7* -> 0.9*
# Then to solve conflicts
r-shiny=1.0* -> 1.2*
r-htmlwidgets=1.0* -> 1.2*
r-nycflights13=0.2 -> 1.0*
r-rsqlite=2.0* -> 2.1*
r-tidyverse=1.1* -> 1.2*
r-rmarkdown=1.8* -> 1.11*
```

PR #789, opened 2018-12-30
Updated R packages in datascience-notebook to match r-notebook packages update in PR #786

romainx:
Updated R packages in `datascience-notebook` to match `r-notebook` packages update in PR #786
Note: only  the `rpy2` package (also updated here) is in the `datascience-notebook` and not in  `r-notebook`, **other packages are identical**. 

We should think of finding a way to help us maintain the coherence of the images between them (at least some documentation). Let me know if you have an idea.

### DIFF
--- a/datascience-notebook/Dockerfile
+++ b/datascience-notebook/Dockerfile
@@ -46,7 +46,7 @@ USER $NB_UID
 
 # R packages including IRKernel which gets installed globally.
 RUN conda install --quiet --yes \
-    'rpy2=2.8*' \
+    'rpy2=2.9*' \
     'r-base=3.5.1' \
     'r-irkernel=0.8*' \
     'r-plyr=1.8*' \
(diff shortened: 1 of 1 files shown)


parente:
There are hooks for running pytest after a build of each image. Maybe extend that concept to do some basic validation before each build, like checking versions by inspecting Dockerfiles?

We considered a Dockerfile code generator in the past but shied away to make contributions easier for newcomers: update version numbers in familiar Dockerfiles not versions in some unknown jinja or custom code setup.

romainx:
I understand. The current solution is simple, robust self documented and do not need a complex setup (Jinja or a shared yaml env. file may be overkill). Your idea of a post-validation sounds good. We could check with a kind of linter if all versions of the same package are the same and raise a warning if not. I will have a look. Thanks for the hint.

PR #792, opened 2019-01-04
conda install raise PermissionError while run container with root

dungdm93:
I run container with `root` user, like:
```shell
docker run --user root \
    -p 8888:8888 \
    -e JUPYTER_ENABLE_LAB="true" \
    jupyter/base-notebook
```
And when I try to install some `conda` packages inside container, I got following error:
```log
jovyan@wrnvfljfs$ conda install psycopg2
Solving environment: failed

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

      File "/opt/conda/lib/python3.7/site-packages/conda/exceptions.py", line 819, in __call__
        return func(*args, **kwargs)
      File "/opt/conda/lib/python3.7/site-packages/conda/cli/main.py", line 78, in _main
        exit_code = do_call(args, p)
      File "/opt/conda/lib/python3.7/site-packages/conda/cli/conda_argparse.py", line 77, in do_call
        exit_code = getattr(module, func_name)(args, parser)
      File "/opt/conda/lib/python3.7/site-packages/conda/cli/main_install.py", line 11, in execute
        install(args, parser, 'install')
      File "/opt/conda/lib/python3.7/site-packages/conda/cli/install.py", line 235, in install
        force_reinstall=context.force,
      File "/opt/conda/lib/python3.7/site-packages/conda/core/solve.py", line 518, in solve_for_transaction
        force_remove, force_reinstall)
      File "/opt/conda/lib/python3.7/site-packages/conda/core/solve.py", line 451, in solve_for_diff
        final_precs = self.solve_final_state(deps_modifier, prune, ignore_pinned, force_remove)
      File "/opt/conda/lib/python3.7/site-packages/conda/core/solve.py", line 180, in solve_final_state
        index, r = self._prepare(prepared_specs)
      File "/opt/conda/lib/python3.7/site-packages/conda/core/solve.py", line 592, in _prepare
        self.subdirs, prepared_specs)
      File "/opt/conda/lib/python3.7/site-packages/conda/core/index.py", line 174, in get_reduced_index
        subdir_datas = tuple(SubdirData(Channel(url)) for url in channel_urls)
      File "/opt/conda/lib/python3.7/site-packages/conda/core/index.py", line 174, in <genexpr>
        subdir_datas = tuple(SubdirData(Channel(url)) for url in channel_urls)
      File "/opt/conda/lib/python3.7/site-packages/conda/core/subdir_data.py", line 69, in __call__
        subdir_data_instance = super(SubdirDataType, cls).__call__(channel)
      File "/opt/conda/lib/python3.7/site-packages/conda/core/subdir_data.py", line 131, in __init__
        self.cache_path_base = join(create_cache_dir(),
      File "/opt/conda/lib/python3.7/site-packages/conda/core/subdir_data.py", line 639, in create_cache_dir
        mkdir_p_sudo_safe(cache_dir)
      File "/opt/conda/lib/python3.7/site-packages/conda/gateways/disk/__init__.py", line 81, in mkdir_p_sudo_safe
        os.chown(path, uid, gid)
    PermissionError: [Errno 1] Operation not permitted: '/opt/conda/pkgs/cache'

`$ /opt/conda/bin/conda install psycopg2`

  environment variables:
                 CIO_TEST=<not set>
                CONDA_DIR=/opt/conda
               CONDA_ROOT=/opt/conda
                     PATH=/opt/conda/bin:/opt/conda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin
                          :/usr/bin:/sbin:/bin
               PYTHONPATH=
       REQUESTS_CA_BUNDLE=<not set>
            SSL_CERT_FILE=<not set>
             SUDO_COMMAND=/opt/conda/bin/jupyter lab
                 SUDO_GID=0
                 SUDO_UID=0
                SUDO_USER=root

     active environment : None
       user config file : /home/jovyan/.condarc
 populated config files : /opt/conda/.condarc
          conda version : 4.5.12
    conda-build version : not installed
         python version : 3.7.1.final.0
       base environment : /opt/conda  (writable)
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/linux-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/pro/linux-64
                          https://repo.anaconda.com/pkgs/pro/noarch
          package cache : /opt/conda/pkgs
                          /home/jovyan/.conda/pkgs
       envs directories : /opt/conda/envs
                          /home/jovyan/.conda/envs
               platform : linux-64
             user-agent : conda/4.5.12 requests/2.21.0 CPython/3.7.1 Linux/4.15.0-43-generic ubuntu/18.04 glibc/2.27
                UID:GID : 1000:100
             netrc file : None
           offline mode : False


An unexpected error has occurred. Conda has prepared the above report.
If submitted, this report will be used by core maintainers to improve
future releases of conda.
Would you like conda to send this report to the core maintainers?
[y/N]:
No report sent. To permanently opt-out, use

    $ conda config --set report_errors false
```
After some investigations, I figure out this error cause from this line of code:
```
sed -r "s#Defaults\s+secure_path=\"([^\"]+)\"#Defaults secure_path=\"\1:$CONDA_DIR/bin\"#" /etc/sudoers | grep secure_path > /etc/sudoers.d/path
```
I know that `sudo` lookup command from `secure_path` instead of current `$PATH` variable for security reason. And side effect of above action is it also protect your `$CONDA_DIR/bin`, so you can't update this path even you are its owner.

So, what is solution?
Simple: DON'T use `sudo`, use [`gosu`](https://github.com/tianon/gosu). It's very common in Docker world :)


rkdarst:
I had a similar problem, and if I try `conda install` a second time, it works.  In the end I found it was that `/opt/conda/pkgs/cache` did not exist, it would try to create and chown, and chown would fail.  Looking at your error message above, it fails there too.  My workaround was to pre-mkdir `/opt/conda/pkgs/cache`, and I thought it had something to do with my particular filesystems and/or user management so didn't make any upstream reports.

I am just wondering... how does protecting `/opt/conda/bin/` relate to `/opt/conda/pkgs/cache`?  Was my problem something else?  And it seems my path isn't protected.  Is this a newer sudo?

dungdm93:
@rkdarst I'm not sure about your problem, but you can try my solution.
And about `secure_path`, no `/opt/conda/bin/` in `secure_path` any more. Why should we protect this path when each person run on his own container, and furthermore his also is the owner of `$CONDA_DIR`?

base-notebook/start.sh
```diff
@@ -95,14 +95,11 @@ if [ $(id -u) == 0 ] ; then
         echo "$NB_USER ALL=(ALL) NOPASSWD:ALL" > /etc/sudoers.d/notebook
     fi
 
-    # Add $CONDA_DIR/bin to sudo secure_path
-    sed -r "s#Defaults\s+secure_path=\"([^\"]+)\"#Defaults secure_path=\"\1:$CONDA_DIR/bin\"#" /etc/sudoers | grep secure_path > /etc/sudoers.d/path
```
parente: If the user wants to allow the unprivileged `jovyan` user to become root by specifying `GRANT_SUDO=yes`, will they still be able to use `sudo conda install` in the container after this change?

base-notebook/test/test_container_options.py
```diff
@@ -74,31 +74,6 @@ def test_sudo(container):
...
-    c = container.run(
-        tty=True,
-        user='root',
-        command=['start.sh', 'which', 'jupyter']
-    )
-    rv = c.wait(timeout=10)
-    assert rv == 0 or rv["StatusCode"] == 0
-    assert c.logs(stdout=True).decode('utf-8').rstrip().endswith('/opt/conda/bin/jupyter')
-
-
```
parente: I vaguely recall we used `su` or `gosu` in the past in these images and ran into problems reported in issues and corrected in past PRs. If we make this switch, I think we need to at least retain the spirit of the tests that were in place to check behavior with and without `GRANT_SUDO=yes`.

base-notebook/start.sh
parente: Also, if I'm missing how the switch to `gosu` removes the need for the `GRANT_SUDO=yes` flag allowing the `jovyan` user in the container to `sudo`, please let me know. As it stands, it looks like the gosu change only impacts how we initially switch to the `jovyan` user when the container starts as root, but the secure path issue is really in place to allow `sudo` back to root later one.

dungdm93: Thank you for your comment.
Now it worked with `sudo conda install` by adding `conda` symbolic link to `/usr/local/bin/`
```bash
ln -t /usr/local/bin/ -s \
  $CONDA_DIR/bin/conda \
  $CONDA_DIR/bin/pip \
  $CONDA_DIR/bin/python
```
In my opinion, this way is more secure than put conda bin to `secure_path`.

base-notebook/test/test_container_options.py
dungdm93: > I vaguely recall we used su or gosu in the past in these images and ran into problems reported in issues and corrected in past PRs.
Could you give me a reference for that issues?

rkdarst:
Um, another question about this... the image is designed so that `sudo` is not required to install packages as a user.  That doesn't work, so we are trying to use sudo.  sudo doesn't work, we are trying to replace sudo.  Aren't we currently working two steps away from the root problem?

I was able to replace the original problem with this run command: `docker run --user=root -e NB_USER=aoeu -e NB_UID=1001 -e GRANT_SUDO=yes -it -p 8888:8888 jupyter/scipy-notebook:latest`.  Then:

```
$ conda install plotly
...
An unexpected error has occurred. Conda has prepared the above report.
If submitted, this report will be used by core maintainers to improve
future releases of conda.

### but it should work... let's try again, right away:
$ conda install plotly
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
### um... it works now.  why?
```

Here is what I found works:
```
$ mkdir /opt/conda/pkgs/cache
$ fix-permissions /opt/conda/
$ conda install plotly
...
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
```

The actual problem is this chown, which happens inside conda as part of it's creation of the cache directory (check the original error message):
```
        os.chown('/opt/conda/pkgs/cache', uid, gid)
```
... it seems like it is being fixed https://github.com/conda/conda/pull/7180, though it does some other things and I'm not sure if it will completely fix things.

If sudo *did* protect `secure_path` from modifications, package managers wouldn't work.  I haven't found any documententation that "And side effect of above action is it also protect your $CONDA_DIR/bin, so you can't update this path even you are its owner." is the case.  Is there?  I'd be very interested.

parente:
Let's start with the simple fix of creating `/opt/conda/pkgs/cache` at startup, or possibly at build time if `fix-permissions` corrects for the `NB_UID` and `NB_USER` at runtime.  The upstream conda issue @rkdarst mentioned might be fixed in conda 4.6 so we can try to update and see if the issue goes away as well.

dungdm93:
@rkdarst You are right.
But all you guys think that we should use `gosu` instead of `sudo` according to [Docker best practice](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user).
> Avoid installing or using sudo as it has unpredictable TTY and signal-forwarding behavior that can cause problems. If you absolutely need functionality similar to sudo, such as initializing the daemon as root but running it as non-root), consider using “gosu”.

parente:
It looks like conda 4.6.x fixed the issue as expected. The commands noted at the top of this PR now succeed with the latest build of base-notebook.

```
docker run --user root \
    -p 8888:8888 \
    -e JUPYTER_ENABLE_LAB="true" \
    jupyter/base-notebook
```

```
jovyan@d4863479b394:~$ conda install psycopg2
Collecting package metadata: done
Solving environment: done


==> WARNING: A newer version of conda exists. <==
  current version: 4.6.7
  latest version: 4.6.8

Please update conda by running

    $ conda update -n base conda



## Package Plan ##

  environment location: /opt/conda

  added / updated specs:
    - psycopg2


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    libpq-11.2                 |       h4770945_0         2.4 MB  conda-forge
    openssl-1.1.1b             |       h14c3975_1         4.0 MB  conda-forge
    psycopg2-2.7.7             |   py37h9d9f1b6_1         304 KB  conda-forge
    ------------------------------------------------------------
                                           Total:         6.7 MB

The following NEW packages will be INSTALLED:

  libpq              conda-forge/linux-64::libpq-11.2-h4770945_0
  psycopg2           conda-forge/linux-64::psycopg2-2.7.7-py37h9d9f1b6_1

The following packages will be UPDATED:

  openssl                                 1.1.1b-h14c3975_0 --> 1.1.1b-h14c3975_1


Proceed ([y]/n)? y
```

Thanks for the discussion about the issue. I'm going to close this PR.

To entertain a switch from `sudo` to `gosu`, I think we first need better test coverage of the base image behaviors that depend on `sudo`. I'll happily review and merge additional tests if someone wants to tackle that first step, and alter follow-up with a PR switching the startup script over to using `gosu`.

PR #795, opened 2019-01-06
r-cran-rodbc is required for ODBC connection

heoa:
Further information [here](https://superuser.com/questions/283272/problem-with-rodbc-installation-in-ubuntu/283274) and [here](https://stackoverflow.com/questions/54057120/odbc-driver-for-jupyter-r-notebook-based-on-jupyter-docker-stack).


heoa:
Fixes in the newest request below.

https://github.com/jupyter/docker-stacks/pull/796

PR #796, opened 2019-01-07
Support for ODBC

heoa:
I recommend to add the support for ODBC.


ericdill:
> ODBC support is generally useful

agreed.

👍 from me here. This should definitely be in either the docs or just in the base image. Since there's already the PR to add it to the base image this seems like the right path forward.

parente:
No objections. In it goes!

Issue #801, opened 2019-01-28
Command in docs outdated (port moved?)


winni2k:
**What docker image you are using?**

`jupyter/datascience-notebook:08e09e60eb05`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

```
docker run --rm -p 10000:8888 -e JUPYTER_ENABLE_LAB=yes -v "$PWD":/work jupyter/datascience-notebook:08e09e60eb05
```

**What steps do you take once the container is running to reproduce the issue?**

1. Visit `http://http://127.0.0.1:8888/?token=71dd69ca0175184cbaa49d395c728e66758e627e2d7a0e56`

**What do you expect to happen?**

Jupyter notebook appears in the browser.

**What actually happens?**

The browser displays the content:
```
This site can’t be reached 127.0.0.1 refused to connect.
Try:

Checking the connection
Checking the proxy and the firewall
ERR_CONNECTION_REFUSED
```

**Proposed fix**

When I change the port command from `-p 10000:8888` to `-p 8888:8888` then I can connect to the jupyter notebook in my browser as expected. My suggestion is to update the command given in *example_2* on https://jupyter-docker-stacks.readthedocs.io/en/latest/index.html to use a recent docker image and the port argument `-p 8888:8888`.

parente:
The mismatch between what Jupyter Notebook prints as the connection URL and the port you configured arises because of how Docker port mapping works. In this case, the notebook process in the  Docker container doesn't know that it's internal port might is 8888 but is in fact mapped externally as 10000.

Visiting http://localhost:10000 should in fact work with that example, but I understand how it is confusing, especially with the notebook server logging that it's on port 10000. You're right that we should probably switch those initial examples to `-p 8888:8888` and delay the confusion until a user tries a different port mapping. At the point that they are changing port mappings, they're more likely to understand what's happening.

Would you like to send a PR to the docs and README? I think the example appears in both.

winni2k:
Sure!

As an aside: It is funny to me how completely confused I was about this example. I did not even consider that the `-p` port mappings work the other way around from what I thought...

parente:
I believe this issue has been addressed by the linked PR. I'm going to close it out. Feel free to send additional PRs if you find other ways to improve the documentation. Thanks again.

PR #802, opened 2019-01-28
Add intermediate example to describe port mapping

winni2k:
See #801


parente:
Thanks for the PR. Let's :shipit: .

FWIW, the same 2 examples appear in the README.md in the root of the repo. If you're feeling up to it and would like to change those to match, feel free to send another PR.

Issue #807, opened 2019-02-08
Build error for base-notebook


ASaragga:
Trying to build 'jupyter/base-notebook' (commit 992d83c) on macOS v10.14, I am getting the following error: 
```
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... failed
ERROR conda.core.link:_execute(544): An error occurred while installing package 'conda-forge::jupyterlab-0.35.4-py37_0'.
PermissionError(13, 'Permission denied')
Attempting to roll back.
```
This happens both locally and with Travis
Thank you

zchunhai:
I have same error with old version:

```
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... failed
ERROR conda.core.link:_execute(544): An error occurred while installing package 'conda-forge::jupyterlab-0.34.12-py36_1000'.
PermissionError(13, 'Permission denied')
Attempting to roll back.

Rolling back transaction: ...working... done

[Errno 13] Permission denied: '/opt/conda0c3esols'
```

parente:
Thanks for reporting. Something regressed about a week back as the daily cron build on Travis started failing too. I haven’t had cycles to investigate. 

Updating to latest conda and jupyterlab are two things to try if someone wants to dig in before I get to it.

CMobley7:
I updated the image to the latest miniconda3 and jupyterlab packages and still got the same error

parente:
Thanks for trying that. Next thing to check would be to spawn a container off the last successful local build layer and see if something is lingering in /opt/conda that is not owned by the `$NB_USER`. The `fix-permissions` script in the previous `RUN` command should be tidying everything up, but perhaps not anymore.

Another possibility is that a directory that the jupyterlab package wants to write into (e.g., /opt/conda/share/jupyter` get created in an earlier step by the `root` user now so the package install fails when it tries to write in there.

CMobley7:
So, earlier in the build process I got 

```chgrp: changing group of '/opt/conda/bin/tini': Operation not permitted
chgrp: changing group of '/opt/conda/conda-meta/pinned': Operation not permitted
chgrp: changing group of '/opt/conda/conda-meta/tini-0.18.0-h14c3975_1001.json': Operation not permitted
chgrp: changing group of '/opt/conda/pkgs/tini-0.18.0-h14c3975_1001': Operation not permitted
[...]

PR #810, opened 2019-02-14
Fix for build error #807

twalcari:
Fix permissions in parent dir of $CONDA_DIR instead of $CONDA_DIR


parente:
@twalcari Thank you for digging into the problem. It's surprising that conda now writes something outside its own directory. I wonder what may have changed.

At any rate, thanks for the PR to fix. 🍰

PR #814, opened 2019-02-20
Made variable reference PYTHONPATH safe in case of set -eu.

javabrett:
I noticed this when a before-notebook hook script `set -eu` and didn't unset them.  `PYTHONPATH` wasn't set, resulting in:

`/usr/local/bin/start.sh: line 105: PYTHONPATH: unbound variable`

If we dereference as `PYTHONPATH=${PYTHONPATH:-}` this problem would be avoided, without any functionality changes.  The other variables in the same place seem less likely to be unset.


parente:
Thanks for the addition. 🍰

PR #829, opened 2019-03-17
[ci skip] Improve manifests, update index page

parente:
Continuation of #828


parente:
More QA over in https://github.com/parente/sandbox/wiki. Merging to see what transpires ...

PR #833, opened 2019-03-20
chown: add logging and a test

javabrett:
This adds logging-output for each `CHOWN_EXTRA` item, as well as mentioning `CHOWN_HOME_OPTS` and `CHOWN_EXTRA_OPTS` in output.

Added a test for `CHOWN_EXTRA`.


parente:
Very nice. Thanks for the test and log improvement.

PR #836, opened 2019-03-29
Added s3 + spark session instructions

ying-w:
Instructions seemed outdated so I've updated it to use spark session.

I couldn't find a way to set hadoopConf from sparksession

Setting `aws-java-sdk` in `--packages` shouldn't be necessary since dependency resolution of `hadoop-aws` should grab the proper version, however, the version required is quite old (march 2014). I tried a newer jar but got an error that looked like [this](https://stackoverflow.com/questions/49278728/pyspark-an-error-occurred-while-calling-o50-parque)

### DIFF
--- a/docs/using/recipes.md
+++ b/docs/using/recipes.md
@@ -156,7 +156,29 @@ A few suggestions have been made regarding using Docker Stacks with spark.
 
 ### Using PySpark with AWS S3
 
+Using Spark session for hadoop 2.7.3
+
+```py
+import os
+# !ls /usr/local/spark/jars/hadoop* # to figure out what version of hadoop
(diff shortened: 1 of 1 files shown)


parente:
LGTM though I don't have a way to verify the steps. Let's merge it and fix over time if people find problems. 🍰

ogierpaul:
Well, thank you so much @ying-w  for adding this comment:
````
 # !ls /usr/local/spark/jars/hadoop* # to figure out what version of hadoop
````

That just saved my life.
I was trying to connect to AWS with jupyter/pyspark-notebook, and everywhere the code found was reference Hadoop 2.7.3. Turned out that my version of Hadoop in the latest docker image was 3.2.0. 

Thank you again!

PR #853, opened 2019-04-25
Update to JupyterLab 0.35.5

clkao:
This contains a fix for chrome download file bug: https://github.com/jupyterlab/jupyterlab/pull/6232


parente:
Thanks @clkao . The build failure is unrelated and fixed in #855.

PR #859, opened 2019-05-01
[Doc] Change Spark Python Notebook entry point sentence

marcraminv:
I have changed the statement used to initialise the spark in _Python Notebook_ to keep up-to-date according to Spark docs. I have followed the same ideas as [**Spark Documentation**](http://spark.apache.org/docs/latest/quick-start.html#self-contained-applications) used for first _Hello World_ example.

### DIFF
--- a/docs/using/specifics.md
+++ b/docs/using/specifics.md
@@ -13,12 +13,10 @@ Spark local mode is useful for experimentation on small data when you do not hav
 #### In a Python Notebook
 
 ```python
-import pyspark
-sc = pyspark.SparkContext('local[*]')
-
+from pyspark.sql import SparkSession
+spark = SparkSession.builder.appName("SimpleApp").getOrCreate()
 # do something to prove it works
-rdd = sc.parallelize(range(1000))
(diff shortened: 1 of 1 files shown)


docs/using/specifics.md
```diff
@@ -13,12 +13,10 @@ Spark local mode is useful for experimentation on small data when you do not hav
...
 ```python
-import pyspark
-sc = pyspark.SparkContext('local[*]')
-
+from pyspark.sql import SparkSession
+spark = SparkSession.builder.appName("SimpleApp").getOrCreate()
 # do something to prove it works
-rdd = sc.parallelize(range(1000))
-rdd.takeSample(False, 5)
+spark.sql('SELECT "Test" as c1').show()
```
parente: Will this select function properly when the user brings up a local SparkContext with no data backing it?

marcraminv: User only need `SparkSession` instance to work with. This select statement doesn't refer to any table because has not `from <table>` in the query, this is like this dummy example `SELECT 1 AS c1`

PR #872, opened 2019-05-12
Update RTD config file to v2

willingc:
- Update to v2
- Create config as a dotfile


parente:
Thanks @willingc! 📚📚📚

Issue #889, opened 2019-06-20
add e1071 to r-notebook Dockerfile


ttimbers:
I have tried training models in an image built from the r-notebook Dockerfile and although I can load `caret` using the `library` without an error, when I run `caret::train()` I get the following error:

```
Error: package e1071 is required
```

When I then install e1071 inside the container via `install.packages('e1071')` in R then I can successfully run `caret::train()`.

```
library(e1071)
sessionInfo()
```

Indicates that the version of `e1071` that was installed is `e1071_1.7-2`. I will add this to the Dockerfile for r-notebook and send a PR.

PR #890, opened 2019-06-20
Added install of e1071 R package

ttimbers:
The e1071 R package is a dependency of functions within the caret package. Without it, `caret::train` (for example) cannot be run. This pull request addresses issue #889 (https://github.com/jupyter/docker-stacks/issues/889). 

note - I first tried installing e1071 in the RUN command started on line 27 but that resulted in e1071 not installing (don't know why) but adding an additional RUN command led to succesful installation... So I just went with it.


minrk:
Thanks!

PR #912, opened 2019-07-28
Use prod instead of dev build for JupyterLab

maresb:
Using a dev build, vendors~main.js is 35.4 MB.  With the prod build it is only 5.0 MB.  This makes a huge difference in load times for non-local sessions.  For more details, see the links in https://github.com/jupyterlab/jupyterlab/issues/6869 .


parente:
Thanks for the PR @maresb. 🍰

PR #896, opened 2019-07-04
conda 4.7.5

minrk:
gets much faster installs

had to move WORKDIR $HOME up due to tempfile permission issues with conda 4.7 (fixes #891, fixes #878). Conda 4.7 creates some tempfiles in the CWD instead of /tmp (actually conda_package_handling).

PR #913, opened 2019-07-29
Switch to matplotlib-base to avoid qt dependency

maresb:
Based on [this discussion](https://github.com/jupyter/docker-stacks/pull/896#issuecomment-515792038).  It seems to work, but I have not tested it thoroughly.


parente:
Let's ship it and see.

PR #917, opened 2019-08-02
Option to enable restart without exiting container

### LINKED ISSUE #880: Feature: restartable Jupyter in container
I do not know the most elegant way to implement this, so I'm posting an issue instead of a PR.  I would like to be able to restart Jupyter within the docker-stacks container without the docker process ending because the ENTRYPOINT/CMD exited.  That is to say, if I ran a notional command `docker run -p 8888:8888 jupyter/base-notebook --restartable`, then when I hit the 'Quit'  button the Jupyter process would exit and restart without the docker container stopping.  

One example of when it is necessary to restart the Jupyter process is after installing Jupyter serverextensions such as voila.  

In many cases, a user can `docker start` a container after it has been stopped (effectively restarting Jupyter and the container) but there are times when a user does not have control over docker orchestration. 

Thank you.

jeffyjefflabs:
Add optional environment variable `RESTARTABLE` to enable Jupyter to be restarted without causing the container to exit.  When the variable is set, `run-one-constantly` (from the Ubuntu [run-one](http://manpages.ubuntu.com/manpages/trusty/man1/run-one.1.html) package) is
inserted into the Jupyter startup command.

The run-one package is negligible in size; apt says it requires 45K and docker reports the image at 700MB both before and after the change.

Fixes #880


parente:
LGTM. Thanks for the PR @jeffyjefflabs. 🍰

PR #918, opened 2019-08-02
Update all package versions

rkdarst:
New versions found with `conda upgrade --all` inside the current image,
which I guess is the right way to do it.  Only scipy-notebook so far.
It builds but I haven't done any live tests yet.

### DIFF
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -23,10 +23,10 @@ USER $NB_UID
 
 # R packages
 RUN conda install --quiet --yes \
-    'r-base=3.5.1' \
-    'r-irkernel=0.8*' \
-    'r-ggplot2=3.1*' \
-    'r-sparklyr=0.9*' \
+    'r-base=3.6.1' \
+    'r-irkernel=1.0*' \
+    'r-ggplot2=3.2*' \
+    'r-sparklyr=1.0*' \
     'r-rcurl=1.95*' && \
     conda clean --all -f -y && \
     fix-permissions $CONDA_DIR && \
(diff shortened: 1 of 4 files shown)


rkdarst:
Now updated in all notebooks.  The missing ones seem to have no updates that are needed.  Also, I only updated the conda packages, not the wget + install add-on software (though from a brief check many are still OK).

I build them all locally but haven't done real tests.

There are some things I'd like to fix in a later PR.  For example, sort the Dockerfile conda installation lines so that it's faster to compare with the sorted list that comes out of conda.  Also there is duplication in r-notebook and datascience-notebook: it looks like the R packages were copied and pasted, but only r-notebook got some extra packages.  Should these two have the same R packages?  This is a slighly smaller concern for all-spark-notebook.

parente:
> I build them all locally but haven't done real tests.

I think it's fine to rely on conda to install versions of packages compatible with one another, and to count on the package releases to be in working order. At one point, I [considered adding notebooks that we'd run with nbconvert](https://github.com/jupyter/docker-stacks/issues/406#issuecomment-352431169) after each build to check for regressions. We never really found a need to do so.

> For example, sort the Dockerfile conda installation lines so that it's faster to compare with the sorted list that comes out of conda.

Sounds reasonable. The lines used to be alphabetized, and might still be, but for no real reason.

> Also there is duplication in r-notebook and datascience-notebook: it looks like the R packages were copied and pasted, but only r-notebook got some extra packages. Should these two have the same R packages? 

I think it's fine to leave the packages lists in those images independent from each other. 

We used to add packages when and where people requested them. Now we prefer to show people how to add packages to their own images derived `FROM <image>` instead of continually growing the size of these images.

parente:
Thanks for doing this work @rkdarst. 🍰

PR #920, opened 2019-08-04
Sort package installation lines alphabetically for easier comparison

rkdarst:
- This would have made updating all package versions easier, so doing it now.
- Differences I noticed:
  - r-notebook and datascience-notebook are almost identical
  - r-notebook has r-odbc and unixodbc, datascience-notebook doesn't
  - all-spark-notebook has r-ggplot2 explicitely added and version
    pinned.  The others pull it in but isn't version pinned.
- I'm conflicted as to if I should skip the CI on this, but since it's
  not pure text changes I won't.

### DIFF
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -24,10 +24,11 @@ USER $NB_UID
 # R packages
 RUN conda install --quiet --yes \
     'r-base=3.6.1' \
-    'r-irkernel=1.0*' \
     'r-ggplot2=3.2*' \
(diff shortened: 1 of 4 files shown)


parente:
The build is failing with `/bin/sh: 1: r-htmltools=0.3*: not found` which is interesting considering it was in the package list before this sorting change. ]

Typo? Package change upstream?

rkdarst:
It was the `&&` endings that also got sorted (despite remembering to handle them before I started).  I just moved them to a new line with a force-push.

CI wins... obviously!

PR #930, opened 2019-08-22
Bump Julia to 1.2.0

arnavs:


arnavs:
@parente Another one of these... `:-P`.

jlperla:
@parente @rkdarst If either of you have a chance to merge and tag, it would be appreciated.  There are jupyterhub solutions effected downstream by this that are hoping to finalize everything before teaching begins next week.

Thanks again for the great resource.

arnavs:
I updated this PR to remove a few packages from Julia: 

* `InstantiateFromURL` --- I develop this package. Added it a while back, but I think pinning the version inside a Jupyter image could confuse people.  

* `RDatasets` and `Gadfly` --- These are useful packages, but probably better to start with just `IJulia` and let people add the precise packages/versions they need. Especially since Julia's resolver will change what operations do, so as to ensure a feasible point for all existing packages. That is, the pre-existing versions may have side-effects for other package operations.

parente:
Sounds reasonable to me. :shipit:

jlperla:
@arnavs  I don't know where the source for it was, but I think that https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#jupyter-datascience-notebook description should be removed to strip out the packages listed there since this has been merged.

PR #938, opened 2019-09-16
Update Miniconda to 4.7.10 and conda to 4.7.12

edurenye:
Update Miniconda to 4.7.10 and conda to 4.7.12 also use a env variable to set the miniconda MD5


parente:
Thanks @edurenye.

PR #940, opened 2019-09-18
Don't upgrade base image

ilveroluca:
This PR removes the `apt-get dist-upgrade` from the build  process of the base-notebook image, as per the [dockerfile best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices).

Upgrading the base image, especially without being able to specify package versions, can break dependencies, reproducibility of builds, and can significantly enlarge the final image.

I also added a call to `apt-get clean` after package installations: this command will remove any debs files left in apt-get's cache.


parente:
Seems right to me, @ilveroluca. I don't recall seeing `dist-upgrade` as an addition in the past. Thanks.

PR #942, opened 2019-09-24
Make fix-permissions executable by all

ilveroluca:
Make the `fix-permissions` script executable by everyone, avoiding permission denied error when `NG_UID` isn't root and doesn't match the file's owner or group id.


parente:
LGTM. Thanks.

PR #950, opened 2019-10-14
With JupyterLab 1.1, --dev-build=False is no longer necessary.

maresb:
Reverts a temporary workaround included in [this commit](https://github.com/jupyter/docker-stacks/commit/eb5c77ee43c8ee193f34c8e23ac3568f1f3f2578#diff-90dd1590ae70de58d60090b55b7ee318), which addressed [this issue](https://github.com/jupyterlab/jupyterlab/issues/6869) which has subsequently been fixed.


parente:
Thanks @maresb.

PR #967, opened 2019-12-05
Update ubuntu base image

AllOtherUserNamesTaken:
Twistlock vulnerability scanning showed OS issues removable by upgrading of package versions.


AllOtherUserNamesTaken:
Will open a new one with single commit.

PR #973, opened 2019-12-11
instructions for python 3.x kernels with conda

mathematicalmichael:
this was surprisingly hard to find help for online that was up-to-date and worked with the stacks, contributing here in hopes it'll help others.

### DIFF
--- a/docs/using/recipes.md
+++ b/docs/using/recipes.md
@@ -98,6 +98,47 @@ USER $NB_USER
 Ref:
 [https://github.com/jupyter/docker-stacks/issues/440](https://github.com/jupyter/docker-stacks/issues/440)
 
+## Add a Python 3.x environment
+
+The default version of Python that ships with conda/ubuntu may not be the version you want.
+To add a conda environment with a different version and make it accessible to Jupyter, the instructions are very similar to Python 2.x but are slightly simpler (no need to switch to `root`):
+
+```
+# Choose your desired base image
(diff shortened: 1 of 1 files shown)


docs/using/recipes.md
```diff
@@ -98,6 +98,47 @@ USER $NB_USER
...
+# COPY environment.yml /home/$NB_USER/tmp/
+# RUN cd /home/$NB_USER/tmp/ && \
+#     conda env create -p $CONDA_DIR/envs/$conda_env -f environment.yml && \
+#     conda clean --all -f -y
+
+
+# create Python 3.x environment and link it to jupyter
+RUN $CONDA_DIR/envs/${conda_env}/bin/python -m ipykernel install --user --name=${conda_env} && \
+    fix-permissions $CONDA_DIR && \
+    fix-permissions /home/$NB_USER
```
mathematicalmichael: actually not sure if it's better to run this as a separate last step. suggestions welcome

parente: I think it's OK if it runs separately. It's still part of the same `RUN` command which means Docker will create only one filesystem layer regardless.

parente:
Thanks for this recipe @mathematicalmichael. Apologies for taking so long to review. I haven't tried the steps myself, but at a glance they look right to me. Let's merge it and iterate if the community identifies improvements or problems.

PR #974, opened 2019-12-12
Add sparkmagic-notebook Image

FabioBatSilva:
This patch add support for sparkmagic.

I understand that the preferred approach would be to include [recipe](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/recipes.html) with steps on how to set sparkmagic
But the setup is a bit more complex than other recipes given that you need to customize sparkmagic.

And i don't think that adding this to all-spark-notebook make sense given the size of the images and that sparkmagic doesn't need spark running on the same host.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,8 @@ ALL_STACKS:=base-notebook \
 	tensorflow-notebook \
 	datascience-notebook \
 	pyspark-notebook \
-	all-spark-notebook
+	all-spark-notebook \
+	sparkmagic-notebook
 endif
 
 ALL_IMAGES:=$(ALL_STACKS)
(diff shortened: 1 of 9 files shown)


parente:
@FabioBatSilva Thanks for the contribution. We've been requesting that all new image definitions take the [community images path](https://jupyter-docker-stacks.readthedocs.io/en/latest/contributing/stacks.html), primarily because this repository is unwieldy and we're already finding it difficult to maintain the images that already exist here. 

If you wind up turning this PR into a separate repository, we'll be happy to link to your image definition [like we have others](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#community-stacks).

parente:
I'm taking a pass through old PRs. I'm going to close this one as it is better suited as a community stack.

PR #965, opened 2019-12-04
Update Julia Version

Nosferican:
Update Julia from 1.2.0 to 1.3.0.

PR #980, opened 2020-01-03
Update to Julia 1.3.0

kjfergu:
This just updates two lines to have the build pull down Julia 1.3.1 (Most current as of Dec 30 2019). The first is the variable JULIA_VERSION, the second is the associated checksum for the tar.gz file.

### DIFF
--- a/datascience-notebook/Dockerfile
+++ b/datascience-notebook/Dockerfile
@@ -23,12 +23,12 @@ RUN apt-get update && \
 # install Julia packages in /opt/julia instead of $HOME
 ENV JULIA_DEPOT_PATH=/opt/julia
 ENV JULIA_PKGDIR=/opt/julia
-ENV JULIA_VERSION=1.2.0
+ENV JULIA_VERSION=1.3.1
 
 RUN mkdir /opt/julia-${JULIA_VERSION} && \
     cd /tmp && \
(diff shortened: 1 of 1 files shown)


kjfergu:
This should have been titled Julia 1.3.1 as that is what this is actually doing, not 1.3.0. PR #965 updates to 1.3.0.

Bidek56:
@parente Do you need help with these PRs? I have noticed that they have been accumulating :)

parente:
Thanks @kjfergu and @Bidek56. I'm taking a pass through the PRs now and merging.

Nosferican:
Could we get this tagged at Docker Hub?

parente:
@Nosferican Images with Julia 1.3.1 are now being tagged on Docker Hub after the build process fix mentioned in https://github.com/jupyter/docker-stacks/issues/992. The tags follow the git sha standard we use for all the images here (https://github.com/jupyter/docker-stacks/wiki).

If you're interested in seeing additional Julia interpreter versions as tags like we have for Python versions and Jupyter notebook, lab, and hub versions, I'd welcome a PR adding them. A script like this one https://github.com/jupyter/docker-stacks/blob/master/base-notebook/hooks/apply_tags needs to be implemented and placed in https://github.com/jupyter/docker-stacks/blob/master/datascience-notebook/hooks/apply_tags.

PR #989, opened 2020-01-23
install pip explicitly

### LINKED ISSUE #988: Current Base-Notebook Image does not have pip
Noticed that pip will not work with latest image [7f696fe2cb4e](https://hub.docker.com/layers/jupyter/base-notebook/latest/images/sha256-7f696fe2cb4ee581262c8c053831188efbe048751bd8e7f0e2011cae41ec555c)

This Fix will work

```
RUN conda install pip
```

minrk:
it should be there automatically, but no harm in being explicit

bump notebook to 6.0.3 while we are here

closes #988

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -70,8 +70,8 @@ RUN mkdir /home/$NB_USER/work && \
     fix-permissions /home/$NB_USER
 
 # Install conda as jovyan and check the md5 sum provided on the download site
-ENV MINICONDA_VERSION=4.7.10 \
-    MINICONDA_MD5=1c945f2b3335c7b2b15130b1b2dc5cf4 \
+ENV MINICONDA_VERSION=4.7.12.1 \
+    MINICONDA_MD5=81c773ff87af5cfac79ab862942ab6b3 \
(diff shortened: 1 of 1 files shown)


consideRatio:
Since `conda` is used alongside `$CONDA_DIR/bin/conda` I assume it is safe to remove `$CONDA_DIR/bin/` as you do in this PR.

```
# instead of this
conda install --quiet --yes conda && \
conda install --quiet --yes pip && \

# perhaps this?
conda install --quiet --yes conda pip
```

I figure that the latter may be quicker without compromise, other than the fact that I need to make sure that there is no compromise and it kinda feels safer to do it one after the other.

Otherwise, this LGTM. I'm a bit sad I still don't understand why this became an issue in the first place.

consideRatio:
Relevant comment: https://github.com/jupyter/docker-stacks/issues/988#issuecomment-577647971

minrk:
I think the package was removed due to a weird conflict in the dependencies of pip. It's not clear exactly what the conflict was, but updating miniconda does fix it, and explicitly requesting pip does also ensure it is pinned into the environment and should not be a candidate for removal to satisfy conflicts.

PR #991, opened 2020-01-25
chore: Replace ADD with COPY

jgwerner:
- Replace `ADD` with `COPY` as its documented as the [preferred option](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)


parente:
Thanks @jgwerner 🍰

PR #995, opened 2020-01-28
Fall back on symlink if mv fails

stevenstetzler:
In reference to jupyter#975.

Changes
```
mv /home/jovyan "/home/$NB_USER"
```
to
```
mv /home/jovyan "/home/$NB_USER" || ln -s /home/jovyan "/home/$NB_USER"
```
to fall back on a symlink for the home directory in the case that mv fails. mv can fail for some volume mounting options. (This was originally documented for AWS EBS-backed volume when running JupyterHub on Kubernetes.)


parente:
Thanks for submitting this PR, @stevenstetzler. 🍰

PR #1006, opened 2020-02-10
[ci skip] Show jupyter/datascience-notebook tags on the wiki

parente:
e.g., https://github.com/jupyter/docker-stacks/wiki


parente:
@Nosferican The `jupyter/datascience-notebook` language tags are showing up on Docker Hub now. https://hub.docker.com/repository/registry-1.docker.io/jupyter/datascience-notebook/tags?page=1

I forgot about the template that renders the tags on the build log wiki page: https://github.com/jupyter/docker-stacks/wiki. I'm adding to that template now. (It's an ugly one-liner because I haven't figured out a way to escape newlines in the github markdown table syntax.)

PR #1010, opened 2020-02-13
Remove quotes from singleuser.sh

dmpe:
Being consistent with other args

Thanks goes to my colleague at work who saw it.

### DIFF
--- a/base-notebook/start-singleuser.sh
+++ b/base-notebook/start-singleuser.sh
@@ -14,7 +14,7 @@ fi
 # These won't be passed from DockerSpawner 0.9,
 # so avoid specifying --arg=empty-string
 if [ ! -z "$NOTEBOOK_DIR" ]; then
-  NOTEBOOK_ARGS="--notebook-dir='$NOTEBOOK_DIR' $NOTEBOOK_ARGS"
+  NOTEBOOK_ARGS="--notebook-dir=$NOTEBOOK_DIR $NOTEBOOK_ARGS"
 fi
 if [ ! -z "$JPY_PORT" ]; then
   NOTEBOOK_ARGS="--port=$JPY_PORT $NOTEBOOK_ARGS"


dmpe:
Haha, it is not so simple. The directory may contain a whitespace and therefore it was implemented. 

In principle, arrays would be better.

parente:
Thanks for submitting this PR. Is there an issue the quoting is causing when specifying a the notebook directory?

dmpe:
Not necessary that I have come across, but i had to build a custom docker image and was "scanning" files. 
Quotes within quotes can be quite confusing and different shells (OS/versions) may treat them differently as well. Maybe it is a question of style too, i am more used to apply in these cases bash arrays.

base-notebook/start-singleuser.sh
```diff
@@ -14,7 +14,7 @@ fi
 # These won't be passed from DockerSpawner 0.9,
 # so avoid specifying --arg=empty-string
 if [ ! -z "$NOTEBOOK_DIR" ]; then
-  NOTEBOOK_ARGS="--notebook-dir='$NOTEBOOK_DIR' $NOTEBOOK_ARGS"
+  NOTEBOOK_ARGS="--notebook-dir=$NOTEBOOK_DIR $NOTEBOOK_ARGS"
```
consideRatio, line 17:
```suggestion
  NOTEBOOK_ARGS="--notebook-dir=\"$NOTEBOOK_DIR\" $NOTEBOOK_ARGS"
```

consideRatio:
The reason that the quote exists is to not crash if the file path contains spaces.

Using `'` quotes instead of `"` quotes (that could be introduced with `\"` will make paths like `~/work` fail to work though, as that require bash to expand them to `/home/jovyan/work`.

I think if it could make sense to quote like this instead `\"` instead though to allow NOTEBOOK_DIR='~/work' to be expanded to the user fold or similar.

Oh wait... This is all about something that is deprecated and I would advocate we remove all together as it was deprecated 3 years ago. I've added a commit to remove this section in #1052 which ended up being a major refactoring of start.sh / start-notebook.sh / start-singleuser.sh.

I suggest we close this PR, @dmpe what do you think?

PR #1013, opened 2020-02-14
Remove installation of unnecessary pandoc and tex OS packages

romainx:
Hello,

I've noticed that `pandoc` was installed 2 times. It's installed in the `minimal-notebook` as a `conda` package, then in the `minimal-notebook` as an OS package.

```bash
# The pandoc used is the one installed by conda
$ which pandoc

# /opt/conda/bin/pandoc

$ find / -name pandoc

# /usr/share/lintian/overrides/pandoc
# /usr/share/doc/pandoc
# /usr/share/pandoc
# /usr/bin/pandoc
# /opt/conda/bin/pandoc

$ ls -alsh  /opt/conda/bin/pandoc /usr/bin/pandoc

# 59M -rwxrwxr-x 1 jovyan users 59M Jan  6 00:01 /opt/conda/bin/pandoc
# 50M -rwxr-xr-x 1 root   root  50M Jan 26  2018 /usr/bin/pandoc
```

So I've removed the OS package installation from the `minimal-notebook` and written a test on the `base-notebook` to check that `pandoc` is working without this package. I've also performed a manual export to PDF test in a `minimal-notebook`.

It will contribute to slightly reduce the size of this image (same effect on downstream images as well).

Here is comparison

```bash
REPOSITORY                 TAG                 IMAGE ID            CREATED             SIZE
jupyter/minimal-notebook   latest              a817cb6de443        37 seconds ago      2.49GB
jupyter/minimal-notebook   <none>              774718b3f9c5        2 weeks ago         2.54GB
```

 Best

### DIFF
--- a/minimal-notebook/Dockerfile
+++ b/minimal-notebook/Dockerfile
@@ -20,12 +20,13 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     lmodern \
     netcat \
     python-dev \
-    texlive-fonts-extra \
+    # ---- nbconvert dependencies ----
+    texlive-xetex \
     texlive-fonts-recommended \
     texlive-generic-recommended \
-    texlive-latex-base \
-    texlive-latex-extra \
-    texlive-xetex \
+    # Optional dependency
+    texlive-fonts-extra \
+    # ----
     tzdata \
     unzip \
     nano \
--- a/minimal-notebook/Dockerfile
+++ b/minimal-notebook/Dockerfile
@@ -19,7 +19,6 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     libxrender1 \
     lmodern \
     netcat \
-    pandoc \
     python-dev \
     texlive-fonts-extra \
     texlive-fonts-recommended \
--- a/base-notebook/test/test_pandoc.py
+++ b/base-notebook/test/test_pandoc.py
@@ -1,0 +1,20 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import logging
+
+import pytest
+
+LOGGER = logging.getLogger(__name__)
+
+
+def test_pandoc(container):
+    """Pandoc shall be able to convert MD to HTML."""
+    c = container.run(
+        tty=True, command=["start.sh", "bash", "-c", 'echo "**BOLD**" | pandoc']
+    )
+    c.wait(timeout=10)
+    logs = c.logs(stdout=True).decode("utf-8")
+    LOGGER.debug(logs)
+    assert "<p><strong>BOLD</strong></p>" in logs
+
--- a/minimal-notebook/test/data/notebook1.ipynb
+++ b/minimal-notebook/test/data/notebook1.ipynb
@@ -1,0 +1,149 @@
+{
+ "metadata": {
+  "name": "notebook1"
+ },
+ "nbformat": 3,
+ "nbformat_minor": 0,
+ "worksheets": [
+  {
+   "cells": [
+    {
+     "cell_type": "heading",
+     "level": 1,
+     "metadata": {},
+     "source": [
+      "A simple SymPy example"
+     ]
+    },
+    {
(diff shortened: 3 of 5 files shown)


minimal-notebook/Dockerfile
```diff
@@ -19,7 +19,6 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     libxrender1 \
     lmodern \
     netcat \
-    pandoc \
```
parente: I don't see pandoc installed elsewhere in minimal-notebook via conda. I'm guessing one of the conda packages in base-notebook now specifies pandoc as a dependency and so it's getting installed there. I don't see it listed in the recipes for notebook, jupyterlab, or jupyterhub on conda-forge, so I'm a bit confused about where it's getting pulled in. We should probably make it an explicit conda install until we know for sure it's not an accidental dependency.

Do we know if tex still needs to be installed at the OS level or if the pandoc conda package installs everything it needs too?

romainx: You're right I had not checked. The pandoc dependency comes form nbconvert.

```
nbconvert 5.6.1 py37_0
----------------------
file name   : nbconvert-5.6.1-py37_0.tar.bz2
name        : nbconvert
version     : 5.6.1
build       : py37_0
build number: 0
size        : 487 KB
license     : BSD-3-Clause
subdir      : linux-64
url         : https://conda.anaconda.org/conda-forge/linux-64/nbconvert-5.6.1-py37_0.tar.bz2
md5         : 7b04496e019d1f922f9718882eb94752
timestamp   : 2019-10-25 01:42:31 UTC
constraints : 
  - python_abi * *_cp37m
dependencies: 
  - bleach
  - defusedxml
  - entrypoints >=0.2.2
  - jinja2
  - jupyter_core
  - mistune >=0.8.1
  - nbformat >=4.4
  - pandoc >=1.12.1
  - pandocfilters >=1.4.1
  - pygments
  - python >=3.7,<3.8.0a0
  - testpath
  - traitlets >=4.2
```
The environment file can be seen [in the github](https://github.com/jupyter/nbconvert/blob/0391f83c95d01844a60e1a74ba41a77a0e20a120/docs/environment.yml).

However it's unclear since the nbconvert documentation still ask to install pandoc. I will create an issue in the GitHub to get a confirmation on that.

For tex sorry but I don't know, I will check.

romainx: For tex the nbconvert installation guide has a [dedicated section for tex](https://nbconvert.readthedocs.io/en/latest/install.html#installing-tex) 

```
sudo apt-get install texlive-xetex texlive-fonts-recommended texlive-generic-recommended
```
In the `base-notebook` the required packages are installed and so IMO they have to be kept.

- `texlive-xetex`
- `texlive-fonts-recommended`
- `texlive-generic-recommended`

These are dependencies of `texlive-xetex`, so they could be removed -- just to be lean, the result will be the same.
- `texlive-latex-base`
- `texlive-latex-extra`

```bash
$ apt-cache depends texlive-xetex
# texlive-xetex
#  Depends: tex-common
#  Depends: texlive-base
#  Depends: texlive-binaries
#  Depends: texlive-latex-base
#  Depends: texlive-latex-extra
[...]
```

And the following extra packages is installed, there necessity could be challenged

- `texlive-fonts-extra`

[Here](https://tex.stackexchange.com/a/504566) is a great answer on SO explaining the content and the dependencies between those packages.

romainx: According to the previous discussion, I've cleaned the tex installation and written a test to check if nbconvert is working properly without pandoc and without removed tex packages.

parente: Awesome. Thank you for doing this extra investigation and updating the PR!

parente:
@romainx I'm cool with merging this PR if you are. You had it marked WIP. Are you planning any other changes or should I ship it?

romainx:
Hey @parente. No sorry I just did not change the name, I will not perform other change on this PR. So let's go!
Thank you.

PR #1020, opened 2020-02-21
limit fix permissions to /opt/conda

rscohn2:
I am building a custom container by using the base-notebook/Dockerfile and overriding BASE_CONTAINER from the command line. My base image has additional software installed in /opt. The dockerfile is trying to fix permissions on the entire /opt tree, which blows up the image size. Is it OK to limit the fix permissions to /opt/conda?

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -59,7 +59,7 @@ RUN echo "auth requisite pam_deny.so" >> /etc/pam.d/su && \
     chown $NB_USER:$NB_GID $CONDA_DIR && \
     chmod g+w /etc/passwd && \
     fix-permissions $HOME && \
-    fix-permissions "$(dirname $CONDA_DIR)"
+    fix-permissions $CONDA_DIR
 
 USER $NB_UID
 WORKDIR $HOME


romainx:
Make sense for me.  I do not see any reason to call the `fix-permissions` in the `opt` dir.  Moreover at this stage and with the original image it only contains `conda`.

```
$ ls -alsh /opt

# total 12K
# 4.0K drwsrwsr-x 1 root   users 4.0K Jan 30 03:41 .
# 4.0K drwxr-xr-x 1 root   root  4.0K Feb 22 07:45 ..
# 4.0K drwsrwsr-x 1 jovyan users 4.0K Jan 30 03:46 conda
```

parente:
There is (was?) a competing use case of people wanting to run the docker image locally, keep the container around for the long term (i.e., no `--rm` or `docker save`), and have the ability to install conda packages as the `jovyan` user since now they have a semi-permanent notebook environment. I understand changing the behavior to disallow writes in `/opt` would be friendly to admin use case like the one you describe, but does represent a breaking change for the individual use case.

parente:
The history of why `fix-permissions` is called on `/opt` is in https://github.com/jupyter/docker-stacks/issues/807. tl;dr: A bug in conda 4.6.2. We probably no longer need the fix since we're on conda 4.7.12.

parente:
conda issue fixed in https://github.com/conda/conda/pull/8253.

parente:
Thanks for the update @rscohn2 🍰 .

PR #1021, opened 2020-02-23
Turning channel priority to strict

romainx:
I'm thinking of speeding up conda (because it's really slow) either during the build of the stacks and for the user when they want to install additional package.

> As of version 4.6.0, Conda has a strict channel priority feature. Strict channel priority can dramatically speed up conda operations and also reduce package incompatibility problems. We recommend it as a default. However, it may break old environment files, so we plan to delay making it conda's out-of-the-box default until the next major version bump, conda 5.0.
>
> -- <cite>https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html#strict</cite>

As we prepend `conda-forge` channel, I think the impact could be limited since I think the goal is to privilege `conda-forge` packages.

> With strict channel priority, packages in lower priority channels are not considered if a package with the same name appears in a higher priority channel.

Here are the differences on `base-notebook` between `flexible` (default) and `strict` channel priority.

```diff
+ ld_impl_linux-64          2.33.1               h53a641e_8    conda-forge
- libedit                   3.1.20181209         hc058e9b_0    defaults
+ libedit                   3.1.20170329      hf8c457e_1001    conda-forge
- python                    3.7.4                h265db76_1    defaults
+ python                    3.7.6                h357f687_2    conda-forge
- readline                  7.0               hf8c457e_1001    conda-forge
+ readline                  8.0                  hf8c457e_0    conda-forge
- sqlite                    3.31.1               h7b6447c_0    defaults
+ sqlite                    3.30.1               hcee41ef_0    conda-forge
```

In terms of improvement what I have observed locally at build is not meaningful let's check with the full stack build if we can see a difference.

What do you think about performing this change? 

Best.

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -84,6 +84,7 @@ RUN cd /tmp && \
     conda config --system --prepend channels conda-forge && \
     conda config --system --set auto_update_conda false && \
     conda config --system --set show_channel_urls true && \
+    conda config --system --set channel_priority strict && \
     if [ ! $PYTHON_VERSION = 'default' ]; then conda install --yes python=$PYTHON_VERSION; fi && \
     conda list python | grep '^python ' | tr -s ' ' | cut -d '.' -f 1,2 | sed 's/$/.*/' >> $CONDA_DIR/conda-meta/pinned && \
     conda install --quiet --yes conda && \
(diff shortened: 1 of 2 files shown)


romainx:
At least we have the answer :sob:
However since `r-notebook` is outdated I will check in depth what is going on.
I let you know.

romainx:
Much better with an up-to-date `r-notebook` 😄 .
The build has run for for 28 min 22 sec, so for sure not a big improvement in term of total build time -- maybe not an improvement at all since some builds have already ran in ~ 28 min.

parente:
Thanks @romainx. I'm fine with enabling strict mode. I updated your branch with master and will merge when the build succeeds.

PR #1027, opened 2020-02-27
Update Dockerfile to R 3.6.2

xiaodaigh:


romainx:
In the PR #1021 I have also bumped the R version in `r-notebook`. If you could have a look there is some additional packages to update. It could be interesting to group all the update in a single PR.

Thanks a lot.

xiaodaigh:
Hopefully can be in soon as moe it's basically unusable because some packages like stringi are not installable

romainx:
:-) Please also note that there is an initiative (#693) in progress to deprecate this notebook.

parente:
Thanks for the PR @xiaodaigh. @romainx has fixed this in #1021 which is building now on Docker Hub.

PR #1028, opened 2020-02-27
cgspatial-notebook

KyriakosZorbas:
cgspatial-notebook is a community Jupyter Docker Stack image. The image includes major geospatial Python & R libraries on top of the datascience-notebook image


parente:
Thanks for sharing @KyriakosZorbas. 🍰

Issue #955, opened 2019-10-22
scipy-notebook Image Missing ipympl?


j123github:
It appears that the scipy-notebook image is missing ipympl, resulting in `%matplotlib widget` failing to run. As a result, interactive plots are not easily available when running Jupyter Lab.

It wasn't clear to me that this was a deliberate omission, since the jupyter-widgets extension is being installed [here](https://github.com/jupyter/docker-stacks/blob/8bfffcaa3f41d8493a71648399bb540e8aa24860/scipy-notebook/Dockerfile#L51). Apologies in advance if this is an explicit choice and I have misunderstood.

I have added the installation of `ipympl=0.3.3` to a local scipy-notebook Dockerfile and have been using `%matplotlib widget` without errors.

**What docker image you are using?**

`jupyter/scipy-notebook` or `jupyter/datascience-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

`docker run -p 8888:8888 jupyter/scipy-notebook:latest start.sh jupyter lab`

or

`docker run -p 8888:8888 jupyter/datascience-notebook:latest start.sh jupyter lab`

**What steps do you take once the container is running to reproduce the issue?**

Example:

1. Visit http://localhost:8888/lab
2. Start a Python notebook in Jupyter Lab
3. Run 
```
%matplotlib widget
import matplotlib.pyplot as plt
```

**What do you expect to happen?**

For the command to complete without error.

**What actually happens?**

The following error is output due to ipympl missing:
```
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-eeaad41e1227> in <module>
----> 1 get_ipython().run_line_magic('matplotlib', 'widget')
      2 import matplotlib.pyplot as plt

/opt/conda/lib/python3.7/site-packages/IPython/core/interactiveshell.py in run_line_magic(self, magic_name, line, _stack_depth)
   2312                 kwargs['local_ns'] = sys._getframe(stack_depth).f_locals
   2313             with self.builtin_trap:
-> 2314                 result = fn(*args, **kwargs)
   2315             return result
   2316 

</opt/conda/lib/python3.7/site-packages/decorator.py:decorator-gen-108> in matplotlib(self, line)

/opt/conda/lib/python3.7/site-packages/IPython/core/magic.py in <lambda>(f, *a, **k)
    185     # but it's overkill for just that one bit of state.
    186     def magic_deco(arg):
[...]

PR #1029, opened 2020-02-27
Try to get ipympl working

mpmdean:
I am proposing to allow the use of ipympl widgets as raised in  #955 

I thought these simple additional packages and build commands would suffice, but somehow it does work. This PR functions correctly within jupyter notebooks, but it does not work with jupyterlab. The cell only responds by returning ‘Loading widget…’ , but it never loads anything. 

Do not merge this!

### DIFF
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -26,6 +26,7 @@ RUN conda install --quiet --yes \
     'h5py=2.10.*' \
     'hdf5=1.10.*' \
     'ipywidgets=7.5.*' \
+    'ipympl=0.3.*'\
     'matplotlib-base=3.1.*' \
     'numba=0.48.*' \
     'numexpr=2.7.*' \
@@ -40,6 +41,7 @@ RUN conda install --quiet --yes \
     'statsmodels=0.11.*' \
     'sympy=1.5.*' \
     'vincent=0.4.*' \
+    'widgetsnbextension=3.5.*'\
     'xlrd' \
     && \
     conda clean --all -f -y && \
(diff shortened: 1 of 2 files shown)


scipy-notebook/Dockerfile
```diff
@@ -50,7 +50,7 @@ RUN conda install --quiet --yes \
     # Also activate ipywidgets extension for JupyterLab
     # Check this URL for most recent compatibilities
     # https://github.com/jupyter-widgets/ipywidgets/tree/master/packages/jupyterlab-manager
-    jupyter labextension install @jupyter-widgets/jupyterlab-manager@^1.0.01 --no-build && \
```
romainx: Updated to `@jupyter-widgets/jupyterlab-manager@^1.1.0`

scipy-notebook/Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN conda install --quiet --yes \
     'h5py=2.10.*' \
     'hdf5=1.10.*' \
     'ipywidgets=7.5.*' \
+    'ipympl=0.3.*'\
```
romainx: Replaced by `ipympl=0.5.*`

mpmdean: Thanks! I've independently verified that this indeed solves it.

mpmdean:
Thanks to @romainx this is now working. 

I would propose that  the status:Needs Debugging label is removed and that the pull request can be considered for merging.

romainx:
Great thank you. I forgot a point in my review. 
Could you also update the [documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html) to mention ipyml?

mpmdean:
@romainx I've now added mention of ipympl to the selecting an image section of the docs.

romainx:
@mpmdean thank you, everything seems to be fine and ready to merge now 👍.

parente:
Thanks for the PR @mpmdean and for reviewing @romainx. I'm merging master into this branch to make sure there are no build errors before merging back to master.

mpmdean:
@romainx

I am still seeing problems with ipympl in jupyterlab (see the image below). The issue is the same one discussed previously. I've reproduced this both by pulling from dockerhub and by building the current version. I'm stuck as to what happened, as it was working before and I cannot see what change is correlated with the problem.

Screen Shot 2020-03-09 at 2 55 25 PM

romainx:
@mpmdean You're right same result on my side and same lack of explanation ... :expressionless:
I keep you posted if I found something.

mpmdean:
@romainx I tried to backtrack to a working version via:

1. git reset --hard 4bff6ab1978f682ab089baa261ce0361190e411d (which is right after I thought I had a working version and before the ubuntu update you mention)
2. build the base notebook
3. edit the tag of the scipy-notebook to that build and then rebuild.

This failed in the way we are discussing.

romainx:
I will try to investigate tonight. In the meantime you could open an issue in their project since it's easily reproducible.

mpmdean:
I've open an issue.

Issue #1012, opened 2020-02-13
datascience-notebook: newer images generate " Illegal instruction (core dumped)" when importing sympy


jlphillipsphd:
`docker pull jupyter/datascience-notebook:9f4983c5d1f3`
`docker run -it --rm --name temp jupyter/datascience-notebook:9f4983c5d1f3`
in other terminal:
`docker exec -it temp /bin/bash`
Inside the container:
```
jovyan@1016eb8fa911:~$ python
Python 3.7.6 | packaged by conda-forge | (default, Jan  7 2020, 22:33:48)
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sympy
Illegal instruction (core dumped)
```

Expected behavior: importing sympy should not cause a core dump, and allow additional code execution
(Note: the kernel dies within the notebook environment as well when importing sympy)

I have no trouble using an image which is a little older: jupyter/datascience-notebook:7a0c7325e470

Of course, I see the illegal instruction message, so here are my CPU specs (from /proc/cpuinfo within the container):
```
processor       : 0
vendor_id       : AuthenticAMD
cpu family      : 21
model           : 2
model name      : AMD Opteron(tm) Processor 6376
stepping        : 0
microcode       : 0x6000852
cpu MHz         : 2598.211
cache size      : 2048 KB
physical id     : 2
siblings        : 16
core id         : 7
cpu cores       : 8
apicid          : 79
initial apicid  : 79
fpu             : yes
fpu_exception   : yes
cpuid level     : 13
wp              : yes
flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm constant_tsc rep_good nopl nonstop_tsc cpuid extd_apicid amd_dcm aperfmperf pni pclmulqdq monitor ssse3 fma cx16 sse4_1 sse4_2 popcnt aes xsave avx f16c lahf_lm cmp_legacy svm extapic cr8_legacy abm sse4a misalignsse 3dnowprefetch osvw ibs xop skinit wdt lwp fma4 tce nodeid_msr tbm topoext perfctr_core perfctr_nb cpb hw_pstate ssbd ibpb vmmcall bmi1 arat npt lbrv svm_lock nrip_save tsc_scale vmcb_clean flushbyasid decodeassists pausefilter pfthreshold
[...]

Issue #966, opened 2019-12-05
pyspark: undefined symbol in pyarrow


dungdm93:
**What docker image you are using?**

`jupyter/pyspark-notebook`

**Steps**

```console
dungdm93@localhost:~$ docker run -it jupyter/pyspark-notebook bash
jovyan@2c64156fbae6:~$ python
Python 3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 21:52:21) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyarrow
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/conda/lib/python3.7/site-packages/pyarrow/__init__.py", line 54, in <module>
    from pyarrow.lib import cpu_count, set_cpu_count
ImportError: /opt/conda/lib/python3.7/site-packages/pyarrow/lib.cpython-37m-x86_64-linux-gnu.so: undefined symbol: _ZNK5arrow6Status8ToStringB5cxx11Ev
>>>
```

The error say that:
```log
ImportError: /opt/conda/lib/python3.7/site-packages/pyarrow/lib.cpython-37m-x86_64-linux-gnu.so: undefined symbol: _ZNK5arrow6Status8ToStringB5cxx11Ev
```

Bidek56:
Please see https://github.com/jupyter/docker-stacks/issues/944 for resolution but because of a Conda issue, PR won't build and the issue can not be resolved at this time.

parente:
Thanks for reporting @dungdm93. I'm going to close this issue as a duplicate of #944.

PR #1037, opened 2020-03-08
Test packages

romainx:
This test module tests if R and Python packages installed can be imported.
It's a basic test aiming to prove that the package is working properly.

The goal is to detect import errors that can be caused by incompatibilities between packages for example:

- #1012: issue importing `sympy`
- #966: isssue importing `pyarrow`

This module checks dynmamically, through the `CondaPackageHelper`, only the specified packages i.e. packages requested by `conda install` in the `Dockerfiles`.
This means that it does not check dependencies. This choice is a tradeoff to cover the main requirements while achieving reasonabe test duration.
However it could be easily changed (or completed) to cover also dependencies `package_helper.installed_packages()` instead of `package_helper.specified_packages()`.

Example:

    $ make test/datascience-notebook

    # [...]
    # test/test_packages.py::test_python_packages 
    # --------------------------------------------------------------------------------------------- live log setup ----------------------------------------------------------------------------------------------
    # 2020-03-08 09:56:04 [    INFO] Starting container jupyter/datascience-notebook ... (helpers.py:51)
    # 2020-03-08 09:56:04 [    INFO] Running jupyter/datascience-notebook with args {'detach': True, 'ports': {'8888/tcp': 8888}, 'tty': True, 'command': ['start.sh', 'bash', '-c', 'sleep infinity']} ... (conftest.py:78)
    # 2020-03-08 09:56:04 [    INFO] Grabing the list of specifications ... (helpers.py:76)
    # ---------------------------------------------------------------------------------------------- live log call ----------------------------------------------------------------------------------------------
    # 2020-03-08 09:56:07 [    INFO] Testing the import of packages ... (test_packages.py:125)
    # 2020-03-08 09:56:07 [    INFO] Trying to import conda (test_packages.py:127)
    # 2020-03-08 09:56:07 [    INFO] Trying to import notebook (test_packages.py:127)
    # 2020-03-08 09:56:08 [    INFO] Trying to import jupyterhub (test_packages.py:127)
    # [...]

### DIFF
--- a/test/test_packages.py
+++ b/test/test_packages.py
@@ -13,7 +13,7 @@ The goal is to detect import errors that can be caused by incompatibilities betw
 - #966: isssue importing `pyarrow`
 
 This module checks dynmamically, through the `CondaPackageHelper`, only the specified packages i.e. packages requested by `conda install` in the `Dockerfiles`.
-This means that it does not check dependencies. This choice is a tradeoff to cover the main requirements while achieving reasonabe test duration.
+This means that it does not check dependencies. This choice is a tradeoff to cover the main requirements while achieving reasonable test duration.
 However it could be easily changed (or completed) to cover also dependencies `package_helper.installed_packages()` instead of `package_helper.specified_packages()`.
 
 Example:
@@ -50,6 +50,7 @@ PACKAGE_MAPPING = {
     "beautifulsoup4": "bs4",
     "scikit-learn": "sklearn",
     "scikit-image": "skimage",
+    "spylon-kernel": "spylon_kernel1",
     # R
     "randomforest": "randomForest",
     "rsqlite": "DBI",
@@ -62,11 +63,11 @@ EXCLUDED_PACKAGES = [
     "tini",
     "python",
     "hdf5",
+    # FIXME: shall be parsed as conda-forge::blas[build=openblas]
     "conda-forge::blas[build",
     "protobuf",
     "r-irkernel",
     "unixodbc",
-    "spylon-kernel",
 ]
 
 
@@ -78,7 +79,7 @@ def package_helper(container):
 
 @pytest.fixture(scope="function")
 def packages(package_helper):
-    """Return the list of specified packages (i.e. packages explicitely installed excluding dependecnies)"""
+    """Return the list of specified packages (i.e. packages explicitely installed excluding dependencies)"""
     return package_helper.specified_packages()
 
 
@@ -124,10 +125,10 @@ def check_import_r_package(package_helper, package):
     )
 
 
-def _import_packages(package_helper, filtered_packages, check_function):
+def _import_packages(package_helper, filtered_packages, check_function, max_failures):
     """Test if packages can be imported
-    
-    Note: using a list of packages instead of a fixture for the list of packages since pytest prevent to use multiple yields
+
+    Note: using a list of packages instead of a fixture for the list of packages since pytest prevents use of multiple yields
     """
     failures = {}
     LOGGER.info(f"Testing the import of packages ...")
@@ -139,8 +140,10 @@ def _import_packages(package_helper, filtered_packages, check_function):
             ), f"Package [{package}] import failed"
         except AssertionError as err:
             failures[package] = err
-    if failures:
+    if len(failures) > max_failures:
         raise AssertionError(failures)
+    elif len(failures) > 0:
+        LOGGER.warning(f"Some import(s) has(have) failed: {failures}")
 
 
 @pytest.fixture(scope="function")
@@ -152,10 +155,10 @@ def r_packages(packages):
     )
 
 
-def test_python_packages(package_helper, python_packages):
+def test_python_packages(package_helper, python_packages, max_failures=0):
     """Test the import of specified python packages"""
     return _import_packages(
-        package_helper, python_packages, check_import_python_package
+        package_helper, python_packages, check_import_python_package, max_failures
     )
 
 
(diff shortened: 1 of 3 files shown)


test/test_packages.py
```diff
@@ -0,0 +1,170 @@
...
+This test module tests if R and Python packages installed can be imported.
+It's a basic test aiming to prove that the package is working properly.
+
+The goal is to detect import errors that can be caused by incompatibilities between packages for example:
+
+- #1012: issue importing `sympy`
+- #966: isssue importing `pyarrow`
+
+This module checks dynmamically, through the `CondaPackageHelper`, only the specified packages i.e. packages requested by `conda install` in the `Dockerfiles`.
+This means that it does not check dependencies. This choice is a tradeoff to cover the main requirements while achieving reasonabe test duration.
```
parente:
```suggestion
This means that it does not check dependencies. This choice is a tradeoff to cover the main requirements while achieving reasonable test duration.
```

test/test_packages.py
```diff
@@ -0,0 +1,170 @@
...
+However it could be easily changed (or completed) to cover also dependencies `package_helper.installed_packages()` instead of `package_helper.specified_packages()`.
```
parente: I agree with the trade-off you've made here, that we should only be testing direct dependencies for the time being. We have a ~50 minute runtime limit on Travis. We'll likely need to finish #693 to reduce the number of images we build, switch CIs, and/or change build strategies if we decide one day to test *everything*.

test/test_packages.py
```diff
@@ -0,0 +1,170 @@
...
+    "rcurl": "RCurl",
+    "rodbc": "RODBC",
+}
+
+# List of packages that cannot be tested in a standard way
+EXCLUDED_PACKAGES = [
+    "tini",
+    "python",
+    "hdf5",
+    "conda-forge::blas[build",
```
parente: Sanity check: is this the full package name?

test/test_packages.py
```diff
@@ -0,0 +1,170 @@
...
+    "protobuf",
+    "r-irkernel",
+    "unixodbc",
+    "spylon-kernel",
```
parente: You should be able to `import spylon_kernel` to test this one. https://github.com/Valassis-Digital-Media/spylon-kernel#using-it-as-an-ipython-magic

test/test_packages.py
```diff
@@ -0,0 +1,170 @@
...
+
+@pytest.fixture(scope="function")
+def package_helper(container):
+    """Return a package helper object that can be used to perform tests on installed packages"""
+    return CondaPackageHelper(container)
+
+
+@pytest.fixture(scope="function")
+def packages(package_helper):
+    """Return the list of specified packages (i.e. packages explicitely installed excluding dependecnies)"""
```
parente:
```suggestion
    """Return the list of specified packages (i.e. packages explicitly installed excluding dependencies)"""
```

test/test_packages.py
```diff
@@ -0,0 +1,170 @@
...
+    """Try to import a R package from the command line"""
+    return _check_import_package(
+        package_helper, ["R", "--slave", "-e", f"library({package})"]
+    )
+
+
+def _import_packages(package_helper, filtered_packages, check_function):
+    """Test if packages can be imported
+    
+    Note: using a list of packages instead of a fixture for the list of packages since pytest prevent to use multiple yields
```
parente:
```suggestion
    Note: using a list of packages instead of a fixture for the list of packages since pytest prevents use of multiple yields
```

test/test_packages.py
romainx: Typo fixed

test/test_packages.py
romainx: Thanks make sense.

test/test_packages.py
romainx: 😅 In fact it's a bug. The package name as retrieved by the `conda` command is `conda-forge::blas[build=openblas]`. The bug is introduced by the way I split package and versions. Normally packages are expressed like this ` cloudpickle=1.3`. I will have to use something more complex than 
```python
packages_list = map(lambda x: x.split("=", 1), dependencies)
```
:weary: RegExp!

test/test_packages.py
romainx: You're right just need a specific mapping.

```python
# Mapping between package and module name
PACKAGE_MAPPING = {
    # Python
    "spylon-kernel": "spylon_kernel",
```

test/test_packages.py
romainx: Typo fixed

test/test_packages.py
romainx: A french guy who writes in English ...

romainx:
@parente Thanks for the positive feedback and for the relevant review. 😄 

To answer to your question, my opinion is to fail if an import error is raised -- if it's not the case users may have a problem and could open an issue. However in some case it could be interesting to ignore some errors. To be ready to do it, I have modified the code to introduce a `max_failures` parameter.
The parameter is equal to zero by default and can be tuned. If some errors are raised but ignored a warning is printed.

```python
if len(failures) > max_failures:
    raise AssertionError(failures)
elif len(failures) > 0:
    LOGGER.warning(f"Some import(s) has(have) failed: {failures}")
```

I have committed all the change except the parsing of the package bug ... So I'm flagging it as **WIP**.

See you.

test/test_packages.py
romainx: I'm really bad at RegExp, so if if anyone has a suggestion, I'm interested :blush:

parente: I wonder if the conda code to parse the version descriptors is exposed in a public API that we could import and call?

romainx: Good idea, I will check. Everything is better than implementing a RegExp in my point of view 😆.

romainx:
@parente I'm waiting for the CI to finish, but I think it's OK now.
If you're agree, we are ready to merge :tada:

test/test_packages.py
parente: You make fewer typos than I do as a English-speaker. 🤣

parente:
Let's give it a shot! Thanks for putting this testing in place.

Issue #1038, opened 2020-03-10
ipympl not working in scipy-notebook


mpmdean:
There appears to be a problem with ipympl not working within jupyterlab when using scipy-notebook.

1. Launch docker run -p 8888:8888 jupyter/scipy-notebook
2. Switch to jupyterlab mode by prepending lab to the URL.
3. Run notebook commands as:
![76247554-1a674700-6216-11ea-8a01-7d268ebb7746 (1)](https://user-images.githubusercontent.com/699991/76316787-821ea000-62b1-11ea-882a-57df897545cd.png)

**What do you expect to happen?**
Plot to appear 

**What actually happens?**
No plot. Error displaying widget: model not found

romainx:
Hey just submitted a PR #1040 to fix it.

I let you check to confirm it's OK on your side.
Thanks
Best

parente:
Fixed by #1039. 🤞

romainx:
Good job, just pulled the latest version of the `scipy-notebook` and it works as expected with the following versions. Good job 👍 

```bash
jovyan@29e533608d50:~$ jupyter-labextension list
JupyterLab v2.0.1
Known labextensions:
   app dir: /opt/conda/share/jupyter/lab
        @bokeh/jupyter_bokeh v2.0.0  enabled  OK
        @jupyter-widgets/jupyterlab-manager v2.0.0  enabled  OK
        jupyter-matplotlib v0.7.2  enabled  OK
```

PR #1039, opened 2020-03-10
update jupyterlab and plotting extensions

mpmdean:
This version update fixes issue #1038 by updating jupyterlab and the jupyter-matplotlib and bokeh extensions.

### DIFF
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -45,6 +45,11 @@ RUN conda install --quiet --yes \
     'xlrd' \
     && \
     conda clean --all -f -y && \
+    # Activate ipywidgets extension in the environment that runs the notebook server
+    jupyter nbextension enable --py widgetsnbextension --sys-prefix && \
+    # Also activate ipywidgets extension for JupyterLab
+    # Check this URL for most recent compatibilities
+    # https://github.com/jupyter-widgets/ipywidgets/tree/master/packages/jupyterlab-manager
     jupyter labextension install @jupyter-widgets/jupyterlab-manager@^2.0.0 --no-build && \
     jupyter labextension install @bokeh/jupyter_bokeh@^2.0.0 --no-build && \
     jupyter labextension install jupyter-matplotlib --no-build && \
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -45,13 +45,8 @@ RUN conda install --quiet --yes \
     'xlrd' \
     && \
     conda clean --all -f -y && \
-    # Activate ipywidgets extension in the environment that runs the notebook server
-    jupyter nbextension enable --py widgetsnbextension --sys-prefix && \
-    # Also activate ipywidgets extension for JupyterLab
-    # Check this URL for most recent compatibilities
-    # https://github.com/jupyter-widgets/ipywidgets/tree/master/packages/jupyterlab-manager
-    jupyter labextension install @jupyter-widgets/jupyterlab-manager@^1.1.0 --no-build && \
-    jupyter labextension install jupyterlab_bokeh@1.0.0 --no-build && \
+    jupyter labextension install @jupyter-widgets/jupyterlab-manager@^2.0.0 --no-build && \
+    jupyter labextension install @bokeh/jupyter_bokeh@^2.0.0 --no-build && \
     jupyter labextension install jupyter-matplotlib --no-build && \
     jupyter lab build && \
     npm cache clean --force && \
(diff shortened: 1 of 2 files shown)


mpmdean:
@romainx 

This would be my preferred way to solve the issue discussed previously.

romainx:
@mpmdean Sorry I've missed this PR :disappointed: and submitted something similar in #1040 ...
The main difference is that I'm using the `jupyterlab` `1.2.x` instead of the `2.x`.
I do not know the status of the new version of jupyterlab, if it's a good idea to switch to the `2.x`. 
I have no insight about this choice, maybe we could have a look at the [release note](https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html)

Whatever the jpyterlab version we want to use I think it should be better to pin the version of the extension to avoid regressions in the future.

```dockerfile
jupyter labextension install jupyter-matplotlib@0.7.2 --no-build
```

parente:
Thank you both, @romainx  and @mpmdean. I think we should roll with upgrading to JupyterLab 2.0.x as it's the latest stable release and it's always been policy here to upgrade when requested. If users of these images are installing extensions that are made incompatible by the upgrade, they can [pin to a prior image tag until they're ready to take the upgrade](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#versioning).

scipy-notebook/Dockerfile
```diff
@@ -50,8 +50,8 @@ RUN conda install --quiet --yes \
     # Also activate ipywidgets extension for JupyterLab
     # Check this URL for most recent compatibilities
     # https://github.com/jupyter-widgets/ipywidgets/tree/master/packages/jupyterlab-manager
-    jupyter labextension install @jupyter-widgets/jupyterlab-manager@^1.1.0 --no-build && \
-    jupyter labextension install jupyterlab_bokeh@1.0.0 --no-build && \
+    jupyter labextension install @jupyter-widgets/jupyterlab-manager@^2.0.0 --no-build && \
+    jupyter labextension install @bokeh/jupyter_bokeh@^2.0.0 --no-build && \
```
parente: If you could pin jupyter-matplotlib on the next line, we can merge this PR. I'm OK with ^ pinning instead of hard pinning for now. https://www.npmjs.com/package/jupyter-matplotlib suggests the latest is 0.7.2 which @romainx confirmed working in #1040 

```jupyter labextension install jupyter-matplotlib@^0.7.2 --no-build && \```

Comment (mpmdean) [reply scipy-notebook/Dockerfile]: Sure. I've pinned jupyter-matplotlib in the latest commit.

Comment (romainx) [reply scipy-notebook/Dockerfile]: Hello, to be homogeneous we should pin with `^` to be sure to install version greater or equal to `0.7.2` as suggested by @parente.
```diff
+jupyter-matplotlib@^0.7.2
-jupyter-matplotlib@0.7.2
````

File: scipy-notebook/Dockerfile [hunk 2]
```diff
@@ -50,9 +50,9 @@ RUN conda install --quiet --yes \
     # Also activate ipywidgets extension for JupyterLab
     # Check this URL for most recent compatibilities
     # https://github.com/jupyter-widgets/ipywidgets/tree/master/packages/jupyterlab-manager
-    jupyter labextension install @jupyter-widgets/jupyterlab-manager@^1.1.0 --no-build && \
-    jupyter labextension install jupyterlab_bokeh@1.0.0 --no-build && \
-    jupyter labextension install jupyter-matplotlib --no-build && \
+    jupyter labextension install @jupyter-widgets/jupyterlab-manager@^2.0.0 --no-build && \
+    jupyter labextension install @bokeh/jupyter_bokeh@^2.0.0 --no-build && \
+    jupyter labextension install jupyter-matplotlib@0.7.2 --no-build && \
```
parente: I'm fine with either pinning (`bokeh@2.0.0` and `jupyter-matplotlib@0.7.2` exactly) or (`bokeh@^2.0.0` and `^jupyter-matplotlib@0.7.2`). Consistency in how we're pinning across packages makes tends to make problems easier to debug when they arise.

```suggestion
    jupyter labextension install jupyter-matplotlib@^0.7.2 --no-build && \
```

parente: @mpmdean I'm going to assume you're OK with the above change and just want the problem fixed, so I'm hitting the "commit suggestion" button on your behalf to get the ball rolling. 😄

mpmdean:
@parente sure. No problem changing to ^ type pinning.

parente:
Thanks @mpmdean and @romainx for debugging and fixing this. 🍰

PR #1040, opened 2020-03-10
Fix ipympl not working in scipy-notebook #1038

romainx:
Hello,

This is a fix for the issue #1038. Unfortunately I've not understood the root cause of the issue. However my guess is that some versions changes could have led to this issue. In fact [several versions of the lab extension](https://www.npmjs.com/package/jupyter-matplotlib) have been released in the last 10 days.

So to try to avoid this situation I've pinned the version of the extension and I had also to bump the version of `jupyterlab` to the latest `1.2.x`.

With this changes it works with both Jupyter notebook and lab.

JupyterLab_-_Mozilla_Firefox

Best

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -112,7 +112,7 @@ RUN conda install --quiet --yes 'tini=0.18.0' && \
 RUN conda install --quiet --yes \
     'notebook=6.0.3' \
     'jupyterhub=1.1.0' \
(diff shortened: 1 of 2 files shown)


mpmdean:
I have independently confirmed that this works and encourage merging either this PR or #1039 soon in order to fix the bug in the main branch.

romainx:
So I'm closing this PR in favor of #1039 to use jupyterlab `2.x`.

Issue #1043, opened 2020-03-12
dvipng


mpmdean:
I suggest adding dvipng to scipy-notebook. This is used by matplotlib (albeit not by default)

<img width="923" alt="Screen Shot 2020-03-12 at 4 19 50 PM" src="https://user-images.githubusercontent.com/699991/76563335-92837600-647d-11ea-9be5-9e317ea318a1.png">

RuntimeError: Failed to process string with tex because dvipng could not be found

**What docker image you are using?**
`jupyter/scipy-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

jupyter-repo2docker --editable --image-name=jupyter-notebook

**What steps do you take once the container is running to reproduce the issue?**

run the commands in the image included above.

**What do you expect to happen?**

a plot with latex axis labels

**What actually happens?**

No output. RuntimeError: Failed to process string with tex because dvipng could not be found.

romainx:
@mpmdean PR #1044 merged. Thanks.
Could you test if it's OK and close the issue?

Many thanks for your contribution :icecream:

mpmdean:
Yes this is fixed in the version now on docker hub.

Thanks!

PR #1044, opened 2020-03-12
include dvipng

mpmdean:
This responds to #1043 adding the dvipng packaged used by matplotlib to render latex.

I don't think this requires an update for the docs since this tends to only outline the python packages, but a note could be added if needed.


romainx:
Thanks. Seems relevant and sounds good for the PR :smile:

Issue #752, opened 2018-11-07
`jupyter lab clean` - something to be utilized?


consideRatio:
Is using `jupyter lab clean` something that makes sense to utilize?

I recall seeing it used like `RUN jupyter lab build && jupyter lab clean` in a Dockerfile.

It sais by using the `--help` flag of the clean command that...

> This will clean the app directory by removing the staging and static directories.

And I also notice how there are statements like the following to be found in the docker-stacks:

https://github.com/jupyter/docker-stacks/blob/610bb8b938dbd3c5f881aa4628af6a6581897413/scipy-notebook/Dockerfile#L57

parente:
Seems reasonable to switch to `clean` if it accomplishes the same thing or does better at cleaning up working files. Might be we added the lab extensions before it existed or just didn't know about it.

consideRatio:
@dhirschfeld wrote this in the gitter chat of jupyterlab btw, I found it to be related and relevant to consider as well. I don't know what differences will come out of it etc though.

> @consideRatio - I use docker multi-stage builds, build the extensions in a throwaway container then copy the lab folder into my final image. This has the benefit of ensuring you don't end up with any npm cache stuff in your final image and minimises layers.

consideRatio:
Related: https://github.com/jupyterlab/jupyterlab/issues/4930

parente:
Clean all the things!

romainx:
Hello,

The command has evolve with the new `2.x` Jupyterlab version.

> Added options for jupyter lab clean to clean specific parts of the build, such as --extensions, --settings, --static, and --all (#7583, #6734)

- <cite>https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html</cite>

We could only use `jupyter lab clean` to remove the `staging` since other options will lead to not usable extension.

> The `staging` directory is used to create the build and then populate the `static` directory.

- <cite>https://jupyterlab.readthedocs.io/en/stable/user/extensions.html#staging-and-static</cite>

I've tested in the `scipy-notebook` we are already removing the entire folder so there is nothing more to do.

```bash
jovyan@3486515cf88f:~$ jupyter lab clean
[LabCleanApp] Cleaning /opt/conda/share/jupyter/lab...
[LabCleanApp] staging not present, skipping...
[LabCleanApp] Success!
```

However we could replace the `rm` by a call to `jupyter lab clean`, it would be cleaner :grin:

```diff
[...]

PR #1046, opened 2020-03-15
Using jupyter lab clean command instead of rm (#752)

romainx:
As discussed in #752 , just replaced a `rm` by a call to `jupyter lab clean` for clarity reasons.
The result is the same the `$CONDA_DIR/share/jupyter/lab/staging` is removed.

### DIFF
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -54,8 +54,8 @@ RUN conda install --quiet --yes \
     jupyter labextension install @bokeh/jupyter_bokeh@^2.0.0 --no-build && \
     jupyter labextension install jupyter-matplotlib@^0.7.2 --no-build && \
     jupyter lab build && \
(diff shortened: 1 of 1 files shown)


s-rog:
Is there a reason this is not applied to all dockers? The base notebook still uses rm.

romainx:
@s-rog nice catch you must be right — I checked the history of issues. Could you please create and issue or directly draft a PR to fix it?

```bash
$ jupyter lab clean --help
# Clean the JupyterLab application

# This will clean the app directory by removing the `staging` directories.
# Optionally, the `extensions`, `settings`, and/or `static` directories, or the entire contents of the app directory, can also be removed.
```

https://github.com/jupyter/docker-stacks/blob/69c382672cb55cb3253f310c4f9b2379e8df7abc/base-notebook/Dockerfile#L140-L140

Many thanks

PR #1049, opened 2020-03-23
Add noninteractive flag for apt

mausch:
In particular tzdata can hang a docker build because of this.

### DIFF
--- a/minimal-notebook/Dockerfile
+++ b/minimal-notebook/Dockerfile
@@ -8,7 +8,7 @@ LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 USER root
 
 # Install all OS dependencies for fully functional notebook server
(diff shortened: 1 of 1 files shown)


mausch:
Thanks for the quick reply! In my case I was using a different BASE_CONTAINER so that ENV from the base dockerfile wasn't in effect... Having said that I think I actually wanted to set ROOT_CONTAINER instead! So up to you if you want to support changing the BASE_CONTAINER ;-)

romainx:
@mausch Thank you for submitting this and for the additional information you just provided.

Knowing that it makes more sense for me to not implement this change and to let the users the responsibility to define all the needed settings in their own `BASE_CONTAINER` if they choose to replace the standard one.

romainx:
@mausch So I'm closing it. Thank you for your contribution.
Really appreciate.

PR #1052, opened 2020-03-27
[WIP] Rework of start*.sh scripts

### LINKED ISSUE #1034: NOTEBOOK_ARGS passed to image, but has no effect

`jupyter/minimal-notebook`

`docker run -p 8888:8888 -e JUPYTER_ENABLE_LAB=yes -e NOTEBOOK_ARGS="--NotebookApp.token='' " jupyter/minimal-notebook`

Example:

1. `docker ps` to get image id
2. `docker exec -it [container-id] bash` to launch terminal within container
3. `echo $NOTEBOOK_ARGS` returns `--NotebookApp.token=''`

I expect jupyter to launch without token required.

**What actually happens?**

It requires the token

**Extra info**
From what I gathered, `jupyter/minimal-notebook` is inherited from `jupyter/base-notebook`.

According to https://github.com/jupyter/docker-stacks/blob/master/base-notebook/start-singleuser.sh , `$NOTEBOOK_ARGS` are passed to `start.sh`

Then, `--NotebookApp.token=''` is a configuration option that should dispable notebook authentication. But it does not.

The same happens if I try passing  `"--NotebookApp.token=''"` or `"\"--NotebookApp.token=''\""`

I am using Ubuntu 19.10, Docker version 18.09.9, build 1752eb3

### LINKED ISSUE #1053: Wrong PATH in start.sh when running as root

```
#!/bin/bash
EOF

$ chmod a+x wrong_python.sh
$ docker run \
    -it \
    --rm \
    --user root \
    jupyter/base-notebook \
        start.sh python \
;
Set username to: jovyan
usermod: no changes
Executing the command: python
```
```
$ docker run \
    -it \
    --rm \
    --user root \
    jupyter/base-notebook \
        start.sh which python \
;
Set username to: jovyan
usermod: no changes
/opt/conda/bin/python
```

consideRatio:
# PR Summary

Fixes #1053, Fixes #1034, Closes #1054

When we transition from a root user to the actual user that runs the command passed to start.sh, we are doing a lot of complicated things to preserve the environment variables, but we currently fail with this while also having a complicated bash script that is hard to understand.

With this PR, we will start to properly preserve the `PATH`, `LD_LIBRARY_PATH`, 
and `PYTHON*` variables. We will also allow the $NB_USER's that were granted sudo to retain its own PATH when running sudo commands.

The original use case for me was that I needed to set `LD_LIBRARY_PATH` and `GRANT_SUDO`, but `LD_LIBRARY_PATH` was stripped by the default configuration when using `sudo` even with the custom configuration to preserve it, because it was part of the default `env_delete` list of environment variables in the sudoers configuration.

### Related extra part of the PR

I've also added a feature in the last commit, which I could extract to a standalone PR later, introduced an environment variable named `JUPYTER_ENV_VARS_TO_UNSET` that allow us to list environment variables to unset just before we exec the command given to start.sh. This allow hooks to use something sensitive to be stripped from the environment later.

### Review suggestions

Read the PR summary, then review the commit changes commit by commit in order.

# Original outdated text below

When we run `sudo -u jovyan`, even with `-E` (`--preserve-env`), we end
up with reset environment variables. By running `sudo -V` as a root user
or `sudo sudo -V` we get information about what environment variables
will be reset when we would do for example `sudo -E -u jovyan`.

```
Environment variables to remove:
        *=()*
        RUBYOPT
        RUBYLIB
        PYTHONUSERBASE
        PYTHONINSPECT
        PYTHONPATH
        PYTHONHOME
        TMPPREFIX
        ZDOTDIR
        READNULLCMD
        NULLCMD
        FPATH
        PERL5DB
        PERL5OPT
        PERL5LIB
        PERLLIB
        PERLIO_DEBUG
        JAVA_TOOL_OPTIONS
        SHELLOPTS
        BASHOPTS
        GLOBIGNORE
        PS4
        BASH_ENV
        ENV
        TERMCAP
        TERMPATH
        TERMINFO_DIRS
        TERMINFO
        _RLD*
        LD_*
        PATH_LOCALE
        NLSPATH
        HOSTALIASES
        RES_OPTIONS
        LOCALDOMAIN
        CDPATH
        IFS
```

A use case I ran into was the need to preserve the `LD_LIBRARY_PATH`
variable. I want to help provide information about where to find
libraries installed after the image is built. If they were installed
before the image was built I could have done...

```dockerfile
RUN echo "/usr/local/nvidia/lib64" > /etc/ld.so.conf.d/nvidia.conf \
 && ldconfig
```

... which would have `ldconfig` search through the folder and summarize
information about libraries into `/etc/ld.so.cache`. But, they are not
there yet, and there will be plenty of files in there later.

Anyhow. This PR ensures we preserve LD_LIBRARY_PATH as well when we
switch from the root user to the jovyan kind of user with sudo
privileges.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -90,8 +90,13 @@ if [ $(id -u) == 0 ] ; then
     # home directory as a fallback if they don't have one mounted already.
     if [[ "$NB_USER" != "jovyan" ]]; then
         if [[ ! -e "/home/$NB_USER" ]]; then
-            echo "Relocating home dir to /home/$NB_USER"
-            mv /home/jovyan "/home/$NB_USER" || ln -s /home/jovyan "/home/$NB_USER"
+            echo "Attempting to move /home/jovyan to /home/${NB_USER}..."
+            if mv /home/jovyan "/home/$NB_USER"; then
+                echo "Done"
+            else
+                echo "Failed. Attempting to symlink /home/jovyan to /home/${NB_USER}..."
+                ln -s /home/jovyan "/home/$NB_USER" && echo "Done"
+            fi
         fi
         # Ensure the current working directory is updated
         if [[ "$PWD/" == "/home/jovyan/"* ]]; then
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -48,6 +48,9 @@ if [ $(id -u) == 0 ] ; then
         echo "Set username to: $NB_USER"
         usermod -d /home/$NB_USER -l $NB_USER jovyan
     fi
+    # Update any environment variables we set during build of the
+    # Dockerfile that contained the home directory path.
+    export XDG_CACHE_HOME=/home/$NB_USER/.cache
 
     # Handle case where provisioned storage does not have the correct permissions by default
     # Ex: default NFS/EFS (no auto-uid/gid)
@@ -88,20 +91,43 @@ if [ $(id -u) == 0 ] ; then
         useradd --home /home/$NB_USER -u $NB_UID -g $NB_GID -G 100 -l $NB_USER
     fi
 
-    # Enable sudo if requested
+    # Conditionally enable passwordless sudo usage for the jovyan user
     if [[ "$GRANT_SUDO" == "1" || "$GRANT_SUDO" == 'yes' ]]; then
-        echo "Granting $NB_USER sudo access and appending $CONDA_DIR/bin to sudo PATH"
-        echo "$NB_USER ALL=(ALL) NOPASSWD:ALL" > /etc/sudoers.d/notebook
+        echo "Granting $NB_USER passwordless sudo rights!"
+        echo "$NB_USER ALL=(ALL) NOPASSWD:ALL" > /etc/sudoers.d/added-by-start-script
     fi
 
-    # Add $CONDA_DIR/bin to sudo secure_path
-    sed -r "s#Defaults\s+secure_path=\"([^\"]+)\"#Defaults secure_path=\"\1:$CONDA_DIR/bin\"#" /etc/sudoers | grep secure_path > /etc/sudoers.d/path
+    # Ensure that the initial environment that this container is started with
+    # is preserved when we run transition from running as root to running as
+    # NB_USER.
+    #
+    # - We use the sudo command to execute the command as NB_USER. But, what
+    #   happens to the environment will be determined by configuration in
+    #   /etc/sudoers and /etc/sudoers.d/* as well as flags we pass to the sudo
+    #   command. The behavior can be inspected with `sudo -V` run as root.
+    #
+    #   ref: `man sudo` - https://linux.die.net/man/8/sudo ref: `man sudoers` -
+    #   https://www.sudo.ws/man/1.8.15/sudoers.man.html
+    #
+    # - We use the `--preserve-env` flag to pass through most environment, but
+    #   understand that exceptions are caused by the sudoers configuration:
+    #   `env_delete`, `env_check`, and `secure_path`.
+    #
+    # - We reduce the `env_delete` list of default variables to be deleted by
+    #   default which would ignore the `--preserve-env` flag and `env_keep`
+    #   configuration.
+    #
+    # - We manage the PATH variable specifically as `secure_path` is set by
+    #   default in /etc/sudoers and would override the PATH variable. So we
+    #   disable that default.
+    echo 'Defaults env_delete -= "PATH LD_* PYTHON*"' >> /etc/sudoers.d/added-by-start-script
+    echo 'Defaults !secure_path' >> /etc/sudoers.d/added-by-start-script
 
-    # Exec the command as NB_USER with the PATH and the rest of
-    # the environment preserved
+    # NOTE: This hook is run as the root user!
     run-hooks /usr/local/bin/before-notebook.d
-    echo "Executing the command: ${cmd[@]}"
-    exec sudo -E -H -u $NB_USER PATH=$PATH XDG_CACHE_HOME=/home/$NB_USER/.cache PYTHONPATH=${PYTHONPATH:-} "${cmd[@]}"
+
+    echo "Running as $NB_USER with preserved environment: ${cmd[@]}"
+    exec sudo --preserve-env --set-home --user $NB_USER "${cmd[@]}"
 else
     if [[ "$NB_UID" == "$(id -u jovyan)" && "$NB_GID" == "$(id -g jovyan)" ]]; then
         # User is not attempting to override user/group via environment
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -38,6 +38,18 @@ run-hooks () {
     echo "$0: done running hooks in $1"
 }
 
+# A helper function to unset env vars listed in the value of the env var
+# JUPYTER_ENV_VARS_TO_UNSET.
+unset_explicit_env_vars () {
+    if [ ! -z "$JUPYTER_ENV_VARS_TO_UNSET" ]; then
+        for env_var_to_unset in $(echo $JUPYTER_ENV_VARS_TO_UNSET | tr ',;:' ' '); do
+            echo "Unset ${env_var_to_unset} due to JUPYTER_ENV_VARS_TO_UNSET"
+            unset ${env_var_to_unset}
+        done
+        unset JUPYTER_ENV_VARS_TO_UNSET
+    fi
+}
+
 
 
 # NOTE: This hook will run as the user the container was started with!
@@ -126,7 +138,9 @@ if [ $(id -u) == 0 ] ; then
 
     # Ensure that the initial environment that this container is started with
     # is preserved when we run transition from running as root to running as
-    # NB_USER.
+    # NB_USER. But, make an exception for environment variables explicitly
+    # listed in the JUPYTER_ENV_VARS_TO_UNSET environment variable, let these be
+    # unset before the transition.
     #
     # - We use the sudo command to execute the command as NB_USER. But, what
     #   happens to the environment will be determined by configuration in
@@ -153,7 +167,8 @@ if [ $(id -u) == 0 ] ; then
     # NOTE: This hook is run as the root user!
     run-hooks /usr/local/bin/before-notebook.d
 
-    echo "Running as $NB_USER with preserved environment: ${cmd[@]}"
+    unset_explicit_env_vars
+    echo "Running (as $NB_USER): ${cmd[@]}"
     exec sudo --preserve-env --set-home --user $NB_USER "${cmd[@]}"
 
 
@@ -199,6 +214,7 @@ else
     fi
 
     run-hooks /usr/local/bin/before-notebook.d
+    unset_explicit_env_vars
     echo "Running: ${cmd[@]}"
     exec "${cmd[@]}"
 fi
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -11,24 +11,26 @@ else
     cmd=( "$@" )
 fi
 
+# The run-hooks function looks for .sh scripts and executable files
+# - .sh scripts will be sourced
+# - executable files (+x) will be executed
 run-hooks () {
-    # Source scripts or run executable files in a directory
     if [[ ! -d "$1" ]] ; then
         return
     fi
-    echo "$0: running hooks in $1"
+    echo "$0: running hooks in $1 as uid / gid: $(id -u) / $(id -g)"
     for f in "$1/"*; do
         case "$f" in
             *.sh)
-                echo "$0: running $f"
+                echo "$0: running script $f"
                 source "$f"
                 ;;
             *)
                 if [[ -x "$f" ]] ; then
-                    echo "$0: running $f"
+                    echo "$0: running executable $f"
                     "$f"
                 else
-                    echo "$0: ignoring $f"
+                    echo "$0: ignoring non-executable $f"
                 fi
                 ;;
         esac
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -12,9 +12,8 @@ else
     cmd=( "$@" )
 fi
 
-# The run-hooks function looks for .sh scripts and executable files
-# - .sh scripts will be sourced
-# - executable files (+x) will be executed
+# The run-hooks function looks for .sh scripts to source and executable files to
+# run within a passed directory.
 run-hooks () {
     if [[ ! -d "$1" ]] ; then
         return
@@ -39,8 +38,8 @@ run-hooks () {
     echo "$0: done running hooks in $1"
 }
 
-# The unset_explicit_env_vars function unset env vars listed in the value of the
-# env var JUPYTER_ENV_VARS_TO_UNSET.
+# The unset_explicit_env_vars function unset environment variables listed in the
+# value of JUPYTER_ENV_VARS_TO_UNSET.
 unset_explicit_env_vars () {
     if [ ! -z "$JUPYTER_ENV_VARS_TO_UNSET" ]; then
         for env_var_to_unset in $(echo $JUPYTER_ENV_VARS_TO_UNSET | tr ',;:' ' '); do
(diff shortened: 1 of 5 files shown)


rkdarst:
This looks like the kind of thing that would be useful to me, too... I have done some big workarounds to set up my users' environments, including setting things like environment variables.

My second thought is "there's a reason these are removed", but it doesn't fully apply here.  When starting the image, we sudo root -> user.  But if sudo is granted to users, then one can do user->root, and security could matter more.  It's not inconceivable to me that someone would allow users to run certain commands with sudo, without wanting them to have full root access.

I don't know the sudoers syntax myself, but could this be limited to just the root -> user direction?

consideRatio:
Hi Richard!

I think as we have it setup right now, the user can escape and become a root user using sudo, and then it can do whatever it pleases, such as edit these files. I think... This PR is mainly a small refactoring and allowing also LD_LIBRARY_PATH and environment variables match PYTHON* and not only PYTHONPATH to be passed from the initial environment to the `jovyan` environment.

consideRatio:
I ran into this error, I wonder if it is a temporary fluke though, hmm...

```
=================== 3 passed, 1 deselected in 15.26 seconds ====================
docker build --build-arg TEST_ONLY_BUILD=1 --rm --force-rm -t jupyter/tensorflow-notebook:latest ./tensorflow-notebook
Sending build context to Docker daemon  8.192kB
Step 1/4 : ARG BASE_CONTAINER=jupyter/scipy-notebook
Step 2/4 : FROM $BASE_CONTAINER
 ---> d0b40dffbc41
Step 3/4 : LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 ---> Running in b2d80307fae2
Removing intermediate container b2d80307fae2
 ---> c2d1fcea414c
Step 4/4 : RUN pip install --quiet     'tensorflow==2.1.0' &&     fix-permissions $CONDA_DIR &&     fix-permissions /home/$NB_USER
 ---> Running in f5c16fadec8d
ERROR: Exception:
  File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/urllib3/contrib/pyopenssl.py", line 313, in recv_into
    return self.connection.recv_into(*args, **kwargs)
  File "/opt/conda/lib/python3.7/site-packages/OpenSSL/SSL.py", line 1840, in recv_into
    self._raise_ssl_error(self._ssl, result)
  File "/opt/conda/lib/python3.7/site-packages/OpenSSL/SSL.py", line 1646, in _raise_ssl_error
    raise WantReadError()
OpenSSL.SSL.WantReadError
During handling of the above exception, another exception occurred:
  File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/urllib3/contrib/pyopenssl.py", line 313, in recv_into
    return self.connection.recv_into(*args, **kwargs)
  File "/opt/conda/lib/python3.7/site-packages/OpenSSL/SSL.py", line 1840, in recv_into
    self._raise_ssl_error(self._ssl, result)
  File "/opt/conda/lib/python3.7/site-packages/OpenSSL/SSL.py", line 1663, in _raise_ssl_error
    raise SysCallError(errno, errorcode.get(errno))
OpenSSL.SSL.SysCallError: (104, 'ECONNRESET')
During handling of the above exception, another exception occurred:
  File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/urllib3/response.py", line 425, in _error_catcher
    yield
  File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/urllib3/response.py", line 507, in read
    data = self._fp.read(amt) if not fp_closed else b""
  File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/cachecontrol/filewrapper.py", line 62, in read
    data = self.__fp.read(amt)
  File "/opt/conda/lib/python3.7/http/client.py", line 457, in read
    n = self.readinto(b)
  File "/opt/conda/lib/python3.7/http/client.py", line 501, in readinto
    n = self.fp.readinto(b)
  File "/opt/conda/lib/python3.7/socket.py", line 589, in readinto
    return self._sock.recv_into(b)
  File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/urllib3/contrib/pyopenssl.py", line 328, in recv_into
    return self.recv_into(*args, **kwargs)
  File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/urllib3/contrib/pyopenssl.py", line 318, in recv_into
    raise SocketError(str(e))
OSError: (104, 'ECONNRESET')
During handling of the above exception, another exception occurred:
  File "/opt/conda/lib/python3.7/site-packages/pip/_internal/cli/base_command.py", line 186, in _main
    status = self.run(options, args)
  File "/opt/conda/lib/python3.7/site-packages/pip/_internal/commands/install.py", line 331, in run
    resolver.resolve(requirement_set)
  File "/opt/conda/lib/python3.7/site-packages/pip/_internal/legacy_resolve.py", line 177, in resolve
    discovered_reqs.extend(self._resolve_one(requirement_set, req))
  File "/opt/conda/lib/python3.7/site-packages/pip/_internal/legacy_resolve.py", line 333, in _resolve_one
    abstract_dist = self._get_abstract_dist_for(req_to_install)
  File "/opt/conda/lib/python3.7/site-packages/pip/_internal/legacy_resolve.py", line 282, in _get_abstract_dist_for
    abstract_dist = self.preparer.prepare_linked_requirement(req)
  File "/opt/conda/lib/python3.7/site-packages/pip/_internal/operations/prepare.py", line 482, in prepare_linked_requirement
    hashes=hashes,
  File "/opt/conda/lib/python3.7/site-packages/pip/_internal/operations/prepare.py", line 287, in unpack_url
    hashes=hashes,
  File "/opt/conda/lib/python3.7/site-packages/pip/_internal/operations/prepare.py", line 159, in unpack_http_url
    link, downloader, temp_dir.path, hashes
  File "/opt/conda/lib/python3.7/site-packages/pip/_internal/operations/prepare.py", line 303, in _download_http_url
    for chunk in download.chunks:
  File "/opt/conda/lib/python3.7/site-packages/pip/_internal/network/utils.py", line 39, in response_chunks
    decode_content=False,
  File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/urllib3/response.py", line 564, in stream
    data = self.read(amt=amt, decode_content=decode_content)
  File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/urllib3/response.py", line 529, in read
    raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
  File "/opt/conda/lib/python3.7/contextlib.py", line 130, in __exit__
    self.gen.throw(type, value, traceback)
  File "/opt/conda/lib/python3.7/site-packages/pip/_vendor/urllib3/response.py", line 443, in _error_catcher
    raise ProtocolError("Connection broken: %r" % e, e)
pip._vendor.urllib3.exceptions.ProtocolError: ('Connection broken: OSError("(104, \'ECONNRESET\')")', OSError("(104, 'ECONNRESET')"))
Removing intermediate container f5c16fadec8d
The command '/bin/sh -c pip install --quiet     'tensorflow==2.1.0' &&     fix-permissions $CONDA_DIR &&     fix-permissions /home/$NB_USER' returned a non-zero code: 2
Makefile:47: recipe for target 'build/tensorflow-notebook' failed
make: *** [build/tensorflow-notebook] Error 2
```

consideRatio:
I'm testing this on my JupyterHub deployment and failed to have both the PATH and LD_LIBRARY_PATH transferred now. Hmm......

consideRatio:
I'm digging in on this issue. I'm currently on my way to resolve it by properly reviewing the start.sh script and learning everything about sudoers.

maresb:
Woah, crazy timing! I just submitted this.  https://github.com/jupyter/docker-stacks/issues/1053

maresb:
I like your approach, actually better than mine. Simply passing through PATH solves my problem and ensures the expected behavior.

The question of security implications seems subtle. The `secure_path` stuff was added by @parente in https://github.com/jupyter/docker-stacks/commit/6fa67cc. If he's available it'd be great to hear his thoughts.

rkdarst:
> I think as we have it setup right now, the user can escape and become a root user using sudo, and then it can do whatever it pleases, such as edit these files.

Only if `GRANT_SUDO=1`, an explicit opt in.  If this was the default, almost certainly somewhere in the world, people could access data not intended for them.  At least nbgrader will have security problems.

Of course this only matters if someone opts-in.  But they could opt-in with sudo permission to only sudo to run certain commands.  But then this change decreases that security level.  Of course we don't support people limiting to certain commands, but it could be done and some expects it to be as secure as regular sudo.   and if something is made to be secure, I'd rather make decisions explicit.

... maybe we already do worse things and this doesn't matter.

base-notebook/start.sh
```diff
@@ -2,6 +2,34 @@
...
+#   have been the intention so far looking in script.sh, even though it turned
+#   out that secure_path had the final say which overrode other parts.
+# - Will linux user ids always be either not coupled, or uniquely coupled, to a
+#   username?
+# - Is it correct that the filesystem will reference permission to user IDs, and
+#   the /etc/passwd file will couple these IDs to usernames?
+# - Can multiple usernames be coupled to the userid 0?
+# - In this code, we see a big if/else clause about "id --uid == 0", but "id
+#   --gid == 0" would be quite powerful as well right? Would it make sense to
+#   check for either "id --uid == 0" or "id --gid == 0"?
```
consideRatio: @rkdarst and @maresb, perhaps you can help me understand some of these questions I've raised above?

maresb:
Fantastic work, @consideRatio! I really like how you improved the documentation and made things more explicit.

As for your questions, I don't know any of the answers off the top of my head. I think I'm at a similar level to you; I would only be able to find the answers by digging into the man pages (and possibly source code) for several hours.

As for the security implications, I am getting confused by the following: 

> [This is me, typing in a block quote for indentation.]
> 
> So `start.sh` is effectively two scripts: one root user, one non-root user. (The non-root-user script we can ignore for this discussion.) As far as I understand, the root-user part of `start.sh` is intended to be run with `docker run --user root`. It modifies `/etc/sudoers` and changes the user to `jovyan`.
>
> As I understand @rkdarst, we should avoid making changes to `jovyan`'s resulting environment which would weaken security, for example, allowing `jovyan` to pass `PATH` for his `sudo` commands, allowing him to effectively replace the system binaries with his own. Thus in any case, we should forbid `jovyan` from passing `PATH` to `sudo`, but maybe we could allow it for `sudo` only from `root`.
>
> Now if `jovyan` can run `sudo start.sh`, and `sudo start.sh` can pass `PATH` to `sudo`, **then we have effectively allowed `jovyan` to pass `PATH` to `sudo`**. 
>
> Maybe this is okay for the following reason. If someone restricts `jovyan` to only be able to use `sudo` with certain commands, the  `start.sh` would presumably not be on the list.

I wish I could be more help here, but I don't actually have experience with configuring `/etc/sudoers`.

@rkdarst, assuming that we can configure `/etc/sudoers` so that `root` can pass any environment variables to `sudo` but other users cannot, would that be an acceptable solution?

maresb:
I realized that it would be good to explicitly decide what is the appropriate default `secure_path` for `jovyan` running with `GRANT_SUDO`. The current settings in the master branch are in my opinion all scrambled: `start.sh` tries to pass `PATH` in the `sudo` command, but this is overridden by `secure_path`. If we want to keep `secure_path` secure, then it shouldn't have any directories owned by `jovyan`. However, `start.sh` currently appends `/opt/conda/bin`! Moreover in this context, appending (as opposed to prepending) does not seem sensible, since the resulting environment has the mixture of `/usr/bin/python` and `/opt/conda/bin/conda`.

Some possibilities:

* From a user's perspective, I would naively expect `sudo` to carry over my full Conda environment and (almost) all environment variables.

* From a security-conscious sysadmin perspective, I might expect `sudo` to have a path consisting only of system binaries.

Unfortunately these two are incompatible.

To argue against the security-conscious sysadmin perspective, the [documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#docker-options) states explicitly in bold, "**You should only enable sudo if you trust the user or if the container is running on an isolated host.**" I feel like if someone enables `GRANT_SUDO` but doesn't fully trust the user, then it's their responsibility to properly configure their `sudoers` to fit their needs, not ours.

maresb:
I am trying to understand a bit about how `sudoers` works. As I understand, in the master branch, `start.sh` writes a new `Defaults secure_path=` line to `/etc/sudoers.d/path`. The `sudoers` file uses `#` as a comment character, _except for the last line `#includedir /etc/sudoers.d`_, which is a directive containing the `#` character (WTF). Thus it seems that the `Defaults secure_path=` in `/etc/sudoers.d/path` overrides the one from `/etc/sudoers`.

If we decide to make a `secure_path` just for `jovyan`, it looks like we can do `Defaults:jovyan secure_path=`...

consideRatio:
@maresb yeah haha #include is a statement which has meaning i think, while # include would be a comment :)

I've started to think quite clearly now about everything:

**Our modification of `secure_path`:**
1. If we don't grant sudo, our choice of secure_path configuration for sudoers will only impact the single user of sudo we make from within start.sh, where we seemingly have wanted to pass along PATH entirely, and seem safe to me.
2. If we do grant sudo, we have the option to limit the jovyan user we start up as by having `secure_path` set, but, this jovyan could then decide to simply do `sudo su` and update the /etc/sudoers` file for example, so it doesn't make sense to me that we trouble the sudo empowered jovyan but instead let this power user not run into such trouble, so, it secure_path should be disabled no matter what!

**Our use of -E / --preserve-env:**
We use `sudo --preserve-env` (`-E`), and that will make all set environment variables when the Dockerfile started (except certain such as LD_LIBRARY_PATH), be passed when we switch from root to the actual user. That this happens is very reasonable to me, since not starting as root would grant the jovyan user that environment directly.

I think there shouldn't be a difference between starting the docker container as root or jovyan with regards to what the user will have in its environment, and we should assume that any environment variables set when starting the container should be presumed accessible to the actual user we transition to run as in the start.sh script. But... If we want to hide something from the user, we can create an exception, for example of variables prefixed with `ROOT_`.

consideRatio:
@maresb I made plenty of changes and made this PR a lot more readable.
@parente I now consider it ready for review if you would have some time to spend on this.

I've updated the initial post with a PR summary and wrote some review suggestions to help you avoid reading too much.

rkdarst:
You know, it just occured to me that #787 is relevant here (which I've been using at least as long as it's been open).  These problems come because we try to start the directly starts the notebook, without a chance for user configuration.  In my cluster, I use #787 and then have an extensive hooks directory that would do things like this after the sudo.  It seems more natural we are doing complex things here, so first the system is booted, then we change to the user, then the user's environment is set up, then the notebook runs.

If I reworked this, I would remove all the conda stuff before the sudo, and have a user hook that does "source activate conda" in there.  This doesn't work if someone starts the image manually and sudos manually as part of testing stuff (but well, then what we see here wouldn't happen either...).

base-notebook/start.sh
rkdarst:
```
# - Can multiple usernames be coupled to the userid 0?
# - In this code, we see a big if/else clause about "id --uid == 0", but "id
#   --gid == 0" would be quite powerful as well right? Would it make sense to
#   check for either "id --uid == 0" or "id --gid == 0"? 
```

I think the user management (`usermod`, `mv /home/jovyan ...`) won't work with just gid=0 ... it needs to be uid=0 to have those permissions.  Backed up by a quick but not very complete test...

consideRatio:
Thanks for linking that PR! I have been thinking about these hooks as well, lets work on this also while making it safe to merge without risk breaking peoples deployments! 

Really hlad to have you thinking about these issues with me @rkdarst and @marrsb! :heart:

maresb:
I don't want to sidetrack this PR, but for the record, my current use case for `start.sh` is to use it as a wrapper for executing commands in the container from the host via `docker exec`.  Thus I would make a "container run" command like
```
alias crun='docker exec my-container start.sh'
```
so that I can do something like `crun ls` to run commands as `jovyan`. I can see two issues with this:

1. We might not want to run all hooks in such a scenario. We might need two types of hooks: startup hooks and environment hooks. Startup hooks are intended to run once and start necessary services, while environment hooks help achieve the correct environment for running commands as above. (Then we would want the cartesian product of these with @rkdarst's user/root hooks, and things get complicated...)

2. Logging is going to cause conflicts. In addition to seeing the output of `ls` the output of the debugging commands would be mixed in. My ideal solution to this would be to have a default log-level of info and output to `stderr`, but make this adjustable via environment variables. It may be useful to use an established logging library. I did a few quick searches and found [logr](https://brettterpstra.com/projects/logr/), but I have never tried it out.

With @consideRatio's major improvements already in this PR, perhaps it's wise to take this one step at a time, deferring these to a follow-up PR.

consideRatio:
I ended up making a larger rework of this again, i really want to get this less error probe and easy to debug and grasp in the future. A refactoring and documentation update is incomming.

rkdarst:
[this is written before @consideRatio's latest stuff, so is probably out of date.  I also forgot to submit the previous comment earlier in the day - oops!]

1. I approve of this PR - it may not be what I would do, and seems like a more compiled solution than direct approaches of explicitly passing the variables.  But if it's what it takes for progress...

2. While studying docker `ENTRYPOINT` vs `CMD`, it seems that ENTRYPOINT can be used for env setup that should "always" be done, and CMD is the overridden thing.  so, must of the work of start.sh can be put there, and leave the default CMD to just the jupyter-related stuff.  This would elegantly handle @maresb's case.  It's worth asking: what happens if we don't want this setup to run?  If everything is controlled by environment variables only, not setting any env vars would leave us the default image.

But when there are startup vs environment hooks - I'm not sure how that affects the above.

3. 
>     2. Logging is going to cause conflicts.
Yeah, I agree logging can be minimal by default, enabled only by env var.  Can be done later.

consideRatio:
It is not clear to me why logging would be problematic, do you have an example in mind @rkdarst?

I'm not sure I understood so much from the latest post.

- docker's ENTRYPOINT and CMD, or Kubernetes equivalent `command` and `args`, are two parts that combine to the final command to run, which will be influenced if one or the other or both are set. I don't think it can provide any special features for us with regards to manipulating the environment.

> It's worth asking: what happens if we don't want this setup to run? If everything is controlled by environment variables only, not setting any env vars would leave us the default image.

What did "this setup" refer to more specifically?

If we want to not run anything relating to this, we should not use any of the `start*.sh` scripts from this docker image.

I'm aiming for that this PR shouldn't introduce any breaking changes. It will make a small influence on a user with sudo rights that use the `sudo` command with `-E` - that environment will be a bit richer as LD_LIBRARY_PATH, PATH, and PYTHON* will be passed along where it wasn't before.

base-notebook/start.sh
```diff
@@ -11,95 +12,169 @@ else
...
+    #   disable that default.
+    echo 'Defaults env_delete -= "PATH LD_* PYTHON*"' >> /etc/sudoers.d/added-by-start-script
+    echo 'Defaults !secure_path' >> /etc/sudoers.d/added-by-start-script
 
-    # Exec the command as NB_USER with the PATH and the rest of
-    # the environment preserved
+    # NOTE: This hook is run as the root user!
     run-hooks /usr/local/bin/before-notebook.d
-    echo "Executing the command: ${cmd[@]}"
-    exec sudo -E -H -u $NB_USER PATH=$PATH XDG_CACHE_HOME=/home/$NB_USER/.cache PYTHONPATH=${PYTHONPATH:-} "${cmd[@]}"
```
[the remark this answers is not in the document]
maresb: > When the PATH=$PATH is here, it actually does the right thing and path is correct (/opt/conda/bin is in the front, not at the end).

I'm pretty sure that `PATH=$PATH` is not doing the right thing here, even though it may look like it is. See #1053 for details. It's very subtle, and even the `which` command gives the "wrong" answer.

maresb:
@consideRatio, the problem with logging is explained in my https://github.com/jupyter/docker-stacks/pull/1052#issuecomment-605630703 regarding using `start.sh` as a wrapper to run commands in the container from the host. For example, I might want to get the output of `ls` from the host:

> In addition to seeing the output of `ls` the output of the debugging commands would be mixed in.

rkdarst:
> > When the PATH=$PATH is here, it actually does the right thing and path is correct (/opt/conda/bin is in the front, not at the end).
> 
> I'm pretty sure that `PATH=$PATH` is not doing the right thing here, even though it may look like it is. See #1053 for details. It's very subtle, and even the `which` command gives the "wrong" answer.

```console
docker run -u root -it --rm aaltoscienceit/notebook-server:2.0.5 bash   # my derivative of scipy-notebook

# Set up the sudoers file - since it probably matters.  Copied from master start.sh.
# sed -r "s#Defaults\s+secure_path=\"([^\"]+)\"#Defaults secure_path=\"\1:$CONDA_DIR/bin\"#" /etc/sudoers | grep secure_path > /etc/sudoers.d/path

# cat /etc/sudoers.d/path
Defaults secure_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin:/opt/conda/bin"

# sudo -E -H -u jovyan env | grep PATH
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin:/opt/conda/bin
# conda bin at the end

# sudo -E -H -u jovyan PATH=$PATH env | grep PATH
PATH=/opt/conda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
# conda bin at the front
```

I also tested this before, and just did more exhaustive tests now.  It has also always worked on my production systems... what do you run to get it to not work?

maresb:
@rkdarst, see my session in #1053. Even if the `PATH` gets passed by `sudo` it still gets ignored by `sudo` itself. (Apologies if that's obvious and if I'm misinterpreting "doing the right thing".)

rkdarst:
On Sun, Mar 29, 2020 at 03:43:57PM -0700, Ben Mares wrote:
> @rkdarst, see my session in #1053. Even if the `PATH` gets passed by `sudo` it still gets ignored by `sudo` itself. (Apologies if that's obvious and if I'm misinterpreting "doing the right thing".)

Ah right, you are directly running `sudo python` - not starting some
other shells and then starting python. (but I guess it would work for
LD_LIBRARY_PATH) Yes, that's a different matter and doesn't work...

Incidentally, running it through both the root and user sides would
work with the existing form, since the bash subshell would have the
right paths...

romainx:
Hello,

I did not have the time to review all the changes. But maybe while performing this big refactoring, it could be interesting (or not) to also implement what has been suggested here: https://github.com/jupyter/docker-stacks/issues/1034#issuecomment-603507044.

What do you think about that? Does it make sense?

Best

consideRatio:
@romainx I added a commit regarding #1034!

This is now no longer a PR to fix a specific thing, but a bigger rework of the startup scripts.

consideRatio:
# Current considerations

I'm letting this PR become a workbench fix all kinds of issues and limitations relating to the startup scripts. Here are some considerations.

### root/user based hooks

@rkdarst have made work in #787 regarding hooks to run as root / user.

- [ ] -user / -root suffixed hooks added
- [ ] tests with mounted hooks added

### exec command in an already started container as a user

`docker exec` or `kubectl exec` can run a command in an already started container. This is also the mechanism used in Kubernetes pod lifecycle hooks.

When a `exec` commands run, it will be run independently from the command that got the container started in the first place, which means any root -> user transitions made in past executions of `start.sh` are irrelevant and we may want to switch to the user we configured in the initial call to `start.sh`.

It would be good to have a documented approach on how to do this. We may need a dedicated helper script since re-executing start.sh could trigger chown and hooks etc to re-execute. At the same time, we have some parts in start.sh that we may want to also have in this script such as the newly added JUPYTER_ENV_VARS_TO_UNSET.

Currently I suggest we extract the "run command" part of `start.sh` to be reused in a standalone manner without triggering user creation, chown, hooks, etc.

```
# 1. Waiting for the initial start.sh to finish can be vital as it creates the user etc.
# <waiting logic>
# Implementation idea: we let file existing/not-existing indicate if start.sh has finished running hooks etc, and let this be created/deleted in start.sh just before running the passed command.

# 2. Run command as user
if [ $(id -u) == 0 ]; then
    exec sudo --preserve-env --set-home --user $NB_USER "${cmd[@]}"
else
    exec "${cmd[@]}"
fi
```

### NB_UMASK implementation adjustment

I think it may be problematic that the NB_UMASK configuration from https://github.com/jupyter/docker-stacks/pull/781 is applied from the `jupyter_notebook_config.py` as compared to in the startup script. Perhaps this should be part of the transition to user space instead of being part of the jupyter_notebook_config.py that assume you only need it for your jupyter server.

---

@rkdarst I think your _initial questions_ section in #1055 is addressed by this discussion. I'm not sure on how to go, but I think it can be suitable to:
- Add `-user.d` and `-root.d` hooks.
- Extract the final launch command logic into a separate script that can be run by itself.
  - Let this script be able to: ensure start.sh has run hooks once by a waiting mechanism.
  - Let this script be able to: adjusts to getting started as root/user.

consideRatio:
@rkdarst I'm struggling with sourcing of scripts. I don't know how the start.sh script started as root can source something that is meant to run as the user, and I dislike to introduce something that will behave inconsistently.

I think one may need something like this in the case where sudo needs to be used to switch to the user environment: https://unix.stackexchange.com/a/269080/257111

rkdarst:
I think, like that , sourcing won't work well.  If it's setting
environment variables (the reason for source instead of executing),
they wont' be propagated back up to the script.

What I did was sourced the root ones during the root part, and the
user parts after the primary sudo to the user.  If this is what I
think it is, it's what we'll have to do anyway...


p.s. I'm sorry I stopped responding to this... it seems like it became
a much larger task than I could focus on (and also had too many
different goals to focus on).  But if it would help, I'm happy to
start paying more attention.  Though I'd sort of rather restarting and
porting over the relevant changes one step at a time, last time I
looked it seemed like this was changing a whole lot at once...

maresb:
Based on my understanding, it now looks to me like `-user.d` and `-root.d` hooks are the wrong approach. (I changed my mind.) They add another layer of complexity to the startup scripts. One can always use `if [ $(id -u) == 0 ]` in order to customize behavior based on if one is root or not. 

The root of the problem (which @rkdarst seems to have been trying to solve with the hooks system) seems to be the lack of a good script to run a command as a user. I think we should do this first, and then we can reevaluate if the `-user.d` and `-root.d` hooks are actually necessary.

(Apologies to @rkdarst if I'm misrepresenting #787.)

parente:
@consideRatio thank you for your time and work on rethinking the startup scripts. I agree they have grown unwieldy over the years as use of these images has grown and we (I) failed to define a clear scope for the project.

I truly appreciate that others have stepped up to review these changes and share their experience (@maresb, @rkdarst) as I have not had much energy to contribute deep thought to open source lately. That said, if there's something specific I can contribute here, please let me know and I'll try to do it.

parente:
I'm taking a pass through old PRs and cleaning up ones that haven't been touched in some time. I'm planning to leave this one along given the depth of conversation and relation to other on-going discussions (e.g., https://github.com/jupyterhub/mybinder.org-deploy/issues/1474)

consideRatio:
I think this PR has a lot of relevant discussion and changes, but is at this point by itself too large and hard to overview what it tries to accomplish. I suggest we close it. It is still referenced from various issues it meant to close for findability.

PR #1054, opened 2020-03-27
Prepend $CONDA_DIR/bin instead of appending

### LINKED ISSUE #1053: Wrong PATH in start.sh when running as root

maresb:
Fixes #1053

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -88,18 +88,18 @@ if [ $(id -u) == 0 ] ; then
 
     # Enable sudo if requested
     if [[ "$GRANT_SUDO" == "1" || "$GRANT_SUDO" == 'yes' ]]; then
-        echo "Granting $NB_USER sudo access and appending $CONDA_DIR/bin to sudo PATH"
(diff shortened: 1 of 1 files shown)


maresb:
Let's put this on hold until https://github.com/jupyter/docker-stacks/pull/1052 gets sorted out.

mathbunnyru:
I think, we can give this a go (please, fix the conflict, though), because the PR you mentioned was closed.

maresb:
Looks like I deleted my fork. Thankfully GitHub kept the ref. I restored my fork, but the link is dead. I will close this and reopen a duplicate.

PR #1058, opened 2020-04-04
Housekeeping

romainx:
Hello,

I've have made some housekeeping :wrench:

- Trying to switch to the latest **bionic** version of the [Travis build environment](https://docs.travis-ci.com/user/reference/bionic/).
- Bump root container version
- Bump  packages of
  - `scipy-notebook`
  - `r-notebook`
  - `datascience-notebook`
- Add `-y` flag to the `jupyter lab` commands in `scipy-notebook`
- Fix a bug in check outdated packages test when packages installed through `pip`

Best


test/helpers.py
```diff
@@ -89,7 +89,9 @@ def _packages_from_json(env_export):
         """Extract packages and versions from the lines returned by the list of specifications"""
         dependencies = json.loads(env_export).get("dependencies")
         packages_dict = dict()
-        for split in map(lambda x: x.split("=", 1), dependencies):
+        for split in map(lambda x: x.split("=", 1) if isinstance(x, str) else None, dependencies):
+            if split is None:
+                break
```
consideRatio, line 94: This code is very hard to read for me! I'd like to make it more readable.

Do you want to abort the looping (`break`) on first encountering a non string item in `dependencies`, or that you want to ignore it (`continue`)?

If you want to ignore some non-string dependencies, then i suggest we pre-process `dependencies` by filtering out those instead.

I'm a bit vary of seeing this change without understanding what situation this happens in. What value is it that we can end up seeing in the `dependencies` list?

romainx: You're right this fix is awful and buggy, so thank you for the review 🐞.
I've replaced it by some more readable code.

```python
# Filtering packages installed through pip in this case it's a dict {'pip': ['toree==0.3.0']}
# Since we only manage packages installed through conda here
dependencies = filter(lambda x:  isinstance(x, str), dependencies)
for split in map(lambda x: x.split("=", 1), dependencies):
[...]
```

So I expect to have in `dependencies` only `conda` packages and so I can skip packages installed through `pip` that are reported as a `dict` (`{'pip': ['toree==0.3.0']}`).

parente:
@romainx Thanks for your continued contributions here. I've not been focusing on open source lately but will take a pass through the pending PRs today. This one looks good to go.

romainx:
@parente I understand, no problem. It's the same for me, I try to help when I can. I have also encouraged some users to make PR. Thank you for your always nice messages.

Issue #779, opened 2018-12-24
vim not installed in minimal-notebook but is mentioned in documentation


wjgeorge:
**What docker image you are using?**
jupyter/minimal-notebook

**What steps do you take once the container is running to reproduce the issue?**
create a notebook terminal and run vim.

**What do you expect to happen?**
vim is started

**What actually happens?**
vim is not found

wjgeorge:
is this just an error in the documentation?

https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#jupyter-minimal-notebook

peterprescott:
surely the fix needed here is to add vim to the Dockerfile for the minimal notebook, rather than to remove it from the docs? developers might be split into rival camps by their choice of minimal text editor, but the jupyter project surely aims to bridge those divides rather than pick sides? i'll go ahead and make the relevant PR...

PR #1059, opened 2020-04-05
add vim to minimal-notebook (cf. issue 779)

peterprescott:
Someone raised [an issue](https://github.com/jupyter/docker-stacks/issues/779) a year or so ago that [vim](https://www.vim.org/) doesn't work in the Notebook console, even though it was mentioned in the Docs. And someone resolved the issue by removing the mention of vim from the docs.

But surely the fix needed here is to add vim to the Dockerfile for the minimal notebook, rather than to remove it from the docs? Developers might be split into rival camps by their choice of minimal text editor, but the Jupyter project surely aims to bridge those divides rather than pick sides?

Anyway, here's the necessary change: I've just added the line into the minimal-notebook's Dockerfile, and it [seems to work](https://hub.docker.com/r/peterprescott/notebook-with-vim).


peterprescott:
Well, I've added vim to the docs.

Four text editors does seem excessive, but when you compare their sizes vim isn't the biggest!

```
$ docker image ls
REPOSITORY                        TAG                 IMAGE ID            CREATED             SIZE
jupyter/no-editors          latest              9028600bad15        About an hour ago   2.41GB
jupyter/just-nano           latest              4f45a3b60448        2 minutes ago       2.41GB
jupyter/just-jed            latest              468858bf486b        13 minutes ago      2.42GB
jupyter/just-vim            latest              6b73ed8075d7        5 seconds ago       2.46GB
jupyter/just-emacs          latest              afd99342844c        15 minutes ago      2.52GB
jupyter/minimal-notebook          latest              c52c88f003ad        2 weeks ago         2.53GB
jupyter/notebook-with-vim   latest              50c04cfec06b        33 hours ago        2.58GB
```

rkdarst:
I don't know much about it, but I see there is a `vim-tiny` package.  Does that have enough features to be useful and small enough to be a "do it now" thing?

peterprescott:
Don't know much about it... https://askubuntu.com/questions/104138/what-features-does-vim-tiny-have

peterprescott:
Okay, I've added vim-tiny, which looks like it's even smaller than nano.

Twenty seconds of playing confirmed that it offers basic vim-like text-editing functionality.

It's run from the command-line with `vi` rather than `vim` (with the latter I get `bash: vim: command not found`) so in the docs I've called it *vi* with a bracket explaining that it's actually vim-tiny.

docs/using/selecting.md
```diff
@@ -43,7 +43,7 @@ repository. The following sections describe these images including their content
 
 * Everything in `jupyter/base-notebook`
 * [Pandoc](http://pandoc.org) and [TeX Live](https://www.tug.org/texlive/) for notebook document conversion
-* [git](https://git-scm.com/), [emacs](https://www.gnu.org/software/emacs/), [jed](https://www.jedsoft.org/jed/), [nano](https://www.nano-editor.org/), tzdata, and
+	* [git](https://git-scm.com/), [emacs](https://www.gnu.org/software/emacs/), [vi](actually [vim-tiny](https://vim.org)), [jed](https://www.jedsoft.org/jed/), [nano](https://www.nano-editor.org/), tzdata, and
```
romainx, line 46: I guess there is an unwanted tab before the bullet. 
`[vi](actually [vim-tiny](https://vim.org))` does not render correctly -> [vi](actually [vim-tiny](https://vim.org))
Replace by ```[vi](https://www.vim.org/) (actually `vim-tiny`)```-> [vi](https://www.vim.org/) (actually `vim-tiny`).

romainx:
Hello

@peterprescott thanks for your work and @rkdarst thanks for your idea, sounds good 🚀.
Adding `vim` sounds reasonable to me. I will create a dedicated issue in order to check if we have to discard one of these editors. 
@peterprescott I've seen a typo in the doc would you mind fix it before approving this PR?

Many thanks.
Best.

romainx:
I've made a check and I end with these sizes.


> The disk space is given as the integer value of the estimated installed size in bytes, divided by 1024 and rounded up.
-- [source](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-installed-size)

```bash
$ packages="vim-common vim-tiny jed jed-common emacs emacs25 emacs25-bin-common emacs25-common emacsen-common nano"
$ apt-cache --no-all-versions show $packages |      awk '$1 == "Package:" {p = $2}
         $1 == "Installed-Size:" {printf("%s: %s\n", p, $2, is)}' 

vim-common: 329
vim-tiny: 1271
---
~1 600 KB

jed: 349
jed-common: 1806
---
~2 155 KB

emacs: 8
emacs25: 19203
emacs25-bin-common: 462
emacs25-common: 65307
emacsen-common: 136
---
~85 116 KB

nano: 760
---
~760 KB
```

For information the same gives for vim "standard"

```bash
vim-common: 329
vim: 2789
vim-runtime: 28420
---
31 538 KB
```

rkdarst:
> emacs: 8
> emacs25: 19203
> emacs25-bin-common: 462
> emacs25-common: 65307
> emacsen-common: 136
> ---
> ~85 116 KB

There's also emacs-nox:

> This package contains a version of Emacs compiled without support
> for X, and provides only a text terminal interface.

No idea about the relative size, though...

Looking at packages on my debian system, it seems the emacsNN packages
are deprecated, and now emacs depends on `emacs-gtk | emacs-lucid |
emacs-nox`.  Would make sense to explicitely do emacs-nox?

romainx:
@rkdarst I think it could be interesting to install `emacs-nox`, however the saving seems to be small. However it's worth doing it since we do not need X support.

```bash
emacs-nox: 8
emacs25-bin-common: 462
emacs25-common: 65307
emacs25-nox: 17432
emacsen-common: 136
---
~ 83 345 KB
```

Thanks

docs/using/selecting.md
```diff
@@ -43,7 +43,7 @@ repository. The following sections describe these images including their content
 
 * Everything in `jupyter/base-notebook`
 * [Pandoc](http://pandoc.org) and [TeX Live](https://www.tug.org/texlive/) for notebook document conversion
-	* [git](https://git-scm.com/), [emacs](https://www.gnu.org/software/emacs/), [vi](actually [vim-tiny](https://vim.org)), [jed](https://www.jedsoft.org/jed/), [nano](https://www.nano-editor.org/), tzdata, and
+* [git](https://git-scm.com/), [emacs](https://www.gnu.org/software/emacs/), [vi](actually [vim-tiny](https://vim.org)), [jed](https://www.jedsoft.org/jed/), [nano](https://www.nano-editor.org/), tzdata, and
```
romainx, line 46: Thanks for the tab, however as I said in my previous comment the vim entry does not render correctly, see the screenshot below. Could you please fix it? I've made a proposal in my previous comment.
add_vim_to_minimal-notebook__cf__issue_779__by_peterprescott_·_Pull_Request__1059_·_jupyter_docker-stacks
Many thanks.

romainx:
@peterprescott, as suggested by @rkdarst would you mind replacing `emacs` by `emacs-nox`. The saving in the size should permit to not increase the size of the global image related to the addition of `vim-minimal`.
After the fix in the doc and this improvement we should be ready to go 👍

peterprescott:
not sure why that failed :/

romainx:
@peterprescott this seems to be a transient error, nothing to do with your change. Thanks I thin we are ready to go 🚀

peterprescott:
```
('Connection broken: OSError("(104, \'ECONNRESET\')")', OSError("(104, 'ECONNRESET')"))
Removing intermediate container 49e20f542d1b
The command '/bin/sh -c conda install --quiet --yes     'r-base=3.6.2'     'r-caret=6.0*'     'r-crayon=1.3*'     'r-devtools=2.2*'     'r-forecast=8.11*'     'r-hexbin=1.28*'     'r-htmltools=0.4*'     'r-htmlwidgets=1.5*'     'r-irkernel=1.1*'     'r-nycflights13=1.0*'     'r-plyr=1.8*'     'r-randomforest=4.6*'     'r-rcurl=1.98*'     'r-reshape2=1.4*'     'r-rmarkdown=2.1*'     'r-rodbc=1.3*'     'r-rsqlite=2.1*'     'r-shiny=1.4*'     'r-tidyverse=1.3*'     'unixodbc=2.3.*'     &&     conda clean --all -f -y &&     fix-permissions $CONDA_DIR' returned a non-zero code: 1
Makefile:47: recipe for target 'build/r-notebook' failed
make: *** [build/r-notebook] Error 1
```

Not sure I understand this... replacing `emacs` with `emacs-nox` in the minimal-notebook is causing the build of the R-notebook to fail?

peterprescott:
> @peterprescott this seems to be a transient error, nothing to do with your change. Thanks I thin we are ready to go

oh good :)

Issue #1064, opened 2020-04-17
start.sh adds conda path to secure_path for sudoers, which fails on centos redhat images


carstenpohllhind:
**What docker image you are using?**

`jupyter/base-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

This does not apply here.

**What steps do you take once the container is running to reproduce the issue?**

If you change the docker image to a redhat / centos based image, the start.sh will fail, and you will see an output similar to:
`Set username to: jovyan
usermod: no changes
Executing the command: jupyterhub-singleuser --ip=0.0.0.0
sudo: jupyterhub-singleuser: command not found`

**What do you expect to happen?**

The notebook should start, and the start should not throw an error.

**What actually happens?**

The notebook does not start.

**Additional info**

The problem lies within the sed command, that tries to get the information of the secure_path from the sudoers file. The regexp inside the sed command expects spaces before and after the equals sign, and double quotes before the path variables. This is the case in the Ubuntu / debian-esque /etc/sudoers file, but not in an centos based image, wich omits the double quotes and has trailing and leading spaces before the parameter name.

romainx:
Hello,

Thank you for the issue and the PR #1065 that seems OK.
I have posted a review in it asking a question about one test supposed to check this behavior.

Best.

PR #1065, opened 2020-04-17
Changed the regular expression in start.sh that reads the /etc/sudoer…

carstenpohllhind:
Fixes https://github.com/jupyter/docker-stacks/issues/1064

…s file and

adds the conda path as a secure path to sudo.

The regular expression expects that the equals sign has no leading and trailing
spaces after the parameter secure_path. Furthermore it expects that the value
is enclosed in double quotes.

This is the case for the sudoers file in debian based distributions, but the
in centos / redhat based distributions this is not the case. The default value
of the secure_path in centos / redhat based distributions has no double quotes
and leading and trailing spaces.

This change adds optional spaces before and after the space, and makes the
double quotes optional.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -93,7 +93,7 @@ if [ $(id -u) == 0 ] ; then
     fi
 
     # Add $CONDA_DIR/bin to sudo secure_path
-    sed -r "s#Defaults\s+secure_path=\"([^\"]+)\"#Defaults secure_path=\"\1:$CONDA_DIR/bin\"#" /etc/sudoers | grep secure_path > /etc/sudoers.d/path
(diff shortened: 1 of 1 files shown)


romainx:
Hello @parente ,

Maybe you have some history about the test `test_container_options.py::test_sudo_path_without_grant` I do not understand the rationale of this test since, or at least its naming, since it only tests IMO that `root` has `jupyter` in its `PATH`. 
Once this open question solved, this PR will be ready to merge.

Many thanks.

parente:
@romainx The test was added in https://github.com/jupyter/docker-stacks/pull/529 which has a link to the corresponding issue. If I recall, the test was added to make containers running as `root` could launch jupyter as well.

romainx:
@parente thank you for the answer. In this case we should rename the test in something like `test_root_path` since it confusing. Nevertheless the fix is OK and the PR can be safely merged.

@carstenpohllhind thanks again for your contribution. :smile:

PR #1078, opened 2020-05-02
Fix #1077

### LINKED ISSUE #1077: Changing `NB_USER` requires explicit `CHOWN_HOME` option to be set
Hello,

Following some issues (#885, #1003) related to home directories, I have performed some tests and figured out that the following recipe in the documentation does not work as expected.

> `-e NB_USER=jovyan` - Instructs the startup script to change the default container username from `jovyan` to the provided value. Causes the script to rename the `jovyan` user home folder. For this option to take effect, you must run the container with `--user root` and set the working directory `-w /home/$NB_USER`. This feature is useful when mounting host volumes with specific home folder.

```bash
$ docker run --rm -p 8888:8888 --user root -e NB_USER=romain --workdir /home/romain jupyter/base-notebook
# [...]
# PermissionError: [Errno 13] Permission denied: '/home/romain/.local'
```

In fact the problem is that the home directory is created but it's owned by `root`.

```bash
# Set username to: romain
# 4.0K drwxr-xr-x 1 root   root  4.0K May  2 07:34 .
# 4.0K drwxr-xr-x 1 root   root  4.0K May  2 07:34 ..
# 4.0K drwsrwsr-x 1 romain users 4.0K Apr 30 10:39 jovyan
# 4.0K drwxr-xr-x 2 root   root  4.0K May  2 07:34 romain
```

So in order to make it work we need to use the option `CHOWN_HOME=yes`, full command is below.

```bash
$ docker run --rm -p 8888:8888 --user root -e NB_USER=romain -e CHOWN_HOME=yes --workdir /home/romain jupyter/base-notebook
````

In order to avoid to complexifie a little bit more the `start.sh` script, I propose to modify the documentation to mention the `CHOWN_HOME` option and to add a test to check this behavior.

Best

romainx:
Documentation + test fix for #1077 (code not changed)

* Documentation `common.md` updated
* `test_nb_user_change` test written


romainx:
Another transient error.

parente:
> Another transient error.

It looks like there's a new package incompatibility in `r-notebook` and likely `data-science-notebook`. I'm OK merging this test change. Those images simply won't build until the conflict is resolved.

romainx:
Hello @parente,

Sorry I had not seen the incompatibility issue I was still looking at [this build log](https://travis-ci.org/github/jupyter/docker-stacks/builds/682325569#L3362) that is a transient error.

I will have look and let you know.
Best.

romainx:
Created a new issue #1090 to reference this case.

romainx:
@parente I'm a bit lost, I do not understand why it works now. I've tested during the day and reproduced the bug, and now it's OK on my side and also on the CI.
So you can merge it if you want.

parente:
Thanks for looking into it more @romainx. I suspect there was a package rebuild or release in conda-forge causing the incompatibility which fixed thereafter.

PR #1080, opened 2020-05-08
Upgrade Tensorflow to 2.2.0

maresb:

PR #1082, opened 2020-05-12
Use pip --no-cache-dir to install tensorflow

mathbunnyru:
Before image size was 5.21GB, after fix: 4.78GB - ~430MB saved.

### DIFF
--- a/tensorflow-notebook/Dockerfile
+++ b/tensorflow-notebook/Dockerfile
@@ -7,6 +7,6 @@ LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 
 # Install Tensorflow
 RUN pip install --quiet --no-cache-dir \
-    'tensorflow==2.1.0' && \
+    'tensorflow==2.2.0' && \
     fix-permissions $CONDA_DIR && \
     fix-permissions /home/$NB_USER
--- a/tensorflow-notebook/Dockerfile
+++ b/tensorflow-notebook/Dockerfile
@@ -6,7 +6,7 @@ FROM $BASE_CONTAINER
 LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 
 # Install Tensorflow
-RUN pip install --quiet \
+RUN pip install --quiet --no-cache-dir \
     'tensorflow==2.1.0' && \
     fix-permissions $CONDA_DIR && \
     fix-permissions /home/$NB_USER


romainx:
Hello @mathbunnyru ,

Thanks very good suggestion I think we should definitively do it. This is an oversight since the `--no-cache-dir` option is already used in another case.

https://github.com/jupyter/docker-stacks/blob/3b1f4f5e6cc1fd81a14bd57e805fbb25daa3063c/all-spark-notebook/Dockerfile#L37-L38

Note: we have another PR #1080 in progress that aims to upgrade the version of `tensorflow`, we have to take both into account. I will make a feedback on this PR in order for the author to include directly the `--no-cache-dir` option.

Thanks for your help.

Best.

maresb:
This is so awesome! My colleagues are always complaining about how large the image is. Thanks so much!!!

I just added a pull request to @mathbunnyru's pull request, so hopefully we can get this done in one (nested) merge, and he gets the proper credit.

mathbunnyru:
> Hello @mathbunnyru ,
> 
> Thanks very good suggestion I think we should definitively do it. This is an oversight since the `--no-cache-dir` option is already used in another case.
> 
> https://github.com/jupyter/docker-stacks/blob/3b1f4f5e6cc1fd81a14bd57e805fbb25daa3063c/all-spark-notebook/Dockerfile#L37-L38
> 
> Note: we have another PR #1080 in progress that aims to upgrade the version of `tensorflow`, we have to take both into account. I will make a feedback on this PR in order for the author to include directly the `--no-cache-dir` option.
> 
> Thanks for your help.
> 
> Best.

Yes, I saw thay you were already doing it other cases :)

As @maresb suggested in his PR, I increased the version here as well (so now bot PRs are the same), but choose whichever PR you want, I don't mind :)

maresb:
Let's go with this one. I closed my PR.

romainx:
@maresb @mathbunnyru thank you guys.
And sorry for the confusion from my side.

Let's go for this PR 🚀

maresb:
@romainx, sorry if this is the wrong place to ask, but I now see several "Ready to Merge" PRs. Earlier most PRs were spontaneously merged. Has something changed in this regard?

romainx:
@maresb No the thing is just that I cannot do it since I've not the necessary permission to merge PR. So I flag PR as "Ready to merge" to make a kind of triage and to tell other collaborators that I've checked  and that they could merge them in priority.

parente:
Thanks for the PR and your collective patience. LGTM!

PR #1088, opened 2020-05-19
Updating Power PC LE Dockerfile

oskrdt:
Updating code from current `Dockerfile` x86 version
Updating miniconda version to `4.8.2`

FYI, current `Dockerfile.ppc64le` doesn't work since current conda version `4.2.12` doesn't accept the `-f` flag


parente:
Thanks for the PR, @oskrdt. I have no way of testing these changes. Can you confirm the build works for you using them?

oskrdt:
> Thanks for the PR, @oskrdt. I have no way of testing these changes. Can you confirm the build works for you using them?

@parente Confirmed! I tested it on a Red Hat 7.7 PPC64LE System and it was able to create and run the container successfully.

parente:
Thanks again 🍰

PR #1094, opened 2020-05-27
doc/index.rst : more readable Issue Tracker's link on sidebar

NickVeld:
# Before:

image

The links is bigger than "Getting Help" and separated visually thus the header draw less attention than the links. 

Links title is written with camel case thus each line can be interpreted as separated. Thus if you do not see "Getting Help" firstly (for reason described before) you see not three but four items: Forum, Stack, Tracker and Website which can be grouped as links to external websites connected with the stack.

Since change Camel Case to something other is question for discussion I select the way of putting the title in one line.

# After:

image


parente:
Thanks, @NickVeld. The fix looks reasonable to me.

PR #1097, opened 2020-05-28
Improve documentation

mathbunnyru:
Fixed some issues:
1. Added ```dockerfile (bash, python) - it makes it look more beautiful on github (especially dockerfile)
2. When using a list, it's better to indent all the ```s


parente:
Love it. Thanks, @mathbunnyru. I resolved a conflict between #1098 and this PR, but think it's ready to merge now.

PR #911, opened 2019-07-27
Remove mesos lib from pyspark-notebook

parente:
* [x] Remove lib and mentions in the doc
* ~[ ] Add recipe for installing mesos~ I have no way to test a mesos setup. Happy to take a contribution to the recipes page in the documentation if someone would like to pick up that work.

PR #1098, opened 2020-05-28
Follow-on PR #911: Spark documentation rework

romainx:
Some changes to the Spark documentation following the PR #911
for local and standalone use cases with the following drivers

* Simplify some of them (removing options, etc.)
* Use the same code as much as possible in each example to be consistent (only kept R different from the others)
* Add Sparklyr as an option for R
* Add some notes about prerequisites (same version of Python, R installed on workers)

Best

### DIFF
--- a/all-spark-notebook/test/data/local_pyspark.ipynb
+++ b/all-spark-notebook/test/data/local_pyspark.ipynb
@@ -1,0 +1,60 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [
+    {
+     "output_type": "error",
+     "ename": "Error",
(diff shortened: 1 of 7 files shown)


romainx:
Thank you @parente 

Yes all the examples are working. I've reworked them a bit to use the same use case for all: "sum of the first 100 whole numbers".

* Test added for all kernels
* Same examples as provided in the documentation (`specifics.md`)

Note: I've not automatically tested `local_sparklyr.ipynb` since it creates by default the `metastore_db` dir and the `derby.log` file in the working directory. Since I mount it in `RO` in the container it's not working (works in `RW` but it's not the solution). I'm struggling to set it elsewhere...

I let you decide but I think we could merge it and fix this after since it's not mandatory and since all kernels are tested (`sparklyr` is just another way to use the R kernel).

Best

parente:
It's great that you got most of the tests working already. Let's merge it! 🎉

PR #1100, opened 2020-05-29
Docker linting -> Hadolint

romainx:
Hello,

Here is a proposal and a draft of using the [Hadolint][LK1] tool to analyse each `Dockerfile` to comply with [Docker best practices][LK2].

I've fixed most of the linting issues however I may have break something. I will take some times to check.

Best

[LK1]: https://github.com/hadolint/hadolint
[LK2]: https://docs.docker.com/develop/develop-images/dockerfile_best-practices


romainx:
I think I'm done, ready for the review :bowtie:

mathbunnyru:
@romainx this seems to be a nice tool to check dockerfiles, thanks!
I will definitely run it on my own dockerfiles :)

May I suggest you create a config file with 2 ignore rules:
DL3006 (version in FROM) and DL3008 (version in apt-get)

I think it makes sense to ignore these 2 rules, because they are ignored quite a lot in these dockerfiles.

romainx:
@mathbunnyru very good suggestion, thank you! 
It has been taken into account (and documented in the `lint.md`.

Many thanks for your feedback.

parente:
Add `hadolint` makes sense to me. The PR looks good, including the new doc sections, and I agree with the exceptions @mathbunnyru suggested. Let's :shipit:.

PR #593, opened 2018-03-23
base image on ubuntu:xenial-20180228

clkao:

PR #1112, opened 2020-06-16
Upgrading miniconda to py38 and Spark to 3.0

Bidek56:


Bidek56:
It looks like it's failing on [Toree](https://github.com/apache/incubator-toree) test b/c of some Java error but [Toree](https://github.com/apache/incubator-toree) has not been released in years. What should we do? Drop Scala support? Spylon, Spark and R work fine. Thx

Bidek56:
@parente How can we move this forward without Toree? Thanks

parente:
I suggest we remove Toree in this PR and let it sit for a bit to see if anyone with a vested interest would like to try packaging a Spark 3.0 compatible version of Toree in all-spark-notebook. If we do, we should tag the last Spark 2.x image containing Toree before merging, much like we're about to do in #593

Bidek56:
Few things:
1. [Toree Makefile](https://github.com/apache/incubator-toree/blob/master/Makefile) references **jupyter/all-spark-notebook:latest** images which will not work without Toree. Catch 22?
2. Databricks is able to support Scala with Spark 3.0 so there must be another way without Toree, I would love to know what it is.
3. Closing this PR and will do 2 PRs, one to upgrade to 2.4.6 and one for 3.0

PR #1114, opened 2020-06-18
Upgrading miniconda to py38

Bidek56:


Bidek56:
This one is failing because of [spylon-kernel](https://github.com/Valassis-Digital-Media/spylon-kernel) package which has not been upgraded in few years.
@parente I believe that you were the last one that released this package.
Let me see if I can upgraded it.

PR #1115, opened 2020-06-19
Miniconda py38 and Spark 3.0

Bidek56:
Upgrading Spark to 3.0
Upgrading Python to 3.8
Removing Toree due to lack of support
Open JDK needs to stay at 11 because of SparkR limitation
Spylon kernel may need to be removed soon since it has not been maintained in years


parente:
We should consider swapping out Toree and spylon for https://almond.sh/

Bidek56:
We could add it but Almond has a [Dockerfile](https://github.com/almond-sh/almond/blob/master/Dockerfile) already which extends from base-notebook, not sure what value does a copy of others code adds here. It will make the image larger than it already is.

parente:
> We could add it but Almond has a Dockerfile already which extends from base-notebook ...

Good call. That it does.

> not sure what value does a copy of others code adds here

Without Toree or Spylon in the all-spark-notebook image, there's no support for using Scala in Jupyter in the image any longer. Maybe it's fine to support Python and R only but it is a significant departure from the original purpose of the image.

Bidek56:
These's a good chance Toree will be upgraded to work with JDK 11+ now that Spark 3.0 is out.
@parente  do you still have access to the Spylong so it can be upgraded? Thx

Makefile
```diff
@@ -88,7 +88,7 @@ lint-build-test-all: $(foreach I,$(ALL_IMAGES),lint/$(I) arch_patch/$(I) build/$
 
 lint-install: ## install hadolint
 	@echo "Installing hadolint at $(HADOLINT) ..."
-	@curl -sL -o $(HADOLINT) "https://github.com/hadolint/hadolint/releases/download/v1.17.6/hadolint-$(shell uname -s)-$(shell uname -m)"
+	@curl -sL -o $(HADOLINT) "https://github.com/hadolint/hadolint/releases/download/v1.18.0/hadolint-$(shell uname -s)-$(shell uname -m)"
```
romainx, line 91: Thank you for this update. I didn't realize that the version was hardcoded -- my bad. It could be an opportunity to put the hadolint version in a variable and to print it in the `echo` to know the version that was used.

base-notebook/Dockerfile
```diff
@@ -76,14 +76,14 @@ RUN mkdir /home/$NB_USER/work && \
 
 # Install conda as jovyan and check the md5 sum provided on the download site
 ENV MINICONDA_VERSION=4.8.2 \
```
romainx, line 78: There is a new version `4.8.3` (`Miniconda3-py38_4.8.3-Linux-x86_64.sh`) in the [miniconda repo](https://repo.anaconda.com/miniconda/). We should also use this new version.

Bidek56: I was trying to limit the change set to avoid breaking things, I figured once this PR gets merged then I can do it in the next PR, unless disagree.

parente:
> @parente do you still have access to the Spylong so it can be upgraded?

I would remove spylon from the image as unmaintained.

Bidek56:
> > @parente do you still have access to the Spylong so it can be upgraded?
> 
> I would remove spylon from the image as unmaintained.

Do you want to do it in this PR or next one? It still works with Spark 3.0.

base-notebook/Dockerfile
romainx: Ok it makes sense. Thanks.

parente:
If it's working with 3.0, go ahead and leave it for now.

maresb:
Looks like Travis CI failed due to a web error? It builds locally for me.

Note the warning:
```
Solving environment: ...working... WARNING conda.core.solve:_add_specs(601): pinned spec conda==4.8.2 conflicts with explicit specs.  Overriding pinned spec.
```

I don't know where the aforementioned conflicting explicit spec is, but perhaps it would be better after all to just bump to 4.8.3:
```
Miniconda3-py38_4.8.3-Linux-x86_64.sh | 88.7M | 2020-06-16 14:57:56 | d63adf39f2c220950a063e0529d4ff74
```

Bidek56:
@parente Can we merge this PR yet? Thx

maresb:
To be more specific, base-notebook builds for me. I didn't try the others...

Bidek56:
The PR builds fine, the failure is a Travis issue not the code.

parente:
@Bidek56 I'm planning on tagging the last images containing Spark 2.x and then merging this PR when I get a few minutes in the next day or two.

parente:
I've tagged the latest pyspark and all-spark images with `spark-2`. Merging now.

lresende:
> These's a good chance Toree will be upgraded to work with JDK 11+ now that Spark 3.0 is out.
> @parente do you still have access to the Spylong so it can be upgraded? Thx

Yes, release coming soon.

PR #1120, opened 2020-07-05
Upgrading Ubuntu, miniconda and jupyterlab

Bidek56:
Upgrading:
Ubuntu to ubuntu:focal-20200606
Miniconda to 4.8.3
Jupyterlab to 2.1.5


mathbunnyru:
focal-20200703 was released several days ago, we should probably up the version again?

PR #1118, opened 2020-06-25
Fix python codestyle

mathbunnyru:

PR #1123, opened 2020-07-11
pre-commit hook + minor fixes

romainx:
# pre-commit

[pre-commit](https://pre-commit.com/) added to check

- `bash` files with bashate see #1109
- `yaml` files with yamlint see #1108
- `docker` files with hadolint see #1100

# hadolint

Version of the linter added as a variable in the `Makefile`

Could be adapted to integrate python file check see #1118


romainx:
Build error has nothing to do with the change. It has occurred in `r-notebook` during the import of `tidymodels`

```
AssertionError: {'tidymodels': AssertionError('Package [tidymodels] import failed\nassert 1 == 0\n  -1\n  +0')}
```

I will have a look.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
+---
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.1.0
+    hooks:
+      #- id: flake8
```
mathbunnyru, line 6: Seems to be some debug info

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - id: check-yaml
+        files: .*\.(yaml|yml)$
+  - repo: https://github.com/adrienverge/yamllint.git
+    rev: v1.23.0
+    hooks:
+      - id: yamllint
+        files: \.(yaml|yml)$
+        exclude: ^.travis
```
mathbunnyru, line 14: I think travis works fine with formatted yaml files, so I don't think we need an exclude

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - repo: https://github.com/openstack-dev/bashate.git
+    rev: 2.0.0
+    hooks:
+      - id: bashate
+  - repo: local
+    hooks:
+      - id: hadolint
```
mathbunnyru, line 21: I see some inconsistency between hook id "hadolint" and the command "lint-all".
Maybe we should rename something?

.pre-commit-config.yaml
romainx: Removed it was to thing about adding flake8 see  #1118

.pre-commit-config.yaml
romainx: You're right, fixed

.pre-commit-config.yaml
romainx: The `.travis.yml` has some lines too long. I've adapted the configuration of `yamlint` to relax this constraint.

romainx:
@mathbunnyru I've made some ajustements following your review. 
I would like to have your opinion on this new version.
Thanks in advance.

mathbunnyru:
Thank you @romainx !
I took a look after your changes.

LGTM, but I can think about 2 things which may be good:
1. Run pre-commit hook server-side as well. So hadolint is checked server-side, but other linters are not.
2. Add some info for developers to make it easier for newcomers to install these tools and run them.

It can be done in a separate PR, up to you to do decide how you want to implement this.

romainx:
@mathbunnyru thank you, once again good suggestions 🥇.

1. Good idea, however I prefer managing that in a different PR since it will involve CI build debugging. I would prefer only bootstrap pre-commit in this PR and improve its usage over time -- no it's not laziness 🤣.
2. I take it in this PR! I was just thinking about it, there is currently no dedicated page, but I intend to adjust page `lint.md` to talk about this.

For the point 2 I let you know when it's done. I will try to do it quickly.

Once again thanks for you constructive feedback 😀

romainx:
@mathbunnyru I'm good with the doc. Please tell me if it's Ok for you.

mathbunnyru:
Good job, @romainx!
Everything is great 👍

parente:
Thanks for your patience on this PR. I've resolved the merge conflict with the master branch. I'll merge when CI succeeds.

parente:
Thanks for the PR, @romainx and the review @mathbunnyru.

PR #1125, opened 2020-07-12
Update ubuntu image

mathbunnyru:


romainx:
Hello,

Thank you, everything is fine. The problem occurring in `r-notebook` with `r-tidymodels` `0.1.0` is fixed since the availability of the version `0.1.1`. So the CI will work on next build.

Best.

Issue #1083, opened 2020-05-15
Do we need python-dev?


rahulpshah:
https://github.com/jupyter/docker-stacks/blob/master/minimal-notebook/Dockerfile#L23
Why do we need python-dev from the base image if we install python from conda?

mathbunnyru:
And, python-dev is python2, not python3, which doesn't make sense at all. Could you please make a PR without this package, maybe it would be easy to not to install it?

romainx:
Hello @rahulpshah,

You're right :scream:

```bash
jovyan@e49d9de062c1:/usr/bin$ /usr/bin/python

# Python 2.7.17 (default, Apr 15 2020, 17:20:14) 
# [GCC 7.5.0] on linux2
# Type "help", "copyright", "credits" or "license" for more information.
# >>> 
```

`python-dev` installs a python version that is not needed.

```bash
jovyan@1238c0821be8:~$ apt-cache rdepends python-dev

# python-dev
# Reverse Depends:
#  python
#  python-minimal
```

We will reduce the image size and the same for all the downstream images

```bash
REPOSITORY                 TAG                 IMAGE ID            CREATED             SIZE    
jupyter/minimal-notebook   latest              bbcc28f4fde5        29 seconds ago      2.468GB
<none>                     <none>              ba8c39810ce2        11 days ago         2.514GB
```

So, as suggested by @mathbunnyru :thumbsup: please could you make a PR?

Many thanks
Best

rahulpshah:
Thanks for the response @mathbunnyru, @romainx. I will make the PR today.

bollwyvl:
Recently got bitten by thinking i could actually bid a fond farewell to
python2: so long as nodejs is supported (e.g. building jupyterlab, so,
yes), python2 will likely have to be on the path, as it is part of the
build chain for packages with native components.

romainx:
Hello @bollwyvl,

Thanks for your feedback, it's interesting. I think all the stacks can be build without this package. However it 's a problem if we break something elsewhere by removing it.
So could you please develop and / or give an example of what you mean by saying "as it is part of the build chain for packages with native components"?

Many thanks for your help.
Best.

bollwyvl:
Specifically nodejs packages. If there are any "jupyterlab build" commands
in the stack, and any of the core lab npm components, any part of its build
chain, or any extensions or their npm dependencies,  require compiling with
the weird node-gyp toolchain, the build will fail if python2 is not
[...]

Issue #243, opened 2016-07-21
nbconvert failed: Inkscape svg to pdf conversion failed


parente:
I tried converting a trivial notebook (1 markdown cell, 1 code cell, ggplot output) on the latest jupyter/datascience-notebook image (tag: `97a5071c5775`) to a PDF and it failed with a different error than the one in this issue: `nbconvert failed: Inkscape svg to pdf conversion failed`.

Reported in #99 but appears to be a new issue, not the same as the original report.

parente:
Manually doing `apt-get install inkscape` fixes the problem. We can throw it into minimal-notebook where pandoc and the other nbconvert prereqs are.  Or we can put this only in images that use SVGs by default like r- and datascience-notebook.

I think the former is easier to reason about even though it bloats the minimal-notebook image by another ~30 MB.

jakirkham:
Adding to minimal seems reasonable. I mean we are already adding TeXLive to it.

PR #1126, opened 2020-07-12
Do not install python-dev

mathbunnyru:
It's an attempt to fix https://github.com/jupyter/docker-stacks/issues/1083
If everything builds fine, then we probably don't need to install python-dev manually.


mathbunnyru:
jovyan@5daef95157f2:~$ apt-cache depends inkscape | grep python
  Depends: <python2:any>
    python2

It seems that inkscape depends on python2, but still, if we don't need something to be installed as a direct package, then let's not install it.

There is an inkscape bug, that it still needs python2, but it won't be fixed in 20.04. But as soon as 22.04 gets released, the size image will automatically shrink (https://bugs.launchpad.net/ubuntu/+source/inkscape/+bug/1735363)
Or we might delete inkscape.
Or we might use https://launchpad.net/~inkscape.dev/+archive/ubuntu/stable?field.series_filter=focal

mathbunnyru:
@romainx @parente could you please take a look?

romainx:
@mathbunnyru sorry for the late answer and thank you for taking care of this topic.
The reason why inkscape is installed was to fix the issue #243.
In short, inkscape is a dependency of [nbconvert](https://github.com/jupyter/nbconvert/) to treat SVG in pdf output.
If this seems to be pretty much clear, I don't know exactly when it is required.

The  test `test_nbconvert.py` in `minimal-notebook` passes without the inkscape dependency installed. However we should extend the test to add a new example file with a plot in order to check if the dependency is needed / work in this condition.
If it's the case, we will probably go further on this PR.
Does that sound reasonable to you?

mathbunnyru:
Yes, thanks, @romainx.

romainx (approved):
Hello,

Just pushed the PR #1182 to test nbconvert with inkscape.
Tests are OK without the explicit dependency but as expected python2 is still installed due to inkscape dependency.

```bash
$ aptitude why python2
# i   inkscape Depends  python2:any
# i A python2  Provides python2:any
```

So as you said **I'm Ok to remove this explicit dependency** even if it does not bring so much improvements as analyzed in this comment https://github.com/jupyter/docker-stacks/issues/1083#issuecomment-629759171

I've asked @parente to review this also in case we missed something.

Thanks for your effort to improve the stacks 👍

mathbunnyru:
Merged the latest master to make sure it still passes all the tests.

parente (approved):

romainx:
@mathbunnyru, @parente thanks for your help. Let's go for the merge.

PR #1128, opened 2020-07-16
Swap home-dir relocation and chown steps

twalcari:
Swap the home-dir relocation and chown steps to allow for proper ownership change of files present in `/home/jovyan` when `NB_USER` and `NB_UID` and/or `NB_GID` are passed as environment variable.

**Open question:** For the ownership change to be done `CHOWN_HOME` must also be passed as environment variable. Shouldn't this always be done when the combination of `NB_USER` and (`NB_UID` and/or `NB_GID`) is passed ?

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -47,19 +47,6 @@ if [ $(id -u) == 0 ] ; then
         usermod -d /home/$NB_USER -l $NB_USER jovyan
     fi
 
-    # Handle case where provisioned storage does not have the correct permissions by default
-    # Ex: default NFS/EFS (no auto-uid/gid)
-    if [[ "$CHOWN_HOME" == "1" || "$CHOWN_HOME" == 'yes' ]]; then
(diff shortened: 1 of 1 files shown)


parente:
@twalcari I've closed and reopened this PR in attempt to get CI to re-run the tests. Your change makes logical sense. I'm trying to remember if the chown is happening first to solve a prior issue, and think through what ramifications the change in semantics will have.

> Open question: For the ownership change to be done CHOWN_HOME must also be passed as environment variable. Shouldn't this always be done when the combination of NB_USER and (NB_UID and/or NB_GID) is passed ?

If I recall, we introduced `CHOWN_HOME` as a way to support disabling the `chown` step when a mounted host volume already has user/group matching what is passed for NB_UID and NB_GID.

romainx:
Hello,

I've started working on a duplicate of this PR #1186 before realizing it was a duplicate. Maybe we should close this one.
Sorry 😕

parente:
I merged #1186 which fixes the same issue with passing tests. Thank you, @twalcari, for raising the issue and submitting a first attempt at a fix.

PR #1130, opened 2020-07-21
Removed optional texlive-fonts-extra

### LINKED ISSUE #1129: Do we need texlive-fonts-extra in the minimal-notebook?
Do we really need the `texlive-fonts-extra` package in the [minimal-notebook](https://github.com/jupyter/docker-stacks/blob/ea01ec4d9f5715ab7dd85d4742ad6416c343c8bd/minimal-notebook/Dockerfile#L29)? It is commented to be optional and I guess `nbconvert`does not need it for basic functionality.
The problem is that it really bloats the image:
```bash
REPOSITORY                    TAG                 IMAGE ID            CREATED             SIZE
minimal-notebook-nofonts      latest              9641ee9dd31c        3 minutes ago       1.7GB
minimal-notebook              latest              e23d7794a410        14 minutes ago      3.01GB
```
That's 1.3 GB difference...

Any thoughts on that?

LustigePerson:
`texlive-fonts-extra` seems not be be a requirement of `nbconvert`.
Reduces the image size by ~ 1.3 GB which will affect all images that build on the `minimal-notebook` image.

```bash
REPOSITORY                    TAG                 IMAGE ID            CREATED             SIZE
minimal-notebook-nofonts      latest              9641ee9dd31c        3 minutes ago       1.7GB
minimal-notebook              latest              e23d7794a410        14 minutes ago      3.01GB
```
I opened a [issues ](https://github.com/jupyter/docker-stacks/issues/1129)questioning if there is any reason to keep this package. See also the comment [here](https://github.com/jupyter/docker-stacks/pull/1013#discussion_r379897106) in [PR 1013](https://github.com/jupyter/docker-stacks/issues/1129).
So this PR closes #1129.

### DIFF
--- a/minimal-notebook/Dockerfile
+++ b/minimal-notebook/Dockerfile
@@ -25,8 +25,6 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     texlive-xetex \
     texlive-fonts-recommended \
     texlive-plain-generic \
-    # Optional dependency
-    texlive-fonts-extra \
     # ----
     tzdata \
     unzip \


parente:
The fonts extra package has been present in the minimal image since its original definition 5 years ago. I very likely seeded the initial list of apt packages based on brief trial and error and what we had in the all-in-one tmpnb image at the time. 

I think the space savings is worth an attempt at removing it.  If there is outcry, we can discuss and evaluate options (e.g., recipe, restoring it).

PR #1144, opened 2020-08-09
Fix GITHUB_REPOSITORY value

parente:
`GITHUB_REPOSITORY` is a built-in env var for GitHub Actions. We cannot override it via env vars when trying to push to the wiki repo. Pass the desired value directly on the command line instead.


parente:
🤞 this PR squashes the last bug in switching to Actions.

PR #1145, opened 2020-08-15
Resolves #1131: Allow alternative Spark version

### LINKED ISSUE #1131: Create both Spark2 and Spark3 docker images
All of a sudden the image upgraded from Spark2 to Spark3.  I'm really glad there's a manually created `spark-2` tag or I would be sunk!  But I'm now stranded on an old, unmaintained version.  Oops.

Can we get the Docker Build to build `latest` using Spark3 as it does now, and also get the build to create `spark-2` tagged images as well?

It appears https://github.com/jupyter/docker-stacks/commit/568708d2794fb534a2a4c141796de308685be1f7 is the commit that made the change.  Is it as easy as two jobs, one that passes in `APACHE_SPARK_VERSION=2.4.5` and `HADOOP_VERSION=2.7`?

romainx:
Allow to build `pyspark-notebook` image with an alternative Spark version.

- Define arguments for Spark installation
- Add a note in "Image Specifics" explaining how to build an image with an alternative Spark version
- Remove Toree related documentation from "Image Specifics" since its support has been dropped in #1115

It resolves to the following issues:

- #626
- #1145 

Best

### DIFF
--- a/pyspark-notebook/Dockerfile
+++ b/pyspark-notebook/Dockerfile
@@ -34,7 +34,7 @@ WORKDIR /tmp
 # hadolint ignore=SC2046
 RUN wget -q $(wget -qO- https://www.apache.org/dyn/closer.lua/spark/spark-${APACHE_SPARK_VERSION}/spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz\?as_json | \
     python -c "import sys, json; content=json.load(sys.stdin); print(content['preferred']+content['path_info'])") && \
-    #echo "${spark_checksum} *spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz" | sha512sum -c - && \
+    echo "${spark_checksum} *spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz" | sha512sum -c - && \
     tar xzf "spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz" -C /usr/local --owner root --group root --no-same-owner && \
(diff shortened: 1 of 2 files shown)


pyspark-notebook/Dockerfile
```diff
@@ -11,20 +11,30 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
...
-# Using the preferred mirror to download Spark
+# Spark installation
 WORKDIR /tmp
-
+# Using the preferred mirror to download Spark
 # hadolint ignore=SC2046
 RUN wget -q $(wget -qO- https://www.apache.org/dyn/closer.lua/spark/spark-${APACHE_SPARK_VERSION}/spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz\?as_json | \
     python -c "import sys, json; content=json.load(sys.stdin); print(content['preferred']+content['path_info'])") && \
-    echo "BFE45406C67CC4AE00411AD18CC438F51E7D4B6F14EB61E7BF6B5450897C2E8D3AB020152657C0239F253735C263512FFABF538AC5B9FFFA38B8295736A9C387 *spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz" | sha512sum -c - && \
+    #echo "${spark_checksum} *spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz" | sha512sum -c - && \
```
parente, line 37: Should this be uncommented?

romainx:
@parente thank you for the feedback, let's check if it's OK with the checksum activated.
I think we could improve the way the Py4J dependencies is added to `PYTHONPATH` however, for now, I did not find a smarter way to do that.

PR #1146, opened 2020-08-16
Resolve Docker lint (hadolint) errors

romainx:
- Exclude `ppc64` `Dockerfiles` from the linting in pre-commit since we cannot check their build easily and so fix the lint errors without breaking something (already excluded in the `Makefile`)
- Fix examples `Dockerfiles` to invert root and jovyan user instructions ("Last user should not be root" `DL3002`)

Best


romainx:
First merge done :bowtie:

Issue #956, opened 2019-11-04
Container startup with custom user id fails on multi-tenant system with namespace isolation


PALuczak:
**What docker image you are using?**

`jupyter/datascience-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

`docker run --userns=host --gpus all --rm -it -p 42424:8888 -e JUPYTER_ENABLE_LAB=yes --user $(id -u) --group-add users -v /home/username/:/files/ --name datascienceLab jupyter/datascience-notebook`

 and 

`docker run --userns=host --gpus all --rm -it -p 42424:8888 -e NB_UID=$(id -u) -e NB_GID=$(id -g) -e JUPYTER_ENABLE_LAB=yes --user root -v /home/username/:/mount/ --name datascienceLab jupyter/datascience-notebook`

**What steps do you take once the container is running to reproduce the issue?**

Just wait for the container execution to fail, watch output in the terminal.

Docker version: `Docker version 19.03.4, build 9013bf583a` with [namespace isolation](https://docs.docker.com/engine/security/userns-remap/) enabled.

**What do you expect to happen?**

A jupyter lab instance with uid matching my host user id so that I could edit files in the mounted home directory.

**What actually happens?**

First command:

```
Container must be run with group "root" to update passwd file
Container must be run with group "users" to update files
Executing the command: jupyter lab
Traceback (most recent call last):
  File "/opt/conda/lib/python3.7/site-packages/traitlets/traitlets.py", line 528, in get
    value = obj._trait_values[self.name]
KeyError: 'runtime_dir'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/bin/jupyter-lab", line 10, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.7/site-packages/jupyter_core/application.py", line 266, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/opt/conda/lib/python3.7/site-packages/traitlets/config/application.py", line 657, in launch_instance
    app.initialize(argv)
  File "</opt/conda/lib/python3.7/site-packages/decorator.py:decorator-gen-7>", line 2, in initialize
  File "/opt/conda/lib/python3.7/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/opt/conda/lib/python3.7/site-packages/notebook/notebookapp.py", line 1676, in initialize
[...]

Issue #885, opened 2019-06-17
Permission denied when mounting local folders


ejgutierrez74:
**What docker image you are using?**

Example: jupyter/all-spark-notebook

**What complete docker command do you run to launch the container (omitting sensitive values)?**
docker run -it --rm -p 8888:8888 -v /home/eduardo/Descargas/DockerSpark/:/home/jovyan/work -w /home/jovyan/work jupyter/all-spark-notebook

**What steps do you take once the container is running to reproduce the issue?**
Permission denied when mounting local folders. I tried everything related to files: upload a file from filesystem, save a notebook, access a dataset in the same folder and i cant...

Example:

1. http://(869fbf0e4bbe or 127.0.0.1):8888/?token=d5f852e4322a7daccd8cc380d6199d2a3d071789d7ef9b60

![imagen](https://user-images.githubusercontent.com/11474846/59591282-0e6ab080-90ee-11e9-8f0b-1c23db0ad575.png)

**What do you expect to happen?**

Access files

**What actually happens?**
Error permission denied

Im using Ubuntu 18.04 LtS

<pre>eduardo@MiPcLinux:~/Descargas/DockerSpark$ docker version
Client:
 Version:           18.06.1-ce
 API version:       1.38
 Go version:        go1.10.4
 Git commit:        e68fc7a
 Built:             Tue May  7 17:57:34 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.06.1-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.4
  Git commit:       e68fc7a
  Built:            Tue May  7 17:57:34 2019
  OS/Arch:          linux/amd64
  Experimental:     false
</pre>

Than ks in advance

ejgutierrez74:
Any help ??

felipebn:
Hey @ejgutierrez74,
Not sure if you managed to solve your issue, but I was having the same.

After reading the docs (https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html) I was able to get it running with the following env options:

```
docker run --rm -p 8888:8888 --name jupyter -e NB_USER=felipebn -e CHOWN_HOME=yes -e CHOWN_EXTRA_OPTS='-R' --user root -w /home/felipebn/ -v "C:\Users\felipe.brandao\eclipse-workspace-2019-01\Project\jupyter":/home/felipebn/work jupyter/base-notebook 
```

Important parts are:
[...]

PR #1152, opened 2020-08-21
Add -f option with groupadd to gracefully exit when adding existing groups

jgwerner:
Starting the container with `--user root`, as indicated by the [Docker Options section of the docs](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#docker-options) in conjunction with a script using the `start-notebook.d` or the `before-notebook.d` hooks that set a custom`NB_UID`, `NB_UID`, and using the standard `NB_USER` (jovyan) environment variables results in the container exiting stating that the groupadd failed because the `jovyan` group already exists.

This fix adds the `-f` flag to the `groupadd` command to either add the [group if it doesn't exist or exit gracefully if it does exist](http://manpages.ubuntu.com/manpages/cosmic/man8/groupadd.8.html).

Another viable option is to use `groupadd --gid $NB_GID --non-unique ${NB_GROUP:-${NB_USER}}` as done in PR #1052 

Might be related to #956 #885

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -80,7 +80,7 @@ if [ $(id -u) == 0 ] ; then
     if [ "$NB_UID" != $(id -u $NB_USER) ] || [ "$NB_GID" != $(id -g $NB_USER) ]; then
         echo "Set user $NB_USER UID:GID to: $NB_UID:$NB_GID"
         if [ "$NB_GID" != $(id -g $NB_USER) ]; then
-            groupadd -g $NB_GID -o ${NB_GROUP:-${NB_USER}}
+            groupadd -f -g $NB_GID -o ${NB_GROUP:-${NB_USER}}
         fi
         userdel $NB_USER
         useradd --home /home/$NB_USER -u $NB_UID -g $NB_GID -G 100 -l $NB_USER


jgwerner:
Hi @romainx ! Thanks for responding. Apologies, I should have been clearer about the use-case.

We are using an image built with the `repo2docker` tool and then copying files from the `jupyter/base-notebook` image to make it compatible with the standard `jupyter/docker-stacks` images for use with the `JupyterHub`. `repo2docker` seems to use the same id [for both the UID and the GID](https://github.com/jupyterhub/repo2docker/blob/0dba6111c6b1f0a5f0208193ca9b16891d4469e6/repo2docker/buildpacks/base.py#L51), hence the error.

I'm not sure if there are any plans on making images created by `repo2docker` and `jupyter/docker-stacks` 100% compatible, but at least in our case, we are creating our images this way to simplify how the base-image is built and the add a layer on top of this base image to make it compatible with docker-stacks.

romainx:
@jgwerner Thanks I understand better now and it is very clear and it makes sense in this context.

Regarding the compatibility between images in the Jupyter ecosystem, I was recently aware of a conversation jupyterhub/mybinder.org-deploy#1474 that is close to this topic, but no more than that  -- [jupyterhub/binderhub](https://github.com/jupyterhub/binderhub) uses [jupyterhub/repo2docker](https://github.com/jupyterhub/repo2docker) behind the scene. However we could also start by performing small steps and drafting a PR to homogenize the user / group management -- but I guess it will not be so easy since each project may be reluctant to change this core part of the stacks for for compatibility reasons.

@parente I have two questions for you

- Are you good to merge this PR?
- What do you think about working on compatibility improvement between those projects?

If it's OK for you I can take care of both actions.

Many thanks

jgwerner:
@romainx thanks for getting back to me so quickly - and totally understand the complexity of standardizing conventions due to the number of users, a good problem as they say 😄

romainx (approved):

parente:
> Are you good to merge this PR?

Yes.

> What do you think about working on compatibility improvement between those projects?

https://github.com/jupyterhub/mybinder.org-deploy/issues/1474#issuecomment-650582047 captures my personal view of this project. I wouldn't personally be interested in making substantial changes to these images. I am not opposed, however, to others wanting to do that work as long as significant breaking changes do not irrevocably disrupt the existing community of users and others feel comfortable maintaining the images after making such changes.

Issue #1160, opened 2020-09-15
cm-super for matplotlib latex plots


mpmdean:
I think scipy-notebook might be missing the cm-super package required for latex plots?

**What docker image you are using?**

`jupyter/scipy-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**
```
docker pull jupyter/scipy-notebook
jupyter-repo2docker --editable --image-name=jupyter/scipy-notebook .
```

**What steps do you take once the container is running to reproduce the issue?**

Example:

1. Switch to jupyterlab appending /lab to brower URL
2. Open notebook
3. Execute
```
import numpy as np
import matplotlib
matplotlib.rcParams['pgf.texsystem'] = 'pdflatex'
matplotlib.rcParams.update({'font.family': 'serif', 'font.size': 18,
    'axes.labelsize': 20,'axes.titlesize': 24, 'figure.titlesize' : 28})
matplotlib.rcParams['text.usetex'] = True
import matplotlib.pyplot as plt

fig, ax = plt.subplots(1,1)
x = [1, 2]
y = [1, 2]
ax.plot(x,y, label='a label')
ax.legend(fontsize=15)
```
**What do you expect to happen?**
Plot with latex labels

**What actually happens?**

error trackback
```
---------------------------------------------------------------------------
CalledProcessError                        Traceback (most recent call last)
/opt/conda/lib/python3.8/site-packages/matplotlib/texmanager.py in _run_checked_subprocess(self, command, tex)
    303         try:
--> 304             report = subprocess.check_output(command,
    305                                              cwd=self.texcache,

/opt/conda/lib/python3.8/subprocess.py in check_output(timeout, *popenargs, **kwargs)
    410 
--> 411     return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
    412                **kwargs).stdout

/opt/conda/lib/python3.8/subprocess.py in run(input, capture_output, timeout, check, *popenargs, **kwargs)
    511         if check and retcode:
--> 512             raise CalledProcessError(retcode, process.args,
    513                                      output=stdout, stderr=stderr)

CalledProcessError: Command '['latex', '-interaction=nonstopmode', '--halt-on-error', '/home/jovyan/.cache/matplotlib/tex.cache/3bbc0f8d536770a62891b7cc8788c317.tex']' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

RuntimeError                              Traceback (most recent call last)
[...]

PR #1158, opened 2020-09-10
Spark 3.0.1 -> fixes #1156

romainx:
Hello,

Upgrade to Spark `3.0.1` in order to fix #1156.
The build is working only while the version of Spark used in the image is available on the [Spark download page](https://spark.apache.org/downloads.html). In this case Spark `3.0.0` was no more available.

Best.

PR #1162, opened 2020-09-16
add cm-super

mpmdean:
Aims to address issue #1160 -- an error when using latex labels in matplotlib inside scipy-notebook.


romainx (changes requested):
Hello @mpmdean ,

Thank you for this proposal. I've a question regarding the location of the installation of this package.
I guess it's useful not only for `matplotlib` plots and this is the reason why you installed it in `minimal-notebook` with other `nbconvert` dependencies -- and it make sense in this case. In the other case, if it's only useful for `matplotlib` plots we should consider installing it in the `scipy-notebook` since this installation adds ~50 MB to the image size.

Then It will be nice to add a test at the following location https://github.com/jupyter/docker-stacks/blob/612aa5710bf9f9394a6b108cda6cbbc88f41f924/scipy-notebook/test/test_matplotlib.py#L12

Since you have already a snippet of code that reproduces the issue, it's worth adding it to the test cases. Let me know if you have trouble adding it.
Finally could you add a comment near the package in order to indicate its purpose?

Note: The build is failing due to a previous issue fixed in the PR #1158. You can fix the build by rebasing your branch with the origin/master.

Many thanks.

mpmdean:
Hi @romainx 

It's a bit marginal whether the change belongs in ``minimal-notebook`` or ``scipy-notebook``. 

I assumed that it would probably be helpful for ``nbconvert`` since that uses latex, but having looked in more detail one probably needs to have written a custom template and added the computer modern font. For matplotlib, however, cm-super is a optional dependency and one can run into problems relatively easily and unexpectedly.

This package:
https://phonopy.github.io/phonopy/
assumes cm-super is installed for example. 

Maybe  ``scipy-notebook` is a better location? Please let me know what you think.

romainx:
Hello @mpmdean,

Thanks for the precision. I'm sorry but I do not know so much about `nbconvert`. Maybe if we are hesitating between the two location let's add it in `scipy-notebook`

- because we know it is useful
- because it will avoid increasing the size of all images

Is it Ok for you?

If so, could you also consider add your test case like it's already done here for `matplotlib`?

https://github.com/jupyter/docker-stacks/blob/612aa5710bf9f9394a6b108cda6cbbc88f41f924/scipy-notebook/test/test_matplotlib.py#L12

If you are not confortable with that, please let me know I will perform it  in a follow-on PR.

Best.

mpmdean:
@romainx 

That rebase seems to have created a total mess. I think I'll close this and create a new PR of the updated master.

mpmdean:
Changes continue in #1169

Issue #1086, opened 2020-05-19
Missing CONDA_PREFIX


maresb:
I conda installed [`code-server`](https://github.com/cdr/code-server) and [`jupyter-vscode-proxy`](https://github.com/betatim/vscode-binder) in order to run VS Code from JupyterLab. The feedstock for `code-server` [generates a eponymous bash script](https://github.com/conda-forge/code-server-feedstock/blob/master/recipe/build.sh) which relies on the `CONDA_PREFIX` environment variable. I get a beautiful "VS Code" launcher icon, but upon pressing it, I get the following error in the console:
```
Error: Cannot find module '/share/code-server/out/node/entry.js'
```
The error is easily explained by the fact that the `/opt/conda` prefix is defined in `CONDA_DIR` while `CONDA_PREFIX` is empty.

As far as I can tell, `CONDA_PREFIX` should be defined in any activated conda environment. (Almost all my conda experience is via these Docker images.) Hence I'm filing the issue here.

Would it be appropriate to rename `CONDA_DIR` to `CONDA_PREFIX` everywhere? And for good measure set it inside `base-notebook` to avoid breaking compatibility? Or is there some reason to instead use `CONDA_DIR`?

bollwyvl:
CONDA_DIR is indeed non-standard. Unfortunately, just setting some env vars
is not sufficient. Not fully activating the  environment is no doubt a
source of subtle problems.... As packages get more complex (e.g anything
that uses Java, or a runtime compiler) they end up relying on activation
scripts.

About the only thing that would work for arbitrary "docker exec" is a
wrapper script in ENTRYPOINT, which might already be used for tini, to do
"source activate".

maresb:
To clarify, I'm not actually using `docker exec`. I have my own `Dockerfile` extending these containers, so I'm still using the standard activation scripts. But perhaps this is yet another factor to consider in https://github.com/jupyter/docker-stacks/pull/1052.

romainx:
Hi @maresb 

`CONDA_DIR` even if not standard is basically used everywhere in these images and probably elsewhere in custom images build by users. So it seems complicated to remove it. 
Moreover the two variables have not the same intend

* `CONDA_DIR`: the root of the `conda` distribution installation.
* `CONDA_PREFIX`: the path to the conda environment [source][LK1]

So they can be different when using different environment.

```bash
[...]

PR #1166, opened 2020-09-20
Activate conda base environment by default

jan-janssen:
Some conda packages rely in the `conda activate` to set the corresponding environment variables, therefore just including the `bin` directory in the `PATH` is not sufficient.

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -84,7 +84,7 @@ WORKDIR /tmp
 RUN wget --quiet https://repo.continuum.io/miniconda/Miniconda3-py38_${MINICONDA_VERSION}-Linux-x86_64.sh && \
     echo "${MINICONDA_MD5} *Miniconda3-py38_${MINICONDA_VERSION}-Linux-x86_64.sh" | md5sum -c - && \
     /bin/bash Miniconda3-py38_${MINICONDA_VERSION}-Linux-x86_64.sh -f -b -p $CONDA_DIR && \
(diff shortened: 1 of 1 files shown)


romainx (changes requested):
Hello,

Thanks for submitting this PR. You're right, there is definitely something to do and I think it's summarized in #1086. However I'm not sure about the way to do it. What you have done is to overwrite the `~/.profile`. This solution could cause problem in certain contexts where users provide their own profile through bind mount or when running these images in K8s. 

I do not have the solution at the time of writing however, I do not think we will merge this PR as is.
If you have a better solution, feel free to propose it.

Best

jan-janssen:
The alternative option would be adding it directly to `/etc/profile` as that should be loaded before any user specific files are loaded.

romainx:
Another option, more consistent with what is already done, is to add the content generated by the command `conda init bash` in the `.bashrc` skeleton `/etc/skel/.bashrc`. It is already customized in the image.

https://github.com/jupyter/docker-stacks/blob/612aa5710bf9f9394a6b108cda6cbbc88f41f924/base-notebook/Dockerfile#L56-L57

This [SO answer](https://stackoverflow.com/a/58081608/4413446) gives an alternative that could be interesting in our case.

Here is the change to perform.

```diff
# Enable prompt color in the skeleton .bashrc before creating the default NB_USER
+ # hadolint ignore=SC2016
- RUN sed -i 's/^#force_color_prompt=yes/force_color_prompt=yes/' /etc/skel/.bashrc
+ RUN sed -i 's/^#force_color_prompt=yes/force_color_prompt=yes/' /etc/skel/.bashrc && \
+    # Add call to conda init script see https://stackoverflow.com/a/58081608/4413446
+    echo 'eval "$(command conda shell.bash hook 2> /dev/null)"' >> /etc/skel/.bashrc 
```

Once added, the `base` environment is activated by default and the variables are set.

```bash
$ docker run -it --rm  jupyter/base-notebook bash

(base) jovyan@c3ee08cdc4ff:~$ env | grep -E '(CONDA_DIR|CONDA_PREFIX)'
CONDA_PREFIX=/opt/conda
CONDA_DIR=/opt/conda
```

Could you please try to implement it this way and check if it answers your needs?

Many thanks

jan-janssen:
@romainx That is a great suggestion - it works for me, so I updated the pull request accordingly.

romainx (approved):

romainx:
@jan-janssen thanks for the feedback.
@parente I have contributed to the solution, si I would like to have your opinion before merging this PR.
Many thanks.

PR #1169, opened 2020-09-25
add cm-super

mpmdean:
This follows on from #1162 and correcting a messed up git history and adding changes requested from @romainx

PR #1170, opened 2020-09-26
Matplotlib test refactoring (Follow-on PR #1169)

romainx:
Just a refactor of the `matplotlib` tests to avoid code duplication.


romainx:
I'm merging it because it's just a refactoring related to a new test added in PR #1169.

PR #1171, opened 2020-09-29
Upgrading JDK to 14

Bidek56:
Upgrading JDK used in Spark to 14.
I am having a Spark K8s [issue](https://github.com/fabric8io/kubernetes-client/issues/2145) which I am hoping will be resolved by JDK 14.


Bidek56:
Closing since SparkR does not work with JDK > 12 due to:
`Error in checkJavaVersion(): Java version, greater than or equal to 8 and less than 12, is required for this package; found version: 14.0.1`

romainx:
@Bidek56 thanks good to know.

PR #1172, opened 2020-09-30
nano -> nano-tiny

romainx:
A tiny improvement, but consistent with what has already been done.


parente (approved):
LGTM. For anyone else who wants to know the difference between nano and nano-tiny: https://manpages.debian.org/stretch/nano-tiny/nano-tiny.1.en.html

Issue #1148, opened 2020-08-16
Run pre-commit hooks server side


mathbunnyru:
We already discussed with @romainx in https://github.com/jupyter/docker-stacks/pull/1123 pre-commit should be run server-side, to make sure everything will look good in master branch.

An easy way to run all pre-commit hooks:
`pre-commit run --all-files`

Actually, there are already many bashate and few hadolint errors.
So it would be great to fix them and enable server-side checks.

romainx:
Hello,

Now that we have moved to Github actions we should have a look to https://github.com/pre-commit/action.
I'm Ok to work on it and have a clearer way to lint all the files.

Best

mathbunnyru:
Great, but first I will have to fix several bashate errors.

romainx:
@mathbunnyru for bashate errors I suggest the following configuration, it will permit to reduce significantly the number of errors by ignoring the rule `E006`: check for lines longer than 79 columns.

```YAML
- repo: https://github.com/openstack-dev/bashate.git
  rev: 2.0.0
  hooks:
    - id: bashate
      # Ignore
      # - E006: check for lines longer than 79 columns
      args: ['--ignore=E006']
```

The remaining errors will not be too hard to fix nor too risky

```bash
base-notebook/fix-permissions:27:1: E003 Indent not multiple of 4
base-notebook/fix-permissions:27:1: E003 Indent not multiple of 4
base-notebook/fix-permissions:27:1: E003 Indent not multiple of 4
base-notebook/fix-permissions:34:1: E003 Indent not multiple of 4
r-notebook/hooks/run_hook:55:1: E004 File did not end with a newline
5 bashate error(s) found
tensorflow-notebook/hooks/run_hook:45:1: E004 File did not end with a newline
1 bashate error(s) found
minimal-notebook/hooks/run_hook:45:1: E004 File did not end with a newline
1 bashate error(s) found
```

mathbunnyru:
All issues are fixed, I will read the docs about github actions and add this feature.

romainx:
@mathbunnyru OK but maybe I was wrong by pointing the reference to pre-commit actions. We have to stick with what has been done by @parente and maybe rely on the `make` tasks more than GitHub actions. This permits to reproduce at the command line the build steps and to keep a consistent behavior. For example
https://github.com/jupyter/docker-stacks/blob/36d857bb01215cd6ebf67e631092b4c1e81c75f0/.github/workflows/docker.yml#L46-L47

[...]

PR #1173, opened 2020-09-30
Fix bashate style issues

mathbunnyru:
Done as suggested in https://github.com/jupyter/docker-stacks/issues/1148

After that there are no pre-commit issues.


romainx (approved):
Hello,

Thank you everything seems OK.
Ready to merge 🚀 
Best.

PR #1176, opened 2020-10-05
Run pre-commit hooks server side

### LINKED ISSUE #1148: Run pre-commit hooks server side

mathbunnyru:
Fix #1148 

I decided to stick with docker.yml - it seems to work fine and I don't want to add another way of running ci instead of using the existing one.


mathbunnyru:
I removed hadolint from pre-commit, because there's another makefile target and I don't want it to run twice on CI.

I also fixed naming, because the word `lint` is too common and I renamed it to `hadolint`.

romainx (approved):
Great 😄 
I think splitting the two kinds of lint is a reasonable solution to avoid too much complexity.

This setup is the conclusion of a long-term work.
Thanks a lot for this contribution 🍰

PR #1182, opened 2020-10-25
nbconvert + inkscape test improvements

romainx:
Hello,

Just an improvement in order to contribute to #1083 and #1126 by testing if

- inkscape is installed and working because it's a dependency of nbconvert
- nbconvert is able to convert a notebook containing SVG (test that fails if inkscape is not installed)

Best


romainx:
I'm going to merge it since it's only test improvement that does not affect user facing features.

PR #1184, opened 2020-11-02
/usr/bin/nano-tiny does not exist

Weihrauch:
/usr/bin/nano-tiny does not exist, the good path is /bin/nano-tiny on ubuntu 18.04 and 20.40 (https://ubuntu.pkgs.org/20.04/ubuntu-universe-arm64/nano-tiny_4.8-1ubuntu1_arm64.deb.html).

### DIFF
--- a/minimal-notebook/Dockerfile
+++ b/minimal-notebook/Dockerfile
@@ -32,7 +32,7 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
 # Create alternative for nano -> nano-tiny
-RUN update-alternatives --install /usr/bin/nano nano /usr/bin/nano-tiny 10
+RUN update-alternatives --install /usr/bin/nano nano /bin/nano-tiny 10
 
 # Switch back to jovyan to avoid accidental container runs as root
 USER $NB_UID


romainx (commented):
Hello,

Thanks for this PR 😄.
I've checked the link and you're right. However it will not change anything, right?

```bash
$ ll /bin
# lrwxrwxrwx 1 root root 7 Sep 25 01:20 /bin -> usr/bin/

$ which nano-tiny
# /usr/bin/nano-tiny
```

What do you think?

Weihrauch:
I'am using the base image nvidia/cuda:10.2-base-ubuntu18.04 to build the base-notebook in the repository. I don't understand how you get a symbolic on /bin. Even on my Ubuntu desktop /bin is not a symbolic link.
What's the base image are you using on your side ?

mathbunnyru:
By default, docker-stacks is built on ubuntu:focal image (https://github.com/jupyter/docker-stacks/blob/master/base-notebook/Dockerfile).

And in ubuntu:focal it's indeed symlink:
```
~ > docker run -it ubuntu:focal
root@e261c632e114:/# ll /bin
lrwxrwxrwx 1 root root 7 Jul  3 01:56 /bin -> usr/bin/
```

It's true in the real ubuntu focal as well.
https://askubuntu.com/questions/1242157/why-are-bin-and-sbin-now-symlinks-in-ubuntu-20-04

In my opinion, this change doesn't affect any existing users and at least for someone makes their life easier.
So, let's go for a merge :)

romainx:
Thank you @Weihrauch & @mathbunnyru for the precisions it makes sense.
@Weihrauch I was not aware that you were using a different base image.
So let's go 🚀

PR #1186, opened 2020-11-06
[Bug fix] Create home directory before changing ownership

stevenstetzler:
There's a bug in the `start.sh` that causes a crash when starting a notebook with `NB_USER=<username>` and `CHOWN_HOME=yes` environment variable. For example:
```
$ mkdir -p pr/base
$ cd pr/base
$ git clone https://github.com/jupyter/docker-stacks.git
$ cd docker-stacks
$ make build/base-notebook
$ docker tag jupyter/base-notebook:latest jupyter/base-notebook:base
$ docker run -e NB_USER=steven -e CHOWN_HOME=yes --user root jupyter/base-notebook:base
Set username to: steven
Changing ownership of /home/steven to 1000:100 with options ''
chown: cannot access '/home/steven': No such file or directory
```
This happens because `chown` gets called in  `start.sh` before moving/creating the directory `/home/$NB_USER`: https://github.com/jupyter/docker-stacks/blob/master/base-notebook/start.sh#L50-L77

I've flipped the order of these so that `/home/$NB_USER` is made before `chown`: https://github.com/stevenstetzler/docker-stacks/blob/master/base-notebook/start.sh#L50-L77

This results in a success:
```
$ mkdir -p pr/head
$ cd pr/head
$ git clone https://github.com/stevenstetzler/docker-stacks.git
$ cd docker-stacks
$ make build/base-notebook
$ docker tag jupyter/base-notebook:latest jupyter/base-notebook:head
$ docker run -e NB_USER=steven -e CHOWN_HOME=yes --user root jupyter/base-notebook:head
Set username to: steven
Relocating home dir to /home/steven
Setting CWD to /home/steven/
Changing ownership of /home/steven to 1000:100 with options ''
Executing the command: jupyter notebook
...
```

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -47,19 +47,6 @@ if [ $(id -u) == 0 ] ; then
         usermod -d /home/$NB_USER -l $NB_USER jovyan
     fi
 
-    # Handle case where provisioned storage does not have the correct permissions by default
-    # Ex: default NFS/EFS (no auto-uid/gid)
-    if [[ "$CHOWN_HOME" == "1" || "$CHOWN_HOME" == 'yes' ]]; then
-        echo "Changing ownership of /home/$NB_USER to $NB_UID:$NB_GID with options '${CHOWN_HOME_OPTS}'"
-        chown $CHOWN_HOME_OPTS $NB_UID:$NB_GID /home/$NB_USER
-    fi
-    if [ ! -z "$CHOWN_EXTRA" ]; then
-        for extra_dir in $(echo $CHOWN_EXTRA | tr ',' ' '); do
-            echo "Changing ownership of ${extra_dir} to $NB_UID:$NB_GID with options '${CHOWN_EXTRA_OPTS}'"
-            chown $CHOWN_EXTRA_OPTS $NB_UID:$NB_GID $extra_dir
-        done
-    fi
-
     # handle home and working directory if the username changed
     if [[ "$NB_USER" != "jovyan" ]]; then
         # changing username, make sure homedir exists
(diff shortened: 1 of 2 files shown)


stevenstetzler:
Think this is a duplicate of https://github.com/jupyter/docker-stacks/pull/1128, I should have checked first!

romainx (changes requested):
Hello,

Thank you for this PR. It looks good to me. However I was wondering why the following test was OK without this fix.

https://github.com/jupyter/docker-stacks/blob/68d2d835fb16f8cb1774174adbe070d4900c197b/base-notebook/test/test_container_options.py#L69-L81

It it due to the `--workdir` option. But in fact the behavior was not exactly the expected one. The home directory was created and not changed as expected.

```bash
$ ls -l /home

# drwsrwsr-x 1 steven users 4096 Nov  5 18:46 jovyan
# drwxr-xr-x 4 steven users 4096 Nov  8 15:51 steven
```

So we need to also to change the test to remove the `working_dir` option that still needs to be added when running commands inside the started container (without this change commands will use the default working dir set in the image that does not exist).

```diff
def test_nb_user_change(container):
    """Container should change the user name (`NB_USER`) of the default user."""
    nb_user = "nayvoj"
    running_container = container.run(
        tty=True,
        user="root",
        environment=[
            f"NB_USER={nb_user}",
            "CHOWN_HOME=yes"
        ],
-        working_dir=f"/home/{nb_user}",
        command=['start.sh', 'bash', '-c', 'sleep infinity']
    )

    # ...
    LOGGER.info(f"Checking {nb_user} id ...")
    command = "id"
    expected_output = f"uid=1000({nb_user}) gid=100(users) groups=100(users)"
-  cmd = running_container.exec_run(command, user=nb_user)
+  cmd = running_container.exec_run(command, user=nb_user, workdir=f"/home/{nb_user}")
    # ...

    LOGGER.info(f"Checking if {nb_user} owns his home folder ...")
    command = f'stat -c "%U %G" /home/{nb_user}/'
    expected_output = f"{nb_user} users"
-  cmd = running_container.exec_run(command)
+  cmd = running_container.exec_run(command, workdir=f"/home/{nb_user}")
```

Could you please take this change into account? If it's not possible for you I will do it in a follow-on PR.

Related:

- #1077 

Many thanks for your contribution.
Best.

stevenstetzler:
Thanks for reviewing this change. I've added a commit to address moving the `workdir` option from container creation to the exec commands in the test.

romainx (approved):
Great, thank you.
LGTM however I would like to have the confirmation of @parente before merging it since he has left a comment in the duplicate #1112 

Best

parente (approved):
Let's try it.

Issue #1174, opened 2020-10-01
Proposal: Reduce the number of editors in the stacks


romainx:
Hello,

We have plenty of editors installed in the stacks + Jupyter editors.

```bash
$ update-alternatives --config editor

# There are 4 choices for the alternative editor (providing /usr/bin/editor).
# 
#   Selection    Path               Priority   Status
# ------------------------------------------------------------
# * 0            /usr/bin/jed        42        auto mode
#   1            /bin/nano-tiny      0         manual mode
#   2            /usr/bin/emacs      0         manual mode
#   3            /usr/bin/jed        42        manual mode
#   4            /usr/bin/vim.tiny   15        manual mode
```

Some of them (in fact emacs) have a significative impact on the images size.

```bash
$ packages="vim-common vim-tiny jed jed-common emacs-nox emacs-bin-common emacs-common emacsen-common nano-tiny"
$ apt-cache --no-all-versions show $packages | awk '$1 == "Package:" {p = $2} $1 == "Installed-Size:" {printf("%s: %s\n", p, $2, is)}'

# vim-common: 375
# vim-tiny: 1530
# ---
# vim-total: ~ 2 MB

# jed: 361
# jed-common: 1806
# ---
# jed-total: ~ 2 MB
 
# emacs-nox: 41291
# emacs-bin-common: 470
# emacs-common: 66816
# emacsen-common: 62
# ---
# emacs-total: ~ 100 MB
 
# nano-tiny: 177
# ---
# nano-total: 177 KB
```

# Proposal

1. Remove jed. Rationale: I don't think it is widely used 👴--but maybe I'm wrong. The last version has been released more than 10 years ago (2009)
2. Remove emacs (I don't want to create a psychodrama 😨). Rationale: It's size 🏋️. The stack is dedicated to Jupyter editors, other editors should be here has minimal helpers--but maybe I'm wrong.

Estimated size reduction: **1.63GB  -> 1.51GB** on `minimal-notebook` and all downstream images.

What is your opinion, please let me know?

Best

Bidek56:
Emacs and Jed have to go, my grandmother stopped using those years ago. :)

mathematicalmichael:
100% agree

PR #1188, opened 2020-11-12
Remove emacs and jed from the minimal image

mathbunnyru:
Fix https://github.com/jupyter/docker-stacks/issues/1174

### DIFF
--- a/docs/using/selecting.md
+++ b/docs/using/selecting.md
@@ -48,8 +48,8 @@ and versioning strategy.
 
 - Everything in `jupyter/base-notebook`
 - [TeX Live](https://www.tug.org/texlive/) for notebook document conversion
-- [git](https://git-scm.com/), [emacs](https://www.gnu.org/software/emacs/) (actually `emacs-nox`),
-  [vi](https://vim.org/) (actually `vim-tiny`), [jed](https://www.jedsoft.org/jed/),
+- [git](https://git-scm.com/),
+  [vi](https://vim.org/) (actually `vim-tiny`),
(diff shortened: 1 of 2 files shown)


romainx:
Thanks @mathbunnyru and goodbye old friends 😄 

However @parente was not in favor of removing jed at least.
> #121 (comment) talks about why jed was proposed and added. Do we gain much by removing it?

Sorry I should have mentioned it in the issue #1174
So I let him approve it or not. If it's not approved we will close the corresponding issue.

Thanks. again

parente:
I've been on the fence about removing `jed` since it takes up very little space, is stable, and offers new users a basic editing experience. I also see the argument for making the "minimal" image smaller, removing overlapping tools, and focusing on Jupyter tools with other software playing a supporting role. I think the latter mindset is consistent with the recent project trajectory, and so I'm OK with merging this PR and removing both editors.

One question: if someone submits a PR removing vim, will we be consistent and accept it?

romainx:
@parente the war between emacs and vim is never over 🤣 
- **My official answer**: Yes but we have to keep at least a minimal editor to perform basic tasks in the terminal say choosing between vim and nano. :innocent:
- **What I am thinking**: NO! :scream:

mathbunnyru:
@parente hopefully, it's not my choice to make 🤣 (I am not a maintainer)

I would personally argue against deleting vim - it makes it more difficult to debug these images when you don't have tools at all.
So, yeas, as @romainx said, it's nice to have at least a minimal editor.

PR #1189, opened 2020-11-13
Switch from Miniconda to Miniforge

romainx:
Hello,

I'm _trying_ to perform a structuring change by switching from [Miniconda](https://docs.conda.io/en/latest/miniconda.html) installer to [Miniforge](https://github.com/conda-forge/miniforge).

> This repository holds a minimal installer for conda specific to conda-forge.

Here are the main reasons

* Switch to **open source** since Anaconda has recently updated its [TOS](https://www.anaconda.com/terms-of-service) to restrict commercial usage of its repositories (including the repo hosting Miniconda installers) -> This could impact users of Jupyter Docker images
* It's built to use `conda-forge` channel by default
  * it is the channel we are using that is not restricted
  * there is no need to prepend it anymore
  * other [default channels](https://repo.anaconda.com/pkgs/) that are concerned by the new TOS are not configured
* Support various CPU architecture (installer can be adapted through `ARGS`), could be a good asset to build [multi-arch](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) images.
* Support for [pypy](https://www.pypy.org/) by using alternative installers `Miniforge-pypy3` (thanks @mathbunnyru)

Some minor pros

* There is almost no change in the installation process
* Base image size seems to identical or slightly reduced
* Speed up the package dependencies resolution (only one channel)

## Implementation

* Installation is pretty much the same 😊 
* Split the installer definition in several `ARGs` to ease overriding and for better clarity (I hope)
* Removed the environment variable `MINICONDA_VERSION` because I'm not sure it makes sense to replace it by the Miniforge equivalent
* Removed the `strict` channel priority of the `conda` installer because 
  * it is no more necessary with only one channel
  * it makes installation of packages from other channels (set by users) harder
* Decision to keep the installer checksum as an `ARG` instead of downloading it for security reasons (@mathbunnyru suggestion)
* Documented in "Common Features" the way to install packages from `defaults` channels

## Notes on multi-arch images

Here is a table displaying the various CPU architecture available

| Ubuntu    | Miniforge       | Miniconda          |
|-----------|-----------------|--------------------|
| `arm64`   | `linux-aarch64` |  -                 |
| `amd64`   | `linux-x86_64`  | `Linux-x86_64.sh`  |
| `ppc64le` | `linux-ppc64le` | `Linux-ppc64le.sh` |

List of `ubuntu:focal` architectures

```bash
curl -s https://hub.docker.com/v2/repositories/library/ubuntu/tags | \
    jq -r '.results[] | select(.name=="focal") | .images[].architecture'

# arm64
# s390x
# amd64
# ppc64le
# arm
```

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -27,8 +27,9 @@ USER root
 ARG conda_version="4.9.0"
 # Miniforge archive to install
 ARG miniforge_version="${conda_version}-3"
-# Miniforge installer
+# Miniforge installer and its checksum
 ARG miniforge_installer="Miniforge3-${miniforge_version}-Linux-x86_64.sh"
+ARG miniforge_checksum="29f0eb17dd02aceb0dfd4dad2654e974b1699baed06ee6d350b0ab4a2ccf3d02"
 
 # Install all OS dependencies for notebook server that starts but lacks all
 # features (e.g., download as all possible file formats)
@@ -93,10 +94,9 @@ RUN mkdir "/home/$NB_USER/work" && \
 WORKDIR /tmp
 
 RUN wget --quiet "https://github.com/conda-forge/miniforge/releases/download/${miniforge_version}/${miniforge_installer}" && \
-    wget --quiet "https://github.com/conda-forge/miniforge/releases/download/${miniforge_version}/${miniforge_installer}.sha256" && \
-    sha256sum -c "${miniforge_installer}.sha256" && \
+    echo "${miniforge_checksum} *${miniforge_installer}" | sha256sum --check && \
     /bin/bash "${miniforge_installer}" -f -b -p $CONDA_DIR && \
-    rm "${miniforge_installer}" "${miniforge_installer}.sha256" && \
+    rm "${miniforge_installer}" && \
     # Conda configuration see https://conda.io/projects/conda/en/latest/configuration.html
     echo "conda ${CONDA_VERSION}" >> $CONDA_DIR/conda-meta/pinned && \
     conda config --system --set auto_update_conda false && \
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -19,16 +19,25 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 USER root
 
-# Miniforge installation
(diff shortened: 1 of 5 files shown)


base-notebook/Dockerfile
```diff
@@ -82,25 +83,22 @@ RUN echo "auth requisite pam_deny.so" >> /etc/pam.d/su && \
...
+# Install conda as jovyan and check the sha256 sum provided on the download site
 WORKDIR /tmp
-RUN wget --quiet https://repo.continuum.io/miniconda/Miniconda3-py38_${MINICONDA_VERSION}-Linux-x86_64.sh && \
-    echo "${miniconda_checksum} *Miniconda3-py38_${MINICONDA_VERSION}-Linux-x86_64.sh" | md5sum -c - && \
-    /bin/bash Miniconda3-py38_${MINICONDA_VERSION}-Linux-x86_64.sh -f -b -p $CONDA_DIR && \
-    rm Miniconda3-py38_${MINICONDA_VERSION}-Linux-x86_64.sh && \
+
+RUN wget --quiet "https://github.com/conda-forge/miniforge/releases/download/${miniforge_version}/${miniforge_installer}" && \
+    wget --quiet "https://github.com/conda-forge/miniforge/releases/download/${miniforge_version}/${miniforge_installer}.sha256" && \
+    sha256sum -c "${miniforge_installer}.sha256" && \
```
mathbunnyru, line 97: I don't think it's safe to compare the sha256 of the installer with the *downloaded* sha256.

Maybe I'm getting it wrong, but if someone will have the access to the repo, they will change sha256 and we will get no error.

mathbunnyru:
Another great advantage is an ability to use pypy - maybe someone wants it might give good speed improvement.

base-notebook/Dockerfile
romainx: @mathbunnyru I asked myself the question. I didn't find any clear recommandation. I will change it you must be right.
Thanks for the review 😄

mathbunnyru:
LGTM

parente (approved):
This also looks very good to me, @romainx. Thanks for staying on top of conda developments. Merging now. We can deal with issues as they arise.

PR #1192, opened 2020-11-24
Improve spark installation

romainx:
Spark installation improved by sourcing the `spark-config.sh` in the `before-notebook.d` hook that is run by `start.sh`. It permits to add automatically the right Py4J dependency version in the `PYTHONPATH`. So it is not needed anymore to set this variable at build time.

Documentation describing the installation of a custom Spark version modified to remove this step. Also updated to install the latest `2.x` Spark version.

`test_pyspark` fixed (was always OK before that).

Best

### DIFF
--- a/pyspark-notebook/Dockerfile
+++ b/pyspark-notebook/Dockerfile
@@ -16,7 +16,6 @@ USER root
 ARG spark_version="3.0.1"
 ARG hadoop_version="3.2"
 ARG spark_checksum="E8B47C5B658E0FBC1E57EEA06262649D8418AE2B2765E44DA53AAF50094877D17297CC5F0B9B35DF2CEEF830F19AA31D7E56EAD950BBE7F8830D6874F88CFC3C"
-ARG py4j_version="0.10.9"
 ARG openjdk_version="11"
 
 ENV APACHE_SPARK_VERSION="${spark_version}" \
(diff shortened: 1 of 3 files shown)


pyspark-notebook/Dockerfile
```diff
@@ -39,14 +38,17 @@ RUN wget -q $(wget -qO- https://www.apache.org/dyn/closer.lua/spark/spark-${APAC
...
 ENV SPARK_HOME=/usr/local/spark
-ENV PYTHONPATH="${SPARK_HOME}/python:${SPARK_HOME}/python/lib/py4j-${py4j_version}-src.zip" \
-    SPARK_OPTS="--driver-java-options=-Xms1024M --driver-java-options=-Xmx4096M --driver-java-options=-Dlog4j.logLevel=info" \
+ENV SPARK_OPTS="--driver-java-options=-Xms1024M --driver-java-options=-Xmx4096M --driver-java-options=-Dlog4j.logLevel=info" \
     PATH=$PATH:$SPARK_HOME/bin
 
+RUN ln -s "spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}" spark && \
+    # Add a link in the before_notebook hook in order to source automatically PYTHONPATH
+    mkdir -p /usr/local/bin/before-notebook.d && \
+    ln -s "${SPARK_HOME}/sbin/spark-config.sh" /usr/local/bin/before-notebook.d/spark-config.sh
```
parente, line 50: Do you know if this script is present in older versions of Spark? I'm wondering if it'll work for users who override `spark_version` during build to something in the 2.x line of releases.

romainx: Yes. I've tested it through the example updated in the documentation with Spark `2.4.7` and it's one of the benefit of using this script since it will adapt the PY4J version.

```bash
$ tail -5 /usr/local/spark-2.4.7-bin-hadoop2.7/sbin/spark-config.sh
# if [ -z "${PYSPARK_PYTHONPATH_SET}" ]; then
#   export PYTHONPATH="${SPARK_HOME}/python:${PYTHONPATH}"
#   export PYTHONPATH="${SPARK_HOME}/python/lib/py4j-0.10.7-src.zip:${PYTHONPATH}"
#   export PYSPARK_PYTHONPATH_SET=1
# fi
```
And I've checked in older versions as well.

```bash
$ tar -tvf ~/Downloads/spark-2.3.0-bin-hadoop2.7.tgz | grep spark-config.sh
# -rwxr-xr-x  0 sameera 1876110778     1429 22 fév  2018 spark-2.3.0-bin-hadoop2.7/sbin/spark-config.sh

$ tar -tvf ~/Downloads/spark-2.2.0-bin-hadoop2.7.tgz | grep spark-config.sh
# -rwxr-xr-x  0 jenkins jenkins     1429  1 jul  2017 spark-2.2.0-bin-hadoop2.7/sbin/spark-config.sh
```

parente (approved):
:shipit:

PR #1201, opened 2020-12-20
Fix .dockerignore by adding default hooks + test dirs

romainx:
Hello,

Just figured out that standards `hooks` and `test` directories that are part of most of the Docker image's folder were not included in the `.dockerignore`.
It's not a big improvement but they will not be anymore send to the daemon now.

Best


romainx:
Since is a straightforward and non user-facing change, I'm merging it directly.

Best.

PR #1206, opened 2020-12-29
Update 2020-12-27

### LINKED ISSUE #1167: Problem with `Restart kernel and run all cells` for R and Julia kernels in `datascience-notebook`?
Hi,

I'm having issues with the `Restart kernel and run all cells` command for **R** and **Julia** kernels in the latest versions of the `datascience-notebook` (both the option in the `Kernel` menu and the shortcut button at the top of my notebooks). Older versions of the image (e.g. `54462805efcb`) work fine, but in the latest versions I have to _separately_ use the "Restart kernel" command - sometimes several times - followed by the `Run all` command. **This only applies to the Julia and R kernels**; `Restart kernel and run all cells` works fine for the Python kernel.

All kernels do work eventually (i.e. my code runs successfully), but the user experience is now a little confusing/frustrating, especially for new users.

`jupyter/datascience-notebook`

`docker run -ti --rm -p 8888:8888 -v ${PWD}:/home/jovyan/work jupyter/datascience-notebook:latest start.sh jupyter lab`

1. Visit http://127.0.0.1:8888 and log-in with the token
2. Create a new **R** or **Julia** notebook and enter some simple example code in the first cell (e.g. `print("Hello world")` for R or `println("Hello world")` for Julia). Run the cell and confirm that the output is printed successfully
3. Choose `Restart kernel and run all cells` from the `Kernel` menu (or use the equivalent button at the top of the notebook)

The kernel should be restarted, the cell should be executed, and the output should be printed again.

**What actually happens?**

The notebook "hangs" indefinitely, with asterisk symbols (`*`) next to each code cell. The terminal running the notebook server simply shows e.g. 

```
Kernel restarted: db776add-f28d-45d3-90ce-06e6ab7efb6e
```

but nothing actually happens.

A workaround is to choose `Kernel > Restart kernel`, then wait until it says `Kernel | Idle` at the bottom left of the screen, then choose `Run > Run all cells`. This works OK, but it's cumbersome.

I'm testing with Linux containers running on Docker for Windows, but I have exactly the same problem when the image is deployed in Google Cloud on my JupyterHub.

I guess something isn't waiting for the kernel to restart properly before trying to run the code, but I'm not sure where to look next to pin down the problem. Any advice gratefully received.

Thanks!

### LINKED ISSUE #1205: [base-notebook] Cannot connect to server after upgrading to Jupyter Lab 3.0.0
After upgrading (by unpinning 3 Jupyter packages in [my fork of docker-stacks](https://github.com/mirekphd/docker-stacks/commit/0e7b985eaaba531e8f4dfd440b990c6a6a12f332)) Jupyter Lab / Server no longer works (only Notebook is still OK).

To reproduce:
1. Clone my fork `mirekphd/docker-stacks` (or reproduce it by making [this change](https://github.com/mirekphd/docker-stacks/commit/0e7b985eaaba531e8f4dfd440b990c6a6a12f332)).
2. build `mirekphd/base-notebook` container (`cd base-notebook` and `docker build --tag mirekphd/base-notebook .`)
3. Try to run Jupyter Lab:
`docker run -d --rm --name base-notebook -p 8989:8888 -e JUPYTER_ENABLE_LAB=1 mirekphd/base-notebook:latest && docker logs -f base-notebook`
4. Navigate to `localhost:8989` and see the "Unable to connect" browser error.
5. Switch to Notebook (e.g. `-e JUPYTER_ENABLE_LAB_NOT`) to see it works correctly.

Note that earlier today I raised a less reproducible (closed-source) version of this issue in https://github.com/jupyterlab/jupyterlab/issues/9520, so you may find some additional hints / details there.

romainx:
The JupyterLab version has been kept in `2.2.x` since important changes have been made starting at the version `3.0.0` on the way to manage extensions, see the [JupyterLab Changelog][4]. The result is that some of the extensions installed in `scipy-notebook` are not yet compliant. [This issue][5] summarizes in a table the extensions currently compliant with the new version. 

The upgrade to JupyterLab `3.0.x` is detailed in an issue #1207. However a change has been made in order to prepare this migration #1205 and to help users who want to perform the upgrade by themselves to do so.

The following changes have been made.

- `base-notebook`
  - `Changed`: Bump `notebook`
  - `Changed`: Bump `jupyterhub`
  - `Added`: Created a new configuration file `jupyter_server_config.py` in order to prepare the migration to JupyterLab `3.0.0`, see also the documentation [Migrating from Notebook Server][3] explaining this change -> fixes #1205
- `scipy-notebook`
  - `Changed`: Bump `dask`
  - `Changed`: Bump `ipywidgets`
  - `Changed`: Bump `numba`
  - `Changed`: Bump `protobuf`
  - `Changed`: Bump `scikit-image`
  - `Changed`: Bump `scikit-learn`
  - `Added`: A test to perform a check on each extension `jupyter labextension check`.
- `r-notebook`
  - `Changed`: Bump `r-rmarkdown`
- `tensorflow-notebook`
  - `Changed`: Bump `tensorflow`
- `datascience-notebook`
  - `Changed`: Bump `r-rmarkdown`

[1]: https://ipywidgets.readthedocs.io/en/stable/changelog.html
[2]: https://github.com/bokeh/bokeh/issues/10569
[3]: https://jupyter-server.readthedocs.io/en/latest/operators/migrate-from-nbserver.html
[4]: https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html#v3-0
[5]: https://github.com/jupyterlab/jupyterlab/issues/9461


romainx:
According to the [JupyterLab changelog][1]

> Extensions can be installed without building JupyterLab with NodeJS
> In JupyterLab 3.0, a new recommended way of distributing and installing extensions as Python pip or conda packages is available. Installing such extensions does not require rebuilding JupyterLab and does not require having NodeJS installed. The previous way of distributing extensions as npm packages requiring rebuilding JupyterLab is still available as well. See the documentation for more details.

I'm wondering if we should remove all the _old way_ of installing extensions in `scipy-notebook`

https://github.com/jupyter/docker-stacks/blob/d113a601dbb83dca72f0bc73c7e5f223426cc112/scipy-notebook/Dockerfile#L50-L59

And if we should merge this PR by dropping the support of the extensions that at not working at the moment or wait until their availability ? A compliance table is given here https://github.com/jupyterlab/jupyterlab/issues/9461.

[1]: https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html#v3-0

romainx:
This PR has been modified to rollback to JupyterLab `2.2.x` in order to avoid breaking changes regarding extensions for the time being. Its description has been updated, and I think it's ready to merge now.

romainx:
Need to check with the last update from @mirekphd in #1205 before merging. Turned back this PR into draft.

romainx:
Hello, I've verified and added a basic test to check extensions (we will see how it behaves with JupyterLab `3.0.x`).
So everything seems OK for me 🚀 .
@parente please could you have a look before the merge?

base-notebook/Dockerfile
```diff
@@ -131,8 +131,8 @@ RUN wget --quiet "https://github.com/conda-forge/miniforge/releases/download/${m
 # Do all this in a single RUN command to avoid duplicating all of the
 # files across image layers when the permissions change
 RUN conda install --quiet --yes \
-    'notebook=6.1.5' \
-    'jupyterhub=1.2.2' \
+    'notebook=6.1.6' \
```
parente, line 134: I've been following the convo about updating to JupyterLab 3.0 and waiting for extension compatibility. I'll add that the newly preferred backend for JupyterLab 3.0 is `jupyter_server` instead of `notebook`: https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html#jupyter-server. This has implications for extensions that have both server and frontend components, as I found out when updating https://github.com/parente/jupyterlab-quickopen for 3.0 (e.g., maintain compatibility with both backends? move to jupyter_server alone? how to support binders which deploy notebook by default and not jupyter_server yet?) Not all of these issues are relevant to docker-stacks, but may suggest we need to make a decision about removing notebook from these images down the line, and pointing people to `nbclassic` and the new *Simple* toggle in JupyterLab for a classic-like experience.

base-notebook/Dockerfile
```diff
@@ -150,10 +150,16 @@ CMD ["start-notebook.sh"]
 
 # Copy local files as late as possible to avoid cache busting
 COPY start.sh start-notebook.sh start-singleuser.sh /usr/local/bin/
+# Need to have both files to be able to use classic and lab
```
parente, line 153:
```suggestion
# Currently need to have both jupyter_notebook_config and jupyter_server_config to support classic and lab
```

parente (approved):
👋 I left one comment about lab 3.0 and one suggestion inline. I'm pro-actively approving so you can merge when you feel it's ready. 

Thanks as always for keeping the images fresh and responding to community questions.

base-notebook/Dockerfile
romainx: I agree is not so simple I will add this information to the dedicated issue #1207. We could work on it together to define a strategy. We have to be careful since images are used in different context (binder, k8s, standalone, community, etc.).

bollwyvl:
Pretty sure they will both read jupyter_config.json/py

PR #1209, opened 2021-01-08
Jupyter Notebook Deprecation Notice

romainx:
Hello,

This is a very first step of the feature #1217 that consists in printing a deprecation notice in the `start-notebook.sh`.

Here is the new behavior

- **With no option (no specific environment variable)**: Jupyter Notebook is launched (no change). However a warning message is printed pointing to https://github.com/jupyter/notebook#notice.
- **With `JUPYTER_ENABLE_LAB`**: Jupyter Lab is launched (no change).

I've written a test that checks this behavior.

Best

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ in working on the project.
 
 Following [Jupyter Notebook notice](https://github.com/jupyter/notebook#notice), we encourage users to transition to JupyterLab.
 This can be done by passing the environment variable `JUPYTER_ENABLE_LAB=yes` at container startup, 
-more information is available in the [documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html).
+more information is available in the [documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#docker-options).
 
 In April 2021 JupyterLab will become the default, however a new environment variable will be introduced to switch back to Jupyter Notebook if needed.
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Following [Jupyter Notebook notice](https://github.com/jupyter/notebook#notice),
 This can be done by passing the environment variable `JUPYTER_ENABLE_LAB=yes` at container startup, 
 more information is available in the [documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#docker-options).
 
-In April 2021 JupyterLab will become the default, however a new environment variable will be introduced to switch back to Jupyter Notebook if needed.
+In April 2021 JupyterLab will become the default for all of the Jupyter Docker stack images, however a new environment variable will be introduced to switch back to Jupyter Notebook if needed.
 
 After the change of default, and according to the Jupyter Notebook project status and its compatibility with JupyterLab, these Docker images may remove the classic Jupyter Notebook interface altogether in favor of another *classic-like* UI built atop JupyterLab.
 
--- a/README.md
+++ b/README.md
@@ -28,6 +28,13 @@ Anyone in the community can jump in and help with these activities at any time. 
 grant additional permissions (e.g., ability to merge PRs) to anyone who shows an on-going interest
 in working on the project.
 
+## Jupyter Notebook Deprecation Notice
+
+Following [Jupyter Notebook notice](https://github.com/jupyter/notebook#notice), we encourage users to transition to JupyterLab.
+This can be done by passing the environment variable `JUPYTER_ENABLE_LAB=yes` at container startup, 
+more information in the [documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html).
+This change is followed in the issue [#1217](https://github.com/jupyter/docker-stacks/issues/1217).
+
 ## Quick Start
 
 You can try a
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ more information is available in the [documentation](https://jupyter-docker-stac
 
 In April 2021 JupyterLab will become the default, however a new environment variable will be introduced to switch back to Jupyter Notebook if needed.
 
-Next, and according to the Jupyter Notebook project status and its compatibility with JupyterLab, it could be abandoned in favor of another *classic-like* UI.
+After the change of default, and according to the Jupyter Notebook project status and its compatibility with JupyterLab, these Docker images may remove the classic Jupyter Notebook interface altogether in favor of another *classic-like* UI built atop JupyterLab.
 
 This change is tracked in the issue [#1217](https://github.com/jupyter/docker-stacks/issues/1217), please check its content for more information.
 
--- a/README.md
+++ b/README.md
@@ -32,8 +32,13 @@ in working on the project.
 
 Following [Jupyter Notebook notice](https://github.com/jupyter/notebook#notice), we encourage users to transition to JupyterLab.
 This can be done by passing the environment variable `JUPYTER_ENABLE_LAB=yes` at container startup, 
-more information in the [documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html).
-This change is followed in the issue [#1217](https://github.com/jupyter/docker-stacks/issues/1217).
+more information is available in the [documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html).
+
+In April 2021 JupyterLab will become the default, however a new environment variable will be introduced to switch back to Jupyter Notebook if needed.
+
+Next, and according to the Jupyter Notebook project status and its compatibility with JupyterLab, it could be abandoned in favor of another *classic-like* UI.
+
+This change is tracked in the issue [#1217](https://github.com/jupyter/docker-stacks/issues/1217), please check its content for more information.
 
 ## Quick Start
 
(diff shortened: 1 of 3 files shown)


romainx:
@parente could you please share your opinion? I'd like to make sure I'm in the right direction before I go any further and spend more time updating the documentation. Thanks 😄

Bidek56:
JupyterLab may be popular but I feel it's a giant web app which sucks up a lot of browser resources, hence I stay away from.
Notebook is much closer to other platforms like Databricks or Colab.
If we do end up defaulting to it, please add some docs on how to un-default it. :) Thanks

jnsdrtlf:
I added some documentation on your personal fork @romainx.

I understand @Bidek56's concerns and I was also thinking that changing the default behavior might result in confusion. But if the old *notebook* frontend is actually going to be deprecated and to be replaced with [jupyterlab-classic](https://github.com/jtpio/jupyterlab-classic), this change will have to come anyway. It might be beneficial to separate some of the documentation for Docker images from before and after the JupyterLab 3.0 transition to avoid confusion.

romainx:
@jnsdrtlf thank you very much and sorry for not taking the time to look at your contribution. I will do it for sure 😄 . I am waiting for feedback before going any further because I do not have enough knowledge to identify the possible impacts of such a change.

jnsdrtlf:
@romainx no worries, this (and the transition to JupyterLab 3.0) is a huge change and should not be rushed. While the documentation clearly states that [you should always use a tagged version](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#versioning), I can imagine that people tend to use `latest` even in production environments. This could cause a lot of headaches.

base-notebook/start-notebook.sh
```diff
@@ -12,8 +12,12 @@ fi
 if [[ ! -z "${JUPYTERHUB_API_TOKEN}" ]]; then
     # launched by JupyterHub, use single-user entrypoint
     exec /usr/local/bin/start-singleuser.sh "$@"
-elif [[ ! -z "${JUPYTER_ENABLE_LAB}" ]]; then
-    . /usr/local/bin/start.sh $wrapper jupyter lab "$@"
-else
```
parente, line 17: Perhaps we should keep the default like it is for a period of time, but emit a deprecation warning that links to https://github.com/jupyter/notebook#notice and suggests the new default will become jupyterlab by some soft date X months out?

Maybe we should mirror that message into the root README of this repo too?

When we hit X months, we can then change the default with some assurance that we made a reasonable attempt at notifying users of the breaking change, above and beyond suggesting "don't use latest if you want stability".

romainx: 👍 Thanks @parente very good suggestion. So I will change the default (no variable) back to Jupyter Notebook and change the deprecation warning to add the link. I will also put this link into the main README and into the doc.
In addition to these suggestions, should we keep and document the variable `JUPYTER_ENABLE_NB`? It will be a way to prepare the next step?

parente: Sure though it might be a bit confusing to new users in the interim. What happens if enable lab and enable nb are both set?

romainx: In this case I will make JupyterLab win. I will add this test case in the test already written to be clear. The advantage of having it available as soon as possible is to permit to users to use it if they do not want to switch to JupyterLab during the next step.

romainx: You're right I will simply drop the `JUPYTER_ENABLE_NB` variable for now, because it's useless in this first phase. So I will update this PR to add the warning messages and adapt the test I have written.

romainx:
@parente I've implement the changes and now this deprecation stream is followed in a dedicated issue #1217.

README.md
```diff
@@ -28,6 +28,13 @@ Anyone in the community can jump in and help with these activities at any time.
 grant additional permissions (e.g., ability to merge PRs) to anyone who shows an on-going interest
 in working on the project.
 
+## Jupyter Notebook Deprecation Notice
+
+Following [Jupyter Notebook notice](https://github.com/jupyter/notebook#notice), we encourage users to transition to JupyterLab.
+This can be done by passing the environment variable `JUPYTER_ENABLE_LAB=yes` at container startup, 
+more information in the [documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html).
+This change is followed in the issue [#1217](https://github.com/jupyter/docker-stacks/issues/1217).
+
```
parente, line 37: Do we want to state a date after which lab will become the default? April or later maybe?

base-notebook/start-notebook.sh
```diff
@@ -15,5 +15,6 @@ if [[ ! -z "${JUPYTERHUB_API_TOKEN}" ]]; then
 elif [[ ! -z "${JUPYTER_ENABLE_LAB}" ]]; then
     . /usr/local/bin/start.sh $wrapper jupyter lab "$@"
 else
+    echo "WARN: Jupyter Notebook deprecation notice https://github.com/jupyter/notebook#notice."
```
parente, line 18: Should this point to the stacks readme section you added instead, where we have more detail about the flag and the link upstream?

README.md
romainx: Yes good idea, April sounds reasonable I think.

base-notebook/start-notebook.sh
romainx: Yes it's much better I will change It.

romainx: Changes done.

README.md
```diff
@@ -28,6 +28,18 @@ Anyone in the community can jump in and help with these activities at any time.
 grant additional permissions (e.g., ability to merge PRs) to anyone who shows an on-going interest
 in working on the project.
 
+## Jupyter Notebook Deprecation Notice
+
+Following [Jupyter Notebook notice](https://github.com/jupyter/notebook#notice), we encourage users to transition to JupyterLab.
+This can be done by passing the environment variable `JUPYTER_ENABLE_LAB=yes` at container startup, 
+more information is available in the [documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html).
```
mathbunnyru, line 35:
```suggestion
more information is available in the [documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#docker-options).
```

I think more specific link is better.

mathbunnyru (approved):
Good job keeping up with Jupyter changes, @romainx 👍

README.md
```diff
@@ -28,6 +28,18 @@ Anyone in the community can jump in and help with these activities at any time.
...
+
+In April 2021 JupyterLab will become the default, however a new environment variable will be introduced to switch back to Jupyter Notebook if needed.
```
parente, line 37:
```suggestion
In April 2021 JupyterLab will become the default for all of the Jupyter Docker stack images, however a new environment variable will be introduced to switch back to Jupyter Notebook if needed.
```

README.md
```diff
@@ -28,6 +28,18 @@ Anyone in the community can jump in and help with these activities at any time.
...
+
+Next, and according to the Jupyter Notebook project status and its compatibility with JupyterLab, it could be abandoned in favor of another *classic-like* UI.
```
parente, line 39:
```suggestion
After the change of default, and according to the Jupyter Notebook project status and its compatibility with JupyterLab, these Docker images may remove the classic Jupyter Notebook interface altogether in favor of another *classic-like* UI built atop JupyterLab.
```

parente (approved):
I left a couple more wording suggestions for the README as did @mathbunnyru. Thanks for working on this @romainx and being thoughtful about the community of people who use these images.

romainx:
@parente and @mathbunnyru thanks a lot for your help and advices. 
Really appreciate working with you guys ❤️.
I will merge it.

mathbunnyru:
> @parente and @mathbunnyru thanks a lot for your help and advices.
> Really appreciate working with you guys ❤️.
> I will merge it 🎉 .

Parce que vous êtes de Toulouse, je pense que vous parlez français bien et je viens de répondre en français.
Désolé pour les erreurs 😂
Merci à vous pour tous les efforts que vous donnes aux images 👍 
J'aime bien aussi travailler avec vous 🤗

romainx:
@mathbunnyru you're right and your french is better than my english 🤣 .
Thanks for the message.

README.md
```diff
@@ -28,6 +28,18 @@ Anyone in the community can jump in and help with these activities at any time.
 grant additional permissions (e.g., ability to merge PRs) to anyone who shows an on-going interest
 in working on the project.
 
+## Jupyter Notebook Deprecation Notice
+
+Following [Jupyter Notebook notice](https://github.com/jupyter/notebook#notice), we encourage users to transition to JupyterLab.
+This can be done by passing the environment variable `JUPYTER_ENABLE_LAB=yes` at container startup, 
+more information is available in the [documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#docker-options).
+
+In April 2021 JupyterLab will become the default for all of the Jupyter Docker stack images, however a new environment variable will be introduced to switch back to Jupyter Notebook if needed.
```
willingc, line 37: @parente I'm not sure that April 2021 gives enough time for folks to adjust to this change. It will likely break anyone who hasn't pinned to a SHA.

PR #1213, opened 2021-01-15
Install mamba and use it to replace for install, list and clean

### LINKED ISSUE #1210: Add mamba to base-notebook?
Thanks for maintaining these images!

Does it make sense to add [mamba](https://github.com/mamba-org/mamba) to the base-notebook image? https://github.com/jupyter/docker-stacks/blob/703d8b2dcb886be2fe5aa4660a48fbcef647e7aa/base-notebook/Dockerfile#L133-L137 
It speeds up installation of `conda` packages and `docker build` time significantly downstream, some rudimentary tests on scipy-notebook makes the build 2x faster using `mamba install` instead of `conda install`.

romainx:
First try to resolves #1210 

- Installation of [mamba][1] through the [Miniforge][2] installer -> efficient way to do it since already shipped in the installer 
- Use it in all `Dockerfile` to replace `conda` commands: `install`, `list` and `clean`.

> Currently, only install, create, list, search, run, info and clean are supported through mamba.

[1]: https://github.com/mamba-org/mamba
[2]: https://github.com/conda-forge/miniforge

### DIFF
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -22,20 +22,20 @@ RUN apt-get update && \
 USER $NB_UID
 
 # R packages
-RUN conda install --quiet --yes \
+RUN mamba install --quiet --yes \
     'r-base=4.0.3' \
     'r-ggplot2=3.3*' \
     'r-irkernel=1.1*' \
     'r-rcurl=1.98*' \
     'r-sparklyr=1.5*' \
     && \
-    conda clean --all -f -y && \
+    mamba clean --all -f -y && \
     fix-permissions "${CONDA_DIR}" && \
     fix-permissions "/home/${NB_USER}"
 
 # Spylon-kernel
-RUN conda install --quiet --yes 'spylon-kernel=0.4*' && \
-    conda clean --all -f -y && \
+RUN mamba install --quiet --yes 'spylon-kernel=0.4*' && \
+    mamba clean --all -f -y && \
     python -m spylon_kernel install --sys-prefix && \
     rm -rf "/home/${NB_USER}/.local" && \
     fix-permissions "${CONDA_DIR}" && \
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -26,19 +26,19 @@ USER root
 # Conda version
 ARG conda_version="4.9.2"
 # Miniforge installer patch version
-ARG miniforge_patch_number="0"
+ARG miniforge_patch_number="5"
 # Miniforge installer architecture
 ARG miniforge_arch="x86_64"
 # Python implementation to use 
 # can be either Miniforge3 to use Python or Miniforge-pypy3 to use PyPy
-ARG miniforge_python="Miniforge3"
+ARG miniforge_python="Mambaforge"
 
 # Miniforge archive to install
 ARG miniforge_version="${conda_version}-${miniforge_patch_number}"
 # Miniforge installer
 ARG miniforge_installer="${miniforge_python}-${miniforge_version}-Linux-${miniforge_arch}.sh"
 # Miniforge checksum
-ARG miniforge_checksum="6321775eb2c02d7f51d3a9004ce0be839099f126f4099c781531428536669560"
+ARG miniforge_checksum="7f0ad0c2f367751f7878d25a7bc1b4aa48b8dcea864daf9bc09acb595102368b"
 
 # Install all OS dependencies for notebook server that starts but lacks all
 # features (e.g., download as all possible file formats)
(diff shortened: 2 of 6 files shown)


mathbunnyru:
@romainx please, write a comparison table with size and build time here.

Actually, it would be great to see this comparison in all the PR. Or at least, not the comparison, but just numbers for the branch in PR.

romainx:
Ok for the size, for the duration I'm not sure because it depends on many things for example the download of external dependencies like Spark.

mathbunnyru:
Yes, you're right, but at the same time, if the mamba is as fast as promised, I think we still might see a big improvement in numbers.

kyri-petrou:
Hello everyone! I've been occasionally following the PRs of this repo for some time as I'm tracking the progress on the JupyterLab 3.x releases!

As a warning for using mamba, I've had quite mixed experiences in the past (using mamba 0.6.* and below). I've noticed the following downsides in environments using data-science heavy packages:

1. Mamba would some times create inconsistent environments, which would result in errors during runtime.
2. Noticeable performance loss in computation time when using environments created using mamba (~20-30%).

I believe that these issues might exist due to difference in behaviour of `mamba clean` from `conda clean` rather than the installation itself. I did not, however, look into it with greater detail due to various work commitments. Apologies for not being able to provide any more help!

Also, please note that these issues might no longer exist with newer versions of mamba, but it might be worth to perform some benchmarking prior to start shipping your images with packages installed using mamba as the default.

romainx:
Here are the results.

No major improvement on the build time that is around **33m**. The last two builds without mamba took the same time.

## Size

Comparison has been done with the build done in the previous PR (BuildKit), packages should be pretty much the same since they have occurred the same day. Results are matching our expectations ~ **+20MB** in the base notebook, it's the penalty coming with the biggest size of the installer shipped with mamba. Put in downstream images we have some surprising results. It seems that images containing R packages are smaller with mamba. It's weird 🤔. We have to deep dive to check if the installation of some packages are not dropped. 

If the anomaly on R packages is confirmed we could choose to install mamba for end users and keep installing the stack with conda since there is no improvement on build time. Waiting for confirmation, but maybe it's a safer option.

| image                  	| conda  	| mamba  	|
|------------------------	|--------	|--------	|
| `base-notebook`        	| 641MB  	| 673MB  	|
| `minimal-notebook`     	| 1.49GB 	| 1.53GB 	|
| `r-notebook`           	| 2.62GB 	| 2.62GB 	|
| `scipy-notebook`       	| 2.56GB 	| 2.67GB 	|
| `tensorflow-notebook`  	| 3.57GB 	| 3.67GB 	|
| `datascience-notebook` 	| 4.12GB 	| 3.94GB 	|
| `pyspark-notebook`     	| 3.24GB 	| 3.26GB 	|
| `all-spark-notebook`   	| 4.23GB 	| 3.95GB 	|

wolfv:
Hi all, thanks for running these experiments. I (as one of the core-mamba developers) would be quite interested to see where the differences in environment size come from! 

I think with images / environments this large, on consumer hardware, you will see that the download bandwidth is the limiting factor.

@kyri-petrou 
Regarding inconsistent environments that do not work: I really haven't seen that happen in quite a while. I'd be happy about any kind of reproducer.

And the 20-30% performance difference is really curious, but there are many reasons why that could have happened:

- MKL-numpy vs non-mkl numpy
- running the test when laptop battery plugged in vs not plugged in (different energy states)
- running the tests right after conda / mamba finishes. conda uses many more CPU cycles & ram, putting the CPU into a higher frequency

If the same packages are installed, the machine code that is running should be absolutely identical. If you can reproduce this I would be really interested in a reproducer.

Btw. the mybinder team is using mamba since a while and afaik had good success with it!

Thanks again, I'll try to look into the differences tomorrow.

romainx:
> Hi all, thanks for running these experiments. I (as one of the core-mamba developers) would be quite interested to see where the differences in environment size come from!

Thanks @wolfv it would be great.
Just guessing but maybe it could have something to do with R packages and / or `openblas` because I see less `openblas` hits in mamba logs than in conda logs.

https://github.com/jupyter/docker-stacks/blob/d979fa1b8c4affdcb44f1955a983e41129d1f3b2/scipy-notebook/Dockerfile#L20

Best.

maresb:
I'm extremely surprised by these results. I build my dev environment on top of `tensorflow-notebook` and `conda` is essentially unusable. (I have to wait about 5 minutes for `conda`'s solver, while `mamba`'s completes in about 10 seconds.)

In any case, I would very much favor including `mamba` in `base-notebook`. For me it has become an essential tool. In contrast I see `conda` as slow and flaky, even if more battle-tested.

romainx:
@maresb thanks for the feedback it's interesting. What we could do, is to keep building images with `conda` to avoid potential issues while providing `mamba` to end users. It could also be a safe starting point before going further. What is your opinion?

romainx:
I'm closing this PR to replace it by #1245 that implements what I have said in my last comment.
> keep building images with conda to avoid potential issues while providing mamba to end users. It could also be a safe starting point before going further.

PR #1222, opened 2021-01-22
JupyterLab 3.0 and other updates

### LINKED ISSUE #1167: Problem with `Restart kernel and run all cells` for R and Julia kernels in `datascience-notebook`?

### LINKED ISSUE #1207: JupyterLab 3.0
Hello,

JupyterLab `3.0.0` introduces major changes that could impact:

- the compatibility between Jupyter Notebook and JupyterLab,
- the way extensions are managed.

# Terminology

Very good status taken from https://github.com/nbgallery/docker-images/issues/17

> Some terminology, to avoid confusion:
> 
> * Jupyter back-end options:
>   
>   * [notebook](https://github.com/jupyter/notebook) server, which supports "serverextensions" (no space) that I'll refer to as old-server extensions
>   * [jupyter_server](https://github.com/jupyter-server/jupyter_server), which supports "server extensions" (space) that I'll refer to as new-server extensions
> * Jupyter front-end options:
>   
>   * [notebook](https://github.com/jupyter/notebook) - the traditional UI, often called "classic" -- but since I'm about to mention two other "classic" projects, I'll call it "Notebook UI" for clarity
>   * [lab](https://github.com/jupyterlab/jupyterlab) - the next-gen UI
>   * [jupyterlab-classic](https://github.com/jtpio/jupyterlab-classic) - the look and feel of the Notebook UI but built with Lab components
> * Compatibility layers:
>   
>   * [nbclassic](https://github.com/jupyterlab/nbclassic) - provides the Notebook UI as a new-server extension and also translates old-server extensions into new-server extensions.  (To be clear, this isn't just something that looks like the Notebook UI, it is literally the same code, running through the compatibility layer.)
> 

# Impact

## Jupyter Notebook and JupyterLab

[Jupyter Notebook][1] is deprecated and it is advised to transition to JupyterLab.

> Please note that this repository is currently maintained by a skeleton crew of maintainers from the Jupyter community. We encourage users to transition to JupyterLab, where more immediate support can occur.

See dedicated issue #1217.

## Extensions

According to the [JupyterLab changelog][2]

> Extensions can be installed without building JupyterLab with NodeJS
> In JupyterLab 3.0, a new recommended way of distributing and installing extensions as Python pip or conda packages is available. Installing such extensions does not require rebuilding JupyterLab and does not require having NodeJS installed. The previous way of distributing extensions as npm packages requiring rebuilding JupyterLab is still available as well. See the documentation for more details.

We should manage extension differently in `scipy-notebook`.

https://github.com/jupyter/docker-stacks/blob/d113a601dbb83dca72f0bc73c7e5f223426cc112/scipy-notebook/Dockerfile#L50-L59

[This issue][5] summarizes in a table the extensions currently compliant with the new version and here is the list of extensions used in `scipy-notebook`.

- [x] `ipywidgets`: https://github.com/jupyter-widgets/ipywidgets/pull/2973 `7.6`
- [x] `jupyter-matplotlib`: https://github.com/matplotlib/ipympl/pull/272 `0.6.1`
- [ ] `bokeh`: https://github.com/bokeh/bokeh/issues/10569

## JupyterHub

There is also something to check with JupyterHub since the stacks permit to launch `jupyterhub-singleuser`. It think it's well summarized in https://github.com/jupyterhub/jupyterhub/issues/2984 and this [comment](https://github.com/jupyterhub/jupyterhub/issues/2984#issuecomment-658177068).

Best

[1]: https://github.com/jupyter/notebook
[2]: https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html#v3-0
[3]: https://github.com/jupyter-server/jupyter_server
[4]: https://github.com/jupyterlab/nbclassic
[5]: https://github.com/jupyterlab/jupyterlab/issues/9461

romainx:
Upgrade to JupyterLab `3.0.x` as described in #1207 and other upgrades.

- `base-notebook`
  - `Changed`: Bump ubuntu
  - `Changed`: Bump `jupyterlab` -> Fixes #1207 and fixes #1167
- `scipy-notebook`
  - `Changed`: Bump `dask`
  - `Changed`: Bump `ipympl`
  - `Changed`: Bump `pandas`
  - `Changed`: Bump `scipy`
  - `Removed`: Jupyter Extension management through `npm` and `jupyter labextension`.
    It is not necessary anymore, and I have checked that both `ipywidgets` and `matplotlib` are working with JupyterLab and Jupyter. It speeds up the build time.
  - `Removed`: Skip extension test, see limitations below for detail
- `datascience-notebook`
  - `Changed`: Bump `rpy2`
- `tensorflow-notebook`
  - `Changed`: Bump `tensorflow`

Even with the limitations (Bokeh extension is the only regression) listed below, I think it is reasonable to merge it.

### Bokeh extension not available

Bokeh extension is not ready for JupyterLab `3.0.x` see https://github.com/bokeh/bokeh/issues/10569.

### Check extension

`jupyter labextension check` seems to not work anymore. We should create an issue in the upstream repo.

```bash
# Installation don't need to use node now
conda install --quiet --yes 'ipywidgets=7.6.*'

# Preparing transaction: ...working... done
# Verifying transaction: ...working... done
# Executing transaction: ...working... Enabling notebook extension jupyter-js-widgets/extension...
      - Validating: OK
      
# List extensions, everything is OK
jupyter labextension list

# JupyterLab v3.0.5
# /opt/conda/share/jupyter/labextensions
#        @jupyter-widgets/jupyterlab-manager v3.0.0 enabled OK (python, jupyterlab_widgets)

# However, the check does not work
jupyter labextension check @jupyter-widgets/jupyterlab-manager

# @jupyter-widgets/jupyterlab-manager: X
```

### Warning at start

There is a warning at the start that seems related to this configuration `c.ServerApp.open_browser = False` in `jupyter_server_config.py`, see https://github.com/jupyterlab/jupyterlab/issues/9360

```bash
# [W 2021-01-22 08:14:06.693 ServerApp] No web browser found: could not locate runnable browser.
```

### DIFF
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -22,43 +22,29 @@ RUN conda install --quiet --yes \
     'bottleneck=1.3.*' \
     'cloudpickle=1.6.*' \
     'cython=0.29.*' \
-    'dask=2020.12.*' \
+    'dask=2021.1.*' \
     'dill=0.3.*' \
     'h5py=3.1.*' \
     'ipywidgets=7.6.*' \
-    'ipympl=0.5.*'\
+    'ipympl=0.6.*'\
     'matplotlib-base=3.3.*' \
     'numba=0.52.*' \
     'numexpr=2.7.*' \
-    'pandas=1.1.*' \
+    'pandas=1.2.*' \
     'patsy=0.5.*' \
     'protobuf=3.14.*' \
     'pytables=3.6.*' \
     'scikit-image=0.18.*' \
     'scikit-learn=0.24.*' \
-    'scipy=1.5.*' \
+    'scipy=1.6.*' \
     'seaborn=0.11.*' \
     'sqlalchemy=1.3.*' \
     'statsmodels=0.12.*' \
     'sympy=1.7.*' \
     'vincent=0.4.*' \
     'widgetsnbextension=3.5.*'\
-    'xlrd=1.2.*' \
-    && \
+    'xlrd=2.0.*' && \
     conda clean --all -f -y && \
-    # Activate ipywidgets extension in the environment that runs the notebook server
-    jupyter nbextension enable --py widgetsnbextension --sys-prefix && \
-    # Also activate ipywidgets extension for JupyterLab
-    # Check this URL for most recent compatibilities
-    # https://github.com/jupyter-widgets/ipywidgets/tree/master/packages/jupyterlab-manager
-    jupyter labextension install @jupyter-widgets/jupyterlab-manager@^2.0.0 --no-build && \
-    jupyter labextension install @bokeh/jupyter_bokeh@^2.0.0 --no-build && \
-    jupyter labextension install jupyter-matplotlib@^0.7.2 --no-build && \
-    jupyter lab build -y && \
-    jupyter lab clean -y && \
-    npm cache clean --force && \
-    rm -rf "/home/${NB_USER}/.cache/yarn" && \
-    rm -rf "/home/${NB_USER}/.node-gyp" && \
     fix-permissions "${CONDA_DIR}" && \
     fix-permissions "/home/${NB_USER}"
 
--- a/scipy-notebook/test/test_extensions.py
+++ b/scipy-notebook/test/test_extensions.py
@@ -7,6 +7,7 @@ import pytest
 LOGGER = logging.getLogger(__name__)
 
 
+@pytest.mark.skip(reason="Not yet compliant with JupyterLab 3")
 @pytest.mark.parametrize(
     "extension",
     [
@@ -16,7 +17,13 @@ LOGGER = logging.getLogger(__name__)
     ],
 )
 def test_check_extension(container, extension):
-    """Basic check of each extension"""
+    """Basic check of each extension
+
+    The list of extensions can be obtained through this command
+
+    $ jupyter labextension list
+
+    """
     LOGGER.info(f"Checking the extension: {extension} ...")
     c = container.run(
         tty=True, command=["start.sh", "jupyter", "labextension", "check", extension]
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -4,7 +4,7 @@
 # Ubuntu 20.04 (focal)
 # https://hub.docker.com/_/ubuntu/?tab=tags&name=focal
 # OS/ARCH: linux/amd64
-ARG ROOT_CONTAINER=ubuntu:focal-20201106@sha256:4e4bc990609ed865e07afc8427c30ffdddca5153fd4e82c20d8f0783a291e241
+ARG ROOT_CONTAINER=ubuntu:focal-20210119@sha256:3093096ee188f8ff4531949b8f6115af4747ec1c58858c091c8cb4579c39cc4e
 
 ARG BASE_CONTAINER=$ROOT_CONTAINER
 FROM $BASE_CONTAINER
@@ -135,7 +135,7 @@ RUN conda install --quiet --yes \
     'jedi=0.17.2' \
     'notebook=6.2.0' \
     'jupyterhub=1.3.0' \
-    'jupyterlab=2.2.9' && \
+    'jupyterlab=3.0.5' && \
     conda clean --all -f -y && \
     npm cache clean --force && \
     jupyter notebook --generate-config && \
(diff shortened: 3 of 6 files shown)


scipy-notebook/Dockerfile
```diff
@@ -22,43 +22,29 @@ RUN conda install --quiet --yes \
...
-    jupyter labextension install @jupyter-widgets/jupyterlab-manager@^2.0.0 --no-build && \
-    jupyter labextension install @bokeh/jupyter_bokeh@^2.0.0 --no-build && \
-    jupyter labextension install jupyter-matplotlib@^0.7.2 --no-build && \
-    jupyter lab build -y && \
-    jupyter lab clean -y && \
-    npm cache clean --force && \
-    rm -rf "/home/${NB_USER}/.cache/yarn" && \
-    rm -rf "/home/${NB_USER}/.node-gyp" && \
     fix-permissions "${CONDA_DIR}" && \
     fix-permissions "/home/${NB_USER}"
```
mathbunnyru, line 49: Is this line needed after the change?
In the `r-notebook` we only run fix-permissions on `${CONDA_DIR}`.
It might be a bug in the r-notebook (and maybe other notebooks as well).

scipy-notebook/test/test_extensions.py
```diff
@@ -7,6 +7,7 @@
 LOGGER = logging.getLogger(__name__)
 
 
+@pytest.mark.skip(reason="Not yet compliant with JupyterLab 3")
```
mathbunnyru, line 10: All of three extensions do not work or only some of them?
It might be better to comment the ones which do not work, rather than all three of them.

mathbunnyru (approved):
I made some small notes, but I approve the change as it is.

scipy-notebook/test/test_extensions.py
romainx: In fact like I said in the limitations paragraph, only `bokeh` is not available and do not work. 
For the others it's the command `jupyter labextension check $extension` that returns an error code whatever the extension. I guess it's a bug in the `check` subcommand because `list` returns OK. However I do not want to recode the test. This is the reason why everything is skipped even if you're right if it had been coded with the `list` subcommand only `bokeh` should have been skipped.

```bash
jupyter labextension list
# JupyterLab v3.0.5
# /opt/conda/share/jupyter/labextensions
#         jupyter-matplotlib v0.8.2 enabled OK
#         @jupyter-widgets/jupyterlab-manager v3.0.0 enabled OK (python, jupyterlab_widgets)
```

mathbunnyru: Ok, thanks for the explanation.

mathbunnyru (approved):
💪

romainx:
As it has been reviewed by @mathbunnyru, I'm merging it. It's time to take the plunge. We will manage issues if any 🙄.

PR #1232, opened 2021-02-07
JLab clean instead of rm in base docker

s-rog:
Follow-up on: https://github.com/jupyter/docker-stacks/pull/1046#issuecomment-774636297


romainx (commented):
Thanks, perfect.

PR #1234, opened 2021-02-08
Add SensiML notebook, based on TensorFlow notebook

jublin:


romainx:
Hello @jublin ,


Sorry for the late answer and thanks for this PR. However we are sorry but we are not extending the core stack with new images, that takes instead the [community images path][1]. Here is a recent comment explaining this process.

> We've been requesting that all new image definitions take the [community images path][1], primarily because this repository is unwieldy and we're already finding it difficult to maintain the images that already exist here. It sounds like you already have a nice project setup for building the images. Would you be open to [submitting a pull request linking to it from the documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#community-stacks) instead of this PR?

Another solution is to integrate from this PR only the "recipe" part.
So if you agree, we could merge the recipe you wrote in `recipes.md` explaining how to build a `sensiml-notebook` image from the `tensorflow-notebook` image. If it's Ok for you, you can modify your PR to include only this description. 

Please keep us informed of what you want to do between both solutions.

Best.

[1]: https://jupyter-docker-stacks.readthedocs.io/en/latest/contributing/stacks.html

romainx (changes requested):
Hello @jublin,

Do you want to change your PR to keep only the `recipes.md` and I will merge it? In the other case I will close it for the reasons mentioned in my previous comment.

Many thanks

docs/using/recipes.md
```diff
@@ -528,3 +528,24 @@ USER $NB_UID
...
+
+# Install SensiML and enable notebook extensions.
+RUN pip install --quiet --no-cache-dir \
+    'sensiml' && \
+    jupyter nbextension enable --py widgetsnbextension --sys-prefix  && \
+    jupyter nbextension enable --py --sys-prefix bqplot  && \
+    jupyter nbextension enable --py --sys-prefix qgrid  && \
+    fix-permissions "${CONDA_DIR}" && \
+    fix-permissions "/home/${NB_USER}"
+
```
mathbunnyru, line 550:
```suggestion
```

mathbunnyru (changes requested):
@jublin are you still interested in merging this?

Please, take a look at the comment above and after the change proposed this will be a good recipe for the community.

If not, please, close the PR.

consideRatio:
Given previous comments, I'll go ahead and close this PR at this point.

Issue #1237, opened 2021-02-12
Images tagging not done during the latest build


romainx:
Hello,

Just noticed that during the last build only the `latest` tag was pushed to DockerHub on all images of the stack.
It has been done through [this build ](https://github.com/jupyter/docker-stacks/runs/1882175244?check_suite_focus=true).
It's weird since the wiki is up to date and the corresponding PR #1236 was only about bumping some packages versions.

Best

romainx:
In fact the last successfully tagged stack was done the **2021-02-08**, after this date only the latest tag is pushed to DockerHub. @mathbunnyru I do not understand what is going on, any idea on your side?

mathbunnyru:
I will take a look. I'm not familiar with the dockerhub maintenance, but I will try my best, having all the tags is important.

mathbunnyru:
Done :)

PR #1243, opened 2021-02-24
Push all tags in docker push command

mathbunnyru:
Fix https://github.com/jupyter/docker-stacks/issues/1237

There was a change in 20.10.0 docker:
https://docs.docker.com/engine/release-notes/#20100

`Add -a/--all-tags to docker push docker/cli#2220`

So, only the latest version is now pushed by default.
Since we don't fix the docker version and always use the latest one, it affected our setup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,7 +154,7 @@ pull/%: ## pull a jupyter image
 
 push/%: DARGS?=
 push/%: ## push all tags for a jupyter image
-	docker push $(DARGS) $(OWNER)/$(notdir $@)
+	docker push --all-tags $(DARGS) $(OWNER)/$(notdir $@)
 
 push-all: $(foreach I,$(ALL_IMAGES),push/$(I) ) ## push all tagged images
 


Makefile
```diff
@@ -154,7 +154,7 @@ pull/%: ## pull a jupyter image
 
 push/%: DARGS?=
 push/%: ## push all tags for a jupyter image
```
mathbunnyru, line 156: Now it perfectly matches the description :)

mathbunnyru:
I also personally think, that we might want to fix the software version we use to avoid issues like this.
At the same time, nothing bad happened this time.

Makefile
romainx: 😄

romainx (approved):
Great, nice catch. Congrats 👏

romainx:
@mathbunnyru pinning the version has pros and cons. I think we should keep it like this to use the latest version and fix problems when they happen.

mathbunnyru:
> @mathbunnyru pinning the version has pros and cons. I think we should keep it like this to use the latest version and fix problems when they happen.

Agree, until know everything was working quite well (or we don't know about problems :) )

mathbunnyru:
Seems to work fine now:
https://hub.docker.com/r/jupyter/datascience-notebook/tags?page=1&ordering=last_updated

romainx:
Yes I confirm thanks again, I would have had a hard time finding this issue.

PR #1245, opened 2021-02-27
Install mamba in base-notebook

### LINKED ISSUE #1210: Add mamba to base-notebook?

romainx:
Hello,

This PR replaces #1213 and resolves #1210: installation of [Mamba][1] in `base-notebook`.
This time Mamba is only installed and not used to install packages of the stack, so it will be available to end users as a _faster_ alternative to Conda if they want to install their own packages. It seems to be a feature required by several users.

- Installation: I have used a flavor of the [Miniforge][2] installer shipping `mamba` by default.
- Docs: Updated, maybe I missed some.
- Test: Added generic test `test_package_managers.py` for all package managers including `mamba`.
- Size: +23MB penalty on all images
```bash
REPOSITORY              TAG               IMAGE ID       CREATED          SIZE
jupyter/base-notebook   latest            bdb5390b61b5   24 minutes ago   704MB
jupyter/base-notebook   <none>            d5b486f8e692   2 days ago       681MB
```
- Mamba versioning: I wondered how to manage this point, I've chosen for now the simplest way
  - I have not coded something complex to manage Mamba version (pin, default vs. defined version). 
  - It will be automatically updated to the latest version because of the following line
https://github.com/jupyter/docker-stacks/blob/67b8fb91f950d0099b179f27b7c6eeccb3179a9f/base-notebook/Dockerfile#L120

Best

[1]: https://github.com/mamba-org/mamba
[2]: https://github.com/conda-forge/miniforge

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -26,24 +26,24 @@ USER root
 # Conda version
 ARG conda_version="4.9.2"
 # Miniforge installer patch version
-ARG miniforge_patch_number="5"
+ARG miniforge_patch_number="7"
 # Miniforge installer architecture
 ARG miniforge_arch="x86_64"
 # Python implementation to use 
 # can be either Miniforge3 to use Python or Miniforge-pypy3 to use PyPy
-ARG miniforge_python="Miniforge3"
+ARG miniforge_python="Mambaforge"
 
 # Miniforge archive to install
 ARG miniforge_version="${conda_version}-${miniforge_patch_number}"
 # Miniforge installer
 ARG miniforge_installer="${miniforge_python}-${miniforge_version}-Linux-${miniforge_arch}.sh"
 # Miniforge checksum
-ARG miniforge_checksum="49dddb3998550e40adc904dae55b0a2aeeb0bd9fc4306869cc4a600ec4b8b47c"
+ARG miniforge_checksum="5a827a62d98ba2217796a9dc7673380257ed7c161017565fba8ce785fb21a599"
 
 # Install all OS dependencies for notebook server that starts but lacks all
 # features (e.g., download as all possible file formats)
 ENV DEBIAN_FRONTEND noninteractive
-RUN apt-get update \
+RUN apt-get -q update \
  && apt-get install -yq --no-install-recommends \
     wget \
     ca-certificates \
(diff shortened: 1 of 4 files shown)


base-notebook/Dockerfile
```diff
@@ -26,24 +26,24 @@ USER root
 # Conda version
 ARG conda_version="4.9.2"
 # Miniforge installer patch version
-ARG miniforge_patch_number="5"
+ARG miniforge_patch_number="7"
 # Miniforge installer architecture
 ARG miniforge_arch="x86_64"
 # Python implementation to use 
 # can be either Miniforge3 to use Python or Miniforge-pypy3 to use PyPy
```
mathbunnyru, line 33: Let's update these lines to include Mambaforge info?

base-notebook/test/test_package_managers.py
```diff
@@ -0,0 +1,32 @@
...
+LOGGER = logging.getLogger(__name__)
+
+
+@pytest.mark.parametrize(
+    "package_manager, cmd",
+    [
+        ("apt", "--version"),
+        ("conda", "--version"),
+        ("mamba", "--version"),
+        ("npm", "--version"),
```
mathbunnyru, line 16: `pip` should be also included, I suppose.

mathbunnyru (changes requested):
Just some small notes.

base-notebook/test/test_package_managers.py
romainx: Yes off course 👍

romainx: Done

base-notebook/Dockerfile
romainx: Comment updated, good suggestion.

mathbunnyru (approved):
👍

romainx:
@mathbunnyru thank you for the review. Before merging it, I would like to have the opinion of @parente, he always has wise comments too.

maresb:
Thanks so much for this, I really appreciate it!!!

parente (approved):
No wisdom other than LGTM. Thanks @romainx 😄

Issue #1168, opened 2020-09-25
java.lang.UnsupportedOperationException using pandas in Spark


angelcervera:
**What docker image you are using?**
jupyter/all-spark-notebook

**What complete docker command do you run to launch the container (omitting sensitive values)?**
docker run -d -p 8888:8888 jupyter/all-spark-notebook

**What steps do you take once the container is running to reproduce the issue?**

1. Visit http://localhost:8888 with the token
2. Create a new `spylon-kernel` notebook.
3. Copy and paste the example from Spark documentation:
    ```python
    %%python
    from pyspark.sql.functions import pandas_udf
    df = spark.createDataFrame([(1, 21), (2, 30)], ("id", "age"))
    def filter_func(iterator):
        for pdf in iterator:
            yield pdf[pdf.id == 1]

    df.mapInPandas(filter_func, df.schema).show()
    ```
4. Enjoy the exception:
```text
Traceback (most recent call last):
  File "python cell", line 7, in <module>
  File "/usr/local/spark/python/pyspark/sql/dataframe.py", line 440, in show
    print(self._jdf.showString(n, 20, vertical))
  File "/usr/local/spark/python/lib/py4j-0.10.9-src.zip/py4j/java_gateway.py", line 1304, in __call__
    return_value = get_return_value(
  File "/usr/local/spark/python/pyspark/sql/utils.py", line 128, in deco
    return f(*a, **kw)
  File "/usr/local/spark/python/lib/py4j-0.10.9-src.zip/py4j/protocol.py", line 326, in get_return_value
    raise Py4JJavaError(
Py4JJavaError: An error occurred while calling o135.showString.
: org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 0.0 failed 1 times, most recent failure: Lost task 0.0 in stage 0.0 (TID 0, a9b99f9c2052, executor driver): java.lang.UnsupportedOperationException: sun.misc.Unsafe or java.nio.DirectByteBuffer.<init>(long, int) not available
	at io.netty.util.internal.PlatformDependent.directBuffer(PlatformDependent.java:490)
	at io.netty.buffer.NettyArrowBuf.getDirectBuffer(NettyArrowBuf.java:243)
	at io.netty.buffer.NettyArrowBuf.nioBuffer(NettyArrowBuf.java:233)
	at io.netty.buffer.ArrowBuf.nioBuffer(ArrowBuf.java:245)
[...]

PR #1198, opened 2020-12-13
Fix spark installation for Java 11 and Arrow

romainx:
Hello,

Fixes #1168.
It consists in applying the fix described in the [Spark downloading instructions](https://spark.apache.org/docs/latest/#downloading):

> For Java 11, `-Dio.netty.tryReflectionSetAccessible=true` is required additionally for Apache Arrow library. This prevents `java.lang.UnsupportedOperationException: sun.misc.Unsafe` or `java.nio.DirectByteBuffer.(long, int) not available` when Apache Arrow uses Netty internally.

The only problem of this solution is how it fits with other versions of Spark / Java that can be changed at build time through `ARGS`. However I have tested with Java 8 and tests are passing, Spark 2.x does not work with Python 3.8, see https://github.com/apache/spark/pull/26194, so I'm not sure it's worth going further on this compatibility point.

I've added a specific test to ensure that the issue is fixed.
Best

PR #1256, opened 2021-03-20
Fix spark config properties syntax in Pyspark notebook Docker image

tbluejeans:
Hello, 

Following the example properties in the spark default configuration, I believe the syntax is incorrect and might lead to these properties being ignored. 

The syntax problem only occured to me when trying to add additional properties following the syntax with inverted comas. I am not sure if my suggested changes actually fix anything, as I don't know the reason for these properties. If so, it still may worth merging to prevent others from having problems adding properties. 

Hope it helps!

### DIFF
--- a/pyspark-notebook/Dockerfile
+++ b/pyspark-notebook/Dockerfile
@@ -52,8 +52,8 @@ RUN ln -s "spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}" spark && \
 # Fix Spark installation for Java 11 and Apache Arrow library
 # see: https://github.com/apache/spark/pull/27356, https://spark.apache.org/docs/latest/#downloading
 RUN cp -p "$SPARK_HOME/conf/spark-defaults.conf.template" "$SPARK_HOME/conf/spark-defaults.conf" && \
-    echo 'spark.driver.extraJavaOptions="-Dio.netty.tryReflectionSetAccessible=true"' >> $SPARK_HOME/conf/spark-defaults.conf && \
-    echo 'spark.executor.extraJavaOptions="-Dio.netty.tryReflectionSetAccessible=true"' >> $SPARK_HOME/conf/spark-defaults.conf
+    echo 'spark.driver.extraJavaOptions -Dio.netty.tryReflectionSetAccessible=true' >> $SPARK_HOME/conf/spark-defaults.conf && \
+    echo 'spark.executor.extraJavaOptions -Dio.netty.tryReflectionSetAccessible=true' >> $SPARK_HOME/conf/spark-defaults.conf
 
 USER $NB_UID
 


romainx:
Hello @tbluejeans ,

Thanks, good remark, however I'm a bit lost. Here is some context.

## Previous fix

The issue #1168 was fixed in the PR #1198.
The behavior of the following code was not the same before and after.

```python
from pyspark.sql import SparkSession
from pyspark.sql.functions import pandas_udf

# Spark session & context
spark = SparkSession.builder.master('local').config("spark.sql.execution.arrow.pyspark.enabled", "true").getOrCreate()
df = spark.createDataFrame([(1, 21), (2, 30)], ("id", "age"))
def filter_func(iterator):
    for pdf in iterator:
        yield pdf[pdf.id == 1]

df.mapInPandas(filter_func, df.schema).show()
```

If we try to run this code before and after the fix as expected

- It does not work before the fix
- It works after the fix

```bash
# Before the fix -> does not work
docker run -it --rm -p 8888:8888 -e JUPYTER_ENABLE_LAB=yes jupyter/pyspark-notebook:a0a544e6dc6e

# After the fix -> works
docker run -it --rm -p 8888:8888 -e JUPYTER_ENABLE_LAB=yes jupyter/pyspark-notebook:d113a601dbb8
```

So the fix — even with the weird syntax — in the `spark-defaults.conf` was working.
The corresponding test has been added in the unit test suite.

https://github.com/jupyter/docker-stacks/blob/9fe5186aba962cfaf4ca9c88868de0b8eee98cc6/all-spark-notebook/test/test_spark_notebooks.py#L12-L15

## Now

**Now even with no  `spark-defaults.conf` it works**! I'm guessing it's related with the bump of Spark, Arrow or Java versions

- Spark: `3.0.1` -> `3.1.1`
- Arrow: `2.0.0 ` -> `3.0.0`
- Java: `11.0.9.1` -> `11.0.10`

However **I was unable to find the reference to a change log or release note**. In consequence, I'm in favor of finding this information before changing anything and if it's not necessary anymore to simply remove the fix made in `spark-defaults.conf`. So if you have any information, please do not hesitate to share it with us, we will greatly appreciate it.

## References

- https://stackoverflow.com/questions/62109276/errorjava-lang-unsupportedoperationexception-for-pyspark-pandas-udf-documenta

tbluejeans:
Hi @romainx , 

Thank you for your detailed response. I spent some time today trying to understand this myself and digged into some release notes of the referenced libraries. I can basically confirm all your points above. It seems your config is read despite the syntax and the lines added were necessary at the time to get it running with Arrow. 

Concerning the question which package update might have made these lines obsolete, I tried testing different package versions in isolation (Arrow, Spark, Java). The Spark version seems to be the relevant change. When using Spark version 3.0.2 , the exception is present without the additional properties being set. With version 3.1.1, I can also confirm that the lines added to the default properties are no longer required for this test case. The question remains why:

What I can see is that netty has been updated from version `4.1.47.Final` to `4.1.51.Final` but I cannot see anything in the release notes that explicitly references this issue. In the Spark release notes of 3.1.1, I can see that the arrow version was [updated](https://spark.apache.org/releases/spark-release-3-1-1.html). Still no explicit reference to the problem. 

**Summary**
As it is still mentioned in the Spark documentation to set these properties with Java 11 for Arrow support and you might want to stay downward-compatible to earlier Spark versions(<3.1.1), it is probably best to keep the additional properties until Spark removes it from the documentation. 
Interestingly, I was having problems with the syntax of the default configs used here. I had copied those lines from here to add my own properties in a new container image.  Here, I ran into problems which may be specific to those properties which I added. Therefore, I would prefer the syntax in this PR but as it seems to have no functional relevance, we can also just close this PR. I also tested the proposed syntax with the old Spark version and it works. 

A more general topic: I am working on a container image which includes the delta packages (https://delta.io/) by default. It is only a few lines required of sparkconfig required. Does it make sense to integrate this into this container image or is it a too specific usecase?

romainx:
@tbluejeans very clear investigation thank for your time, 

I think it is a reasonable choice.

> As it is still mentioned in the Spark documentation to set these properties with Java 11 for Arrow support and you might want to stay downward-compatible to earlier Spark versions(<3.1.1), it is probably best to keep the additional properties until Spark removes it from the documentation.

I also agree that your proposal is better, because it has a cleaner syntax.
So thanks again, I will merge it :rocket:.

For the second point.

> A more general topic: I am working on a container image which includes the delta packages (https://delta.io/) by default. It is only a few lines of spark config required. Does it make sense to integrate this into this container image or is it a too specific usecase?

The answer is that is too specific — we try to keep the images as small and maintainable as possible —, however it could be very interesting to describe it in a [recipe](https://jupyter-docker-stacks.readthedocs.io/en/latest/contributing/recipes.html).

Many thanks

romainx (approved):
See discussion.

Issue #1250, opened 2021-03-07
Enhance tagging by making it consistent and well documented.


romainx:
Hello,

## Make image tags consistent

The images will be tagged as described below

- `base-notebook`:
  - `SHA`: Commit short SHA
  -  `python-$VER`: Python version
  -  `notebook-$VER`: Jupyter Notebook version
  -  `lab-$VER`: JupyterLab version
  -  `hub-$VER`: JupyterHub version
 - `minimal-notebook`:
   - all upstream tags
 - `scipy-notebook`:
   - all upstream tags
 - `r-notebook`:
   - all upstream tags
   - `r-$VER`: R version
- `tensorflow-notebook`:
   - all upstream tags
   - `tensorflow-$VER`: Tensorflow version
- `datascience-notebook`:
  - all upstream tags
  - `julia-$VER`: Julia version
  - `r-$VER`: R version
- `pyspark-notebook`:
  - all upstream tags
  - `spark-$VER`: Spark version
  - `hadoop-$VER`: Hadoop version
  - `java-$VER`: Java version
- `allspark-notebook`:
  - all upstream tags
  - `r-$VER`: R version

It's related to #1140

## Documentation

As suggested in #823 by @ABCurado it could be interesting to talk about the build history and to put the link to the wiki in

- The `README.md`,
- The documentation.

We also need to describe the tags that are defined for each image.

Best.

Issue #1140, opened 2020-08-08
Refactor duplicative, opaque run_hook scripts


parente:
Follow-on work from #1132 ...

We wrote a set of [bash scripts for Docker Hub](https://docs.docker.com/docker-hub/builds/advanced/#custom-build-phase-hooks) to update the project wiki with build metadata. We stuck with bash, sed, and heredocs as we found use of Python and other languages unreliable in the Docker Hub build environment.

Now that we're running builds on GitHub Actions, we should make the entire post-build hook process more contributor friendly using Python. We can replace most if not all of the docker commands in the heredoc templates with calls using the Python docker client, which is already in the dev requirements for this project. We can also reduce code duplication by having a singular post build script and reasonable markdown templates for populating image specific values.

Issue #823, opened 2019-03-08
Track Python version used in the Docker images


rvernica:
**What docker image you are using?**

`jupyter/datascience-notebook`

**What do you expect to happen?**

I would be nice to have a way to know which Docker image is using which Python version.

Here is a use-case: Current image is using Python `3.7.1`. I'm using PyArrow from `pyarrow-0.9.0-cp36-cp36m-manylinux1_x86_64.whl `. So, I would need Python `3.6`. How can I tell which Docker image is using which Python version?

parente:
Understanding what's in each build is a common request. At the moment, the image tags are the prefix of the git SHA that triggered their build. This provides a link from binary to source, but lacks the exact manifest of packages and versions installed.

The Docker Cloud logs capture everything installed. I'm not sure if they're publicly available (https://cloud.docker.com/u/jupyter/repository/registry-1.docker.io/jupyter/base-notebook/builds/f4a7f579-5af0-4bbb-a587-79721afdd0ed) and they're not very convenient.

We talked about extending the image tag to capture key package version numbers in the past. I think of that as a partial solution since what counts as "key" varies from person to person. I don't think we can encode the version numbers of every package into the docker image tag.

We used to have a post-build hook that would update a wiki page here on GitHub with links to both the build and merge commit that triggered it. Maybe we can take that idea a step further by having a post-build processing step that scrapes the logs or runs commands (conda list, apt-get, etc.) to get a list of packages that are posted / committed back here and indexed by the image tag / git SHA?

rvernica:
I see. Makes sense. The Docker Cloud link did not work for me. 

Having a page that lists the version info of the packages in the image would be a good start. Then we can turn that around and have a page with various versions of a package and the image that contains them.

Still, the image tag could contain the version numbers of the core languages, like Python, R, and Scala, since they will dictate what versions of other packages might be available.

parente:
For tagging language versions in an image like `jupyter/all-spark-notebook`, do you envision it would look something like:

`jupyter/all-spark-notebook:7db1bd2a7511-py37-r31-sc211`?

I think for building, what I'll call here, *package manifests*, we could do something like:

1. In a Docker Cloud `post_build` hook, run `apt list --installed`, `conda list`, `julia -E 'import Pkg; Pkg.status()'`, and `R --silent -e 'installed.packages(.Library)[, c(1,3)]'`.
[...]

PR #1257, opened 2021-03-20
Apply tags from the python

mathbunnyru:
All run_hook scripts are replaced with python code both to tag images and create manifest files.

Implementation note: shellcode if needed is replaced with `plumbum` (it's super easy and easy in adoption) inside python.
So, we're isolating commands like `docker run` or `git log` in small functions which run shell under the hood.

`Tagger` - something, that can run in the image and give a tag (for example, python version)
`Manifest` - also runs in the image and gives a markdown file piece to create a manifest file.
These are 2 different types and have different interfaces (and the list of taggers differs a little bit from the list of manifests).

All images are located in the hierarchy and we inherit taggers and manifests (but still re-evaluate them to get the correct results for the particular image).

This PR aims to fix several issues:
Fix: https://github.com/jupyter/docker-stacks/issues/1250
Fix: https://github.com/jupyter/docker-stacks/issues/1140
Fix: https://github.com/jupyter/docker-stacks/issues/823

Update:
Added ubuntu version tagger to make it easy to update Ubuntu (no manual tagging anymore!):
https://github.com/jupyter/docker-stacks/pull/1257/commits/260fb7352308445ec510392d1b5c5a68711db6b1


Makefile
```diff
@@ -106,7 +106,7 @@ hook/%: ## run post-build hooks for an image
 	DOCKER_REPO="$(OWNER)/$(notdir $@)" \
 	IMAGE_NAME="$(OWNER)/$(notdir $@):latest" \
 	IMAGE_SHORT_NAME="$(notdir $@)" \
-	$(SHELL) $(notdir $@)/hooks/run_hook
```
mathbunnyru, line 109: I deleted this line to make it easier to run `make hook/base-notebook` for example.

tagging/apply_tags.py
```diff
@@ -0,0 +1,62 @@
...
+}
+
+
+def get_all_taggers(short_image_name):
+    taggers = []
+    while short_image_name is not None:
+        image_description = ALL_IMAGES[short_image_name]
+        taggers = image_description.taggers + taggers
+        short_image_name = image_description.parent_image
+    return taggers
```
mathbunnyru, line 44: So, we're inheriting all the taggers of the base images, but not the values of tags.

We are going to reexecute the taggers on image, because this will be helpful for manifest files (python packages are changing between different, so if we use taggers for manifest file creation, we have to reexecute the taggers).

tagging/tagger_interface.py
```diff
@@ -0,0 +1,13 @@
...
+
+class TaggerInterface:
+    """HooksInterface for all hooks common interface"""
+    @staticmethod
+    def tag_value(image):
+        raise NotImplementedError
+
+    @staticmethod
+    def tag_name():
+        raise NotImplementedError
```
mathbunnyru, line 13: This is just to make logs look good, and I think it's nice to use classes for Taggers instead of pure functions - this will give an ability to add manifests in these taggers without a lot of changes.

tagging/apply_tags.py
romainx: Nice!

tagging/taggers.py
```diff
@@ -0,0 +1,39 @@
...
+from tagger_interface import TaggerInterface
+from git_helper import GitHelper
+from plumbum.cmd import docker
+
+
+def _get_full_name(short_image_name, owner):
+    return f"{owner}/{short_image_name}:latest"
+
+
+def _get_program_version(short_image_name, owner, program):
```
romainx, line 12: For this I have drafted another convenient approach that is using `with` do run and remove a Docker image in order to be able to execute commands while it's running. It could be more efficient since the commands we are running are not altering the container. Here is an example.

```python
import docker
import logging

logger = logging.getLogger(__name__)

class DockerRunner:
    def __init__(self, image_name, docker_client=docker.from_env(), command="sleep infinity"):
        self.container = None
        self.image_name = image_name
        self.command = command
        self.docker_client = docker_client

    def __enter__(self):
        logger.info(f"Creating container for image {self.image_name} ...")
        self.container = self.docker_client.containers.run(
            image=self.image_name, command=self.command, detach=True,
        )
        logger.info(f"Container {self.container.name} created")
        return self.container

    def __exit__(self, exc_type, exc_value, traceback):
        logger.info(f"Removing container {self.container.name} ...")
        if self.container:
            self.container.remove(force=True)
            logger.info(f"Container {self.container.name} removed")
```

Now we can use it just like this.

```python
def test_echo(text="test"):
    cmd = f"echo {text}"
    logger.info(f"Checking if echo command is working ...")
    with DockerRunner("ubuntu") as cont:
        out = cont.exec_run(cmd)
        # another command on the same running container
    assert out.exit_code == 0, f"Command: {cmd} failed"
    result = out.output.decode("utf-8")
    logger.info(f"Command result: {result}")
    assert text in result, f"Command result not expected: {result}"
```

romainx (commented):
Hello,

Very good work I think it's the right direction. I can work with you on this if you need. I've not checked in depth since it's a work in progress, however I left a suggestion on a way to run the commands in a Docker container.

Please tell me if I can do something or if you want me to check a specific point.
Many thanks!

mathbunnyru:
> Hello,
> 
> Very good work I think it's the right direction. I can work with you on this if you need. I've not checked in depth since it's a work in progress, however I left a suggestion on a way to run the commands in a Docker container.
> 
> Please tell me if I can do something or if you want me to check a specific point.
> Many thanks!

Thanks!
For now, I know what to do and how to do it :)
If I'm stuck and/or need a piece of advise, I will definitely ask you to help 👍

> For this I have drafted another convenient approach that is using with do run and remove a Docker image in order to be able to execute commands while it's running. It could be more efficient since the commands we are running are not altering the container. Here is an example.

If you don't mind, I will use your code in my PR :)

romainx:
@mathbunnyru 👍 
> If you don't mind, I will use your code in my PR :)
Sure it was the base I intended to use for the same need. So very happy to see that you consider reusing it 😊

tagging/taggers.py
mathbunnyru: Done

mathbunnyru:
Some things left to do:

- add all the images and all the taggers (easy, some copy-paste is needed)
- delete `docker tag` command in the run_hook (easy)
- actually, I think I will have to implement manifests and wiki updates in this PR, because all the variables used for tagging are used in index and SHA is used even in manifest file, and without doing this we have to keep 99% of run_hook script (difficult)
- do some testing (medium, I think testing common things like `DockerRunner` would be enough for now)

romainx:
@mathbunnyru for the generation of the manifest it could be a good idea to use Jinja templates. They will be populated with the values extracted from the images.

tagging/tagger_interface.py
mathbunnyru: Decided to delete this and use `__name__` property - still looks good and much less boilerplate code.

mathbunnyru:
> @mathbunnyru for the generation of the manifest it could be a good idea to use Jinja templates. They will be populated with the values extracted from the images.

I'll take a look, thanks!

base-notebook/hooks/run_hook
```diff
@@ -1,17 +1,11 @@
...
-# Apply tags
 GIT_SHA_TAG=${GITHUB_SHA:0:12}
-docker tag $IMAGE_NAME "$DOCKER_REPO:$GIT_SHA_TAG"
 PY_VERSION_TAG="python-$(docker run --rm ${IMAGE_NAME} python --version 2>&1 | awk '{print $2}')"
-docker tag $IMAGE_NAME "$DOCKER_REPO:$PY_VERSION_TAG"
 NB_VERSION_TAG="notebook-$(docker run --rm -a STDOUT ${IMAGE_NAME} jupyter-notebook --version | tr -d '\r')"
-docker tag $IMAGE_NAME "$DOCKER_REPO:${NB_VERSION_TAG%% }"
 LAB_VERSION_TAG="lab-$(docker run --rm -a STDOUT ${IMAGE_NAME} jupyter-lab --version | tr -d '\r')"
-docker tag $IMAGE_NAME "$DOCKER_REPO:${LAB_VERSION_TAG%%\r}"
 HUB_VERSION_TAG="hub-$(docker run --rm -a STDOUT ${IMAGE_NAME} jupyterhub --version | tr -d '\r')"
```
mathbunnyru, line 8: These variables are used below, that's why I can't easily get rid of them, so I'll have to deal with manifests in this PR.

tagging/images_hierarchy.py
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+from dataclasses import dataclass, field
+from typing import Optional, List
+from taggers import TaggerInterface, \
+    SHATagger, \
+    PythonVersionTagger, \
+    JupyterNotebookVersionTagger, JupyterLabVersionTagger, JupyterHubVersionTagger, \
+    RVersionTagger, TensorflowVersionTagger, JuliaVersionTagger, \
+    SparkVersionTagger, HadoopVersionTagger, JavaVersionTagger
```
mathbunnyru, line 10: I haven't yet checked that these values are correct, just added some copy-paste code to make it look like the final version would look like.

I will check the values.

tagging/apply_tags.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def apply_tags(short_image_name, owner):
+    logger.info(f"Applying tags for image: {short_image_name}")
+    taggers = get_all_taggers(short_image_name)
+
+    image = f"{owner}/{short_image_name}:latest"
+
+    with DockerRunner(image) as container:
```
mathbunnyru, line 29: @romainx this allowed me to reuse the image created so right now I'm only creating container once and then reusing it to calculate tag values.

tagging/images_hierarchy.py
mathbunnyru: Should work now

romainx:
@mathbunnyru are you good with this PR, do you want me to have a look now?

mathbunnyru:
> @mathbunnyru are you good with this PR, do you want me to have a look now?

I'm going on vacation for 2.5 weeks and I won't be able to work during that magnificent time😂

I will continue when I return. If you want to, feel free to continue the work I've done so far 👍

romainx:
> I will continue when I return. If you want to, feel free to continue the work I've done so far 👍

No urge, it will wait for your return 😅 

> I'm going on vacation for 2.5 weeks and I won't be able to work during that magnificent time😂

Enjoy your vacation even in this specific context :smile:

mathbunnyru:
I added ubuntu version tagger to make it easy to update Ubuntu (no manual tagging anymore!):
https://github.com/jupyter/docker-stacks/pull/1257/commits/260fb7352308445ec510392d1b5c5a68711db6b1

mathbunnyru:
I'm going to address the manifest file creation in the upcoming week.

tagging/create_manifests.py
```diff
@@ -0,0 +1,69 @@
...
+        logger.info(f"Using manifests: {manifest_names}")
+
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.DEBUG)
+
+    arg_parser = argparse.ArgumentParser()
+    arg_parser.add_argument("--short-image-name", required=True, help="Short image name to apply tags for")
+    arg_parser.add_argument("--owner", required=True, help="Owner of the image")
+    arg_parser.add_argument("--wiki-path", required=True, help="Path to the wiki pages")
```
mathbunnyru, line 64: This allows super-easy local testing.
I just downloaded the existing wiki git to some folder and checked that the files are changed in the way I wanted them to change.

mathbunnyru:
What I've done for today - [the homepage](https://github.com/jupyter/docker-stacks/wiki) is now updated using Python.

You can see that even the new Ubuntu version tag is now in the list.

Screenshot 2021-04-24 at 20 35 17

mathbunnyru:
The idea I used for taggers is working really well for the manifests.

So, we have an image hierarchy and we will collect all the manifests for the given image.
And we collect not values, but the classes, so we will reevaluate these classes on an image (to get the exact parameters of the particular image).
Each manifest evaluation is gonna just give a valid piece of markdown file.
We will just concatenate these pieces all together and it will create a manifest file like this (https://github.com/jupyter/docker-stacks/wiki/all-spark-notebook-584f43f06586).

So, what is left - to actually write these manifests classes and check that they work.
AFAIK it is not checked in the CI, but now I know how to test it locally without any problems.
After that all `run_hook` files could be completely deleted.

tagging/create_manifests.py
```diff
@@ -0,0 +1,69 @@
...
+MARKDOWN_NEWLINE = "<br />"
+
+
+def append_build_history_line(short_image_name, owner, wiki_path, all_tags, container):
+    date_column = f"`{BUILD_TIMESTAMP}`"
+    image_column = MARKDOWN_NEWLINE.join(
+        f"`{owner}/{short_image_name}:{tag_value}`" for tag_value in all_tags
+    )
+    commit_sha_tag = SHATagger.tag_value(container)  # first 12 letters of commit hash
+    commit_hash = GitHelper.commit_hash()  # full commit hash
```
mathbunnyru, line 27: I have no idea why we use only 12 letters for tagging.
It makes the code a bit more difficult.

But, for now, I don't want to change the behavior (it would be too difficult, but in the future, it might be easier to use full hash as a tag).

mathbunnyru:
Hope I'll have time tomorrow to finish this :)

tagging/images_hierarchy.py
```diff
@@ -0,0 +1,60 @@
...
+        parent_image="minimal-notebook",
+        taggers=[RVersionTagger]
+    ),
+    "tensorflow-notebook": ImageDescription(
+        parent_image="scipy-notebook",
+        taggers=[TensorflowVersionTagger]
+    ),
+    "datascience-notebook": ImageDescription(
+        parent_image="scipy-notebook",
+        taggers=[JuliaVersionTagger]
```
mathbunnyru, line 50: RVersionTagger should also be used here.

mathbunnyru:
I updated the PR description to match the PR.

tagging/manifests.py
```diff
@@ -0,0 +1,108 @@
...
+        ])
+
+
+class SparkInfoManifest(ManifestInterface):
+    @staticmethod
+    def markdown_piece(container) -> str:
+        return "\n".join([
+            "## Apache Spark",
+            "",
+            quoted_output(container, "/usr/local/spark/bin/spark-submit --version"),
```
mathbunnyru, line 107: I hardcoded the path here and changed a little bit how this is run (no `bash -c`).
It works fine, so I don't think we'll have troubles with this.

mathbunnyru:
@parente @romainx I'm done here

What's gonna change (from the markdown/tags point of view):

- tagging and manifests are consistent (if you add something to some parent image, you will also get it for all dependent images). 
- The order of sections in the build manifest is not controllable (I don't think that's big trouble)
- I added the header to the build manifests - it looks better
- Ubuntu version tag

From the code point of view:

- no more bash / sed / tr (no single line, run_hook scripts are deleted, `plumbum` is used just in a few places)
- code is reused, almost zero copy-paste
- easy to test locally (no magic with env variables at all)
- DockerRunner (once again thanks @romainx) class for easy docker handling and `run_simple_command` function to run something inside docker

Please, review my changes.

Makefile
mathbunnyru: Not needed anymore

romainx:
@mathbunnyru sounds great, very good job. Let me some time to check it since I would want to take the time to review it closely. I'm not sure to be able to do it during this week, but I will try to do it at the latest next WE. I hope it's Ok for you.

mathbunnyru:
> @mathbunnyru sounds great, very good job. Let me some time to check it since I would want to take the time to review it closely. I'm not sure to be able to do it during this week, but I will try to do it at the latest next WE. I hope it's Ok for you.

Sure, take your time 👍  I understand that this change is quite big, so it will take more time.

Makefile
```diff
@@ -98,15 +98,10 @@ git-commit: ## commit outstading git changes and push to remote
...
-hook/%: export WIKI_PATH?=../wiki
+hook/%: WIKI_PATH?=../wiki
 hook/%: ## run post-build hooks for an image
-	BUILD_TIMESTAMP="$$(date -u +%FT%TZ)" \
-	DOCKER_REPO="$(OWNER)/$(notdir $@)" \
-	IMAGE_NAME="$(OWNER)/$(notdir $@):latest" \
-	IMAGE_SHORT_NAME="$(notdir $@)" \
-	$(SHELL) $(notdir $@)/hooks/run_hook
+	./tagging/tag_image.py --short-image-name "$(notdir $@)" --owner "$(OWNER)" && \
+	./tagging/create_manifests.py --short-image-name "$(notdir $@)" --owner "$(OWNER)" --wiki-path "$(WIKI_PATH)"
```
romainx, line 104: It would be cleaner to use this tool as a module. To turn it into a module just need to add an `__init__.py` and to turn import into relative imports for example `from docker_runner` -> `from .docker_runner`.

```suggestion
	python -m tagging.tag_image --short-image-name "$(notdir $@)" --owner "$(OWNER)" && \
	python -m tagging.create_manifests --short-image-name "$(notdir $@)" --owner "$(OWNER)" --wiki-path "$(WIKI_PATH)""$(WIKI_PATH)"
```

tagging/create_manifests.py
```diff
@@ -0,0 +1,84 @@
...
+    image = f"{owner}/{short_image_name}:latest"
+
+    with DockerRunner(image) as container:
+        all_tags = [tagger.tag_value(container) for tagger in taggers]
+        append_build_history_line(short_image_name, owner, wiki_path, all_tags)
+        create_manifest_file(short_image_name, owner, wiki_path, manifests, container)
+
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.DEBUG)
```
romainx, line 74:
```suggestion
    logging.basicConfig(level=logging.INFO)
```

romainx (changes requested):
Hello,

You've done a great job, it's a big improvement. Congrats!
I tested it locally and everything works fine.

The objective was not to review each single method however things are well separated in distinct modules. The maintenance -- if any -- will be super easy.

I left a couple of minor suggestions.
Tell me if it's OK for you.

Thanks again.

Makefile
mathbunnyru: I applied your suggestion.
The only difference is that I wrote `python3`. It won't work on my mac with simple python.

mathbunnyru: @romainx now I can't run my script as usual.
Is there a way to have both things?

```
~/D/o/d/tagging > ./tag_image.py --short-image-name datascience-notebook --owner jupyter
  File "/Users/asalikhov/Documents/oss_repos/docker-stacks/tagging/./tag_image.py", line 7, in <module>
    from .docker_runner import DockerRunner
ImportError: attempted relative import with no known parent package
```

romainx: @mathbunnyru unfortunately I don't think it is possible to mix module and script at least with a relative import.

romainx (approved):
I think we are good now 🎉 
@mathbunnyru are we waiting for @parente review to merge it?

Makefile
mathbunnyru: Ok, that's a bit unfortunate.
Could you please explain me why the module approach is better here?
Sorry, I don't know well the python packaging model.

mathbunnyru:
> I think we are good now 🎉
> @mathbunnyru are we waiting for @parente review to merge it?

What would you suggest?
I prefer to have this merged, because now I have some time to fix the problems if there will be any.
But I'm not against waiting for a review as well :)

consideRatio:
@mathbunnyru woooow nice work on this!!! :heart:
@romainx I'm very thankful for your thorough maintainer work on this repo as well, consistently helping PRs get merged etc!!! :heart:

:tada: :heart: :sunflower: :party:

### Documentation

I looked through these changes and they make sense to me given what I understand. I didn't have a great way to onboard myself about this though so I suggest a very brief README.md file is added to the tagging folder.

I think that this README.md file would describe:
- what the source code in that folder purposed for
- what a Tagger class does
- what a Manifest class does
- what a Runner class does
- a code snippet of how to use it for its purpose

In addition to this it would be nice to have a brief docstring for classes or python files as well describing their role.

### Technical

I saw `run_simple_command` defined outside DockerRunner, but it invokes `exec_run` on the container which is a function I think is defined as part of the `docker` library. Due to that, would it make sense for that to be part of the DockerRunner rather than defined outside the DockerRunner class?

mathbunnyru:
Thanks for your kind words @consideRatio ❤️ 

All your suggestions make sense, so I'm going to try to implement them in a few days.
Though the documentation will need to be thoroughly checked - I'm not a native speaker, but I'll try my best.

mathbunnyru:
@consideRatio I fixed the technical part.

I also added `README.md` briefly describing all the processes.
Please, check the wording, it might be a bit awkward.

If you don't mind, I don't want to add brief docstrings, because `README.md` should be enough to get started.

tagging/README.md
```diff
@@ -0,0 +1,115 @@
+# Docker stacks tagging and manifest creation
+
+The main purpose of the source code in this folder is to properly tag all the images and to update [build manifests](https://github.com/jupyter/docker-stacks/wiki).
+These 2 processes are closely related, so the source code is widely reused.
```
consideRatio, line 4:
```suggestion
These two processes are closely related, so the source code is widely reused.
```

tagging/README.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+- `quoted_output` simply runs the command inside container using `DockerRunner.run_simple_command` and wraps it to triple quotes to create a valid markdown piece of file.
+- `manifests.py` contains all the manifests.
+- `create_manifests.py` is a python executable which is used to create the build manifest for an image.
+
+### Images Hierarchy
+
+Hierarchy of all the images, taggers and manifests is defined in `images_hierarchy.py` file.
+`get_taggers_and_manifests.py` file helps to get all the taggers and manifests for a specific image.
```
consideRatio, line 115:
```suggestion
All images dependencies on each other and what taggers and manifest they make use of is defined in `images_hierarchy.py`.

`get_taggers_and_manifests.py` defines a helper function to get the taggers and manifests for a specific image.
```

I wasn't confident about the meaning of hierarchy so I rephrased the sentence to be a bit explicit about that.

consideRatio (approved):
Wooooooooow thank you soo much for writing up this README.md file, it matters A LOT to me trying to learn and understand this PR and is now an amazing resource for any future maintenance work!

:heart: I didn't see anything crucial to fix but gave a suggestion or two.

consideRatio:
@mathbunnyru this involves the publishing of images etc, is there something that we should update in hub.docker.com or the CI/CD system etc to your knowledge?

If not, I figure the next step is to have this merged and follow up that it works all the way with the CI/CD automation we have to publish new versions of the images.

mathbunnyru:
No, everything will work in a fully automatic way.

The images are tagged using GitHub Actions and I almost didn't change this part (because it works well). Only deleted unused variables.
And this step is check on every commit in PR.

Only pushing (for images and docs) is done in `master`.
And this part didn't change at all.

https://github.com/jupyter/docker-stacks/blob/master/.github/workflows/docker.yml#L53-L57

romainx:
@mathbunnyru thanks I will merge it and we will figure out if everything is working.

mathbunnyru:
Great. I think everything should be ok, but I'm online now (for the case when something goes wrong)

romainx:
Same here 😄

mathbunnyru:
So, the tagging worked fine.

Examples:
https://hub.docker.com/r/jupyter/base-notebook/tags?page=1&ordering=last_updated
https://hub.docker.com/r/jupyter/all-spark-notebook/tags?page=1&ordering=last_updated

All the tags are there.

But pushing the docs didn't work well.
And it's not my fault, because there was the same problem in the previous PR.
I thought it was spurious, but it seems that there is a problem with our setup.

I will create a dedicated issue that should be fixed ASAP.

romainx:
Right even manifests seems to be correctly generated.

mathbunnyru:
I created an issue with the full description: https://github.com/jupyter/docker-stacks/issues/1280

romainx:
@mathbunnyru congrats and thank you for this work it's a great improvement!

consideRatio:
@mathbunnyru wieeeee very happy to see this work! Thanks for going the extra mile and not only providing a great improvement on how this is done, but also documenting it to help anyone in the future contribute to its maintenance better!

:heart: :tada: :+1: :sunflower: :100:

PR #1258, opened 2021-03-20
Add MySQL connector jar to pyspark Dockerfile to be able to connect to host MySQL instance

tmadar:
# Description

I want to be able to connect pyspark to the host instance of MySQL without having to mount a mysql-connector jar on container start

# Steps to Reproduce

1. Create a `SparkSession`

```python
from pyspark.sql import SparkSession

spark = SparkSession\
    .builder\
    .appName("Test App")\
    .getOrCreate()
```
2. Attempt to connect to your host machine's MySQL instance

```python
actor_frame = spark.read.format("jdbc") \
    .option("url", "jdbc:mysql://host.docker.internal:3306/sakila") \
    .option("driver", "com.mysql.cj.jdbc.Driver") \
    .option("dbtable", "actor") \ # actor is a table name
    .option("user", "root") \ # mysql username here
    .option("password", "") \ # mysql password here
    .load()
```

3. Attempt to load the frame created from step 2

```python
actor_frame.show()
```

# Screenshots

Error
Screen Shot 2021-03-20 at 11 41 42 AM

Success
Screen Shot 2021-03-20 at 11 53 16 AM

### DIFF
--- a/pyspark-notebook/Dockerfile
+++ b/pyspark-notebook/Dockerfile
@@ -17,6 +17,7 @@ ARG spark_version="3.1.1"
 ARG hadoop_version="3.2"
 ARG spark_checksum="E90B31E58F6D95A42900BA4D288261D71F6C19FA39C1CB71862B792D1B5564941A320227F6AB0E09D946F16B8C1969ED2DEA2A369EC8F9D2D7099189234DE1BE"
 ARG openjdk_version="11"
+ARG mysql_connector_version="8.0.23"
 
 ENV APACHE_SPARK_VERSION="${spark_version}" \
     HADOOP_VERSION="${hadoop_version}"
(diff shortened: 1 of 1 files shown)


romainx (changes requested):
@tmadar very sorry for the late answer :sweat_smile:.

Thanks to share this, however I'm afraid we won't merge it, because we try to keep the images as small and maintainable as possible and we had only very obvious packages for the corresponding stack. Even if it's interesting and helpful for others, I'm not sure it's worth adding it by default in the `pyspark-notebook` image. 

But it could be very interesting to describe it in a [recipe](https://jupyter-docker-stacks.readthedocs.io/en/latest/contributing/recipes.html). If you want to turn your code into a recipe, I will be happy to merge it.

Thanks a lot.

tmadar:
> @tmadar very sorry for the late answer 😅.
> 
> Thanks to share this, however I'm afraid we won't merge it, because we try to keep the images as small and maintainable as possible and we had only very obvious packages for the corresponding stack. Even if it's interesting and helpful for others, I'm not sure it's worth adding it by default in the `pyspark-notebook` image.
> 
> But it could be very interesting to describe it in a [recipe](https://jupyter-docker-stacks.readthedocs.io/en/latest/contributing/recipes.html). If you want to turn your code into a recipe, I will be happy to merge it.
> 
> Thanks a lot.

Thanks for taking a look at this. I will close this pull request and open up a new one with a recipe

PR #1259, opened 2021-03-21
Added recipe for Delta lake

tbluejeans:
Hi everyone!

As more and more people are using Databricks and Delta Lake engine, this recipe may be useful for others. 

Best!

### DIFF
--- a/docs/using/recipes.md
+++ b/docs/using/recipes.md
@@ -538,7 +538,7 @@ FROM jupyter/pyspark-notebook:latest
 USER root
 
 RUN echo 'spark.jars.packages io.delta:delta-core_2.12:0.8.0' >> $SPARK_HOME/conf/spark-defaults.conf && \
- echo 'spark.sql.extensions io.delta.sql.DeltaSparkSessionExtension' >> $SPARK_HOME/conf/spark-defaults.conf &&\
+    echo 'spark.sql.extensions io.delta.sql.DeltaSparkSessionExtension' >> $SPARK_HOME/conf/spark-defaults.conf && \
  echo 'spark.sql.catalog.spark_catalog org.apache.spark.sql.delta.catalog.DeltaCatalog' >> $SPARK_HOME/conf/spark-defaults.conf
 
 USER $NB_UID
--- a/docs/using/recipes.md
+++ b/docs/using/recipes.md
@@ -528,3 +528,18 @@ USER $NB_UID
 
 RUN pip install --quiet --no-cache-dir auto-sklearn
 ```
+
+## Enable Delta Lake in Spark notebooks
+Please note that the Delta lake packages are only available for Spark version > 3.0. By adding the properties to `spark-defaults.conf`, the user no longer needs to enable Delta support in each notebook. 
+
+```dockerfile
+FROM jupyter/pyspark-notebook:latest
+
+USER root
+RUN echo 'spark.jars.packages io.delta:delta-core_2.12:0.8.0' >> $SPARK_HOME/conf/spark-defaults.conf && \
+ echo 'spark.sql.extensions io.delta.sql.DeltaSparkSessionExtension' >> $SPARK_HOME/conf/spark-defaults.conf &&\
+ echo 'spark.sql.catalog.spark_catalog org.apache.spark.sql.delta.catalog.DeltaCatalog' >> $SPARK_HOME/conf/spark-defaults.conf
+
+USER $NB_UID
+
+```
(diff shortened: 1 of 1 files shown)


joshuacook:
Would be good to trigger the pull of these jars as part of the image build. Otherwise the pull will happen with the first spark session created in the container. This could add minutes to the time before the spark session is available on each container.

docs/using/recipes.md
```diff
@@ -528,3 +528,18 @@ USER $NB_UID
...
+## Enable Delta Lake in Spark notebooks
+Please note that the Delta lake packages are only available for Spark version > 3.0. By adding the properties to `spark-defaults.conf`, the user no longer needs to enable Delta support in each notebook. 
+
+```dockerfile
+FROM jupyter/pyspark-notebook:latest
+
+USER root
+
+RUN echo 'spark.jars.packages io.delta:delta-core_2.12:0.8.0' >> $SPARK_HOME/conf/spark-defaults.conf && \
+ echo 'spark.sql.extensions io.delta.sql.DeltaSparkSessionExtension' >> $SPARK_HOME/conf/spark-defaults.conf &&\
```
romainx, line 541: Indentation and end of line to be fixed `&& \`
```suggestion
    echo 'spark.sql.extensions io.delta.sql.DeltaSparkSessionExtension' >> $SPARK_HOME/conf/spark-defaults.conf && \
```

docs/using/recipes.md
```diff
@@ -528,3 +528,18 @@ USER $NB_UID
...
+ echo 'spark.sql.catalog.spark_catalog org.apache.spark.sql.delta.catalog.DeltaCatalog' >> $SPARK_HOME/conf/spark-defaults.conf
+
+USER $NB_UID
```
romainx, line 544: As suggested by @joshuacook it's a way to trigger the download of the jars at build time.

```suggestion
USER $NB_UID

# Run pyspark and exit to trigger the download of the delta lake jars
RUN echo "quit()" > /tmp/init-delta.py && \
    spark-submit /tmp/init-delta.py && \
    rm /tmp/init-delta.py
```

docs/using/recipes.md
```diff
@@ -528,3 +528,18 @@ USER $NB_UID
...
+
+## Enable Delta Lake in Spark notebooks
+Please note that the Delta lake packages are only available for Spark version > 3.0. By adding the properties to `spark-defaults.conf`, the user no longer needs to enable Delta support in each notebook. 
+
+```dockerfile
+FROM jupyter/pyspark-notebook:latest
+
+USER root
+
+RUN echo 'spark.jars.packages io.delta:delta-core_2.12:0.8.0' >> $SPARK_HOME/conf/spark-defaults.conf && \
```
romainx, line 540: Manage the version as an `ARG` is a good practice.

```suggestion
ARG delta-core_version="0.8.0"

USER root

RUN echo "spark.jars.packages io.delta:delta-core_2.12:${delta-core_version}" >> $SPARK_HOME/conf/spark-defaults.conf && \
```

romainx (changes requested):
Hello,

Good recipe. It let some suggestions that you might consider.

Thanks.
Best.

docs/using/recipes.md
joshuacook: The Scala version will bump over time as well:

```
ARG scala_version="2.12"
ARG delta-core_version="0.8.0"
USER root
RUN echo "spark.jars.packages io.delta:delta-core_${scala_version}:${delta-core_version}" >> $SPARK_HOME/conf/spark-defaults.conf && \ 
```

joshuacook:
@tbluejeans @romainx I think that over time it may be worth revisiting whether this should be part of the core image. Right now there are competing projects as to which will be the de facto storage medium to Spark's compute, in particular, Delta Lake, Apache Iceberg, and Apache Hudi. I would imagine that sometime in the next year or two, one of these will emerge as the industry standard for working with Spark. At that point, I'd say including that format in the default image is the right thing to do.

romainx:
@joshuacook very interesting and you're right it could make sense to include the de facto standard. Totally agree.
If you have more information on that do not hesitate to share it with us.
Many thanks.

tbluejeans:
Hi @romainx , 
hi @joshuacook ,

Thank you for your reviews and the suggestions! Good point, adding the dependency fetching into the build. I committed your suggestions and tested the recipe. After a small fix concering the build arg, it now works fine. 😄

romainx:
@tbluejeans and @joshuacook thank you. Merged 🚀

PR #1267, opened 2021-04-16
Update Dockerfile

alexander-manley:
v3.0.14
Clean up browser check (#10080)
Loosen pin on jupyter-packaging (#9998)

v3.0.13
Add cell id per notebook format 4.5 (#10018)
Fix label for “Create Console for Editor” (#9794)
Use blobs to set the svg source of an image in the image viewer (#10029)
Fix copy shareable link command (#10021)
Clarify where the overrides.json file should be in the docs (#9996)
Do not make unnecessary npm registry requests (#9974)
Fix escaping of urls and paths (#9978)


romainx (approved):
Hello @alexander-manley,

Thanks for this update.

Merging it.

PR #1270, opened 2021-04-23
Prepend $CONDA_DIR/bin instead of appending

### LINKED ISSUE #1053: Wrong PATH in start.sh when running as root

maresb:
Replaces #1054 (because I deleted my fork)
Fixes #1053

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -150,8 +150,8 @@ if [ "$(id -u)" == 0 ] ; then
     # Update potentially outdated environment variables since image build
     export XDG_CACHE_HOME="/home/${NB_USER}/.cache"
 
-    # Add ${CONDA_DIR}/bin to sudo secure_path
-    sed -r "s#Defaults\s+secure_path\s*=\s*\"?([^\"]+)\"?#Defaults secure_path=\"\1:${CONDA_DIR}/bin\"#" /etc/sudoers | grep secure_path > /etc/sudoers.d/path
+    # Prepend ${CONDA_DIR}/bin to sudo secure_path
+    sed -r "s#Defaults\s+secure_path\s*=\s*\"?([^\"]+)\"?#Defaults secure_path=\"${CONDA_DIR}/bin:\1\"#" /etc/sudoers | grep secure_path > /etc/sudoers.d/path
 
     # Optionally grant passwordless sudo rights for the desired user
     if [[ "$GRANT_SUDO" == "1" || "$GRANT_SUDO" == "yes" ]]; then
@@ -168,6 +168,12 @@ if [ "$(id -u)" == 0 ] ; then
         PATH="${PATH}" \
         PYTHONPATH="${PYTHONPATH:-}" \
         "${cmd[@]}"
+        # Note on the purpose of "PATH=${PATH}":
+        #   In case "${cmd[@]}" is "bash", then PATH will be used by this bash shell.
+        #   However, PATH is irrelevant to how the above sudo command resolves the
+        #   path of "${cmd[@]}". Sudo's path resolution is done via the "secure_path"
+        #   variable set above in /etc/sudoers.d/path.
+
 
 # The container didn't start as the root user, so we will have to act as the
 # user we started as.
--- a/base-notebook/test/test_container_options.py
+++ b/base-notebook/test/test_container_options.py
@@ -315,7 +315,7 @@ def test_secure_path(container: TrackedContainer, tmp_path: pathlib.Path) -> Non
     d = tmp_path / "data"
     d.mkdir()
     p = d / "wrong_python.sh"
-    p.write_text("""#!/bin/bash\necho "Wrong python executable invoked!" """)
+    p.write_text('#!/bin/bash\necho "Wrong python executable invoked!"')
     p.chmod(0o755)
 
     logs = container.run_and_wait(
--- a/base-notebook/test/test_container_options.py
+++ b/base-notebook/test/test_container_options.py
@@ -306,3 +306,24 @@ def test_jupyter_env_vars_to_unset_as_root(
         **root_args,  # type: ignore
     )
     assert "I like bananas and stuff, and love to keep secrets!" in logs
+
+
+def test_secure_path(container: TrackedContainer, tmp_path: pathlib.Path) -> None:
+    """Make sure that the sudo command has conda's python (not system's) on path.
+    See <https://github.com/jupyter/docker-stacks/issues/1053>.
+    """
+    d = tmp_path / "data"
+    d.mkdir()
+    p = d / "wrong_python.sh"
+    p.write_text("""#!/bin/bash\necho "Wrong python executable invoked!" """)
+    p.chmod(0o755)
+
+    logs = container.run_and_wait(
+        timeout=5,
+        tty=True,
+        user="root",
+        volumes={p: {"bind": "/usr/bin/python", "mode": "ro"}},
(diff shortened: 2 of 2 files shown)


maresb:
Thanks @mathbunnyru!

mathbunnyru (changes requested):
@maresb could you add a test, which checks that everything is set up properly?

I think this change is good, but I have to say I'm not experienced in bash to be 100% sure.

maresb:
@mathbunnyru I'll see what I can do. It may take me several days. (Please do nag me if I don't make progress!)

romainx:
Hello thanks for the work and @mathbunnyru thanks for the comment. @maresb I converted the PR to draft to let you the required time to add the test 😆 .

base-notebook/start.sh
```diff
@@ -88,18 +88,18 @@ if [ $(id -u) == 0 ] ; then
...
-    sed -r "s#Defaults\s+secure_path\s*=\s*\"?([^\"]+)\"?#Defaults secure_path=\"\1:$CONDA_DIR/bin\"#" /etc/sudoers | grep secure_path > /etc/sudoers.d/path
+    sed -r "s#Defaults\s+secure_path\s*=\s*\"?([^\"]+)\"?#Defaults secure_path=\"$CONDA_DIR/bin:\1\"#" /etc/sudoers | grep secure_path > /etc/sudoers.d/path
 
-    # Exec the command as NB_USER with the PATH and the rest of
+    # Exec the command as NB_USER with the PATH from secure_path and the rest of
     # the environment preserved
     run-hooks /usr/local/bin/before-notebook.d
     echo "Executing the command: ${cmd[@]}"
-    exec sudo -E -H -u $NB_USER PATH=$PATH XDG_CACHE_HOME=/home/$NB_USER/.cache PYTHONPATH=${PYTHONPATH:-} "${cmd[@]}"
+    exec sudo -E -H -u $NB_USER XDG_CACHE_HOME=/home/$NB_USER/.cache PYTHONPATH=${PYTHONPATH:-} "${cmd[@]}"
```
consideRatio, line 102: I wonder about removing `PATH=$PATH` here, what is the motivation for doing so? I'm quite confused in general about this.

https://github.com/jupyter/docker-stacks/pull/1052/files#diff-41f90d7afcdae13f8516195e078d0a203972b5c5105851eaecfc0a98e9739107R147-R185

consideRatio: Trying things out locally on my computer, I note:
- using `-E` / `--preserve-env` doesn't preserve PATH, but setting it explicitly with `PATH=$PATH` works, and explicitly setting it to be preserved with `--preserve-env=PATH` works.

I was thinking perhaps this was removed because it served no purpose, but perhaps it does? I'm generally confused.

Should this be part of this PR?

mathbunnyru:
> @mathbunnyru I'll see what I can do. It may take me several days. (Please do nag me if I don't make progress!)

@maresb you asked me to nag you. Doing it right now ;)

base-notebook/start.sh
maresb: @consideRatio thanks for reminding me of this, I had forgotten. It's really painful to come back to something like this a whole year later.

I even wrote [here](https://github.com/jupyter/docker-stacks/issues/1053#issuecomment-605349169) that I like your solution better. I'll see if I can dig in and figure out a reasonable solution...

maresb: I think I have an idea about what's going on and why we're so confused, although I need to do further experiments to be sure...

There are really two separate things going on. First `sudo` executes `"${cmd[@]}"`, and then this execution inherits environment variables. Often `"${cmd[@]}"` will be `bash`, which will then resolve paths according to the `$PATH` passed to it. However, `sudo` does its own path resolution, independently of `$PATH`. For example, in #1053, `sudo` runs `python`, which `sudo` resolves as `/usr/bin/python` according to `secure_path`. However, if `sudo` runs `bash` and then you type `python` into the shell, then `bash` resolves `python` according to `$PATH`.

I wonder if your #1052 solves #1053... I need to test it, but I'm optimistic.

maresb:
@mathbunnyru thanks for the reminder. Actually, please continue to nag me.

I would like to request help from @consideRatio. He reminded me above of his superior solution.

My current plan is actually to extract the `PATH`-relevant pieces from his [`start.sh`](https://github.com/jupyter/docker-stacks/pull/1052/files#diff-41f90d7afcdae13f8516195e078d0a203972b5c5105851eaecfc0a98e9739107) and to replace my current PR with that. I think that @consideRatio is probably better-suited for this job than me. And if he does it, it's easy for him to get direct attribution.

consideRatio:
I'm feeling a bit overwhelmed with things to work on at the moment and won't pick this up for a while I think. If anyone wants to use past work in other PRs etc, I have no concerns at all about attribution!

In general, I found things I attempted ended up becoming quite significant reworks that became hard to merge. If I'd get time to spend on this again I'd try to think what can be done to incrementally go towards a solid solution.

Thanks @maresb for your work to resolve these matters! :heart: :tada:

mathbunnyru:
@maresb one more try here :)

mathbunnyru:
@maresb one more friendly ping here :)

maresb:
Thanks @mathbunnyru, I have actually been working on this again. Not finished yet though.

mathbunnyru:
> @mathbunnyru I'll see what I can do. It may take me several days. (Please do nag me if I don't make progress!)

@maresb I don't see progress here! 😂 😂 😂

maresb:
@mathbunnyru, how's this?

In reference to https://github.com/jupyter/docker-stacks/pull/1270#discussion_r625870619, I reverted my previous removal of `PATH=$PATH`, and I tried to instead summarize my explanation https://github.com/jupyter/docker-stacks/pull/1270#discussion_r633001198 as an inline comment.

Ping @consideRatio

maresb:
I should mention that I did indeed test my test: it fails before my first commit and succeeds after.

base-notebook/test/test_container_options.py
```diff
@@ -306,3 +306,24 @@ def test_jupyter_env_vars_to_unset_as_root(
...
+
+
+def test_secure_path(container: TrackedContainer, tmp_path: pathlib.Path) -> None:
+    """Make sure that the sudo command has conda's python (not system's) on path.
+    See <https://github.com/jupyter/docker-stacks/issues/1053>.
+    """
+    d = tmp_path / "data"
+    d.mkdir()
+    p = d / "wrong_python.sh"
+    p.write_text("""#!/bin/bash\necho "Wrong python executable invoked!" """)
```
mathbunnyru, line 318:
```suggestion
    p.write_text('#!/bin/bash\necho "Wrong python executable invoked!"')
```

mathbunnyru (approved):
This looks awesome. And thank you for adding the test.

consideRatio (approved):
Wieee nice work! These kinds of changes has been rough to make because of the small details have such importance, and its hard to overview how everything works and what details influences what etc.

I've learned a lot, thanks!

PR #1271, opened 2021-04-26
Make documentation work with the latest version of sphinx

mathbunnyru:
The issue I solve:
If I run `make dev-env && make docs`, the latest sphinx is installed and the docs do not work properly.

The problems:

`source_parsers` are long time deprecated: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-source_parsers and they've been removed in 3.0

`MyST-Parser` is now recommended way to handle markdown: https://www.sphinx-doc.org/en/master/usage/markdown.html

Sphinx 1.8.5, which is actually used in production, was released more than 2 years ago.


docs/conf.py
```diff
@@ -25,22 +25,22 @@
 
 # If your documentation needs a minimal Sphinx version, state it here.
 #
-needs_sphinx = '1.4'
+needs_sphinx = '2.1'
```
mathbunnyru, line 28: myst_parser requires at least `2.1` (as the docs say)

mathbunnyru:
@parente I think the configuration on how to build the docs is actually hardcoded somewhere in https://readthedocs.org/projects/jupyter-docker-stacks.

Let's take a look at the latest build:
https://readthedocs.org/projects/jupyter-docker-stacks/builds/13540099/

Somehow, there is a line like `/home/docs/checkouts/readthedocs.org/user_builds/jupyter-docker-stacks/envs/latest/bin/python -m pip install --upgrade --no-cache-dir mock==1.0.1 pillow==5.4.1 alabaster>=0.7,<0.8,!=0.7.5 commonmark==0.8.1 recommonmark==0.5.0 sphinx<2 sphinx-rtd-theme<0.5 readthedocs-sphinx-ext<2.2`.
I don't know where it comes from and you might need to update some setting in the `readthedocs.org` profile.

romainx:
It's a good remark I'm not sure I can help here.

consideRatio:
@mathbunnyru I have struggled a bit with the RTD build environment before and figure this could be relevant knowledge:

1. RTD build environments have packages pre-installed, so if the docs requirements.txt specifies `sphinx` it won't install a more recent version than was already pre-installed unless the requirements.txt files specifies `sphinx>=3` or similar. See comment in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/.readthedocs.yml.
2. The order of how you specify requirements in requirements.txt actually matters, see the comment in this file: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/doc/requirements.txt

From what I can tell, this PR LGTM!

### Related action point
Enable RTD PR preview's as documented on how to do [here](https://docs.readthedocs.io/en/stable/pull-requests.html). This can only be done by a maintainer of the RTD project though. @parente? Feel free to add me as well if you want to offload some future maintenance work!

requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 docker
+myst-parser
 pre-commit
 pytest
-recommonmark
 requests
 sphinx>=1.6
```
mathbunnyru, line 6:
```suggestion
sphinx>=2.1
```

mathbunnyru:
Thanks @consideRatio for a review.
1. In my PR we're installing `myst-parser`, so sphinx will be automatically updated. I also updated `sphinx` requirement to use at least `2.1`.
2. In our case we already have `myst-parser` quite high, so it should work fine, right?

Having RTD PR preview would be awesome, thank you for this suggestion.

consideRatio:
> * In my PR we're installing `myst-parser`, so sphinx will be automatically updated. I also updated `sphinx` requirement to use at least `2.1`.
> * In our case we already have `myst-parser` quite high, so it should work fine, right?

I'm not confident about this in general, but it is my understanding that as long as you have myst-parser that sphinxrequires >=2.1 before a package that just requires sphinx, or sphinx perhaps explicitly listed, you should be fine!

consideRatio:
The CI failure is unrelated and fixed in #1307!

Issue #1268, opened 2021-04-17
Tensorflow 2 is finally on conda-forge!


maresb:
As of about 4 days ago. It is now finally possible to do `mamba install tensorflow` or `conda install tensorflow`.

It would be great to switch tensorflow-notebook from `pip` to `conda` or `mamba`. I have not yet had the chance to test it out though.

mathbunnyru:
That's great news!

I also don't like using pip and conda together (better to install all the things using conda).
But, in this case, we're already using `2.4.1`, but for now only 2.4.0 is available in conda-forge.

https://anaconda.org/conda-forge/tensorflow/files

I hope, they will update soon and we will switch to conda.

mathbunnyru:
There is a PR attempt to update the version:
https://github.com/conda-forge/tensorflow-feedstock/pull/114

PR #1275, opened 2021-04-30
Install tensorflow from conda instead of pip

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1268

### DIFF
--- a/tensorflow-notebook/Dockerfile
+++ b/tensorflow-notebook/Dockerfile
@@ -6,7 +6,7 @@ FROM $BASE_CONTAINER
 LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 
 # Install Tensorflow
-RUN conda install --quiet --yes \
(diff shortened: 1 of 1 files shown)


mathbunnyru:
Here is the thing - `conda` doesn't work with tensorflow :(
It doesn't seem to crash, but it takes ages to install tensorflow.

I think something related was happening here:
https://github.com/jupyter/docker-stacks/pull/1213#issuecomment-774719575

Mamba works fine.

romainx (approved):
@mathbunnyru not our usage to install packages with `mamba` however I'm in favor of that

- it seems cleaner than using `pip` in our context 
- we have either a tensorflow test and a test that loads all packages
- this image has no downstream so we limit the side effect (at least on this repo)
- it's easy to revert in case of issue

So I'm in favor of merging it.

mathbunnyru:
And we will also check, that `mamba` works in real life :)

Issue #1163, opened 2020-09-17
Allow pyspark-notebook build with archived Spark version


romainx:
Hello,

Two recent issues #1156 and #1161 have the same root cause. At this time the build of `pyspark-notebook` will only work with the set of versions available on the [Apache Spark download page](https://spark.apache.org/downloads.html), so it will not work with the archived versions.

To fix it, it sounds easy to switch to the [Spark release archives](https://archive.apache.org/dist/spark/) however the download mechanism used in the image is more clever because it uses the preferred mirror defined by a script before downloading the archive. This mechanism has been put in place in the PR #1009 in order to fix transient build errors and to avoid to put pressure on a specific download site. We should stick to the same principle if we want to support builds from Spark archives.

Another solution is simply not to change anything and only support Spark latest versions.

Best.

PR #1279, opened 2021-05-02
Install spark from archive.apache.org to be able to use old versions

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1163

@romainx as you suggested, in your issue, we could use a spark archive to be able to build the old spark versions.

You said, that we will put pressure on a specific download site.
I think that's true, at the same time it's `https://spark.apache.org`.
The pressure we create is only a few builds per day (at max).
And I think there are not so many people, who rebuild our images from scratch (I think it's not needed in most cases).

I agree that this way is less clever than using the preferred site.
At the same time, it's much easier to use only one `wget` command.

If I'm right, we haven't tried `https://archive.apache.org` as a mirror so it might be reliable to use (I hope it is).

I think - let's give `archive` a try, if everything goes well for us (and users), we will get simpler dockerfile and ability to customize spark version.
If not, it should be easy to revert the commit.

### DIFF
--- a/docs/using/specifics.md
+++ b/docs/using/specifics.md
@@ -12,7 +12,7 @@ This page provides details about features specific to one or more images.
 
 You can build a `pyspark-notebook` image (and also the downstream `all-spark-notebook` image) with a different version of Spark by overriding the default value of the following arguments at build time.
 
-* Spark distribution is defined by the combination of the Spark and the Hadoop version and verified by the package checksum, see [Download Apache Spark](https://spark.apache.org/downloads.html) for more information.
+* Spark distribution is defined by the combination of the Spark and the Hadoop version and verified by the package checksum, see [Download Apache Spark](https://spark.apache.org/downloads.html) and the [archive repo](https://archive.apache.org/dist/spark/) for more information.
   * `spark_version`: The Spark version to install (`3.0.0`).
   * `hadoop_version`: The Hadoop version (`3.2`).
   * `spark_checksum`: The package checksum (`BFE4540...`).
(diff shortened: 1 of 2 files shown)


docs/using/specifics.md
```diff
@@ -12,7 +12,7 @@ This page provides details about features specific to one or more images.
 
 You can build a `pyspark-notebook` image (and also the downstream `all-spark-notebook` image) with a different version of Spark by overriding the default value of the following arguments at build time.
 
-* Spark distribution is defined by the combination of the Spark and the Hadoop version and verified by the package checksum, see [Download Apache Spark](https://spark.apache.org/downloads.html) for more information. At this time the build will only work with the set of versions available on the Apache Spark download page, so it will not work with the archived versions.
+* Spark distribution is defined by the combination of the Spark and the Hadoop version and verified by the package checksum, see [Download Apache Spark](https://spark.apache.org/downloads.html) for more information.
```
romainx, line 15:
```suggestion
* Spark distribution is defined by the combination of the Spark and the Hadoop version and verified by the package checksum, see [Download Apache Spark](https://spark.apache.org/downloads.html) and the [archive repo](https://archive.apache.org/dist/spark/) for more information.
```

romainx (commented):
Hello @mathbunnyru I understand and agree that it is worth trying this solution for the sake of simplicity and to permit to install old versions. Before that change, we had frequent failed builds. Let's check with the archive URL if it works fine.
I made a simple suggestion to mention the archive repo in the doc.

So let's try it!

Issue #1280, opened 2021-05-03
Permission to jupyter/docker-stacks.wiki.git denied to github-actions[bot].


mathbunnyru:
Pushing to the `https://github.com/jupyter/docker-stacks.wiki.git/` seems to be broken right now.

It means that no build manifests are generated.

Fir PR merge where it happened: https://github.com/jupyter/docker-stacks/commit/dfcb043bd88798a9419a5c7a38eb1218f7aac941
The Github Actions run: https://github.com/jupyter/docker-stacks/commit/dfcb043bd88798a9419a5c7a38eb1218f7aac941

It also happened during the last merge: https://github.com/jupyter/docker-stacks/commit/fde8f1eec3e70edf67560b5430545a22382334bf

So, unfortunately, it's not spurious.

The thing is that we didn't change this part of the process (in the first PR).
And it's the `remote`, which gives an error:

```
remote: Permission to jupyter/docker-stacks.wiki.git denied to github-actions[bot].
25
fatal: unable to access 'https://github.com/jupyter/docker-stacks.wiki.git/': The requested URL returned error: 403
```

So it means that something is wrong with the internal GitHub setup.

consideRatio:
Is this an action related to secrets.GITHUB_TOKEN? Perhaps the jupyter org has reduced the permissions by default now?

See https://github.com/jupyterhub/team-compass/issues/404

mathbunnyru:
Yes, it is.

https://github.com/jupyter/docker-stacks/blob/master/.github/workflows/docker.yml#L63-L69

Probably this is the reason why it doesn't work.

PR #1281, opened 2021-05-03
Add write permissions to github jobs

mathbunnyru:
**NOTE: someone more experienced in GitHub actions should tell that this is fine.**

Also, there is no way I could check this (because my permissions are very limited)

Fix: https://github.com/jupyter/docker-stacks/issues/1280

The advice I used: https://github.com/jupyterhub/team-compass/issues/404#issue-863678103

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -20,6 +20,8 @@ jobs:
   build:
     name: Build Docker Images
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     if: >
       !contains(github.event.head_commit.message, 'ci skip') &&
       !contains(github.event.pull_request.title, 'ci skip')
--- a/.github/workflows/sphinx.yml
+++ b/.github/workflows/sphinx.yml
@@ -16,6 +16,8 @@ jobs:
   build:
     name: Build Sphinx Documentation
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     if: >
       !contains(github.event.head_commit.message , 'ci skip') &&
       !contains(github.event.pull_request.title, 'ci skip')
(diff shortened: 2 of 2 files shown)


mathbunnyru:
@consideRatio could you tell me, what do you think? Is this OK or not?

mathbunnyru:
I suggest we don't make quick assumptions and take time because skipping manifests for a few builds is not really a big issue.
Security issues are much more important in this case.

So please, take your time.

romainx:
@mathbunnyru seems to be the right thing to do. Permissions cannot be set at lower level so I do not see a better way of fixing the issue.

mathbunnyru:
I also searched for setting permissions on a lower level and it is indeed seems to be impossible.
Also I thought about having a separate job for pushing, but that is not something easily done, because we need the artefacts from the upper build commands in the job.

consideRatio (approved):
Adding write permissions for `conents` seems correct. I don't think there is a more appropriate scoped permission than this based on [these docs](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token).

I suggest merge and see what happens :)

romainx:
Strike while the iron is hot 🔨

mathbunnyru:
Haha, ok, let's try it out :)

mathbunnyru:
At least it worked for the strings pushing.
https://github.com/jupyter/docker-stacks/commit/ccc1470b027dc72ce207b9f3d9cb3db06224398f
So there are some chances it will work for the wiki as well.

mathbunnyru:
It works now 👍 
https://github.com/jupyter/docker-stacks/runs/2495304110

romainx:
It worked 🎉 Great
Just a line feed to fix 😄 
Home_·_jupyter_docker-stacks_Wiki

mathbunnyru:
Yes, I also noticed it :)

Give me a sec, I will take a look why it's wrong :)

romainx:
Enough merge for tonight :sweat_smile:
We will fix it later.

mathbunnyru:
> Enough merge for tonight 😅
> We will fix it later.

Sorry, already done.
https://github.com/jupyter/docker-stacks/pull/1284

PR #1277, opened 2021-05-01
Use `markdown-cli` to check markdown for the style errors

mathbunnyru:

PR #1286, opened 2021-05-04
Regular update 2021-05-04

romainx:
Hello,

- Few packages updated
- Fix Hadolint warnings
- Style

Not updated Mambaforge version since it brings a new version of Python `3.9` thanks @mathbunnyru for the heads-up.

- Added a test to check the Python version to avoid this kind of surprise in the future.

Best.

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -24,7 +24,7 @@ USER root
 # (ARGS are in lower case to distinguish them from ENV)
 # Check https://github.com/conda-forge/miniforge/releases
 # Conda version
-ARG conda_version="4.10.1"
+ARG conda_version="4.10.0"
 # Miniforge installer patch version
 ARG miniforge_patch_number="0"
 # Miniforge installer architecture
@@ -39,7 +39,7 @@ ARG miniforge_version="${conda_version}-${miniforge_patch_number}"
 # Miniforge installer
 ARG miniforge_installer="${miniforge_python}-${miniforge_version}-Linux-${miniforge_arch}.sh"
 # Miniforge checksum
-ARG miniforge_checksum="d4065b376f81b83cfef0c7316f97bb83337e4ae27eb988828363a578226e3a62"
+ARG miniforge_checksum="c56cc2da96043688c6bdb521d825de27754de0a342d5228ba3155cd94532ff75"
 
 # Install all OS dependencies for notebook server that starts but lacks all
 # features (e.g., download as all possible file formats)
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -39,7 +39,7 @@ ARG miniforge_version="${conda_version}-${miniforge_patch_number}"
 # Miniforge installer
 ARG miniforge_installer="${miniforge_python}-${miniforge_version}-Linux-${miniforge_arch}.sh"
 # Miniforge checksum
-ARG miniforge_checksum="2eb2226214aeec849d9fc935ac092e669993a666a6b25c613b1efc9590441777"
+ARG miniforge_checksum="d4065b376f81b83cfef0c7316f97bb83337e4ae27eb988828363a578226e3a62"
 
 # Install all OS dependencies for notebook server that starts but lacks all
 # features (e.g., download as all possible file formats)
@@ -51,10 +51,9 @@ RUN apt-get -q update \
     sudo \
     locales \
     fonts-liberation \
-    run-one \
- && apt-get clean && rm -rf /var/lib/apt/lists/*
-
-RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \
+    run-one && \
+    apt-get clean && rm -rf /var/lib/apt/lists/* && \
+    echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \
     locale-gen
 
 # Configure environment
@@ -159,9 +158,8 @@ USER root
 
 # Prepare upgrade to JupyterLab V3.0 #1205
 RUN sed -re "s/c.NotebookApp/c.ServerApp/g" \
-    /etc/jupyter/jupyter_notebook_config.py > /etc/jupyter/jupyter_server_config.py
-
-RUN fix-permissions /etc/jupyter/
+    /etc/jupyter/jupyter_notebook_config.py > /etc/jupyter/jupyter_server_config.py && \
+    fix-permissions /etc/jupyter/
 
 # Switch back to jovyan to avoid accidental container runs as root
 USER $NB_UID
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -24,7 +24,7 @@ USER root
 # (ARGS are in lower case to distinguish them from ENV)
 # Check https://github.com/conda-forge/miniforge/releases
 # Conda version
-ARG conda_version="4.10.0"
+ARG conda_version="4.10.1"
 # Miniforge installer patch version
 ARG miniforge_patch_number="0"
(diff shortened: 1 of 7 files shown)


base-notebook/Dockerfile
```diff
@@ -51,10 +51,9 @@ RUN apt-get -q update \
     sudo \
     locales \
     fonts-liberation \
-    run-one \
- && apt-get clean && rm -rf /var/lib/apt/lists/*
-
-RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \
+    run-one && \
```
mathbunnyru, line 54: Please, fix the style for `&&` several lines above as well.

base-notebook/Dockerfile
```diff
@@ -51,10 +51,9 @@ RUN apt-get -q update \
...
+    apt-get clean && rm -rf /var/lib/apt/lists/* && \
+    echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \
```
mathbunnyru, line 56: I think having separate steps is actually good in this case (there is a clear distinction in the processes), so I would prefer hadolint ignore for this line.

base-notebook/Dockerfile
```diff
@@ -24,7 +24,7 @@ USER root
 # (ARGS are in lower case to distinguish them from ENV)
 # Check https://github.com/conda-forge/miniforge/releases
 # Conda version
-ARG conda_version="4.10.0"
+ARG conda_version="4.10.1"
```
mathbunnyru, line 27: This change is not actually that simple.
MiniForge has started using python 3.9, so with this change we will switch to python 3.9 as well.
You can see it in the tagging step :)

There is a small release note as well.
https://github.com/conda-forge/miniforge/releases

mathbunnyru (changes requested):

base-notebook/Dockerfile
mathbunnyru: Maybe we should wait at least till anaconda ships python 3.9 by default?

base-notebook/Dockerfile
mathbunnyru: After a closer look, it actually makes sense both ways, because several lines above we install locales. So, I'm ok with this change.

mathbunnyru (changes requested):
Let's think about updating python.

base-notebook/Dockerfile
romainx: Hello,

I do not understand what need to be fixed here regarding the style, sorry.

```dockerfile
fonts-liberation \
run-one && \
apt-get clean && rm -rf /var/lib/apt/lists/* && \
echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \
locale-gen
```

tensorflow-notebook/Dockerfile
```diff
@@ -7,8 +7,7 @@ LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 
 # Install Tensorflow
 RUN mamba install --quiet --yes \
-    'tensorflow=2.4.1' \
-    && \
```
romainx, line 11: @mathbunnyru is it the style you want to use? If it's the case I have fixed it.

base-notebook/Dockerfile
romainx: So if you are Ok :-)

base-notebook/Dockerfile
romainx: I did not even imagine that the version of Python could change in a patch release. If you want to master the python version and avoid an unwanted change in the future I think we have to write a test. I can do it.

romainx: For information, [Miniconda](https://docs.conda.io/en/latest/miniconda.html) (the previous installer is shipped also with Python 3.9 `Miniconda3-py39_4.9.2-Linux-x86_64.sh`.
What was applied here until now was to always use the latest versions of the software and to rely on tags if users want to use a previous version.

romainx:
I'm converted it do draft while waiting for @mathbunnyru approval.

base-notebook/Dockerfile
mathbunnyru: Several lines above the line start from `&&`like this: `&& apt-get install -yq --no-install-recommends \`

tensorflow-notebook/Dockerfile
mathbunnyru: No, I added the comment about the line I didn't like.

base-notebook/Dockerfile
mathbunnyru: I was also surprised by changing of the Python version in a patch release.

Yes, let's write a test, which detects the current python version and compares it with hardcoded constant - this way we will always know when python is updated.

Ok, let's write a test and update.

base-notebook/Dockerfile
romainx: Ok got it, I though the comment was related to my change, but Ok I've seen it.

tensorflow-notebook/Dockerfile
romainx: Ok because I've rolled back your previous style change and I made the link with your other remark on style with `&&` at the beginning of a line. So I would like to be sure.

base-notebook/Dockerfile
mathbunnyru: Thanks :)
Sorry for being not 100% clear.

tensorflow-notebook/Dockerfile
mathbunnyru: Thanks, right now the style looks good and consistent :)

base-notebook/Dockerfile
romainx: Sure I will do that 👍

romainx: Test implemented but I rolled back the version of Python to stick to `3.8.x`. We will update it later.

romainx:
@mathbunnyru I've implemented all the changes and also reverted to keep Python `3.8.x`. Please tell me if it's Ok for you now and I will merge it. Thanks.

base-notebook/test/test_python.py
```diff
@@ -0,0 +1,19 @@
...
+# Distributed under the terms of the Modified BSD License.
+import logging
+
+from packaging import version
+
+LOGGER = logging.getLogger(__name__)
+
+
+def test_python_version(container, python_max_version="3.9"):
+    """Check that python version is not higher than a max version"""
```
mathbunnyru, line 11:
```suggestion
    """Check that python version is lower than a max version"""
```

I would also change the word max to the word next.

mathbunnyru (changes requested):

base-notebook/test/test_python.py
romainx: Docstring changed, also change the word max to next as you suggested. I hope it will be Ok now.

mathbunnyru (approved):
👍

romainx:
Ok thanks, I will merge it. 
@mathbunnyru I'm gonna take a week break or a week and a half break, I will be able to spend a limited time here before so if something needs to be reviewed and merged before, tell me.

mathbunnyru:
> Ok thanks, I will merge it.
> @mathbunnyru I'm gonna take a week break or a week and a half break, I will be able to spend a limited time here before so if something needs to be reviewed and merged, tell me.

Could you please take a look in the PR: https://github.com/jupyter/docker-stacks/pull/1277?

mathbunnyru:
Another PR, which is ready (you already took a look and it should be good to go): https://github.com/jupyter/docker-stacks/pull/1288

So, two reviews in total, which can be easily merged.

PR #1288, opened 2021-05-05
Use hadolint from pre-commit

mathbunnyru:
Thanks to https://github.com/hadolint/hadolint/pull/523 we can use hadolint directly in `pre-commit`.

Downsides:
- being able to run hadolint on a specific image with `make` command. But it is actually still possible to run it via `pre-commit --run` flags.
- need to be running docker to run hadolint. But this project uses docker for basically everything, so this should not be a problem.

Upsides:
- 21 deletions in Makefile
- approximately 50 deletions in docs
- The setup is easier because there are no specific things for hadolint
- No hadolint binary in HOME (I didn't like this on my mac)
- Pre-commit hooks versions could be updated automatically (I will try to do it later)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,11 +6,11 @@ repos:
       - id: check-yaml
         files: .*\.(yaml|yml)$
   - repo: https://github.com/hadolint/hadolint.git
-    rev: v2.1.0
+    rev: v2.3.0
     hooks:
       - id: hadolint-docker
         # FIXME: remove after https://github.com/hadolint/hadolint/issues/628 is resolved
-        entry: hadolint/hadolint:v2.1.0 hadolint
+        entry: hadolint/hadolint:v2.3.0 hadolint
         exclude: Dockerfile.ppc64le|Dockerfile.ppc64le.patch
   - repo: https://github.com/adrienverge/yamllint.git
     rev: v1.26.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,9 +6,11 @@ repos:
       - id: check-yaml
         files: .*\.(yaml|yml)$
   - repo: https://github.com/hadolint/hadolint.git
-    rev: v2.3.0
+    rev: v2.1.0
     hooks:
       - id: hadolint-docker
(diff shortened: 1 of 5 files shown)


mathbunnyru:
This is waiting for https://github.com/jupyter/docker-stacks/pull/1286 to be merged because there are same errors fixed in that PR and I don't want to have conflicts / do some work twice.

mathbunnyru:
I also noted a bug in the hadolint-docker, which won't allow us to use it because it is better to fix the version of hadolint docker image.

https://github.com/hadolint/hadolint/issues/628

.pre-commit-config.yaml
```diff
@@ -5,6 +5,13 @@ repos:
     hooks:
       - id: check-yaml
         files: .*\.(yaml|yml)$
+  - repo: https://github.com/hadolint/hadolint.git
+    rev: v2.1.0
```
romainx, line 9:
```suggestion
    rev: v2.3.0
```

.pre-commit-config.yaml
```diff
@@ -5,6 +5,13 @@ repos:
...
+    hooks:
+      - id: hadolint-docker
+        # FIXME: remove after https://github.com/hadolint/hadolint/issues/628 is resolved
+        entry: hadolint/hadolint:v2.1.0 hadolint
```
romainx, line 13:
```suggestion
        entry: hadolint/hadolint:v2.3.0 hadolint
```

.pre-commit-config.yaml
mathbunnyru: I will merge master after your PR merge and fix the version in two places.
Unfortunately, right now we need to specify version in two places.
One - for revision of the hook, second - for the docker hadolint used.

romainx (changes requested):
Hello,

Continuation of what we started together :smile:.
It's a good improvement and definitely the right solution.
I understand that for now we have to specify the version in two places, however it's not a big deal in comparison of all the stuff removed.

Thanks 🍰

romainx (approved):

romainx:
Merged thanks a lot ❤️

PR #1289, opened 2021-05-06
Update miniforge to 4.10.1 and python version to 3.9.2

mathbunnyru:
This is a follow up on https://github.com/jupyter/docker-stacks/pull/1286


romainx:
😆 You were against moving to python `3.9`.

mathbunnyru:
> 😆 You were against moving to python `3.9`.

No! :) :)

I didn't want us to move to python `3.9` without even noticing it.
I just wanted it to be explicitly visible.

I'm absolutely ok with moving, in a separate commit, which is easy to revert :)

romainx (approved):
Nothing to say, this test is perfect 😆

PR #1290, opened 2021-05-06
Removal of outdated ppc64le arch Dockerfile

consideRatio:
This PR reverts the addition of the way we at one point supported manual builds of a ppc64le architecture version of the Dockerfile's in this repo.

The key motivation for removing these are that these resources haven't been easy to keep maintained alongside the other files. By removing them, we can reduce maintenance burden and potentially enable a new solution to support building multiple architectures that is from a maintenance standpoint more sustainable.


mathbunnyru (changes requested):
https://github.com/jupyter/docker-stacks/blob/696d48cf898dfd7691a964cd1c487f7d6986c318/.pre-commit-config.yaml#L14

Please, remove this line as well.

mathbunnyru:
We need to be sure, that these files are not used by anyone else.
At least, they are not built right now by us in the CI, which is a good thing (for removal).

mathbunnyru:
The maintainer @ilsiyar-gaynutdinov-ibm hasn't been maintaining this since Dec 8, 2016 (this is his last commit).
There were some changes though by other people.

This is the original addition:
https://github.com/jupyter/docker-stacks/pull/305

consideRatio:
> We need to be sure, that these files are not used by anyone else.

I think this is more stringent than is reasonable to aim for, we just can't be sure. But, let's consider the content of the files.

This is from the Dockerfile.ppc64le.patch file removed.

```
+    'notebook==5.2.*' \
+    'jupyterhub==0.7.*' \
+    'jupyterlab==0.18.*'
```

This is from the Dockerfile.ppc64le

```
    'notebook=6.1.3' \
    'jupyterhub=1.1.0' \
    'jupyterlab=2.2.5' && \
```

This is from the main Dockerfile

```
    'notebook=6.3.0' \
    'jupyterhub=1.4.0' \
    'jupyterlab=3.0.14' && \
```

---

### History of changes

I've looked at the [git blame information of Dockerfile.ppc64le](https://github.com/jupyter/docker-stacks/blame/master/base-notebook/Dockerfile.ppc64le). Besides https://github.com/jupyter/docker-stacks/pull/1088 which was made a year ago, this file hasn't been updated for several years and it was non-functional at the time #1088 was created.

I suggest we settle for a heads up notification to the contributors to this file that we are cleaning up the repos logic in preparation to attempt implement a solution of providing additional architecture builds of the dockerfiles that is sustainable to maintain.

mathbunnyru:
Thanks for this investigation @consideRatio, it makes me much more sure we can (almost) safely delete these file 👍 

My opinion:
I do believe that having code like this that is not built/tested, uses patching, and not pushed to the official docker hub is not a good way to go if we want to maintain the project and I'm ok with deleting these files.

At the same time I do believe that we take some responsibility to try our best to reimplement multi-arch images.
I won't be able to write much of code myself (honestly, because, I'm not that interested in the arm images), but I will do PR reviews / try to help where I can and so on.

I want to hear from @romainx and @parente how they feel about this.

mathbunnyru (approved):

mathbunnyru:
And, to be honest, I wanted to send a PR exactly like this myself several days ago, but didn't have much time 😆

oskrdt:
I updated and tested the `base-notebook/Dockerfile.ppc64le` file a year ago when I had to make it work for Power architecture, and then pushed to DockerHub to `oskrdt/jupyter-base-notebook`. At least at that point I can confirm it was working... after that, I haven't done any update. It was useful for me at that time, but I'm not actively working on updating to the latest versions, so I understand the motivation to remove it. 
In the other hand, I didn't work with `base-notebook/Dockerfile.ppc64le.patch`, so I believe it might be broke. 
I would like to keep the ppc64le Dockerfile version for a open collaboration team, but we can keep the support internally in case you decide to remove it.
PD. Having multiple architecture support would be awesome.

consideRatio:
Thanks for the followup @oskrdt!

consideRatio:
Friendly ping to @romainx, what do you think?

romainx:
@consideRatio and @mathbunnyru sorry for the late answer.

My opinion

- This `Dockerfile` (and everything related) is obsolete and not maintained for a while
- This feature is undocumented
- If someone needs it, it could be restored from the history
- We will _shortly_ replace it with a great multi-arch image 😄 

So let's go for the big cleanup! 🚀

romainx (approved):
See my [comment](https://github.com/jupyter/docker-stacks/pull/1290#issuecomment-846201837).

consideRatio:
Wiee thanks for input @romainx!

Go for merge?

romainx:
Merged thanks !

PR #1292, opened 2021-05-06
Make css, json and yaml style identical

mathbunnyru:
As suggested by @consideRatio https://github.com/jupyter/docker-stacks/pull/1277#issuecomment-830701961 I tried `prettier` hook and it looks good and reasonable to add.
I will create a separate PR for the markdown file and after both PRs get merged it will be easy to add the hook itself.


consideRatio:
LGTM and make sense for me to have in this repo as well.

---

Off topic but while considering autoformatters: we have started using `black` systematically for python autoformatting within the JupyterHub org. Example [here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/7a3d18606d7bd4573216302a8a8a7bab804defed/.pre-commit-config.yaml#L22-L27)

mathbunnyru:
> Off topic but while considering autoformatters: we have started using black systematically for python autoformatting within the JupyterHub org. Example here

I know about `black` but sometimes I don't like how it does things.
And, right now we already check the code with flake8, so everything is already quite beautiful.

So, I probably won't be adding it as a must myself, but if there are some serious advantages, then I think it might be OK to add `black`.

consideRatio:
> So, I probably won't be adding it as a must myself, but if there are some serious advantages, then I think it might be OK to add black.

I think the biggest advantage is less time spent on reviewing formatting aspects and a reduced risk for merge conflicts because someone formatted something unrelated in a PR and which another PR was changing as well. I don't consider it crucial at all so let's do nothing at this point.

parente (approved):

PR #1294, opened 2021-05-06
Make markdown files look prettier using prettier :)

mathbunnyru:
This makes markdown files look very consistent across this repo.


mathbunnyru:
You might think that I'm slowly going crazy with all the pre-commit hooks and linters.
You're wrong.
Not slowly 🏎️ 

Actually, as for me, we have quite good docs (from the content point of view), but they are not as good-looking markdown files as they should be.
The same was true about python (though testing scripts) are quite big and maybe we need to take a close look.

I only have several hooks in mind, which will seriously affect how the code is written and read in this repo.
After that, I'll be done with these PRs :)
Maybe.

examples/openshift/README.md
```diff
@@ -1,10 +1,8 @@
-OpenShift example
-=================
+# OpenShift example
```
consideRatio, line 1: :heart: I saw this before and was thinking that prettier would probably manage to fix this :)

consideRatio (approved):
This LGTM as well!

consideRatio:
@mathbunnyru this LGTM but now have merge conflicts that needs to be resolved

Issue #1306, opened 2021-05-16
jupyter/base-notebook: Hardcoded PATH Environment Variable in Dockerfile


lima4658:
Some developers need to change the PATH Environment Variable (e.g. create their own virtual environments and registrate/add their private notebooks)
You can create virtual environments (conda create ...; python -m venv ...) but the activation will fail due to the hardcoded PATH environment variable.

Solution:
Remove Line  'ENV PATH=$CONDA_DIR/bin:$PATH' from Dockerfile
But this will cause several **problems**:

- tini will not be found in ENTRYPOINT command line
- conda/jupyter commands will not be found in RUN command lines
- jupyter command will not be found during startup sequence

Workarounds:

- Install tini from Ubuntu repository (same Version as from conda) and remove lines
<'tini=0.18.0' && \> and 
<conda list tini | grep tini | tr -s ' ' | cut -d ' ' -f 1,2 >> $CONDA_DIR/conda-meta/pinned && \> from Dockerfile

- Add <export PATH=$CONDA_DIR/bin:$PATH && \> to the corresponding RUN statement sections in Dockerfile
- Regarding the 'run-hooks /usr/local/bin/before-notebook.d' command in the start.sh script:
Place a script conda_path.sh in directory /usr/local/bin/before-notebook.d containing the command 'export PATH=$CONDA_DIR/bin:$PATH'

Diff Dockerfile:

    49,50d48
    <     build-essential binutils binfmt-support \
    <     tini \
    70c68,69
    < ENV HOME=/home/$NB_USER \
    ---
    > ENV PATH=$CONDA_DIR/bin:$PATH \
    >     HOME=/home/$NB_USER \
    112d110
    <     export PATH=$CONDA_DIR/bin:$PATH && \
    120c118,119
    <     'pip' && \
    ---
    >     'pip' \
    >     'tini=0.18.0' && \
    121a121
    >     conda list tini | grep tini | tr -s ' ' | cut -d ' ' -f 1,2 >> $CONDA_DIR/conda-meta/pinned && \
    133,134c133
    < RUN export PATH=$CONDA_DIR/bin:$PATH && \
    <     conda install --quiet --yes \
    ---
    > RUN conda install --quiet --yes \
    159,163d157
    <
    < RUN mkdir -p /usr/local/bin/before-notebook.d && \
    <     printf '#!/bin/bash\n\nexport PATH=$CONDA_DIR/bin:$PATH\n\n' > \
[...]

PR #1308, opened 2021-05-19
Install tini from the ubuntu package

romainx:
Hello,

According to @lima4658 suggestion in #1306, `tini` is now installed from the ubuntu package.
The installation is cleaner because we can remove the corresponding pin of the conda package.

Also added a test to check that `tini` is used as the `PID` 1 when launching the `base-notebook` container.

Best


mathbunnyru:
LGTM. Also I would like to merge to get rid of the style error in master.

PR #1314, opened 2021-05-21
Update links

mathbunnyru:


mathbunnyru:
Some links were redirecting, some of them were not HTTPS (though HTTPS version is available), so I changed all these places.

PR #1315, opened 2021-05-21
Fix shellcheck warnings

mathbunnyru:


examples/docker-compose/notebook/up.sh
```diff
@@ -6,12 +6,12 @@ set -e
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-USAGE="Usage: `basename $0` [--secure | --letsencrypt] [--password PASSWORD] [--secrets SECRETS_VOLUME]"
+USAGE="Usage: $(basename "$0") [--secure | --letsencrypt] [--password PASSWORD] [--secrets SECRETS_VOLUME]"
```
consideRatio, line 9: quotes within quotes probably causing an issue here.

base-notebook/start.sh
```diff
@@ -142,6 +143,6 @@ else
 
     # Execute the command
     run-hooks /usr/local/bin/before-notebook.d
-    echo "Executing the command: ${cmd[@]}"
+    echo "Executing the command:" "${cmd[@]}"
```
consideRatio, line 146: This was a surprising change to me, do you understand what motivates it in shellcheck?

examples/docker-compose/notebook/up.sh
mathbunnyru: It seems to be ok to use quotes in quotes for command substitution in a new style.
https://stackoverflow.com/questions/17984958/what-does-it-mean-in-shell-when-we-put-a-command-inside-dollar-sign-and-parenthe

base-notebook/start.sh
```diff
@@ -39,12 +40,12 @@ run-hooks () {
 run-hooks /usr/local/bin/start-notebook.d
 
 # Handle special flags if we're root
-if [ $(id -u) == 0 ] ; then
+if [ "$(id -u)" == 0 ] ; then
```
consideRatio, line 43: I think this will fail unless...

```suggestion
if [ "$(id -u)" == "0" ] ; then
```

base-notebook/start.sh
mathbunnyru: This is the error with some explanation:
https://github.com/koalaman/shellcheck/wiki/SC2145

examples/docker-compose/notebook/up.sh
consideRatio: Ah thanks for clarifying this

base-notebook/start.sh
mathbunnyru: I created a simple script:

```
#!/bin/bash

if [ "$(id -u)" == 0 ] ; then
    echo "YES"
else
    echo "NO"
fi
```

It works both with or without quotes.

consideRatio: Oh, okay

mathbunnyru:
I will investigate why build fails later.
I'm no bash expert :(

base-notebook/start.sh
```diff
@@ -67,23 +68,23 @@ if [ $(id -u) == 0 ] ; then
     # Ex: default NFS/EFS (no auto-uid/gid)
     if [[ "$CHOWN_HOME" == "1" || "$CHOWN_HOME" == 'yes' ]]; then
         echo "Changing ownership of /home/$NB_USER to $NB_UID:$NB_GID with options '${CHOWN_HOME_OPTS}'"
-        chown $CHOWN_HOME_OPTS $NB_UID:$NB_GID /home/$NB_USER
+        chown "$CHOWN_HOME_OPTS" "$NB_UID:$NB_GID" "/home/$NB_USER"
```
consideRatio, line 71: Hmmm, if for example CHOWN_HOME_OPTS are "--flag1 --flag2" then it is wrong to add quotes here I think. Note sure what CHOWN_HOME_OPTS may be though.

base-notebook/start.sh
```diff
@@ -67,23 +68,23 @@ if [ $(id -u) == 0 ] ; then
...
     fi
-    if [ ! -z "$CHOWN_EXTRA" ]; then
-        for extra_dir in $(echo $CHOWN_EXTRA | tr ',' ' '); do
+    if [ -n "$CHOWN_EXTRA" ]; then
+        for extra_dir in $(echo "$CHOWN_EXTRA" | tr ',' ' '); do
             echo "Changing ownership of ${extra_dir} to $NB_UID:$NB_GID with options '${CHOWN_EXTRA_OPTS}'"
-            chown $CHOWN_EXTRA_OPTS $NB_UID:$NB_GID $extra_dir
+            chown "$CHOWN_EXTRA_OPTS" "$NB_UID:$NB_GID" "$extra_dir"
```
consideRatio, line 76: Same comment about this, it could be a mistake to add quotes to CHOWN_EXTRA_OPTS.

base-notebook/start.sh
mathbunnyru: Yes, I think you're right. Changed the code and added `shellcheck disable`.

consideRatio:
Nice LGTM!

consideRatio (approved):

PR #1331, opened 2021-05-25
Updating the recipe for 1.0.0 release of delta

tbluejeans:
Hello everyone,

Seeing that delta is now available through pypy and includes python methods for downloading the delta packages, I have updated the recipe.

Best! 
Tony

### DIFF
--- a/docs/using/recipes.md
+++ b/docs/using/recipes.md
@@ -524,9 +524,11 @@ Please note that the [Delta Lake](https://delta.io/) packages are only available
 FROM jupyter/pyspark-notebook:latest
 
 ARG DELTA_CORE_VERSION="1.0.0"
+RUN pip install --quiet --no-cache-dir delta-spark==${DELTA_CORE_VERSION} && \
+     fix-permissions "${HOME}" && \
+     fix-permissions "${CONDA_DIR}"
 
 USER root
-RUN pip install delta-spark==${DELTA_CORE_VERSION}
 
 RUN echo 'spark.sql.extensions io.delta.sql.DeltaSparkSessionExtension' >> $SPARK_HOME/conf/spark-defaults.conf && \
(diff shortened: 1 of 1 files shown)


docs/using/recipes.md
```diff
@@ -523,18 +523,20 @@ Please note that the [Delta Lake](https://delta.io/) packages are only available
 ```dockerfile
 FROM jupyter/pyspark-notebook:latest
 
-ARG DELTA_CORE_VERSION="0.8.0"
+ARG DELTA_CORE_VERSION="1.0.0"
 
 USER root
+RUN pip install delta-spark==${DELTA_CORE_VERSION}
```
mathbunnyru, line 529:
```suggestion
ARG DELTA_CORE_VERSION="1.0.0"
RUN pip install --quiet --no-cache-dir delta-spark==${DELTA_CORE_VERSION}

USER root
```

If there's no reason to install under root (I don't think there is one), we should install everything from jovyan.

mathbunnyru: Also, let's not cache these files, because they might be quite big and caching will significantly increase docker image size.

mathbunnyru (changes requested):

docs/using/recipes.md
romainx: I agree with @mathbunnyru.

tbluejeans: Not sure, why we are receiving an error on the checks. I think it is unrelated to the actual pull request as it seems to be missing docker.  

```
Executable `docker` not found
```

mathbunnyru:
I asked to add pre-commit.ci. Unfortunately, it seems that it doesn't work with docker-related pre-commit hooks by default. It has nothing to do with your PR and I will merge it.

mathbunnyru:
I fixed the issue in https://github.com/jupyter/docker-stacks/pull/1338

PR #1335, opened 2021-05-26
Unify bash variables usage and add quotes where needed

mathbunnyru:


base-notebook/Dockerfile
```diff
@@ -60,14 +60,14 @@ RUN apt-get update --yes && \
 # Configure environment
 ENV CONDA_DIR=/opt/conda \
     SHELL=/bin/bash \
-    NB_USER=$NB_USER \
-    NB_UID=$NB_UID \
-    NB_GID=$NB_GID \
+    NB_USER="${NB_USER}" \
+    NB_UID=${NB_UID} \
+    NB_GID=${NB_GID} \
```
consideRatio, line 65: In a way, this difference between quotes or not on NB_USER vs NB_UID is not making sense, as both will be understood as strings in the end as they are set as environment variables can only be string values. I'm not sure what happens if someone sets NB_UID to "123 HELLO", would quote have mattered then?

I'd be fine leaving this as it is or adding quotes here as well.

base-notebook/Dockerfile
```diff
@@ -106,20 +106,20 @@ WORKDIR /tmp
...
+    /bin/bash "${miniforge_installer}" -f -b -p "${CONDA_DIR}" && \
     rm "${miniforge_installer}" && \
     # Conda configuration see https://conda.io/projects/conda/en/latest/configuration.html
-    echo "conda ${CONDA_VERSION}" >> $CONDA_DIR/conda-meta/pinned && \
+    echo "conda ${CONDA_VERSION}" >> "${CONDA_DIR}/conda-meta/pinned" && \
     conda config --system --set auto_update_conda false && \
     conda config --system --set show_channel_urls true && \
-    if [ ! $PYTHON_VERSION = 'default' ]; then conda install --yes python=$PYTHON_VERSION; fi && \
-    conda list python | grep '^python ' | tr -s ' ' | cut -d '.' -f 1,2 | sed 's/$/.*/' >> $CONDA_DIR/conda-meta/pinned && \
+    if [ ! ${PYTHON_VERSION} = 'default' ]; then conda install --yes python=${PYTHON_VERSION}; fi && \
```
consideRatio, line 115: I'm not confident about either if this should have quotes or not. I'm also not confident about the difference between `[` and `[[` in the context of `sh` and `bash` respectively.

Without clear understanding, I would have guessed something like this would make sense in a bash script where we enforce quotes around env vars consistently.

```suggestion
    if [[ "${PYTHON_VERSION}" != "default" ]]; then conda install --yes python="${PYTHON_VERSION}"; fi && \
```

I'm not confident enough about this to say though, I bet I may have even introduced a syntax error above.

base-notebook/Dockerfile
consideRatio: I suggest we don't think about this further and just go with the PR as it is. No change requested!

consideRatio (approved):
Thorough work on this @mathbunnyru! This LGTM!

base-notebook/Dockerfile
mathbunnyru: > In a way, this difference between quotes or not on NB_USER vs NB_UID is not making sense

There is actually a difference.
NB_USER - is a username, it can be sth like "my best username ever". It can be with spaces so quoting it makes sense.
While NB_UID - is a Linux uid, which is numerical value. Quoting doesn't help (but doesn't make it worse).

```
# usermod -u 100 some_user
usermod: user 'some_user' does not exist
# usermod -u 100text some_user
usermod: invalid user ID '100text'
```

consideRatio: I was thinking that NB_UID will hold a numerical value we assume, but it is a string representing a numerical value whenever we hold it in an environment variables, so in that sense, it is the same as NB_NAME even though we expect NB_UID to only contain numbers and no spaces etc.

mathbunnyru: Yes, but, we can safely assume that NB_UID will have no spaces, while NB_USER easily might.
I don't like too excessive quoting where it won't help, because it might make reading a bit more difficult.
That's why I quoted only partially based on context and the meaning of variable.

consideRatio: I agree!

base-notebook/Dockerfile
mathbunnyru: Let's try this one out :)

mathbunnyru: It seems to work.

mathbunnyru:
I'm going to merge this one - it makes the style consistent and bash variables usage safer (the spaces problem).

PR #1319, opened 2021-05-21
Update packages

mathbunnyru:

PR #1340, opened 2021-06-01
Replace vincent with altair

mathbunnyru:
This was already done in https://github.com/jupyter/docker-stacks/pull/1319 (and approved there), but it's better to have a separate commit, which is easier to revert if something bad happens. Also, it's not an update, so also doesn't make a lot of sense there.


mathbunnyru:
As this was already approved in another PR by @consideRatio and @romainx, merging.

Issue #1203, opened 2020-12-27
Improve monolithic build


romainx:
Hello,

As discussed in two recents PR (#1196 and #1202) the current build seems to be too monolithic.

>  If we add additional stacks, the build will grow and will become error prone and it will become a pain to publish all the stacks. The build takes between 40 min and 1 hour and if a single thing fails (a test, a package installation) everything fails. 

We should benefit from GitHub actions that have been introduced by @parente in PR #1132 and / or by proposal made by @bollwyvl in his [comment](https://github.com/jupyter/docker-stacks/pull/1202#issuecomment-751359223). My first idea is to split the treatment of each image in different jobs that can be run in parallel (at least for some of them) with dependencies between them to achieve more flexibility. We could benefit from advanced features like [artifact sharing](https://docs.github.com/en/free-pro-team@latest/actions/guides/storing-workflow-data-as-artifacts) to achieve this.

A job could be responsible of the whole stuff for an image:
- Lint
- Build
- Test
- Hook / Manifest -> see also another improvement on this point #1140
- Tag -> support more flexible tagging mechanism like tagging images with main components version (#976)
- Push

All these cool new stuff with the ability to build things locally.
So this is not so simple, there is a lot of things to solve / decide but there is room for improvement 😄.

Best.

trallard:
I recently offered to help with some tasks for #1407, after reading @mathbunnyru  proposal and additional comments it is clear that this issue - i.e. improving the monolithic build should be tackled first to ensure we have a solid foundation. 

@manics  I would be happy to explore and work on a POC leveraging GH actions + doit (for x86):
1. I can start with a smaller scoped POC to gauge how doit + GH actions would work for our scenario 
2. If this is indeed the best path moving forward, this can be then further expanded to allow for images to be built/tested/tagged/etc. in parallel

Some advantages I see from this approach are:
1. We can leverage the existing GH actions work introduced by @parente 
2. Exploring the use of doit vs Make - this should help with multiple automation tasks, also might make it easier for other folks to contribute to since doit is Python-based, and in general, help make thee workflows a bit more flexible
3. Having this sorted out should also allow us to identify the best path moving forward to leverage cache or the use of artefacts to start optimising the builds in subsequent runs 


thoughts?

romainx:
@trallard In a recent PR (#1604) I've explored a way to standardise by using `docker buildx bake`.
> Buildx also aims to provide support for high-level build concepts that go beyond invoking a single build command. [...]
[...]

PR #1341, opened 2021-06-01
Run pre-commit in separate workflow

mathbunnyru:
Several reasons for this change:
- PRs like this shouldn't trigger docker build. https://github.com/jupyter/docker-stacks/pull/1339
- A bit faster CI (1 minute)
- Unfortunately, we can't fully rely on pre-commit.ci, because we use Docker-based linter and docker is not available in pre-commit.ci

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -7,14 +7,17 @@ on:
       - "binder/**"
       - "docs/**"
       - "examples/**"
+      - ".pre-commit-config.yaml"
   push:
     branches:
       - master
+      - main
     paths-ignore:
       - "*.md"
       - "binder/**"
       - "docs/**"
       - "examples/**"
+      - ".pre-commit-config.yaml"
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


mathbunnyru:
Related to https://github.com/jupyter/docker-stacks/issues/1203

mathbunnyru:
@consideRatio please, do not merge, while there is something in the master which is not yet built.
This will cause errors.
The reason for that is that in the docker builds we clone the docs, and at the end of the build update the docs.
And then there will be two commits, which change the docs in the different way, and only one will succeed.
So, the second build will not update manifests.

consideRatio:
Wooops ah, did i merge this PR and trigger a workflow running alongside another workflow triggered via some other PR that got merged independently?

mathbunnyru:
Yes, https://github.com/jupyter/docker-stacks/pull/1340 was merged 16 minutes before you merged, and that's why there are two simultaneous runs right now.

mathbunnyru:
@consideRatio I updated the description of the situation and created a mitigation PR.
The build failed as I described: https://github.com/jupyter/docker-stacks/runs/2717699712

PR #1342, opened 2021-06-01
Use python-version: 3.x in github actions

mathbunnyru:
This is already used in sphinx, works fine.
And it will be easier to start using new features for the new code.

(Changing the python version here doesn't change the content of the docker image).


mathbunnyru:
We're using the latest versions of python packages (not fixing versions), so not fixing python versions for building also seems to be fine.
Merging this to have a successful build.

PR #1346, opened 2021-06-01
Use docker/login-action instead of manual docker login

mathbunnyru:
This is more idiomatic on GitHub, doesn't create a warning, supports plenty of registries (https://github.com/marketplace/actions/docker-login) and does automatic logout in the end of the job.


mathbunnyru:
Unfortunately, this is not checked in the PR, only in master.

mathbunnyru:
Let's then check it in master.

mathbunnyru:
It worked just fine.

Снимок экрана 2021-06-01 в 17 42 00

Issue #1352, opened 2021-06-01
Python version changes in pyspark-notebook (and all-spark-notebook)


mathbunnyru:
Right now base image installs python 3.9.2.
But, when installing pyarrow, actually, python is updated to 3.9.4!

I'm not sure this is expected at all.

https://github.com/jupyter/docker-stacks/runs/2719734258#step:5:6118

Also, it's quite easy to see in the wiki https://github.com/jupyter/docker-stacks/wiki, that `pyspark-notebook` `all-spark-notebook` get tag `python-3.9.4`, while everything else gets `python-3.9.2`.

mathbunnyru:
We have quite a difficult way to pin python version.
We should probably fix the version in a more strict way, shouldn't we?

https://github.com/jupyter/docker-stacks/blob/master/base-notebook/Dockerfile#L116

mathbunnyru:
Found the place where it was added: https://github.com/jupyter/docker-stacks/pull/778#discussion_r243851245

PR #1351, opened 2021-06-01
Use mamba instead of conda in spark images

mathbunnyru:
I decided to try mamba separately and I will take a look what changes when using mamba.

Fix: https://github.com/jupyter/docker-stacks/issues/1352

PR #778, opened 2018-12-22
Upgrade Python in scipy-notebook to 3.7

alasdairtran:
This PR pins the Python version to prevent Conda from downgrading (e.g. from 3.7 to 3.6). It also updates two packages, `numba` and `h5py`, to make them compatible with Python 3.7.  Fixes issue #768

PR #1354, opened 2021-06-01
Pin python version in a strict way

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1352

This is another approach to fix the python version (mamba doesn't change the python version in https://github.com/jupyter/docker-stacks/pull/1351).

For now, I'm not 100% sure about this PR, because @parente wanted the python patch version to be floating (https://github.com/jupyter/docker-stacks/pull/778#discussion_r243851245).

I want to see how it's gonna work for pyspark image, which was updating python version, even it might not be needed.
Also, it's nice to see the same tags in all the images and to control python version via a variable (it was unexpected for me that two images had different python versions at all).


mathbunnyru:
Seems to work fine.
This PR:
https://github.com/jupyter/docker-stacks/pull/1354/checks?check_run_id=2722386028#step:5:6153
Built image size: 3.32GB

Master:
https://github.com/jupyter/docker-stacks/runs/2719734258#step:5:6155
Built image size: 3.41GB

Aslo saves 90 MB of information.

base-notebook/Dockerfile
```diff
@@ -113,7 +113,7 @@ RUN wget --quiet "https://github.com/conda-forge/miniforge/releases/download/${m
     conda config --system --set auto_update_conda false && \
     conda config --system --set show_channel_urls true && \
     if [ ! $PYTHON_VERSION = 'default' ]; then conda install --yes python=$PYTHON_VERSION; fi && \
-    conda list python | grep '^python ' | tr -s ' ' | cut -d '.' -f 1,2 | sed 's/$/.*/' >> $CONDA_DIR/conda-meta/pinned && \
+    conda list python | grep '^python ' | tr -s ' ' | cut -d ' ' -f 1,2 >> $CONDA_DIR/conda-meta/pinned && \
```
consideRatio, line 116: This is too technical for me to parse, can you describe what is changed in what was previously written to the pinned file vs what is now written to the pinned file?

mathbunnyru: For example, the current version of python is `3.9.2`.
Currently, in the master branch we will pin the version like this `python 3.9.*`.
So, during conda update (and even install), we might update (and even downgrade) python version to `3.9.x`.

The change: write exactly `python 3.9.2`. This way we pin python versions in a strict way and it won't change during conda update / install.

consideRatio: Excellent thank you!

consideRatio:
### Questions in my mind
- Are we in agreement to want to pin our Python version while installing the docker image?
- Should we have an entry in the `$CONDA_DIR/conda-meta/pinned` file after our build is complete?
  - If so, how tight should that version pin be?

mathbunnyru:
> ### Questions in my mind
> * Are we in agreement to want to pin our Python version while installing the docker image?
>   
>   * I think we should enforce a single python version to be used across all our build steps, and if we need to rely on adding a line to the pinned file when using `mamba`, then I support that at least during the build.

As this review shows, we can rely on adding a line to the pinned file (python version is not changing).

> * Should we have an entry in the `$CONDA_DIR/conda-meta/pinned` file after our build's various updates to conda packages are complete?
>   
>   * If so, how tight should that version pin be?

I think, it's ok to pin the version in a strict way - we provide our docker images with complete python tag like `jupyter/all-spark-notebook:python-3.9.4`, so it won't be obvious for users that they might change the python version after some conda installs / updates.

mathbunnyru:
And, I think we're updating quite a lot our mambaforge installation (at least we try), so we will be always with quite recent version of python).

mathbunnyru:
Let's try this one out.
Now, after switching to mamba, the python version is not changing.
But let's try to merge this to see how it goes for the users. It's easy to revert in case something goes wrong.

PR #1367, opened 2021-06-14
Fix hadolint docker version

mathbunnyru:


mathbunnyru:
@consideRatio you were too fast merging https://github.com/jupyter/docker-stacks/pull/1366

Unfortunately we have to pin docker version manually.
Even though I fixed the upstream issue (https://github.com/hadolint/hadolint/issues/628), @lorenzo forgot to update https://github.com/hadolint/hadolint/blob/master/.pre-commit-hooks.yaml before release.

So, for now, we have to do it manually :(

mathbunnyru:
Btw, pre-commit.ci is quite a nice tool, isn't it? :)

mathbunnyru:
Another thing we could do is we could use hadolint hook, which uses system-wide hadolint.
But, we will have to install it (it means pinning the version in a different place and it won't probably be automatically updated).

I do like pre-commit updates, so for now I think it's better to stick with docker version and update the tag manually.

consideRatio:
Ah, the duplicated version that I had to update as well was just out of the range to be previewed in the change preview.

consideRatio:
> Another thing we could do is we could use hadolint hook, which uses system-wide hadolint.

I'm not sure if that implies increasing the complexity of the developer setup by adding a "please also install ...", but if that is the case then I'm very favorable of using whatever we can install with pre-commit automatically. I don't think we will run into issues by having an old linter, but I can see myself or others struggling to install tools on various platforms.

> Btw, pre-commit.ci is quite a nice tool, isn't it? :)

Yeah I appreciate the automatic PRs updating things, I'd love for it to be part of dependabot's automation in the future to have one less tool for that kind of job.

Issue #1019, opened 2020-02-19
Possible docker image for raspberry pis


jiwidi:
Hi!

I was working recently with some Docker images for jupyterlab to run in raspberry pis and I was wondering if an image like that one could be worth including in this notebook or if its of value for the Jupyter team.  I will very happy to work towards a contribution for this but first want to check wether there is value for it.

Thanks!

parente:
@jiwidi Thanks for your interest in extending the Jupyter ecosystem. https://github.com/jupyter/docker-stacks/issues/899 has some discussion and movement toward supporting rpi images as community-maintained docker stacks. Perhaps you'd be interested in contributing to that effort?

jiwidi:
> @jiwidi Thanks for your interest in extending the Jupyter ecosystem. #899 has some discussion and movement toward supporting rpi images as community-maintained docker stacks. Perhaps you'd be interested in contributing to that effort?

I think its a good idea if we both join forces, I left him a comment and lets see how it goes :) 

Thanks for the referal!

AvverbioPronome:
@jiwidi don't these images run there? 

What's the issue? pip takes too long compiling? (add piwheels)

jiwidi:
> @jiwidi don't these images run there?
> 
> What's the issue? pip takes too long compiling? (add piwheels)

What do you mean? If you mean if the normal images can be run in raspberry pi, no they can't. Because of the ARM architecture of rpis there is need for some special tricks or bypass to compile nodejs and other libraries

AvverbioPronome:
I've started looking into it and the first obvious problem is that `base-notebook` uses miniconda x86_64 explicitly. https://github.com/jupyter/docker-stacks/blob/master/base-notebook/Dockerfile#L79

Latest miniconda armv7 is `2015-08-24 11:01:14` no good. https://repo.continuum.io/miniconda/

So, yeah, making this image "platform independent" means rebuilding on ubuntu packages or pip (pip is fragile, is fragile a lot.)

step21:
You can check my work here https://hub.docker.com/repository/docker/step21/jupyter-minimal-notebook (base and minimal so far, soon also scipy-notebook once scikit-image gets merges)
It has both aarch64 and armv7l images, but unless someone gets armv7l packages into conda-forge, I do not think it makes sense to maintain armv7l images.

sakuraiyuta:
I'm interested this topic too.
`aarch64` docker images builded with my repository are here.

[...]

PR #1368, opened 2021-06-16
Make the base & minimal notebook containers not amd specific (e.g. support building for arm64)

holdenk:
I run a small Kubernetes cluster with a mix of arm64 & x86_64 hosts so I can use Jupyter I changed the Dockerfile to support being cross-built. (for additional context see https://scalingpythonml.com/2020/12/12/deploying-jupyter-lab-notebook-for-dask-on-arm-on-k8s.html )

Even without the cross-build, this is still useful for anyone who wants to build the Jupyter container on any other non-x86_64 platform.

Right now this currently has two lists of containers: those that support cross-building & those who's Dockerfiles need to be updated to support cross-building. `make build-all` "does the right thing" in that it cross-builds containers which support cross-building and builds single arch containers for those that are not yet ported.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -31,27 +31,6 @@ jobs:
       !contains(github.event.head_commit.message, 'ci skip') &&
       !contains(github.event.pull_request.title, 'ci skip')
     steps:
-      - name: Should we push this image to a public registry?
-        run: |
-          if [ "${{ startsWith(github.ref, 'refs/tags/') || (github.ref == 'refs/heads/main') }}" = "true" ]; then
-              # Empty => Docker Hub
-              echo "REGISTRY=" >> $GITHUB_ENV
-          else
-              echo "REGISTRY=localhost:5000/" >> $GITHUB_ENV
-          fi
-      - uses: actions/checkout@v2
-      # Setup docker to build for multiple platforms, see:
-      # https://github.com/docker/build-push-action/tree/v2.4.0#usage
-      # https://github.com/docker/build-push-action/blob/v2.4.0/docs/advanced/multi-platform.md
-
-      - name: Set up QEMU (for docker buildx)
-        uses: docker/setup-qemu-action@25f0500ff22e406f7191a2a8ba8cda16901ca018 # associated tag: v1.0.2
-
-      - name: Set up Docker Buildx (for multi-arch builds)
-        uses: docker/setup-buildx-action@2a4b53665e15ce7d7049afb11ff1f70ff1610609 # associated tag: v1.1.2
-        with:
-          # Allows pushing to registry on localhost:5000
-          driver-opts: network=host
       - name: Clone Main Repo
         uses: actions/checkout@v2
         with:
@@ -88,3 +67,6 @@ jobs:
         with:
           username: ${{secrets.DOCKERHUB_USERNAME}}
           password: ${{secrets.DOCKERHUB_TOKEN}}
+      - name: Push Images to DockerHub
+        if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main'
+        run: make -C main push-all
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -31,6 +31,28 @@ jobs:
       !contains(github.event.head_commit.message, 'ci skip') &&
       !contains(github.event.pull_request.title, 'ci skip')
     steps:
+      - name: Should we push this image to a public registry?
+        run: |
+          if [ "${{ startsWith(github.ref, 'refs/tags/') || (github.ref == 'refs/heads/main') }}" = "true" ]; then
+              # Empty => Docker Hub
+              echo "REGISTRY=" >> $GITHUB_ENV
+          else
+              echo "REGISTRY=localhost:5000/" >> $GITHUB_ENV
+          fi
+      - uses: actions/checkout@v2
+
+      # Setup docker to build for multiple platforms, see:
+      # https://github.com/docker/build-push-action/tree/v2.4.0#usage
+      # https://github.com/docker/build-push-action/blob/v2.4.0/docs/advanced/multi-platform.md
+
+      - name: Set up QEMU (for docker buildx)
+        uses: docker/setup-qemu-action@25f0500ff22e406f7191a2a8ba8cda16901ca018 # associated tag: v1.0.2
+
+      - name: Set up Docker Buildx (for multi-arch builds)
+        uses: docker/setup-buildx-action@2a4b53665e15ce7d7049afb11ff1f70ff1610609 # associated tag: v1.1.2
+        with:
+          # Allows pushing to registry on localhost:5000
+          driver-opts: network=host
       - name: Clone Main Repo
         uses: actions/checkout@v2
         with:
@@ -67,6 +89,3 @@ jobs:
         with:
           username: ${{secrets.DOCKERHUB_USERNAME}}
           password: ${{secrets.DOCKERHUB_TOKEN}}
-      - name: Push Images to DockerHub
-        if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main'
-        run: make -C main push-all
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -40,7 +40,6 @@ jobs:
               echo "REGISTRY=localhost:5000/" >> $GITHUB_ENV
           fi
       - uses: actions/checkout@v2
-
       # Setup docker to build for multiple platforms, see:
       # https://github.com/docker/build-push-action/tree/v2.4.0#usage
       # https://github.com/docker/build-push-action/blob/v2.4.0/docs/advanced/multi-platform.md
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ build-cross/%: ## build the latest image for a stack on x86 and ARM
 	@docker images $(OWNER)/$(notdir $@):latest --format "{{.Size}}"
 
 build-all: $(foreach I,$(CROSS_IMAGES), build-cross/$(I) ) $(foreach I,$(X86_IMAGES), build/$(I) ) ## build all stacks
-build-test-all: build-all $(foreach I,$(ALL_IMAGES), test/$(I) ) ## build and test all stacks
+build-test-all: build-all test-all ## build and test all stacks
 
 check-outdated/%: ## check the outdated conda packages in a stack and produce a report (experimental)
 	@TEST_IMAGE="$(OWNER)/$(notdir $@)" pytest test/test_outdated.py
--- a/Makefile
+++ b/Makefile
@@ -5,20 +5,8 @@
 # Use bash for inline if-statements in arch_patch target
 SHELL:=bash
 OWNER?=jupyter
-# By default we still target single image amd64 & docker output type.
 
 # Need to list the images in build dependency order
-# These are images we can cross-build
-CROSS_IMAGES:= base-notebook \
-	minimal-notebook
-# These images that aren't currently supported for cross-building, your help is welcome.
-X86_IMAGES:= r-notebook \
-	scipy-notebook \
-	tensorflow-notebook \
-	datascience-notebook \
-	pyspark-notebook \
-	all-spark-notebook
-# All of the images
 ALL_IMAGES:=base-notebook \
 	minimal-notebook \
 	r-notebook \
@@ -41,18 +29,12 @@ help:
 
 build/%: DARGS?=
 build/%: ## build the latest image for a stack
-	docker buildx build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@) --platform "linux/amd64" --build-arg OWNER="${OWNER}" --push
+	docker build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@)
 	@echo -n "Built image size: "
 	@docker images $(OWNER)/$(notdir $@):latest --format "{{.Size}}"
 
-build-cross/%: DARGS?=
-build-cross/%: ## build the latest image for a stack on x86 and ARM
-	docker buildx build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@) --platform "linux/amd64,linux/arm64" --build-arg OWNER="${OWNER}" --push
-	@echo -n "Built image size: "
-	@docker images $(OWNER)/$(notdir $@):latest --format "{{.Size}}"
-
-build-all: $(foreach I,$(CROSS_IMAGES), build-cross/$(I) ) $(foreach I,$(X86_IMAGES), build/$(I) ) ## build all stacks
-build-test-all: build-all test-all ## build and test all stacks
+build-all: $(foreach I,$(ALL_IMAGES), build/$(I) ) ## build all stacks
+build-test-all: $(foreach I,$(ALL_IMAGES), build/$(I) test/$(I) ) ## build and test all stacks
 
 check-outdated/%: ## check the outdated conda packages in a stack and produce a report (experimental)
 	@TEST_IMAGE="$(OWNER)/$(notdir $@)" pytest test/test_outdated.py
@@ -113,6 +95,12 @@ pull/%: ## pull a jupyter image
 	docker pull $(DARGS) $(OWNER)/$(notdir $@)
 
 pull-all: $(foreach I,$(ALL_IMAGES),pull/$(I) ) ## pull all images
+
+push/%: DARGS?=
+push/%: ## push all tags for a jupyter image
+	docker push --all-tags $(DARGS) $(OWNER)/$(notdir $@)
+
+push-all: $(foreach I,$(ALL_IMAGES),push/$(I) ) ## push all tagged images
 
 run/%: DARGS?=
 run/%: ## run a bash in interactive mode in a stack
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@
 
 # Use bash for inline if-statements in arch_patch target
 SHELL:=bash
+ARCH:=$(shell uname -m)
 OWNER?=jupyter
 
 # Need to list the images in build dependency order
@@ -29,7 +30,7 @@ help:
 
 build/%: DARGS?=
 build/%: ## build the latest image for a stack
-	docker build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@)
+	docker buildx build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@) --push --platform linux/arm64,linux/amd64
 	@echo -n "Built image size: "
 	@docker images $(OWNER)/$(notdir $@):latest --format "{{.Size}}"
 
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,6 @@
 
 # Use bash for inline if-statements in arch_patch target
 SHELL:=bash
-ARCH:=$(shell uname -m)
 OWNER?=jupyter
 
 # Need to list the images in build dependency order
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,10 @@
 # Use bash for inline if-statements in arch_patch target
 SHELL:=bash
 OWNER?=jupyter
+# By default we still target single image amd64 & docker output type.
+# To cross build run make PLATFORM="linux/amd64,linux/arm64" BUILDX_OUTPUT_TYPE="registry"
+PLATFORM ?= "linux/amd64"
+BUILDX_OUTPUT_TYPE ?= "docker"
 
 # Need to list the images in build dependency order
 ALL_IMAGES:=base-notebook \
@@ -29,7 +33,7 @@ help:
 
 build/%: DARGS?=
 build/%: ## build the latest image for a stack
-	docker buildx build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@) --push --platform linux/arm64,linux/amd64
+	docker buildx build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@) --output=type="${BUILDX_OUTPUT_TYPE}" --platform "${PLATFORM}"
 	@echo -n "Built image size: "
 	@docker images $(OWNER)/$(notdir $@):latest --format "{{.Size}}"
 
--- a/Makefile
+++ b/Makefile
@@ -6,11 +6,19 @@
 SHELL:=bash
 OWNER?=jupyter
 # By default we still target single image amd64 & docker output type.
-# To cross build run make PLATFORM="linux/amd64,linux/arm64" BUILDX_OUTPUT_TYPE="registry"
-PLATFORM ?= "linux/amd64"
-BUILDX_OUTPUT_TYPE ?= "docker"
 
 # Need to list the images in build dependency order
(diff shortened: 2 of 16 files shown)


mathbunnyru:
Let's run the CI first to make sure it works.

mathbunnyru:
Ok, so there is a problem, please, fix it.

Makefile
```diff
@@ -4,6 +4,7 @@
 
 # Use bash for inline if-statements in arch_patch target
 SHELL:=bash
+ARCH:=$(shell uname -m)
```
mathbunnyru, line 7: I don't think this line is needed.
It can also be a source of confusion, because make might be run in the host environment, not the target.

base-notebook/Dockerfile
```diff
@@ -2,9 +2,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 # Ubuntu 20.04 (focal)
-# https://hub.docker.com/_/ubuntu/?tab=tags&name=focal
```
mathbunnyru, line 5: This line is ok, it doesn't hardcode arch

base-notebook/Dockerfile
```diff
@@ -2,9 +2,7 @@
...
-# OS/ARCH: linux/amd64
-ARG ROOT_CONTAINER=ubuntu:focal-20210416@sha256:86ac87f73641c920fb42cc9612d4fb57b5626b56ea2a19b894d0673fd5b4f2e9
+ARG ROOT_CONTAINER=ubuntu:focal
```
mathbunnyru, line 5:
```suggestion
ARG ROOT_CONTAINER=ubuntu:focal-20210416
```

base-notebook/Dockerfile
```diff
@@ -105,8 +103,7 @@ WORKDIR /tmp
 
 # Prerequisites installation: conda, mamba, pip, tini
 RUN wget --quiet "https://github.com/conda-forge/miniforge/releases/download/${miniforge_version}/${miniforge_installer}" && \
-    echo "${miniforge_checksum} *${miniforge_installer}" | sha256sum --check && \
```
mathbunnyru, line 108: I'm OK with not checking sha, because we don't do it in other places.

base-notebook/Dockerfile
```diff
@@ -105,8 +103,7 @@ WORKDIR /tmp
...
-    /bin/bash "${miniforge_installer}" -f -b -p "${CONDA_DIR}" && \
```
mathbunnyru, line 109: Please, leave this line as it was, because it's safer from the bash perspective.

mathbunnyru (changes requested):

mathbunnyru (changes requested):
Overall, I do approve this PR, it's a first step in making jupyter images work under arm64 and I think the direction is right.

There are several things I would like to ask though:

1. Does it break something for current users (I mean, simple `docker pull` or `FROM jupyter/some_image`) works as it was, right? I mean, it's ok to ask people building our images to have `dockerx`. But I don't want to ask all the users to install `dockerx` and hope it's still gonna work.

2. Let's update docs or add the information to the docs, that we have some support of arm64-based images.

3. Does the tagging work well with 2 arches?

4. Let's not forget about build times - we need to be careful here, because we will be building for the not-host platform.

5. Please, fix my comments and I would like this PR to have at least 2 approvals. So, @consideRatio @romainx @parente tell us what you think.

mathbunnyru:
https://github.com/jupyter/docker-stacks/pull/1202 might be a great help because very similar thing was done by @romainx in the past (but it was'nt merged).

Makefile
```diff
@@ -29,7 +30,7 @@ help:
 
 build/%: DARGS?=
 build/%: ## build the latest image for a stack
-	docker build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@)
+	docker buildx build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@) --push --platform linux/arm64,linux/amd64
```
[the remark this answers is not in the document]
manics: Currently if you use `docker buildx` to build for multiple architectures there's no way of accessing any of the images without pushing them somewhere, but it has the big advantage of creating a _manifest_ for the tag: this means anyone running `docker pull image:tag` will automatically pull a matching architecture. You could for example push it to a local registry, which is what we do in JupyterHub:
- https://github.com/jupyterhub/jupyterhub/blob/6988d740015fb947f4de21b613f5255ae277f89c/.github/workflows/release.yml#L75-L80

The alternative is to build each architecture one at a time, and create the docker manifest separately.
- https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

manics:
In case anyone missed it there's a discussion in https://github.com/jupyter/docker-stacks/issues/1019 about ARM64 support

Makefile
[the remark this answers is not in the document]
mathbunnyru: Thanks! So, I think this PR will also have to add a local registry to make it work.

Makefile
holdenk: Good call, thanks I'll drop it

base-notebook/Dockerfile
holdenk: Sounds good, I'll restore it.

base-notebook/Dockerfile
holdenk: Ok makes sense, so were pinning to a date rather than a specific sha, I dig it :)

Makefile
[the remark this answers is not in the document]
holdenk: So provided the user sets the "owner" variable to an account that they own it should work without needing to setup a local registry.

base-notebook/Dockerfile
```diff
@@ -37,9 +33,11 @@ ARG miniforge_python="Mambaforge"
...
 ARG miniforge_version="${conda_version}-${miniforge_patch_number}"
 # Miniforge installer
-ARG miniforge_installer="${miniforge_python}-${miniforge_version}-Linux-${miniforge_arch}.sh"
-# Miniforge checksum
-ARG miniforge_checksum="9eb335cb559644a6e462c077ebc129af51b7329817574fb707b994dafdddf2af"
+RUN export arch=$(uname -m) && \
+    if [ "$arch" == "aarm64" ]; then \
+      arch="arm64"; \
+    fi; \
+RUN export miniforge_installer="${miniforge_python}-${miniforge_version}-Linux-${arch}.sh"
```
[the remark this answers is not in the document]
holdenk: Sure

base-notebook/Dockerfile
holdenk: ah my bad I didn't notice I dropped the quotes, thanks for catching that.

holdenk:
> Overall, I do approve this PR, it's a first step in making jupyter images work under arm64 and I think the direction is right.
> 
> There are several things I would like to ask though:
> 
> 1. Does it break something for current users (I mean, simple `docker pull` or `FROM jupyter/some_image`) works as it was, right? I mean, it's ok to ask people building our images to have `dockerx`. But I don't want to ask all the users to install `dockerx` and hope it's still gonna work.
vanilla docker pull "does the right thing" and pulls the matching arch.
> 2. Let's update docs or add the information to the docs, that we have some support of arm64-based images.
Makes sense, I assume I should update the wiki would be done after we're all on board with the PR? 
Are there other places I should update docs? I did some poking around in the DOCS folder and added some parts about setting the OWNER env variable for contributing.
> 3. Does the tagging work well with 2 arches?
Docker tagging does if that's what we're referencing?
> 4. Let's not forget about build times - we need to be careful here because we will be building for the not-host platform.
True, one option would be to let users specify which archs they want to build for instead of just forcing cross-build arm64 & amd64, what do you think?
> 5. Manifest creation needs to be fixed - I think if tags stay the same between images (they will stay the same if we use the same versions of python packages), then it makes sense to only change `Build manifest` sections here https://github.com/jupyter/docker-stacks/wiki because they are definitely gonna be different.
I'm a little fuzzy on this part can you elaborate? Sorry relatively new to the project, thanks for your help :)
> 6. Please, fix my comments and I would like this PR to have at least 2 approvals. So, @consideRatio @romainx @parente tell us what you think.

Makefile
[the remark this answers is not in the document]
mathbunnyru: But we want to be able to build images without pushing them to the real docker hub.
That's how it's working in the CI right now.
We only push to the hub, after the merge to master (you can see it in github docker workflow).

So, to be able to do that, I think we will have to setup a local registry.

docs/contributing/features.md
```diff
@@ -42,6 +42,7 @@ If there's agreement that the feature belongs in one or more of the core stacks:
    If you use `make`, call:
 
    ```bash
+   OWNER={yourdockerhubusername}
    make build/somestack-notebook
```
mathbunnyru, line 46: This won't work, because make will still use `OWNER?=jupyter` line instead of a variable you set.

mathbunnyru (changes requested):

consideRatio:
@holdenk you wrote a reply quoting a long response, i cant see what you have written inside of that. Can you update that comment so that it is clear what you say vs what @mathbunnyru sais?

Makefile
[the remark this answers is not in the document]
mathbunnyru: @manics how do manage local docker image?
You have to set up the localhost registry manually, right?

If I'm right, we will have to ask our developers, if they want to run the build locally, to run the registry first (it's quite easy I suppose).
That's fine by me, but I would like to know all the implications of this change before merging it.
And I think it's much better that having to force everyone to push to the hub, that's why I'm asking for changes.

holdenk:
Thanks, sorry I forgot about the github formatting, updated the comment.

Makefile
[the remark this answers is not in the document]
mathbunnyru: An example given by @manics is really useful.

https://github.com/jupyterhub/jupyterhub/blob/6988d740015fb947f4de21b613f5255ae277f89c/.github/workflows/release.yml#L75-L80

They're dealing with the same problem as we.

docs/contributing/features.md
holdenk: Ah yeah I'll need to update the owner refs (per https://stackoverflow.com/questions/24263291/define-a-makefile-variable-using-a-env-variable-or-a-default-value ) as well.

mathbunnyru: We probably won't need that after dealing with local registry as suggested above.

mathbunnyru:
> 2. Makes sense, I assume I should update the wiki would be done after we're all on board with the PR?
> Are there other places I should update docs? I did some poking around in the DOCS folder and added some parts about setting the OWNER env variable for contributing.

I think we were not actually telling anything about the arch of our images.
The only things I would add are:
- in `README.md` and `docs/index.rst`: `For now, we support these architectures:...`
- if we need to setup local registry, we need to add some info on how to do that (probably a command in `Makefile` and some info in docs)

> 3. Docker tagging does if that's what we're referencing?

Yes, I meant docker tagging.
If it works well, then it's great.

> 4. True, one option would be to let users specify which archs they want to build for instead of just forcing cross-build arm64 & amd64, what do you think?

Actually, for now, let's first try to make it work on GitHub - for that you will probably have to change our docker github workflow. Then we'll see how much it takes to build the images.
There are different strategies on dealing with docker build time problem and I'm not sure which one is the best.
So, for now, let's keep it in mind but focus on the more important stuff.

> 5. I'm a little fuzzy on this part can you elaborate? Sorry relatively new to the project, thanks for your help :)

https://github.com/jupyter/docker-stacks/tree/master/tagging
There is a short README, I think it should be helpful

mathbunnyru:
The most important things for now is to make it work in GitHub actions.

holdenk:
Makes sense, I'll give it a shot with getting it working in GHA :)

mathbunnyru:
@holdenk please, tell me, if you need a hand with this PR.
A lot of people are requesting arm images from time to time, so, if you have some questions or troubles, feel free to ask.

holdenk:
Thanks @mathbunnyru I hurt my hand on Friday so I've been running a little slower than planned. I'm going to make this default to local registry as we discussed.

holdenk:
Actually, now that I look at the "build docker image" CI step which naturally doesn't have mixed arch support, I think I'll switch this is use buildx with single image mode by default and then for "publishing" people can switch to registry.

mathbunnyru:
> Thanks @mathbunnyru I hurt my hand on Friday so I've been running a little slower than planned. I'm going to make this default to local registry as we discussed.

Good luck with the recovery process.

mathbunnyru (changes requested):
Please, tell me when it's ready to review, I will convert it to a draft for now.

holdenk:
@mathbunnyru  I think it should be ready for review, it works for me locally and it looks like it works in CI.

mathbunnyru (changes requested):
Ok, then I will review the current state.

This PR seems to add arm support, but it actually doesn't add it.

What I mean by adding arm support:
- images are built both for x86_64 and arm
- it's possible both locally and in the GitHub CI
- images are pushed to the Docker hub
- wiki build manifest is updated
- docs are updated to represent that we support 2 arches

I mean, it's great, that with this small change we can build arm images.
But until we make it work ass I told I can't even tell if your approach is gonna work or not.

holdenk:
Thanks for the reminder about the documentation. I've updated it. The language isn't quite what you suggested since I don't have permission to push to the docker hub, and I don't want to get people confused trying to use the images before they are pushed.

So it can push if you follow the instructions in the `Makefile`, for example running `make build/base-notebook OWNER="holdenk" PLATFORM="linux/amd64,linux/arm64" BUILDX_OUTPUT_TYPE="registry"` produces the cross-image container at https://hub.docker.com/layers/123936974/holdenk/base-notebook/latest/images/sha256-7a7ca53585413772d1b6798df959d84a47b6d187919ef2dfe541fa7c01392a0c?context=explore (note you can see both ARM & x86 images).

For "wiki build manifest is updated," I'd be happy to do that, but I don't know what you are asking for to be updated in the build manifest wiki.


For "images are pushed to the Docker hub," I don't have permission to push images to your docker hub account, but I'm hoping showing that they can be pushed to my account as above does the trick.


Building ARM images inside x86 containers on GithubActions is probably doable, but I'm not the person for that task. Similarly, it looks like it's possible to get ARM GitHub action runners, but I don't work with Azure cloud, so I'm not the person for that task either. Also,  both these tasks sound painful and do not spark joy for me.

I want to think this PR is useful in and of itself, e.g., people can use it to build cross-platform images, and the existing single arch still works as is, but if you want to wait for someone better at qemu magic with GitHub actions, I totally get it.

holdenk:
It sounds like you might not be interested in this PR @mathbunnyru and I totally get that. I'm happy to maintain this in my own fork although of course, I'd rather upstream it.

mathbunnyru:
> For "images are pushed to the Docker hub," I don't have permission to push images to your docker hub account, but I'm hoping showing that they can be pushed to my account as above does the trick.

After the merge to master in this repo, the commit is built once again and it's pushed to the DockerHub in a fully automatic way.
You can take a look at the docker.yml file to see exactly how it works.
So, you don't have to have the access to our account (I don't have it either for now).
We don't push manually at all and I think we haven't done it at least for a year.

And, to make it easy for our end users to use these images it would be best to push it and not force them to clone this repo and build images ourselves.

Also, if we update some image (for example, base image) and it stops working for arm, we won't know about it.
That's why I'm asking a bit more, than "the image builds on my arm machine" to say that arm is fully supported.

Hope you will understand my point of view.

> For "wiki build manifest is updated," I'd be happy to do that, but I don't know what you are asking for to be updated in the build manifest wiki.

For each image, we create a build manifest here: https://github.com/jupyter/docker-stacks/wiki
This manifest is gonna be different for arm image, at least because system packages version and installed packages will differ.
Here is the description of this process: https://github.com/jupyter/docker-stacks/tree/master/tagging

> Building ARM images inside x86 containers on GithubActions is probably doable, but I'm not the person for that task. Similarly, it looks like it's possible to get ARM GitHub action runners, but I don't work with Azure cloud, so I'm not the person for that task either. Also, both these tasks sound painful and do not spark joy for me.

Yes, I think we have to build these images on GithubActions using runners / or probably qemu (as other people do).

> It sounds like you might not be interested in this PR @mathbunnyru and I totally get that.

That's a bit sad to hear, given the time I spent to review these files and thinking about what to do best :(

mathbunnyru:
> I want to think this PR is useful in and of itself, e.g., people can use it to build cross-platform images, and the existing single arch still works as is, but if you want to wait for someone better at qemu magic with GitHub actions, I totally get it.

@parente @consideRatio how do you feel about this change?
I mean, it's a good start point, but I'm not sure it's good enough to be merged while it's not actually running in CI.

manics:
> Building ARM images inside x86 containers on GithubActions is probably doable, but I'm not the person for that task

This is exactly what `docker buildx` is designed to do, see the jupyterhub workflow for an example:
https://github.com/jupyterhub/jupyterhub/blob/f7b26c02dc8c1a633c85c33be19c2b167bb2f2ba/.github/workflows/release.yml#L83-L119
Note since this uses a local registry (depending on the branch) it runs fine in PRs

mathbunnyru:
> > Building ARM images inside x86 containers on GithubActions is probably doable, but I'm not the person for that task
> 
> This is exactly what `docker buildx` is designed to do, see the jupyterhub workflow for an example:
> https://github.com/jupyterhub/jupyterhub/blob/f7b26c02dc8c1a633c85c33be19c2b167bb2f2ba/.github/workflows/release.yml#L83-L119
> Note since this uses a local registry (depending on the branch) it runs fine in PRs

 Yes, thanks, I've take a look the last time you gave this link and I was hoping to see something simial in this PR as well.

holdenk:
Yeah I get that you don't want to do the partial step forward of adding build support without having CI for the change. I'm probably not the right person to try and work on your CI though.

To be clear: this also only adds support for the base-image, the other images have their own issues that probably need some fixing (I tried building the R stack for example and the packages had some weird conflicts).

Do you want me to close this PR?

mathbunnyru:
As I said, I would like to hear from @parente @consideRatio @romainx

holdenk:
Sounds good 👍

I can look at the CI based on the example above but it'll only support the base container, is that ok or do y'all want to support all the containers instead of incremental?

mathbunnyru:
> Sounds good 👍
> 
> I can look at the CI based on the example above but it'll only support the base container, is that ok or do y'all want to support all the containers instead of incremental?

I think the easiest way, for now, is to do it like this:
Try to build all the images.
If some image builds fine or it's easy to fix it so it builds fine, then great.

If not, put something like `if not x86_64 do nothing for steps which don't work`.
I have no idea if there are many steps like this (I hope not).

holdenk:
What about if I group the images into x86 only build targets and cross build targets? I say this because three of them don't build out of the box and I don't have the skills to debug R or Binder issues and it would be a shame to let perfect be the enemy of progress. But once again I completely understand if you've got higher standards (eg only want to do the port if all of them work), for my own use case I have a lower bar.

mathbunnyru:
> What about if I group the images into x86 only build targets and cross build targets?

Could you explain more, please? Whay do you mean by grouping?

consideRatio (commented):
> What about if I group the images into x86 only build targets and cross build targets?

For the maintenance of this repo long term, I think it is quite critical that we don't make our Dockerfile's x86 specific or arm64 specific etc - we have had such files before that failed to be maintained and synced with updates etc so I'm very cautious for that.

---

From reading a bit in this PR comments I perceive that the scope of the PR may grow beyond what @holdenk may be comfortable doing as it expands to learning about the docker-stacks project's CI system etc. I've been in similar situation to this both as a contributor and a maintainer. At those times I appreciate when the work to finish the PR transitions by a maintainer asks if adding code commits to complement the PR is okay and if the original PR author would like to help with review of the final parts being added.

It is always complicated though, one never knows how much work a PR will be - as a contributor or maintainer! Thank you @holdenk, @mathbunnyru, and @manics for your work on this PR!

holdenk:
Yeah I also do not want to have separate x86 and ARM dockerfiles. I was thinking of keeping just one set of files but having it only cross-build the docker containers where the porting was done as a way of allowing incremental development because I'm not interested in doing all of the things, just some of the things.

That being said, I would love to hand this over to someone else to continue forward who has more experience with the CI & R & binder 👍 

Happy to leave the PR open or closed whatever's best.

holdenk:
So I've redone it a little bit to hopefully be more amenable to incremental development. The github actions change is mostly copy-pasta though and I need someone to sign-off on that running to see if it does what y'all want.

I know there was the hope that porting all of the containers could be easily done in one go, but there are package conflicts in the base R container & the scipy container on ARM that need to be resolved.

If anyone is looking for (unofficial) cross-built arm64/amd64 containers you can get them from https://hub.docker.com/repository/docker/holdenk/minimal-notebook :)

.github/workflows/docker.yml
```diff
@@ -67,6 +88,3 @@ jobs:
         with:
           username: ${{secrets.DOCKERHUB_USERNAME}}
           password: ${{secrets.DOCKERHUB_TOKEN}}
-      - name: Push Images to DockerHub
-        if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main'
-        run: make -C main push-all
```
consideRatio, line 72: Hmmm I have a hard time reviewing this PR. I think we all have struggled a bit on finding a scope for a PR to work towards publishing arm64 images. Perhaps we can narrow this PR to just update the base-notebook Dockerfile to be buildable on x86 and arm64?

If this PR is narrowed to the changes in the base-notebook Dockerfile, I'd be happy to approve this PR quickly, but for changes to the CI system I'm far more hesitant. I consider the change above for example - I think we would end up no longer doing the previous CI sequence of build->test->push which should be unchanged no matter if we build / publish a new kind of architecture.

---

holdenk: OMG I'd love to backout the CI changes.

holdenk: How about for the makefile I'll switch build & build-all back to their original behaviour, but leave in build-cross for folks who want to build cross-compiled images?

consideRatio: Haha ;D If you are okay with it @holdenk, I'd be happy to strip this PR for relevant changes in another PR focused on CI stuff - making it come to use in the end. I have a local copy of this PR checked out on my computer atm.

For this PR though, I think it makes sense to not let the scope be broader than just making the base-notebook/Dockerfile be buildable on an arm64 architecture. It could be a single change, a single file.

Btw I've not read up on the purpose of OWNER arg or if its in scope for arm64/x86 stuff, but if it isn't I suggest excluding that as well at this point.

holdenk: Yeah so the point of the OWNER arg is to allow people to build to their own dockerhub account. It's already in the MAKEFILE it just only affects the tagging not the FROM (so e.g. if you wanted to build your own base-notebook & minimal notebook on top of it that isn't supported without the owners ARG), but I'm happy to back it out too.

consideRatio: Thanks for clarifying that! Let's go minimalistic and remove it from this PR as well, and let's get this PR approved and merged!

consideRatio (approved):
Thank you @holdenk! This LGTM!

consideRatio:
@holdenk @mathbunnyru I suggest we squash this PRs commits on merge, as 16 commits going back and forth can make git history a bit hard to follow when using `git blame` for example to track some change. Is that okay with you @holdenk?

holdenk:
Yeah I think a squash merge makes sense here 👍

mathbunnyru:
Ok, let's squash merge this.

Hope someone will finish this work soon :)

@holdenk thanks for your efforts, great job 👍
@consideRatio I hope you will make this happen ❤️

PR #1373, opened 2021-06-19
Add openssh-client (needed for git-over-ssh)

### LINKED ISSUE #1372: Consider adding openssh-client to all images containing git
**What docker images this feature is applicable to?**

All images derived from `jupyter/minimal-notebook`.

**What changes do you propose?**

I think pretty much everyone moves to git-over-ssh these days. It'd be great to have `openssh-client` included in all images, where git comes pre-installed (I think it's everything derived from minimal-notebook).
It's just around 4 MiB of extra packages.

**How does this change will affect users?**

In Kubernetes environment, it's normally not allowed to run containers as root, so the only way to install ssh is through conda, which is known for not being too fast. Plus, it's just inconvenient to install ssh upon every restart.
So, I'd say having ssh-client preinstalled would make docker-stacks more user-friendly out-of-the-box.

weisdd:
As discussed in https://github.com/jupyter/docker-stacks/issues/1372, this PR adds `openssh-client` to the list of packages installed in the `minimal-notebook`. - It's needed for the git-over-ssh support.

Closes #1372

### DIFF
--- a/minimal-notebook/Dockerfile
+++ b/minimal-notebook/Dockerfile
@@ -19,6 +19,7 @@ RUN apt-get update --yes && \
     libxrender1 \
     lmodern \
     netcat \
+    openssh-client \
     # ---- nbconvert dependencies ----
     texlive-xetex \
     texlive-fonts-recommended \


consideRatio (approved):
LGTM and i think its worth adding it to this image alongside git.

mathbunnyru:
Thanks. Merging this!

consideRatio:
Thanks @weisdd !!

PR #1374, opened 2021-06-20
Make docs build process less obscure

mathbunnyru:


.github/workflows/sphinx.yml
```diff
@@ -8,6 +8,7 @@ on:
   push:
     branches:
       - master
+      - main
```
mathbunnyru, line 11: If we ever rename to main, nothing should actually break.

.github/workflows/sphinx.yml
```diff
@@ -35,11 +36,10 @@ jobs:
       - name: Build Documentation
         run: make docs
       - name: Extract Source Strings
-        if: github.ref == 'refs/heads/master'
```
mathbunnyru, line 38: It's ok to change the cloned repo, we will check that everything works in the PRs.
We will only push to master when already in master.

mathbunnyru:
I don't like makefiles, it is much more simple to run `sphinx-build` manually than copy-paste Makefile (and make.bat).

.github/workflows/sphinx.yml
mathbunnyru: Well, almost, there are some other places, where it's gonna break, I will address it after this PR.

.github/workflows/sphinx.yml
```diff
@@ -35,11 +36,10 @@ jobs:
       - name: Build Documentation
         run: make docs
```
mathbunnyru, line 37: The idea to keep `make docs` in the main Makefile is simple - it's an easy way to let users build the docs.
They don't need to type `sphinx-build` command or remember it.

At the same time the process of updating source strings hasn't changed for a looooong time, so let's keep the command here, instead of putting it in Makefile.

consideRatio (approved):
LGTM. It makes sense to me given the little use and few commanda in the current docs/Makefile file.

Issue #1153, opened 2020-08-25
An easy way to update packages versions


mathbunnyru:
I see a lot of commits just updating several packages.

It would be great to have a tool, which automatically finds all the versions which are old (with some blacklist maybe) and tries to update these versions and if tests pass, it will automatically commit the changes.

So we will always have latest (and greatest packages), and reduce a lot of manual work.

parente:
One implementation thought: We could set up a GitHub Action workflow to run that tool on some cadence and open / close PRs based on test status.

romainx:
Good idea, note that this tool already exist and can be used at least as a base. It is described in the [contributor documentation](https://jupyter-docker-stacks.readthedocs.io/en/latest/contributing/packages.html#notes).

```
$ make check-outdated/base-notebook

# INFO     test_outdated:test_outdated.py:80 3/8 (38%) packages could be updated
# INFO     test_outdated:test_outdated.py:82
# Package     Current    Newest
# ----------  ---------  --------
# conda       4.7.12     4.8.2
# jupyterlab  1.2.5      2.0.0
# python      3.7.4      3.8.2
```

romainx:
I think one of the best way to do it would be to use a tool like [dependabot](https://github.com/dependabot/dependabot-core). However it does not support conda, and its Docker support will not help on our use cases.

trallard:
I think the idea of having a GitHub action to update the dependencies would be a good approach (and overcomes the limitations from dependabot). 

I would be happy to give this a go

mathbunnyru:
> I think the idea of having a GitHub action to update the dependencies would be a good approach (and overcomes the limitations from dependabot).
> 
> I would be happy to give this a go

I think no one has been working on this issue for the past few months, so I say, git this a go, if you want to, it would be awesome :)

trallard:
Great - will get working on this and create a draft PR as soon as possible

mathbunnyru:
> This is perhaps a silly question, but why pin the dependencies in the first place? Especially for scipy-notebook.
> 
> I've been gaining more experience with conda-forge, and it seems like a good strategy is to leave most things unpinned. When something unexpected happens, it's probably because of some error in the upstream dependencies, which can (and should) be fixed in the respective feedstock.

That's not a silly question at all.

I see several positive things in pinning versions:

1. Reproducibility. If we don't fix versions and build the same code at different times, it will give different results. It's not something that I expect (I have some background in C++). This is really important.

[...]

PR #1378, opened 2021-06-20
Automatically deduce package versions

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1153

For now, this is a test, that it's building right now.
There are other things to do, to make this work well (if we will go this way).


mathbunnyru:
Things, which still have manual versioning:

- [ ] base ubuntu focal image - easy to unpin, just using `focal` tag should work fine (`latest` - is another choice, but it will update when LTS updates, which should also be fine)
- [ ] miniforge (We're downloading from GitHub, so probably there's an API to get the latest version easily)
- [ ] julia (It's possible to get the version from Github and then use existing code)
- [ ] spark (Not sure about this one)

We will also have to add some once-a-week trigger, which will rebuild everything, to make this work.

mathbunnyru:
For now, I'm ready to implement the `once-a-week trigger`.
I think, the best way to do that would be to have a cron workflow, which creates an empty commit and then launches docker workflow.

Empty commit is needed because we push SHA-based tag and it's not good to overwrite this tag with another content.
And because our builds would not be reproducible anymore, we will have to create some commit to distinguish the builds.

consideRatio:
I think this is what we need to have in the long run I think. 

We don't pin things manually, but instead we document clearly how the version changes over time via the automated creation of image manifests reporting that.

.github/workflows/cron-docker.yaml
```diff
@@ -0,0 +1,26 @@
...
+        uses: actions/checkout@v2
+      - name: Create an empty commit
+        uses: stefanzweifel/git-auto-commit-action@5dd17c3b53a58c1cb5eaab903826abe94765ccd6 # dependabot updates to latest release
+        with:
+          commit_message: Empty commit to update sha tag
+          commit_options: "--allow-empty"
+      - name: Repository Dispatch
+        uses: peter-evans/repository-dispatch@ce5485de42c9b2622d2ed064be479e8ed65e76f4 # dependabot updates to latest release
+        with:
+          token: ${{secrets.REPO_ACCESS_TOKEN}}
```
mathbunnyru, line 25: @parente could you please give me the necessary rights for managing secrets?

mathbunnyru: @parente may I ask you to respons here?

mathbunnyru:
@consideRatio this PR seems to work fine.
I made several merges of master to make sure it works reliably and it worked quite well.
Arm also seems to be working just fine.

Please, review once again.

Things left to be done:
- [ ] Create `secrets.REPO_ACCESS_TOKEN`. This can only be done by @parente and I'm still waiting.
- [ ] Test that the current scheme with automatic commit creation works fine - I will do it in my fork.

consideRatio (commented):
Excellent work on this @mathbunnyru! I'll email @parente about the permissions

mathbunnyru:
Thanks for good suggestions, I applied them 👍

consideRatio:
@mathbunnyru I'm unhappy about how we need to use a PAT for our user accounts for the dispatcher action. If I would for example grant my PAT restricted as much as possible, I would in a single token still expose the entire JupyterHub org if the token was leaked.

Due to this I'm considering the following:
1. Do we really need to do it like this? Isn't there another way to do this?
2. If we really need to do this, I suggest we create a dedicated account for the purpose, which is a hassle still...

I think I have an idea that can help us avoid it without using any permissions... Hmmm...

We make our regular build-and-push job have a step that only runs if triggered by a cron job, letting it make the dummy commit. Then, we have a step in the end of the build-and-push job that pushes the commit we made during this job back to master. Or hmm... perhaps we don't push our commit back to master at all? Then we avoid cluttering our git history with dummy commits but at the same time then our git image tags will be unrelated to the git history. We could also perhaps make our image tagger that defines the image tag sha do something custom for our weekly builds, such as `<latest sha here>-<date here>` or similar.

Hmm... I'm not sure what makes sense, but I appreciate:
- if we avoid using a PAT no matter what account its coupled to
- if we don't have dummy commits in the git history
- if our image tags associated with these scheduled rebuilds reference actual git commits and/or dates

.github/workflows/cron-docker.yaml
parente: @mathbunnyru I never received a GitHub notification for your @ here. Odd. I've updated your and @consideRatio's permissions on this repo to admin from maintainer.

parente:
The scope of PATs is something I've struggled with in other projects. The best approach I've found, which [mirrors GitHub's guidance for SSH keys](https://docs.github.com/en/developers/overview/managing-deploy-keys#machine-users), is to create a dedicated "machine user" and invite it as an external collaborator with least privilege access to repositories where it is needed.

.github/workflows/cron-docker.yaml
mathbunnyru: Thanks, @parente!
Maybe you have accidentally disabled notifications for the project or for this particular PR?

Anyway, thanks for the admin permissions ❤️ Now, I will be able to make changes like this easier👍

mathbunnyru:
@consideRatio I think your suggestion sounds good, but I'm not sure it will work with GitHub limitations without PAT.

I will try to look into what can be done here, but I'm quite busy now, so I won't be able to do that before mid-August.

What I can do now, is to add a `-{date}` tag in a separate PR.
That's easy and sounds useful if we will use cron.

consideRatio:
@mathbunnyru what do you think about instead of having a `{sha}-{date}` format, we just go for a pure `{date}` tag?

Hmm I'll outline a work plan below, let me know what you think - if you agree on this we can start working in parallel when we find time.

### Work plan

1. We add a DateTagger that use a `yyyy-mm-dd` format.
2. We either a) let it always tag images and risk override previous versions, or b) only when our cronjob runs
   I'm not sure what makes sense, but I figure it may be relevant to consider how the manifests we create are published on the wiki.
3. We refactor this PRs cron triggered workflow to dispatch another job, but instead relies on that other job that currently we use to build images etc, to have steps specific for when building as part of a cronjob trigger (if needed). By doing this, we avoid all PATs and maintenance of a separate workflow.

### Concrete action points
- [ ] Decide if DateTagger can tag like `YYYY-MM-DD` (example 2021-08-05, zeroes meant to be included)
- [ ] Decide if 1-3 above is ok
- [ ] Decide if 2a or 2b should be done
- [ ] In separate PR: Add DateTagger like class with logic to the choice about 2a or 2b. (Do you want to have a go at this?)
- [ ] In this PR: Integrate the cronjob CI logic in this PR into the pre-existing workflow logic (I'd be happy to do this)

consideRatio:
@mathbunnyru I pushed two commits that you can revert or force-push remove if you think it went in the wrong directly, but I figure I'd try get this to move along more concretely.

I conclude that we have had two kinds of taggers in the past. The `SHATagger` and the various `<something>VersionTagger`. The VersionTaggers got updated and overridden, while the SHATagger didn't get overridden.

With this PR and the most recent commits, we will have three kinds of taggers. `SHATagger`, `DateTagger`, and the various `<something>VersionTagger`. With this PR, `SHATagger` tags will be overridden weekly if no new commits have been made in a week, and the `DateTagger` could be overridden if multiple builds are made in one day.

I think this behaviour is fine and acceptable, I think the desired behavior for anyone who wants to lock onto a specific version should refer to a date based tag as they won't change after that day has passed.

consideRatio:
@mathbunnyru I think this is could be ready to merge, before or after #1424. We will not need a PAT any more with this setup.

mathbunnyru:
Yes, this setup looks fine to me.
But our docs need to be updated.
Right now, we declare something like - pin SHA and you will be fine.
It won't be true after this change, because after this change SHA will be overwritten.

So, I think we have to update the docs.

consideRatio:
@mathbunnyru ah, I don't overview the docs so well. Do you know where the docs needs to be updated?

mathbunnyru:
I think this section has to be updated:
https://github.com/jupyter/docker-stacks/blob/master/docs/using/selecting.md#versioning

mathbunnyru:
LGTM as well!

consideRatio:
@mathbunnyru I gave it a shot. I didn't update anything under docs/locale and I'm not sure how to go about that in general.

I would say it is out of scope for this projects maintenance capacity to try keep those updated if it requires mirroring our changes to multiple files.

docs/using/selecting.md
```diff
@@ -193,23 +193,21 @@ diagram](../images/inherit.svg)](http://interactive.blockdiag.com/?compression=d
...
-`latest` is a moving target, by definition, and will have backward-incompatible changes regularly.
+Whenever a docker image is pushed to the container registry, it is tagged with:
 
-Every image on Docker Hub also receives a 12-character tag which corresponds with the git commit SHA that triggered the image build.
-You can inspect the state of the `jupyter/docker-stacks` repository for that commit to review the definition of the image
-(e.g., images with tag `33add21fab64` were built from <https://github.com/jupyter/docker-stacks/tree/33add21fab64>.
+- a `latest` tag
+- a 12-character git commit SHA like `b9f6ce795cfc`
+- a date formatted like `2021-08-29`
+- a set of software version tags like `python-3.9.6`, and `lab-3.0.16`
```
consideRatio, line 205:
```suggestion
- a set of software version tags like `python-3.9.6` and `lab-3.0.16`
```

mathbunnyru:
Don't worry @consideRatio.
I have never touched `docs/locale` as well - it should be done automatically by transifex, to support multi-language docs, but I'm not sure that it works at all.

consideRatio:
Okay from my end, this LGTM and can be merged as well. The tests have passed besides the tests of docker build for the recent docs changes.

If this LGTY i suggest we go for a merge whenever the current build jobs finish following merge of #1424.

mathbunnyru:
Agreed, also merged latest master in here to make sure everything works with this change combined.

bsikander:
Hi, I build an image on top of images provided by docker-stacks. I also install my own R-packages.
Recently, when I was building my image on top of 1.4.2 image, i had alot of conflicts. I noticed this change where you removed all the versions.

Since I am very new to R, could you please add a reason to PR why it was necessary to remove the version and why pinning is not prefered? Further, should the downstream images also consider not pinning?

consideRatio:
@bsikander removing the pinnings in general was in my mind a key step towards making maintenance of this project more sustainable. Manual version bumps would otherwise need to happen regularly, and still result in breaking changes in the same way as if they were automated.

If you want to keep using an older pinned version of an R package, you can pin the version of the image, and you can observe the various package versions by inspecting https://github.com/jupyter/docker-stacks/wiki and clicking on build manifest for a specific image build.

> Further, should the downstream images also consider not pinning?

I would recommend not pinning versions also in downstream images if you rebuild your image on top of the latest from jupyter/docker-stacks. If you use latest jupyter/docker-stacks image + additional pinned installations, you will in time get a mismatch between what versions are compatible. If you pin the base image from jupyter/docker-stacks, then you can pin your package versions in your derived images without problem as well.

My recommendation would be: don't pin jupyter/docker-stacks image in your `FROM` statement in a Dockerfile you maintain, and also don't pin your own versions, and then before assuming your image functions well - make some quick checks by trying to run some basic stuffs in the image and verify it doesn't error.

mathbunnyru:
A bit more info:

1. Pinning some packages version doesn't mean, that you will always have these versions. New time you run `mamba install new-package`, you might change other package versions as well.

2. The version appeared because quite a long time ago, automatic version deducing was not as reliable as it is right now. Now, we don't have the issues anymore.

> Further, should the downstream images also consider not pinning?

I have another opinion on this one :)
If you have lots of users / many-many lines of code / code, which highly relies on 3rd party libraries and not so many testing, then you might want to pin both the image your image builds on and the packages you install on top.
But, it will be even more difficult to update in the future.

If you are actually OK, when something breaks and you need to fix some code working with 3rd party libs, then I would not recommend pinning versions and use latest (presumably the most security safe, fastest and overall usable) packages.

PR #1393, opened 2021-06-29
Remove microbadge usage

mathbunnyru:
MicroBadger.com will shutdown on 1st July 2021.
https://microbadger.com/shutdown


mathbunnyru:
We will lose information about layers size, but I don't think that's really important.
The size will stay thanks to https://shields.io.

consideRatio:
:tada: :heart: =) Nice work keeping things updated @mathbunnyru!

PR #1398, opened 2021-07-08
Introduce owner to Dockerfiles to make it easy to test locally

mathbunnyru:
This is a follow-up on https://github.com/jupyter/docker-stacks/pull/1368

This should make it easy for someone to have local builds pushed to their docker hub and pulled from it.


holdenk (approved):
looks good to me :)

PR #1399, opened 2021-07-10
Support arm builds

mathbunnyru:
This is a POC, **do not merge**.

Based on @holdenk PR, with some fixes and additions.


Makefile
```diff
@@ -23,18 +35,23 @@ export DOCKER_BUILDKIT:=1
...
+	@echo "Replace % with a stack directory name (e.g., make build-cross/minimal-notebook)"
 	@echo
 	@grep -E '^[a-zA-Z0-9_%/-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
-build/%: DARGS?=
-build/%: ## build the latest image for a stack
-	docker build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest --build-arg OWNER=$(OWNER) ./$(notdir $@)
+build-x86/%: DARGS?=
+build-x86/%: ## build the latest image for a stack on x86 only
+	docker buildx build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@) --build-arg OWNER=$(OWNER) --platform "linux/amd64" --push
```
manics, line 44: I think you need to include the `$REGISTRY` environment variable as part of the tag argument so that it gets pushed to `localhost:5000`

consideRatio:
I'd like to have a clear scope for this PR defined, even as a draft, as I think it would help me contribute in general to this work - either directly in this PR or in surrounding PRs.

### Main scope suggestion

I suggest that if we can make this PR able to do the following things, we go for a merge.

- [ ] To update the Makefile to be able to, in an opt-in fashion, to _build_ our Dockerfiles for _all compatible architectures_ (like in: [Z2JH](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/b804bac3ad6acf8b656f6108a5a18b95d9d22c10/.github/workflows/publish.yml#L60-L68), [JupyterHub](https://github.com/jupyterhub/jupyterhub/blob/3fec19d191119e4203c243f44c43e74c8fe62f04/.github/workflows/release.yml#L96-L142), [ConfigurableHTTPProxy](https://github.com/jupyterhub/configurable-http-proxy/blob/9bcabb140e7cc534ca0b9a31ebb146d1e6c82786/.github/workflows/publish.yml#L31-L79))
- [ ] To update the Makefile to be able to _test_ our built Dockerfiles _specifically in the CI systems architecture_.
- [ ] To update the Makefile to be able to, in an opt-in fashion, to _publish_ all our built Dockerfiles for one or all platforms.

### Out of scope suggestion

I suggest we consider it out of scope for this PR to look to actually change the Dockerfiles, but instead focus on building whatever dockerfiles we currently can build using --platform linux/arm64.

consideRatio:
@mathbunnyru I went ahead and rebased this on current master with the OWNER change, and started working to have something that can work locally as before, and on the CI system as well to test build multi-arch images, and to test the amd64 images, and to publish the multi-arch images.

Work remains but I'm on it.

consideRatio:
I think this may successfully be able to build and publish the images we have made support multiple architectures so far, base + minimal to my knowledge.

Some insights:
1. __Emulating architectures is slow__
   It is far slower to build a dockerfile via the emulated architecture setup. The base image took 90 seconds to build I think but the arm64 took several minutes. If we build all images with multi-arch, it will take more than an hour to verify our PRs etc.
2. It feels quite unsustainable to let a single GitHub job do everything from build, test, tag, push for all images and all arches.
3. We don't do tagging for arm64, but assume the amd64 tags.

There are plenty of things that can be improved in general to avoid slowing down the CI build duration etc, but it is also a lot of work and will require significant changes.

### Review suggestions

To make our CI system build, test, and publish images compatible with amd64 and arm64 is complicated and bound to include some trade offs.
- slowing down our CI system
- retaining support for using the Makefile locally
- being thorough in what we test
- making arm64 based images accessible

I have outlined a scope in https://github.com/jupyter/docker-stacks/pull/1399#issuecomment-877714949 that this PR address. Is the scope reasonable is the first question, and the second if the technical implementation and its associated trade offs is reasonable.

I've considered some tweaks to avoid slowing down our CI system too much, such as letting our PRs triggered CI jobs retain the quicker behavior by omitting multi-arch details. At the moment, I think I'd prefer having a slow but thorough CI system while we try get our images arm64 compatible, and perhaps later make it quick and less thorough.

I have some ideas on how we can make the CI system a lot more fancy, but it also involves a lot more work. I don't think I can find time to put in that much work at the moment so I hope we can settle for something not yet perfect. Ideas include:
- splitting apart our massive workflow with a massive job to build, test, tag, and push all of our images
  - use of `needs`
- running only the jobs and steps within that makes sense for any given PR
- using github's container registry to cache images:
  - from nighly main branch builds
  - from PR test builds

mathbunnyru:
@consideRatio I will definitely take a look at this, hope today or in the next few days.

@manics may I also ask you to take a look? You seem to be experienced with this kind of changes.

Makefile
```diff
@@ -19,37 +36,92 @@ ALL_IMAGES:=base-notebook \
...
 	@echo "jupyter/docker-stacks"
 	@echo "====================="
-	@echo "Replace % with a stack directory name (e.g., make build/minimal-notebook)"
+	@echo "Replace % with a stack directory name (e.g., make build-multi/minimal-notebook)"
 	@echo
 	@grep -E '^[a-zA-Z0-9_%/-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
+
+
 build/%: DARGS?=
```
mathbunnyru, line 51: I think I would like to call this build-amd64, it makes more sense.

Makefile
```diff
@@ -19,37 +36,92 @@ ALL_IMAGES:=base-notebook \
...
+	@echo "::group::Build $(OWNER)/$(notdir $@) (amd64)"
+	docker buildx build $(DARGS) --rm --force-rm -t $(OWNER)$(notdir $@):latest ./$(notdir $@) --build-arg OWNER=$(OWNER) --platform "linux/amd64" --load
+	@echo -n "Built image size: "
+	@docker images $(OWNER)/$(notdir $@):latest --format "{{.Size}}"
+	@echo "::endgroup::Build $(OWNER)/$(notdir $@) (amd64)"
+
+	@echo "::group::Build $(OWNER)/$(notdir $@) (amd64,arm64)"
+	docker buildx build $(DARGS) --rm --force-rm -t build-multi-tmp-cache/$(notdir $@):latest ./$(notdir $@) --build-arg OWNER=$(OWNER) --platform "linux/amd64,linux/arm64"
+	@echo "::endgroup::Build $(OWNER)/$(notdir $@) (amd64,arm64)"
+build-all-multi: $(foreach I,$(MULTI_IMAGES), build-multi/$(I)) $(foreach I,$(AMD64_ONLY_IMAGES), build/$(I)) ## build all stacks
```
mathbunnyru, line 105: This should be called build-all, because you're also building non-multi images.

Makefile
```diff
@@ -59,49 +131,63 @@ dev/%: ## run a foreground container for a stack
...
-
+	@echo "::endgroup::Push $(OWNER)/$(notdir $@) (amd64)"
 push-all: $(foreach I,$(ALL_IMAGES),push/$(I) ) ## push all tagged images
 
+push-multi/%: DARGS?=
+push-multi/%: ## push all tags for a jupyter image that support multiple architectures
+	@echo "::group::Push $(OWNER)/$(notdir $@) (amd64,arm64)"
+	docker buildx build $(DARGS) --rm --force-rm $($(subst -,_,$(notdir $@))_EXTRA_TAG_ARGS) -t $(OWNER)/$(notdir $@):latest ./$(notdir $@) --build-arg OWNER=$(OWNER) --platform "linux/amd64,linux/arm64"
+	@echo "::endgroup::Push $(OWNER)/$(notdir $@) (amd64,arm64)"
+push-all-multi: $(foreach I,$(MULTI_IMAGES),push-multi/$(I)) $(foreach I,$(AMD64_ONLY_IMAGES),push/$(I)) ## push all tagged images
```
mathbunnyru, line 187: This is push-all

Makefile
```diff
@@ -59,49 +131,63 @@ dev/%: ## run a foreground container for a stack
...
+
 pull/%: DARGS?=
 pull/%: ## pull a jupyter image
 	docker pull $(DARGS) $(OWNER)/$(notdir $@)
-
-pull-all: $(foreach I,$(ALL_IMAGES),pull/$(I) ) ## pull all images
+pull-all: $(foreach I,$(ALL_IMAGES),pull/$(I)) ## pull all images
 
 push/%: DARGS?=
 push/%: ## push all tags for a jupyter image
```
mathbunnyru, line 176: This is push-amd64

mathbunnyru (commented):
I think we should improve naming - I'm absolutely fine breaking `make build/something` command, it's our implementation detail.

mathbunnyru:
Nicely done!!

Снимок экрана 2021-07-14 в 22 28 00

Makefile
```diff
@@ -19,37 +36,92 @@ ALL_IMAGES:=base-notebook \
...
+#    test the build on the different platform and provide cached layers for
+#    later.
+#
+# 3. We let push-multi refer to rebuilding a multi image with `--push`.
+#
+#    We now rely on the cached layer.
+#
+# Outcomes of the workaround:
+#
+# 1. We can keep using the previously defined Makefile commands that doesn't
```
mathbunnyru, line 87: I don't think this is really necessary.

mathbunnyru (commented):
Another thing - let's not forget about docs:

- let's explicitly say, that now we support arm for some images
- we need to list all the caveats there
- also, updating tagging/README.md would be nice

mathbunnyru:
> Is the scope reasonable is the first question

Yes, I think it is well described.

> the second if the technical implementation and its associated trade offs is reasonable.

I have a question for you.
For example, let's consider a situation, where we get rid off of amd64-only images (I think it's quite possible to do so).
How do you think, would it be possible to get over current limitations easily?

And also, I think we should create at least 2 issues and a new `arm` tag:
- tagging is done from amd64 images
- testing is not done for arm images

consideRatio:
> How do you think, would it be possible to get over current limitations easily?

I think the limitations should be resolved by having dedicated self-hosted GitHub arm64-based runners where the performance isn't crap building arm64 images, because then it is reasonable to have two jobs side by side and we don't lose any performance - one for amd64, one for arm64.

If I recall correctly, the `build` Makefile command doesn't specify either amd64 or arm64, and that was a reason for me motivating not calling them `build-amd64` - they would be `build-arm64` if it was ran on an arm64 machine as an explicit platform wasn't specified - right?

---

Thanks for the review @mathbunnyru! I'll go through it more properly later, I think I agree with your points but I need to deliberate a bit more on the naming of makefile commands.

consideRatio:
@mathbunnyru I made a commit (44dfb03) about the naming of the Makefile commands.

I created two new issues:
- #1401
- #1402

> I think we should create [...] and a new arm tag:

I don't understand what you mean with this. Do you mean that we should tag our multi-platform images with `arm`?

> updating tagging/README.md would be nice

I looked through it and couldn't figure out what kinds of updates you would like there. Perhaps you can push a commit for that?

mathbunnyru:
> And also, I think we should create at least 2 issues and a new arm tag

Sorry, I meant a label tag in github. I created it and added it to your newly created issues.

mathbunnyru:
I agree with your point about naming, you're right about system's arch, so let's keep it simple `build/` etc.

mathbunnyru:
About merging procedure:
I think in this case we also would like to have all commits squashed, what do you think?
@consideRatio 

I also want to merge Makefile refactoring separately, because it looks better and it makes this PR shorter.
https://github.com/jupyter/docker-stacks/pull/1403

consideRatio:
> I think in this case we also would like to have all commits squashed, what do you think?

Sure!

> I also want to merge Makefile refactoring separately, because it looks better and it makes this PR shorter.

Okay! I'll get this merge-able again after that is merged.

mathbunnyru:
> Okay! I'll get this merge-able again after that is merged.

Because this is my intention to merge Makefile separately, I will also deal with merge conflicts here to make your life easier :)

consideRatio:
@mathbunnyru ah thanks!

I suggest solving merge conflicts by resetting all the files to the state of this branch than trying to do a set of "keep current" or "keep incoming" choices etc.

I think it can be done by:

1. merge the other PR and update master branch locally
2. checkout this PRs branch and do `git reset --soft master`
3. now I think you should see files as if they were from this branch, but a git index like master, and then you can make a single commit to reflect these changes adding me as a co-author by ending the commit with `Co-authored-by: Erik Sundell <erik.i.sundell@gmail.com>`

Makefile
consideRatio: I want to have two separate options, one `build-all` like we have had it, doing exactly what it did - and `build-all-multi` for example like above, where it builds all in as many archs it can.

Makefile
consideRatio: Like for `build-all` I think it's important to separate this from `push-all` that doesn't use `docker buildx`. I think `push-all` shouldn't be used any more by this repo officially any more, but I think it can be useful still if you maintain your own image or similar using this infrastructure and only want to build/push system images instead of the dual-platform images.

.github/workflows/docker.yml
```diff
@@ -22,51 +22,76 @@ on:
...
+      # https://github.com/docker/build-push-action/tree/master#usage
+      # https://github.com/docker/build-push-action/blob/master/docs/advanced/multi-platform.md
+      - name: Set up QEMU (for docker buildx)
+        uses: docker/setup-qemu-action@27d0a4f181a40b142cce983c5393082c365d1480 # dependabot updates to latest release
+
+      - name: Set up Docker Buildx (for multi-arch builds)
+        uses: docker/setup-buildx-action@0d135e0c2fc0dba0729c1a47ecfcf5a3c7f8579e # dependabot updates to latest release
+        with:
+          # Allows pushing to registry on localhost:5000
+          driver-opts: network=host
```
consideRatio, line 45:
```suggestion
        with:
          # Allows pushing to registry on localhost:5000
          driver-opts: network=host
```

Ah we ended up not using this btw.

mathbunnyru:
Did the merging, give me 20 minutes to do cleanup.

Makefile
```diff
@@ -26,19 +42,73 @@ export DOCKER_BUILDKIT:=1
...
 	@grep -E '^[a-zA-Z0-9_%/-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 
 
 build/%: DARGS?=
-build/%: ## build the latest image for a stack
-	docker build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest --build-arg OWNER=$(OWNER) ./$(notdir $@)
+build/%: ## build the latest image for a stack using the system's architecture
+	@echo "::group::Build $(OWNER)/$(notdir $@) (system's architecture)"
+	docker build $(DARGS) --rm --force-rm -t $(OWNER)/$(notdir $@):latest ./$(notdir $@) --build-arg OWNER=$(OWNER)
```
mathbunnyru, line 54: I will create a separate PR for this as well, because it's not actually a part of arm support.

mathbunnyru: I don't think that's easy though :(

mathbunnyru:
@consideRatio I think this change is great and we're almost ready to merge it.

You've already seen that I tried to merge style and not related improvements separately.
May I ask you to create a PR for adding groups to workflow run (without changing env)?

The reason for asking:
1. The scope of this PR will be much easier to understand (even in the future).
2. If we decide to go with Github self-hosted runners, we might want to more-or-less revert this change. It's absolutely discussible and right now I don't know at all what's gonna happen, but if this review will be focused on exactly one thing, it will be much easier.

> Nicely done!!
> 
> Снимок экрана 2021-07-14 в 22 28 00

tagging/github_workflow_commands.py
```diff
@@ -0,0 +1,55 @@
...
+
+    if params:
+        comma_sep_params = ",".join([f"{k}={v}" for k, v in params.items()])
+        print(f"::{command_name} {comma_sep_params}::{command_value}")
+    else:
+        print(f"::{command_name}::{command_value}")
+
+
+@contextmanager
+def _gwc_group(group_name):
```
mathbunnyru, line 39: This doesn't seem to be used, is it needed?

mathbunnyru:
Maybe this should do it?
https://github.com/jupyter/docker-stacks/pull/1406

Makefile
mathbunnyru: Seems to be easy :)

tagging/github_workflow_commands.py
consideRatio: I found myself needing it only from the Makefile rather than Python script. It can be deleted from this PR.

tagging/github_workflow_commands.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    Entering the context prints the group command, and exiting the context
+    prints the endgroup command.<<
+    """
+    try:
+        yield _gwc("group", group_name)
+    finally:
+        _gwc("endgroup", group_name)
+
+
```
consideRatio, line 49:
```suggestion
```

tagging/github_workflow_commands.py
```diff
@@ -0,0 +1,41 @@
...
+The functions defined in this file will only emit such messages if found to be
+in a GitHub CI environment.
+"""
+
+import json
+import os
+
+
+def _gwc(command_name, command_value="", **params):
```
mathbunnyru, line 20: @consideRatio is this function used at all?
I think I can safely delete it as well?

consideRatio:
A note to myself that this can be relevant code to add back in the future so it feels fine removing it here.

```python
import json
from contextlib import contextmanager


def _gwc(command_name, command_value="", **params):
    if not os.environ.get("GITHUB_ACTIONS"):
        return
    # Assume non-string values are meant to be dumped as JSON
    if not isinstance(command_value, str):
        command_value = json.dumps(command_value)
        print(f"dumped json: {command_value}")
    if params:
        comma_sep_params = ",".join([f"{k}={v}" for k, v in params.items()])
        print(f"::{command_name} {comma_sep_params}::{command_value}")
    else:
        print(f"::{command_name}::{command_value}")


@contextmanager
def _gwc_group(group_name):
    """
    Entering the context prints the group command, and exiting the context
    prints the endgroup command.<<
    """
    try:
        yield _gwc("group", group_name)
    finally:
        _gwc("endgroup", group_name)
```

tagging/github_workflow_commands.py
consideRatio: Yepp, and the `import json` statement then.

mathbunnyru: I also renamed the file and removed the docs, because they seem to be irrelevant

mathbunnyru:
@consideRatio I think I'm done here and I simplified this PR as much as possible :)
Please, take a look that I didn't miss anything.

@manics could you take a look, please?

consideRatio (approved):
This LGTM! Thanks for your work to extract unrelated changes from this PR @mathbunnyru!

(I introduced these changes to simplify the arm64 feature development, as grouping and line breaks helped me read code/logs and think more clearly.)

Btw @mathbunnyru regarding:

> If we decide to go with Github self-hosted runners, we might want to more-or-less revert this change.

I think with self-hosted runners, we can change almost nothing.
- For testing, we can instead of using `-multi` we just exclude the suffix and voila we have build + test that is architecture native and high performance running in parallel.
- For publishing, we should probably run the workflow on a single arch and rely on docker buildx to do a multi-platform build from scratch.

Note that the only change required to use a self-hosted runner, I think, is to declare it as part of a job. I figure we configure our jobs to run twice using a matrix strategy for the job.

manics (approved):
Nice work, LGTM :smiley:

FYI if you wanted to avoid building amd64 twice you could tag each architecture separately and use `docker manifest` to create the multi-architecture manifest list https://docs.docker.com/engine/reference/commandline/manifest/`
e.g. `docker manifest base-notebook:<tag> base-notebook:<tag>-amd64 base-notebook:<tag>-arm64`
but if your long term plan is to support even more architectures your current approach makes sense.

consideRatio:
@manics is it `docker manifest create <name of image> <amd64 only image> <arm64 only image>`? I figure it makes sense then to build each arch from a native platform, push to githubs container registry, then in a final github job that `needs` the separate build jobs pull both and combine them.

```
docker manifest create --help

Usage:  docker manifest create MANIFEST_LIST MANIFEST [MANIFEST...]

Create a local manifest list for annotating and pushing to a registry

EXPERIMENTAL:
  docker manifest create is an experimental feature.
  Experimental features provide early access to product functionality. These
  features may change between releases without warning, or can be removed from a
  future release. Learn more about experimental features in our documentation:
  https://docs.docker.com/go/experimental/

Options:
  -a, --amend      Amend an existing manifest list
      --insecure   Allow communication with an insecure registry
```

consideRatio:
We did it @mathbunnyru @manics!!! :tada: :sunflower: :heart:

mathbunnyru:
Great news!

@consideRatio awesome work 👍 
@holdenk I also wanted to thank you for starting this work.

consideRatio:
Yes thank you @holdenk!!! :heart: :tada:

consideRatio:
Complemented this PR with #1408

consideRatio:
image

holdenk:
Thank y'all for carrying this all the way to the finish line, I am so thankful and yay arm64 images for all :D

PR #1404, opened 2021-07-16
Make GitHub Workflows steps look beatiful

mathbunnyru:


consideRatio:
I have wanted to do that before but ignored it to avoid noise :D

Issue #1229, opened 2021-02-04
Volumes deleted when entering a container


pylang:
**What docker image you are using?**

`jupyter/scipy-notebook:latest`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

First, the setup:

```bash
> mkdir test_vols
> cd test_vols
> mkdir data
> touch data/foo.txt
```

Now the command:

```bash
  sudo docker run \
              -p 8888:8888 \
              -w /home/ \
              -e NB_USER=user \
              -e CHOWN_HOME=yes \
              -v `pwd`/data:/home/jovyan/data \
              --user root \
              --name test_vols \
              -it --rm jupyter/scipy-notebook:latest \
              start.sh \
              jupyter lab \
              --ip=0.0.0.0 \
              --no-browser
```

Note: `sudo` may be optional on your system.  Replace `user` with your appropriate username.

**What steps do you take once the container is running to reproduce the issue?**

No steps required.  

**What do you expect to happen?**

We are focusing on the volume outside the container:

```bash
> pwd
~/path/to/test_vol
```

1. After closing the container, I expect the contents of the volume folder to persist.

```bash
> tree
.
└──  data
    └── foo.txt
```

**What actually happens?**

1. After ~~closing~~ starting the container, the contents of the volume folder is deleted.

```bash
> tree
.
└── data
```

----

**Details**

*Problems*

Having the volume deleted is serious, and that seems like a bug.  The contents seem to be deleted once the container starts.

There is also an additional minor problem that occurs internally.   `/home/joyvan/` folder does not get properly renamed to `/home/user`.  You can see this while inside the containers via `ls` and `ls -al user`.

*Debugging*

I've invoked many combinations of parameters with a range of results.  I've narrow this particular bug to how mounted volumes are internally mapped:

Mapping | Rename `/home/jovyan/` to `/home/user`? (internal) | Volume preserved? (external)
---|---|---
` -v .../data:/home/jovyan/data ` | No | No 
` -v .../data:/home/user/data `     | No | Yes
` -v .../data:/home/data `             | Yes | Yes

If we map to either user internally, we get undesired results.  If we map elsewhere (e.g. the last row),  we get desired results.

**Pending work-around (so far)**

1. Avoid volume deletions - don't map mounted volumes to `jovyan`.
1. Fully rename jovyan to user homedir - don't include the `user` in any volume mapping `-v` or working directory `-w`.

[...]

PR #1421, opened 2021-08-06
Fix: volumes/files deleted when bind mount to /home/jovyan

rigzba21:
Hi all! Potential fix for https://github.com/jupyter/docker-stacks/issues/1229 for review. Thanks!

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -42,7 +42,6 @@ RUN apt-get update --yes && \
     wget \
     ca-certificates \
     sudo \
-    rsync \
     locales \
     fonts-liberation \
     run-one && \
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -42,6 +42,7 @@ RUN apt-get update --yes && \
     wget \
     ca-certificates \
     sudo \
+    rsync \
     locales \
     fonts-liberation \
     run-one && \
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -55,7 +55,7 @@ if [ "$(id -u)" == 0 ] ; then
         if [[ ! -e "/home/${NB_USER}" ]]; then
             echo "Relocating home dir to /home/${NB_USER}"
             mkdir "/home/${NB_USER}"
-            rsync -avx /home/jovyan "/home/${NB_USER}" || ln -s /home/jovyan "/home/${NB_USER}"
+            cp -R /home/jovyan "/home/${NB_USER}" || ln -s /home/jovyan "/home/${NB_USER}"
         fi
         # if workdir is in /home/jovyan, cd to /home/${NB_USER}
         if [[ "${PWD}/" == "/home/jovyan/"* ]]; then
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -54,7 +54,8 @@ if [ "$(id -u)" == 0 ] ; then
         # (it could be mounted, and we shouldn't create it if it already exists)
         if [[ ! -e "/home/${NB_USER}" ]]; then
             echo "Relocating home dir to /home/${NB_USER}"
-            mkdir "/home/${NB_USER}" && rsync -avx /home/jovyan "/home/${NB_USER}" || ln -s /home/jovyan "/home/${NB_USER}"
+            mkdir "/home/${NB_USER}" 
+            rsync -avx /home/jovyan "/home/${NB_USER}" || ln -s /home/jovyan "/home/${NB_USER}"
         fi
         # if workdir is in /home/jovyan, cd to /home/${NB_USER}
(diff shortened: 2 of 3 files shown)


base-notebook/test/test_container_options.py
```diff
@@ -187,3 +187,30 @@ def test_group_add(container, tmpdir):
...
+        working_dir="/home/",
+        environment=[
+            "NB_USER=user",
+            "CHOWN_HOME=yes",
+        ],
+        volumes={d: {"bind": "/home/jovyan/data", "mode": "rw"}},
+        command=["start.sh", "jupyter", "lab"],
+    )
+
+    time.sleep(10)
```
mathbunnyru, line 213: Let's wait with timeout here to not make this test always run 10 seconds?

It can be seen in the examples above I suppose.

base-notebook/Dockerfile
```diff
@@ -42,6 +42,7 @@ RUN apt-get update --yes && \
     wget \
     ca-certificates \
     sudo \
+    rsync \
```
mathbunnyru, line 45: Let's not add more dependencies here to make images as small as possible.
I think `cp -R` or bare cp could be enough.

mathbunnyru (changes requested):
Thanks for the fix!

Did you check that your test fails without your fix?

base-notebook/test/test_container_options.py
rigzba21: Hi! Yes, I did check that the test does fail with the original ```mv /home/jovyan "/home/${NB_USER}"``` command, and passes with the fix. As for the timeout I was following the above examples and following the pattern.

base-notebook/Dockerfile
rigzba21: I can definitely make that change. Thanks for the feedback!

base-notebook/test/test_container_options.py
mathbunnyru: I think the pattern above waits for 10 seconds in the worst case.
In general it works super-fast.
You're always waiting for 10 seconds.

base-notebook/test/test_container_options.py
```diff
@@ -187,3 +187,28 @@ def test_group_add(container, tmpdir):
...
+        user="root",
+        working_dir="/home/",
+        environment=[
+            "NB_USER=user",
+            "CHOWN_HOME=yes",
+        ],
+        volumes={d: {"bind": "/home/jovyan/data", "mode": "rw"}},
+        command=["start.sh", "jupyter", "lab"],
+    )
+    time.sleep(2)
```
mathbunnyru, line 212: Could you please check rv here as it was done in test_group_add test?
```
c = container.run(...)
rv = c.wait(timeout=5)
assert rv == 0 or rv["StatusCode"] == 0
```

mathbunnyru (changes requested):

base-notebook/test/test_container_options.py
rigzba21: Added 👍

mathbunnyru: Thanks 👍

mathbunnyru:
@consideRatio could you take a look here please?
I didn't have much of an experience with start.sh file.

mathbunnyru:
I think this is good to go. I will merge as soon as tests pass.
Thanks @rigzba21 and sorry for making you wait 🙏

mathbunnyru (approved):

base-notebook/start.sh
```diff
@@ -54,7 +54,8 @@ if [ "$(id -u)" == 0 ] ; then
         # (it could be mounted, and we shouldn't create it if it already exists)
         if [[ ! -e "/home/${NB_USER}" ]]; then
             echo "Relocating home dir to /home/${NB_USER}"
```
consideRatio, line 56:
```suggestion
            echo "Copying home dir to /home/${NB_USER}"
```

This code is quite scary in general for anyone that has mounted something, which is what this PR has concluded and is mitigating.

I'm not feeling confident to say `mv`, `cp`, or `ln` is an acceptable practice. Perhaps `cp` is better than `mv` but yikes this feels very dangerous no matter what.

mathbunnyru: I agree this looks scary.
Is there any better way though? Especially if we don't want to break existing users.

consideRatio: Yeah its a bad situation no matter what is done I feel.

This will incur a performance hit as well now that didn't happen before as well, and its all hidden behind this layer of scripts doing things for users in a way that few really understands - even among us maintainers.

I'm inclined to think about removing a lot of scripts entirely and making what remain very stringently defined, tested, and documented.

mathbunnyru:
@rigzba21 please, merge master to your branch for final check.

PR #1424, opened 2021-08-09
Add DateTagger for tags like 2021-08-09

consideRatio:


mathbunnyru (approved):
Great, let's merge this as soon as tests pass.

PR #1427, opened 2021-08-09
Fix failure to tag correctly

### LINKED ISSUE #1426: Docker image version tags are incorrect
It appears that #1378 successfully has build and pushed an image with jupyterhub 1.4.2 which is the latest version. But from looking at [docker hub](https://hub.docker.com/r/jupyter/base-notebook/tags?page=1&ordering=last_updated) I conclude that it tagged `hub-1.4.1` instead of `hub-1.4.2` even though the image itself contain jupyterhub 1.4.2.

```
docker run -it --rm docker.io/jupyter/base-notebook:hub-1.4.1 jupyterhub --version
1.4.2
```

@mathbunnyru do you have any ideas about what has gone wrong? Did the tagger script download the `:latest` image from dockerhub instead of using the locally built image perhaps?

consideRatio:
Closes #1426 by ensuring that our ci system, after having built an image, makes it accessible to `docker` so that our later steps in our ci system don't go ahead and download an entirely different image and tags/tests using that image instead of the one we built.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ build-all: $(foreach I, $(ALL_IMAGES), build/$(I)) ## build all stacks
 build-multi/%: DARGS?=
 build-multi/%: ## build the latest image for a stack on both amd64 and arm64
 	@echo "::group::Build $(OWNER)/$(notdir $@) (system's architecture)"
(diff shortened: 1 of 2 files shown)


mathbunnyru:
Nice 👍

PR #1431, opened 2021-08-12
Fix example 2 in readme

fortierq:
Example 2 claims that JupyterLab is loaded but JUPYTER_ENABLE_LAB=yes is not set.


mathbunnyru:
Thanks, I will merge this, when tests pass.

Issue #1436, opened 2021-08-23
Type in Dockerfile affecting Arm64 builds?


consideRatio:
This looks like a typo, and that it shouldn't be `aarm64` but `aarch64`. But, why hasn't this failed earlier then? I note that the image is built and pushed to docker hub.

https://github.com/jupyter/docker-stacks/blob/70178b8e48d7825ceade6e2decf374bc3e792d90/base-notebook/Dockerfile#L103-L105

@holdenk what do you think about this?

mathbunnyru:
I think I solved the mystery.

```
    export miniforge_arch && \
    if [ "$miniforge_arch" == "aarm64" ]; then \
      miniforge_arch="arm64"; \
    fi; \
    miniforge_installer="${miniforge_python}-${miniforge_version}-Linux-${miniforge_arch}.sh" && \
```

`uname -m` returns `aarch64`.
So, `if` condition is wrong.
But, fortunately for us, miniforge naming is already good, so we didn't need any replacements in the first place.

https://github.com/conda-forge/miniforge/releases/
For example, the file is called `Mambaforge-4.10.3-5-Linux-aarch64.sh`

mathbunnyru:
I will fix this.
I will also add downloading the latest miniforge version, to make it follow our automatic updates principles.

PR #1438, opened 2021-08-24
Use latest mambaforge release

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1436

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -17,22 +17,6 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 USER root
 
-# ---- Miniforge installer ----
-# Default values can be overridden at build time
-# (ARGS are in lower case to distinguish them from ENV)
-# Check https://github.com/conda-forge/miniforge/releases
-# Conda version
-ARG conda_version="4.10.3"
-# Miniforge installer patch version
-ARG miniforge_patch_number="3"
-# Package Manager and Python implementation to use (https://github.com/conda-forge/miniforge)
-# - conda only: either Miniforge3 to use Python or Miniforge-pypy3 to use PyPy
-# - conda + mamba: either Mambaforge to use Python or Mambaforge-pypy3 to use PyPy
-ARG miniforge_python="Mambaforge"
-# Miniforge archive to install
-ARG miniforge_version="${conda_version}-${miniforge_patch_number}"
-
-
 # Install all OS dependencies for notebook server that starts but lacks all
 # features (e.g., download as all possible file formats)
 ENV DEBIAN_FRONTEND noninteractive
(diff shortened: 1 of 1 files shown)


mathbunnyru:
As we switched to not hardcoding version, I think it would be great to always use latest Mambaforge.
We will see how this works out.

base-notebook/Dockerfile
```diff
@@ -95,17 +79,18 @@ RUN mkdir "/home/${NB_USER}/work" && \
...
     miniforge_arch=$(uname -m) && \
-    export miniforge_arch && \
-    if [ "$miniforge_arch" == "aarm64" ]; then \
-      miniforge_arch="arm64"; \
-    fi; \
-    miniforge_installer="${miniforge_python}-${miniforge_version}-Linux-${miniforge_arch}.sh" && \
-    export miniforge_installer && \
-    wget --quiet "https://github.com/conda-forge/miniforge/releases/download/${miniforge_version}/${miniforge_installer}" && \
+    miniforge_installer="Mambaforge-Linux-${miniforge_arch}.sh" && \
+    wget --quiet "https://github.com/conda-forge/miniforge/releases/latest/download/${miniforge_installer}" && \
```
mathbunnyru, line 93: After this change, the only thing that will be needed to change version would be the "latest" word here.
For now, I don't add build variable, because I don't think users need this.
Anyway, it will be easy to do this in the future.

consideRatio: +1 for complexity reduction!

base-notebook/Dockerfile
```diff
@@ -95,17 +79,18 @@ RUN mkdir "/home/${NB_USER}/work" && \
 # Install conda as jovyan and check the sha256 sum provided on the download site
 WORKDIR /tmp
 
+# ---- Miniforge installer ----
+# Check https://github.com/conda-forge/miniforge/releases
+# Package Manager and Python implementation to use (https://github.com/conda-forge/miniforge)
+# - conda only: either Miniforge3 to use Python or Miniforge-pypy3 to use PyPy
+# - conda + mamba: either Mambaforge to use Python or Mambaforge-pypy3 to use PyPy
```
consideRatio, line 86: You are outlining the miniforge release binaries, but we are using mambaforge at all time and that's it right? I got a bit confused about this listing as we have a hardcoded opinion about mambaforge atm, and this didn't mention it while listing various options.

consideRatio: Concretely, I suggest just mentioning that we have opined to use the classic mambaforge installation.

consideRatio:
This LGTM besides a wish for a tweak to a comment!

base-notebook/Dockerfile
mathbunnyru: Yes, we're using mambaforge at all times, this is a small description of options available.

mathbunnyru: Improved comment

consideRatio (approved):
Oh, I didn't notice the test failure until now. I assume it is caused by a version bump of miniforge somehow, but it could also be due to a package has been bumped separately and influenced dependencies etc somehow.

consideRatio:
```
  E           AssertionError: {'skimage': AssertionError('Package [skimage] import failed\nassert 1 == 0\n  +1\n  -0')}
```

I observe that we do have `scikit-image` explicitly installed in the Dockerfile though...

image

https://github.com/jupyter/docker-stacks/blob/83ab5a86cb5b1c663d77cd824053054dfe454304/scipy-notebook/Dockerfile#L39

This is installed with `conda` though btw, I thought we had transitioned to `mamba` systematically. Is that related?

mathbunnyru:
No, this is not related.
The reason is in the pooch package.

Latest version doesn't work well with skimage.
It is already fixed in the skimage https://github.com/scikit-image/scikit-image/pull/5529.
And the release was made just a few minutes ago.

But we will have to wait for feedstock update.

So I pinned pooch version for now, so it should work now.
As soon as new version is released, I will delete this temporary fix.

Снимок экрана 2021-08-24 в 16 00 47

mathbunnyru:
Merged https://github.com/jupyter/docker-stacks/pull/1439, made tests locally.

mathbunnyru:
Merged master to this branch to check it.

consideRatio:
@mathbunnyru thanks for your work on this!!! :heart:

PR #1439, opened 2021-08-24
Pin pooch version to make skimage work, waiting for feedstock release

mathbunnyru:


.pre-commit-config.yaml
```diff
@@ -18,7 +18,7 @@ repos:
 
   # Lint: Dockerfile
   - repo: https://github.com/hadolint/hadolint.git
-    rev: v2.6.1
```
mathbunnyru, line 21: Hadolint 2.6.1 is also not yet available.
https://hub.docker.com/r/hadolint/hadolint/tags?page=1&ordering=last_updated

PR #1442, opened 2021-08-25
Do not pin pooch version

mathbunnyru:


mathbunnyru:
Updated scikit-image is already released, so we don't need this hotfix anymore.

Issue #1410, opened 2021-07-16
Don't request the absolute latest buildx version


consideRatio:
#1409 requested the latest version of buildx instead of accepting the one pre-installed in the GitHub runner's environment. We can remove that request and save a few seconds when https://github.com/actions/virtual-environments/blob/HEAD/images/linux/Ubuntu2004-README.md#tools lists version 0.6.0 of buildx.

PR #1443, opened 2021-08-25
Do not explicitly ask latest docker buildx version

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1410


mathbunnyru:
Now, Docker-Buildx 0.6.0 is listed, so we can use the default version.

consideRatio (approved):
Thanks!

PR #1456, opened 2021-09-03
Simplify Makefile

mathbunnyru:
Deleted DARGS and `--rm --force-rm` when building with dockerx.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,10 +48,9 @@ help:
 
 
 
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

consideRatio:
I've kept track of the GitHub workflows, and feel quite confident that this shouldn't cause wiki-issues given the order of execution of steps in the job etc.

Issue #1465, opened 2021-09-15
Recent changes to start.sh (mkdir; cp -R instead of mv) makes use-case for NB_USER too narrow (breaks existing z2jh deployments)


jlphillipsphd:
**What docker image you are using?**

Example: `jupyter/minimal-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

I am running z2jh on k3s, and typically use NB_USER to change jovyan to a user id obtained from LDAP, AzureAD, or Github OAuth.

The relevant values.yaml would be:
```
singleuser:
  uid: 0
  cmd: start-notebook.sh
```
The environment variable NB_USER is set using a custom authenticator (see above) which then allows the user to utilize their username in the container rather than jovyan. I currently -do not- change NB_UID nor NB_GID since this requires chown -R of /opt/conda if the user would like to make any changes (these are, of course, ephemeral changes, but this is nice for users since they can simply restart their container if they make a mess of anything).

**What steps do you take once the container is running to reproduce the issue?**

Example:

1. Install jhub using the z2jh helm chart.
2. Log into the hub.
3. Container fails to start due to incorrect permissions on /home/${NB_USER}/.local

**What do you expect to happen?**

The old behavior of start.sh is anticipated: first it trys moving /home/jovyan to /home/${NB_USER}, which actually fails due to permissions, but then falls back to ln -s which symlinks /home/${NB_USER} to /home/jovyan and the container can then start up as usual.

**What actually happens?**

The start.sh script tries to make the directory `/home/${NB_USER}` and then `cp -R /home/jovyan` to `/home/${NB_USER}` (which actually results in the files actually residing in `/home/${NB_USER}/jovyan` which doesn't appear to be what is intended). However, the permissions of .local are being as set root instead of $NB_USER and so the permissions issue arises and causes the container to stop. I read through the new documentation, which does indicate that you should basically specify -three- other pieces of information: 1. CHOWN_HOME="yes", 2. CHOWN_HOME_OPTS="-R" (probably always going to be needed just based on the above behavior), and 3. set the -working directory- to /home/${NB_USER} << (this feature is probably hard to get using the helm chart right? maybe there is a way to set it in z2jh but I can't think of how to do that at the moment).

[...]

PR #1466, opened 2021-09-15
Also copy hidden files/directories

twalcari:
`cp -R` does not copy hidden files/directories, causing issues (cfr #1465) .

Using `-a` instead of `-R` makes `cp` work more closely as `rsync -avx` (which the `cp -R` replaced in 0dbce4526e165ce716e4ca41a7384e8dce0bfd0b) . This solves the problem that `CHOWN_HOME` was required to change file permissions from root  to the local user. Setting the `CHOWN_HOME` environment variable to 'yes' in our JupyterHub deployment caused issues like:

> chown: cannot access '/home/custom-username': No such file or directory

when older images were started

Added by @mathbunnyru 
Fix: https://github.com/jupyter/docker-stacks/issues/1465

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -55,7 +55,7 @@ if [ "$(id -u)" == 0 ] ; then
         if [[ ! -e "/home/${NB_USER}" ]]; then
             echo "Copying home dir to /home/${NB_USER}"
             mkdir "/home/${NB_USER}"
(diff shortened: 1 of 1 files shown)


mathbunnyru:
This seems reasonable to copy everything as an archive.

In a normal situation I would have asked to add a test this kind of situation (adding hidden file/dir to the current test is fine).
@twalcari I would be glad, if you added this as a separate PR.

Thanks for the bugfix!

PR #1467, opened 2021-09-15
Add test for copying of hidden folder

twalcari:
Test if the hidden `.jupyter` folder is properly copied when the user is changed via `NB_USER` environment variable. This folder should exist as `jupyter notebook --generate-config` is executed to generate `/home/jovyan/.jupyter/jupyter_notebook_config.py` .

Test for fix #1466 for issue #1465.


mathbunnyru:
@twalcari is this still a draft?

twalcari:
You can go ahead now. I hadn't tested it locally, hence why I kept it as a draft for testing first.

Issue #1472, opened 2021-09-27
Error while creating/editing files under "work" directory


rehanguha:
Hi,
I have using the below command from the docs

```
docker run --rm -p 10000:8888 -e JUPYTER_ENABLE_LAB=yes -v "${PWD}":/home/jovyan/work jupyter/datascience-notebook:33add21fab64
```

I have an issue to create any file under the work directory. I am getting permission denied when trying to save or create any new file with any extension.

Bidek56:
Please try this command, it works for me, as per [GH](https://github.com/jupyter/docker-stacks/issues/199#issuecomment-217327748)

docker run --rm -p 10000:8888 -e JUPYTER_ENABLE_LAB=yes --user=root  -e NB_UID=\`id -u\`  -v "${PWD}":/home/jovyan/work jupyter/datascience-notebook:33add21fab64

PR #1474, opened 2021-09-29
Updated the Example 3 bash script

rehanguha:
Existing script was not able to write contents on the mounted volume.
https://github.com/jupyter/docker-stacks/issues/1472

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,6 @@ Docker destroys the container after notebook server exit, but any files written 
 docker run --rm -p 10000:8888 -e JUPYTER_ENABLE_LAB=yes --user=root -e NB_UID=`id -u` -v "${PWD}":/home/jovyan/work jupyter/datascience-notebook:33add21fab64
 ```
 
-
 ## Contributing
 
 Please see the [Contributor Guide on ReadTheDocs](https://jupyter-docker-stacks.readthedocs.io/) for
--- a/README.md
+++ b/README.md
@@ -48,8 +48,9 @@ where `hostname` is the name of the computer running docker and `token` is the s
 Docker destroys the container after notebook server exit, but any files written to `~/work` in the container remain intact on the host.
 
 ```bash
-docker run --rm -p 10000:8888 -e JUPYTER_ENABLE_LAB=yes -v "${PWD}":/home/jovyan/work jupyter/datascience-notebook:33add21fab64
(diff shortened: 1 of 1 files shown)


mathbunnyru:
@rehanguha thanks for the PR.

I have a few comments about this PR:

1. Do you know the reason, why the default command doesn't work for you? (it works for me fine) It has definitely to do something with host and docker permissions, they might be wrong somewhere.
2. We can't just fix this example, because there are many other examples as well and it won't be consistent. Also, we need to update the docs.

rehanguha:
Note: I am used to docker but I am new to this Jupyter Docker stack.
When I was setting up a new system, I have installed the docker as per the docker website then tried to run the image and I getting an error when ever I was trying to write something inside the work folder.

I tired it multiple times and I was facing the same error.

So, I want to clarify the setting for docker was:
1. https://docs.docker.com/engine/install/ubuntu/
2. https://docs.docker.com/engine/install/linux-postinstall/

I was expecting a simple pull and run like Tensorflow(tensorflow/tensorflow:latest-gpu-jupyter) dockers which I have used before. I just wanted to try Jupyter docker stacks. 😄 

Thank you for commenting.

mathbunnyru:
Ok, then I think we will have to wait for some explanation why this was happening for you.
I'm glad to hear that your issue is resolved and I hope there are not so many people with similar problems.

consideRatio:
So this is a PR to suggest starting the container as root, because the user otherwise may not have access to write to the mounted home directory?

That sounds like a troublesome workaround. If you would then write a file, you would end up making the ownership of that file, within the /home/my-user directory, be root. Which in turn, makes you unable to work with that file later, unless you are root.

I'm sorry to be boring, but I suggest this PR isn't accepted as it seems like a forceful workaround that will cause problems in the end.

rehanguha:
> So this is a PR to suggest starting the container as root, because the user otherwise may not have access to write to the mounted home directory?

Yes. That is the case which is happening. 

> That sounds like a troublesome workaround.

I agree. But I cannot find a better way to write a file to the mounted directory. Please do suggest me a way so that this problem can be solved.

We can closed the PR as it even feel this not a great workaround. But a solutions needs to be in place so that this can be solved.

consideRatio:
> But a solutions needs to be in place so that this can be solved.

My guess is that you can check the user id of the current linux system's user, then set NB_UID to that, mount the current linux system's user's home folder. Then, you will have the correct permissions I think.

What may be in scope of this project is to improve documentation about this.

PR #1483, opened 2021-10-15
Rename `blacklist` -> `blocklist`

NickCrews:
A simple change, but also quite easy.

See https://en.wikipedia.org/wiki/Master/slave_(technology)


NickCrews:
lol, so that was a complete overhaul of my PR, thanks @mathbunnyru !

Issue #1480, opened 2021-10-13
jupyter/tensorflow-notebook: TypeError: Parameter to MergeFrom() must be instance of same class: expected tensorflow.TensorShapeProto got tensorflow.TensorShapeProto.


piernov:
**What docker image you are using?**

Example: `jupyter/tensorflow-notebook`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

Example: `docker run -it --rm -p 8888:8888 -e JUPYTER_ENABLE_LAB=yes -v "$(pwd)"/d1-notebooks:/home/jovyan/d1-notebooks jupyter/tensorflow-notebook`

**What steps do you take once the container is running to reproduce the issue?**

Example:

1. Visit <http://localhost:8888>
2. Start a Python notebook
3. Write in a code cell:
```
from keras.models import Sequential
model = Sequential()
```
4. Run the cell

**What do you expect to happen?**

No output, a new Sequential model is created.

**What actually happens?**
Error output:
```
TypeError: Parameter to MergeFrom() must be instance of same class: expected tensorflow.TensorShapeProto got tensorflow.TensorShapeProto.
```

**Additional information**
This is an upstream bug. See : https://github.com/tensorflow/tensorflow/issues/50545 .
Please apply https://github.com/tensorflow/tensorflow/pull/51450 if possible.
As of now, it is impossible to use this Docker image to train Deep Neural Networks.

mathbunnyru:
I've fixed the issue, I will revert the patch part when the upstream project releases a new version (it will actually break and we will have to revert the fix).

Waiting for the PR CI, then I will merge.

PR #1485, opened 2021-10-16
Fix tensorflow keras bug

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1480


mathbunnyru:
Merging without CI, it worked fine on my machine.

PR #1506, opened 2021-11-01
Add custom fonts loading option

ggaaooppeenngg:
Add a build argument CUSTOM_FONT to load subset fonts.

### DIFF
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -60,6 +60,17 @@ RUN git clone https://github.com/PAIR-code/facets.git && \
 # Import matplotlib the first time to build the font cache.
 ENV XDG_CACHE_HOME="/home/${NB_USER}/.cache/"
 
+ARG CUSTOM_FONT=""
+RUN if [ "$CUSTOM_FONT" == "Source Han Sans CN" ]; then \
+       PYV=$(ls /opt/conda/lib | grep ^python); \
+       wget --quiet -P /opt/conda/lib/$PYV/site-packages/matplotlib/mpl-data/fonts/ttf/ https://mirrors.cloud.tencent.com/adobe-fonts/source-han-sans/SubsetOTF/CN/SourceHanSansCN-Normal.otf; \
+       sed -i 's/#font.family/font.family/g' /opt/conda/lib/$PYV/site-packages/matplotlib/mpl-data/matplotlibrc; \
+       sed -i 's/#font.sans-serif:/font.sans-serif: Source Han Sans CN,/g' /opt/conda/lib/$PYV/site-packages/matplotlib/mpl-data/matplotlibrc; \
+       sed -i 's/#axes.unicode_minus: True/axes.unicode_minus: False/g' /opt/conda/lib/$PYV/site-packages/matplotlib/mpl-data/matplotlibrc; \
+       rm -rf ~/.cache/matplotlib; \
+       python -c 'import matplotlib.font_manager;print("Font loaded: ",("Source Han Sans CN" in [f.name for f in matplotlib.font_manager.fontManager.ttflist]))'; \
+    fi
+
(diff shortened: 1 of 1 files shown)


scipy-notebook/Dockerfile
```diff
@@ -60,6 +60,17 @@ RUN git clone https://github.com/PAIR-code/facets.git && \
...
 
+ARG CUSTOM_FONT=""
+RUN if [ "$CUSTOM_FONT" == "Source Han Sans CN" ]; then \
+       PYV=$(ls /opt/conda/lib | grep ^python); \
+       wget --quiet -P /opt/conda/lib/$PYV/site-packages/matplotlib/mpl-data/fonts/ttf/ https://mirrors.cloud.tencent.com/adobe-fonts/source-han-sans/SubsetOTF/CN/SourceHanSansCN-Normal.otf; \
+       sed -i 's/#font.family/font.family/g' /opt/conda/lib/$PYV/site-packages/matplotlib/mpl-data/matplotlibrc; \
+       sed -i 's/#font.sans-serif:/font.sans-serif: Source Han Sans CN,/g' /opt/conda/lib/$PYV/site-packages/matplotlib/mpl-data/matplotlibrc; \
+       sed -i 's/#axes.unicode_minus: True/axes.unicode_minus: False/g' /opt/conda/lib/$PYV/site-packages/matplotlib/mpl-data/matplotlibrc; \
+       rm -rf ~/.cache/matplotlib; \
+       python -c 'import matplotlib.font_manager;print("Font loaded: ",("Source Han Sans CN" in [f.name for f in matplotlib.font_manager.fontManager.ttflist]))'; \
```
mathbunnyru, line 71: You have to use `CONDA_DIR` variable here, because people might build their images with another conda directory.

mathbunnyru (changes requested):
I actually think this is a customization which not many people want.
Also it will work only with one font only.

Maybe it would be easier to you just to create your image based on the original image and customize it the way you want.
And, please, consider sending this as a recipe - this way everyone will be able to customize their images the same way.

Here is the guide:
https://jupyter-docker-stacks.readthedocs.io/en/latest/contributing/recipes.html

scipy-notebook/Dockerfile
```diff
@@ -60,6 +60,17 @@ RUN git clone https://github.com/PAIR-code/facets.git && \
 # Import matplotlib the first time to build the font cache.
 ENV XDG_CACHE_HOME="/home/${NB_USER}/.cache/"
 
+ARG CUSTOM_FONT=""
```
mathbunnyru, line 63: Better variable name would be CUSTOM_MATPLOTLIB_FONT, because I think this only changes the font for matplotblib, doesn't it?

scipy-notebook/Dockerfile
```diff
@@ -60,6 +60,17 @@ RUN git clone https://github.com/PAIR-code/facets.git && \
...
+RUN if [ "$CUSTOM_FONT" == "Source Han Sans CN" ]; then \
+       PYV=$(ls /opt/conda/lib | grep ^python); \
+       wget --quiet -P /opt/conda/lib/$PYV/site-packages/matplotlib/mpl-data/fonts/ttf/ https://mirrors.cloud.tencent.com/adobe-fonts/source-han-sans/SubsetOTF/CN/SourceHanSansCN-Normal.otf; \
```
mathbunnyru, line 66: Creating a variable MPL_DATA `/opt/conda/lib/$PYV/site-packages/matplotlib/mpl-data/` will make the code easier to read.

consideRatio:
This is a feature proposal that requires a rebuild of the image, from scratch, and it won't be run automatically by jupyter/docker-stacks. So, doesn't it make sense for @ggaaooppeenngg and the jupyter/docker-stacks project if this is instead built a layer on top of jupyter/docker-stacks.

So...

```Dockerfile
FROM jupyter/base-notebook:latest

# additional changes
```

PR #1508, opened 2021-11-02
Add recipe for installing custom fonts

ggaaooppeenngg:
ref #1506 

### DIFF
--- a/docs/using/recipes.md
+++ b/docs/using/recipes.md
@@ -557,12 +557,12 @@ The example below is a Dockerfile to load Source Han Sans with normal weight whi
 ```dockerfile
 FROM jupyter/scipy-notebook:latest
 
-RUN PYV=$(ls $CONDA_DIR/lib | grep ^python) && \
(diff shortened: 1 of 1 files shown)


docs/using/recipes.md
```diff
@@ -549,3 +549,20 @@ RUN echo "from pyspark.sql import SparkSession" > /tmp/init-delta.py && \
...
 ```
+
+## Add Custom Font in Scipy notebook
+
+The example below is a Dockerfile to load Source Han Sans with normal weight which is usually used for web.
+
+```dockerfile
+FROM jupyter/scipy-notebook:latest
+
+RUN PYV=$(ls "${CONDA_DIR}/lib" | grep ^python) && \
```
mathbunnyru, line 560: This is a bit dangerous because some package might be called python-something, for example, but let's keep it this way - if someone faces a problem then, the fix will be easy enough.

mathbunnyru:
Thanks, @ggaaooppeenngg. I added some bash quotes to match it to the current codestyle of our bash scripts and this is good to go.

PR #1510, opened 2021-11-02
Refactor github workflows

mathbunnyru:


.github/workflows/docker.yml
```diff
@@ -5,36 +5,53 @@ on:
...
   pull_request:
-    paths-ignore:
-      - ".github/ISSUE_TEMPLATE/**"
-      - "*.md"
-      - "binder/**"
-      - "docs/**"
-      - "examples/**"
-      - ".pre-commit-config.yaml"
-      - ".readthedocs.yaml"
+    paths:
```
mathbunnyru, line 8: I think this approach worked better for me.

.github/workflows/docker.yml
```diff
@@ -5,36 +5,53 @@ on:
...
+      - "requirements-dev.txt"
 
 jobs:
   build-test-publish-images:
     name: Build, test, and publish Docker Images
     runs-on: ubuntu-latest
     permissions:
       contents: write
-    if: >
-      !contains(github.event.head_commit.message, 'ci skip') &&
```
mathbunnyru, line 36: Automatic pushes do not need to be checked, because they won't trigger github workflows, as far as I know :)

.github/workflows/docker.yml
consideRatio: :+1: a bit messier to maintain this file, but for this repo where CI workflows are taking a long time, it is really really appreciated to not have to run them unless absolutely needed.

PR #1511, opened 2021-11-04
Install binary tini instead of debian package

ggaaooppeenngg:
The tini package is not available on ubuntu18.04.
Use binary installtion instead of apt-get.

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -21,15 +21,19 @@ USER root
 # features (e.g., download as all possible file formats)
 # Install tini: init for containers
 ENV DEBIAN_FRONTEND noninteractive
+
+ENV TINI_VERSION v0.19.0
+
 RUN apt-get update --yes && \
     apt-get install --yes --no-install-recommends \
-    tini \
     wget \
     ca-certificates \
     sudo \
     locales \
     fonts-liberation \
     run-one && \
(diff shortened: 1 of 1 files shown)


base-notebook/Dockerfile
```diff
@@ -21,15 +21,19 @@ USER root
 # features (e.g., download as all possible file formats)
 # Install tini: init for containers
 ENV DEBIAN_FRONTEND noninteractive
+
+ENV TINI_VERSION v0.19.0
```
mathbunnyru, line 25: Could you please use `latest` version and do not even create this env variable?
Github allows it and we've been using the latest versions where it is easily possible and it worked well.

mathbunnyru (changes requested):

consideRatio:
Hmmm, I would rather avoid adding complexity like this. I think it is a slippery slope if we make this, then perhaps there are more changes we need to make in order to support Ubuntu 18.04 as well.

After a while, there will be 22.04, and then are we working to have a Dockerfile that builds correctly on all three LTS versions, even though we only test against one - or do we add tests for all even though our test suite is very heavy at the moment?

I suggest this PR isn't accepted to avoid adding complexity, not just in this PR itself, but in other PRs about making changes to support multiple versions of ubuntu.

ggaaooppeenngg:
ubuntu18.04 is not a release that is too old to give up. Just leave it here for people to find the solution.

PR #1514, opened 2021-11-08
Check workflow works

mathbunnyru:


mathbunnyru:
This launched needed workflow, so the issue is fixed.

PR #1519, opened 2021-11-10
Update secure_path and environment variables

maresb:


base-notebook/start.sh
```diff
@@ -130,8 +130,32 @@ if [ "$(id -u)" == 0 ] ; then
...
+    # - We use the `--set-home` flag to set the HOME variable appropriatly.
+    #
+    # - We reduce the `env_delete` list of default variables to be deleted. It
+    #   has higher priority than the `--preserve-env` flag and `env_keep`
+    #   configuration.
+    #
+    # - We disable the `secure_path` which is set by default in /etc/sudoers as
+    #   it would override the PATH variable.
+    echo 'Defaults !secure_path' > /etc/sudoers.d/added-by-start-script
+    echo 'Defaults env_delete -= "PATH LD_* PYTHON*"' >> /etc/sudoers.d/added-by-start-script
```
consideRatio, line 158: I'm not able to rule out this isn't a big security mistake. Until I am, I'm opposed merging this part specifically. I've lost track a bit on the goal and what strategies are available to reach the goal, and what drawbacks the various strategies have. I think to move this onwards, I'd like to have it outlined what we attempt to accomplish with this, and how this is the best strategy to do so, and why that isn't a big security no-no.

mathbunnyru:
@maresb should we close this issue?
I think it's totally fine to close this one.

I have no idea what it achieves and if that is the correct way to do it :)

maresb:
To me it's also not so clear that anything here serves a necessary purpose.

The comment however looks extremely useful. It may be worthwhile to merge it, possibly after adjustments.

mathbunnyru:
I agree! Could you maybe draft a separate PR with the comment only?

Issue #1517, opened 2021-11-09
Make the start script fail loudly with an error where it makes sense


consideRatio:
This issue represents the following comments from another PR, so we can mark them as resolved without forgetting them.

https://github.com/jupyter/docker-stacks/pull/1512#discussion_r745520508
https://github.com/jupyter/docker-stacks/pull/1512#discussion_r745526275

PR #1512, opened 2021-11-07
Rework of start.sh script, take 2

maresb:
Towards fixing #1270 I am convinced by @consideRatio's approach in #1052.

His #1052 grew quite large, and generated a lot of interesting discussion. It also introduced some very valuable enhancements to `start.sh` to improve comments, messages, and make it more robust.

I have rebased a subset of his commits which I consider less controvertial. In theory, they should not significantly change anything except for the messages, but several edge cases should now be handled better.

I hope that this can be tested and approved. Then I hope to
- bring in the remaining commits from #1052
- implement #1270 based on an approach which reverts 4ef8a78, likely with a few additional subtle modifications.
- Try to close #1478

PR #1523, opened 2021-11-10
Fail if creating symlink fails

mathbunnyru:


mathbunnyru:
I gave this a thought: this was actually original behaviour.

https://github.com/jupyter/docker-stacks/pull/1512/files#diff-41f90d7afcdae13f8516195e078d0a203972b5c5105851eaecfc0a98e9739107L58

mathbunnyru:
`cp -a /home/jovyan/. "/home/${NB_USER}/" || ln -s /home/jovyan "/home/${NB_USER}"`
If the second command failed, it did fail the entire script.
So, this fix makes sense.

I will remove one of two comments from here.
https://github.com/jupyter/docker-stacks/issues/1517

consideRatio (approved):
LGTM! I like strict failures a lot over having assumptions break without noticing warnings etc.

PR #1524, opened 2021-11-11
start.sh: JUPYTER_ENV_VARS_TO_UNSET introduced

maresb:
Cherry-pick of b44b7ab. I'm working through https://github.com/jupyter/docker-stacks/pull/1512#issuecomment-962672049

> Allow you to define variables to be unset before running the command
> that start.sh is supposed to start. These variables will still be
> available in the hooks run before.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -39,6 +39,18 @@ run-hooks () {
     echo "${0}: done running hooks in ${1}"
 }
 
+# A helper function to unset env vars listed in the value of the env var
+# JUPYTER_ENV_VARS_TO_UNSET.
+unset_explicit_env_vars () {
+    if [ -n "${JUPYTER_ENV_VARS_TO_UNSET}" ]; then
+        for env_var_to_unset in $(echo "${JUPYTER_ENV_VARS_TO_UNSET}" | tr ',;:' ' '); do
+            echo "Unset ${env_var_to_unset} due to JUPYTER_ENV_VARS_TO_UNSET"
+            unset "${env_var_to_unset}"
+        done
+        unset JUPYTER_ENV_VARS_TO_UNSET
+    fi
+}
+
 
 # NOTE: This hook will run as the user the container was started with!
 run-hooks /usr/local/bin/start-notebook.d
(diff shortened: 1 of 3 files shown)


base-notebook/start.sh
```diff
@@ -39,6 +39,18 @@ run-hooks () {
     echo "${0}: done running hooks in ${1}"
 }
 
+# A helper function to unset env vars listed in the value of the env var
+# JUPYTER_ENV_VARS_TO_UNSET.
+unset_explicit_env_vars () {
+    if [ -n "${JUPYTER_ENV_VARS_TO_UNSET}" ]; then
+        for env_var_to_unset in $(echo "${JUPYTER_ENV_VARS_TO_UNSET}" | tr ',;:' ' '); do
```
mathbunnyru, line 46: Let's only separate by comma - I don't think we have to give the users the choice of separator here.

base-notebook/test/test_container_options.py
```diff
@@ -240,7 +240,6 @@ def test_container_not_delete_bind_mount(container, tmp_path):
     assert len(list(tmp_path.iterdir())) == 1
 
 
-@pytest.mark.skip(reason="not yet implemented; TODO: cherry-pick b44b7ab")
```
mathbunnyru, line 243: I have read the test:
- `SECRET_LOVE` only appears once is this right?
- we should probably check this for non-root user as well?

maresb: Hmm, I'm just looking carefully for the first time myself, and I also notice that `SECRET_ANIMAL` is never tested. I'm not sure about the intention of this test.

We should definitely cover the non-root user with a test since that's a separate execution thread in the script.

maresb:
Feel free to commit to my branch, since I won't be able to work on this for a while. I mainly wanted to cherry-pick the remaining commits into separate PRs before I forgot how everything went.

base-notebook/test/test_container_options.py
mathbunnyru: First part is fixed

mathbunnyru (changes requested):
> We should definitely cover the non-root user with a test since that's a separate execution thread in the script

Only this one is left here.

mathbunnyru:
I added non-root case, let's hope it's gonna work.

Issue #1233, opened 2021-02-07
Consider redoing all scripts with Xonsh


maresb:
I've become a big fan of the [Xonsh](https://xon.sh/) shell scripting language. It allows for almost seamless mixing between Bash and Python. Thus you get the sanity of Python, but you can still simply run a command without needing `subprocess`.

I think it may provide a viable way forward for initiatives which have stalled out, such as #1052 and #1140.

consideRatio:
I like the idea of sticking with Makefile/bash/python even though I don't yet know of Xonsh. I have grown fond whatever tech that can be understood using well known tech as it reduce the threshold to maintain and contribute to a project.

mathbunnyru:
Hi @maresb!
Would you like to work on this feature? 😄

I think the most significant issues in our images reside in shell scripts currently (especially, in `start.sh`), and it would be nice to move to pure Python in the future, so to me, `xonsh` is an excellent intermediate step.
I think it might make fixing some issues easier.

maresb:
Hey @mathbunnyru, thanks for thinking of me! I'm going to pass on this one though. (I have too much other stuff going on already.)

mathbunnyru:
> Hey @mathbunnyru, thanks for thinking of me! I'm going to pass on this one though. (I have too much other stuff going on already.)

I totally get it ❤️

mathbunnyru:
Unfortunately, I think `Xonsh` won't help us very much.
It will definitely look nicer in some parts, like logging.
But we'll still have all the things with renaming user, and these are just calling linux user management utils like `usermod`.
On the other hand, we'll have to deal with `xonsh` vs bash differences and it'll be more difficult for the contributors to change `start.sh` (or `start.xsh`).
Here is the translation guide: https://xon.sh/bash_to_xsh.html

https://github.com/jupyter/docker-stacks/issues/1140 mentioned in the issue is solved though by using pure python scripts in [tagging dir](https://github.com/jupyter/docker-stacks/tree/main/tagging).  (and I think it works beautifully now)

So, I'm closing this issue as not planned.

PR #1526, opened 2021-11-11
Refactor logging in start.sh and add JUPYTER_DOCKER_STACKS_QUIET

maresb:
Cherry-pick of 63295ba, working through  https://github.com/jupyter/docker-stacks/pull/1512#issuecomment-962672049

> It is meant to allow you to opt out from non-error non-warning logs
> generated by start.sh.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -4,11 +4,11 @@
 
 set -e
 
-# The _log function is passed everything this script wants to log. It will
+# The _log function is used for everything this script wants to log. It will
 # always log errors and warnings, but can be silenced for other messages
 # by setting JUPYTER_DOCKER_STACKS_QUIET environment variable.
 _log () {
-    if [[ "$*" == "ERROR:"* ]] || [[ "$*" == "WARNING:"* ]] || [[ "$JUPYTER_DOCKER_STACKS_QUIET" == "" ]]; then
+    if [[ "$*" == "ERROR:"* ]] || [[ "$*" == "WARNING:"* ]] || [[ "${JUPYTER_DOCKER_STACKS_QUIET}" == "" ]]; then
         echo "$@"
     fi
 }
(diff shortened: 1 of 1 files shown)


mathbunnyru (approved):
Looks awesome:

Possible future improvements (just a note for someone, who will want to improve our docker stacks):
1. Always add log_level like this: `_log "INFO" "my_message"
2. Based on log level color can be added.
3. Exit on fatal errors (or maybe just errors)
4. Add timestamps, like `INFO[2021-08-23 11:00:00.000000] my_message`
5. Stop using `bash`, haha 👍 

@maresb maybe you could add a test for this variable?

mathbunnyru:
@maresb 
I added an issue to improve log function, I think, if you added a test, then it would be perfect.
Another thing to add is a line-or-two of documentation.

maresb:
> 5. Stop using `bash`, haha :+1:

:arrow_up: :heavy_check_mark:  :arrow_up: :heavy_check_mark:  :arrow_up: :heavy_check_mark: 

I'm convinced that this right here is the root of most of our problems.

If we get a comprehensive test suite for `start.sh` running, I think it'd be a piece of cake to introduce #1233 and bring it up to feature parity. Once we're convinced it's reliably working, then ditch `start.sh` and maintenance and features become *so* much easier and more sane.

maresb:
Hmm, seems to be a CI error, I'll restart.

mathbunnyru:
@maresb @consideRatio I gave this a thought - do we really need JUPYTER_DOCKER_STACKS_QUIET?
I think this script generates only a couple of lines of logs in the worst case and it's not a big deal.

I personally think it's easier to remove this variable and merge this PR without adding a test case.
If someone will need this variable, then we will gladly accept the PR in the future.

consideRatio:
I recall adding it initially because there was some noise about "if this emits logs, it means my automation runs into trouble that parses the output" or similar. I've lost track of the change motivation at this point, but I recall it wasn't stemming from solving a problem of my own.

mathbunnyru:
I updated the issue, let's merge this 👍

PR #1531, opened 2021-11-15
Fix NOTEBOOK_ARGS pass

mathbunnyru:


mathbunnyru:
Merging this, because it might be important for some users.

PR #1534, opened 2021-11-16
Build and test amd64 docker images in a dedicated workflow for quick feedback

mathbunnyru:
The idea behind this is simple - let's build x86 images in a different workflow.
It's enough for most PRs and we can skip long x86+arm builds and merge quicker.

### DIFF
--- a/.github/workflows/docker-x86.yml
+++ b/.github/workflows/docker-x86.yml
@@ -1,0 +1,74 @@
+name: Build and test x86 Docker Images
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/docker-x86.yml"
+
+      - "all-spark-notebook/**"
+      - "base-notebook/**"
+      - "datascience-notebook/**"
+      - "minimal-notebook/**"
+      - "pyspark-notebook/**"
+      - "r-notebook/**"
+      - "scipy-notebook/**"
+      - "tensorflow-notebook/**"
+
+      - "tagging/**"
+      - "test/**"
+      - "conftest.py"
+      - "Makefile"
+      - "pytest.ini"
+      - "requirements-dev.txt"
+  push:
+    branches:
+      - main
+      - master
+    paths:
+      - ".github/workflows/docker-x86.yml"
+
+      - "all-spark-notebook/**"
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-x86.yml
```diff
@@ -0,0 +1,74 @@
...
+      - "r-notebook/**"
+      - "scipy-notebook/**"
+      - "tensorflow-notebook/**"
+
+      - "tagging/**"
+      - "test/**"
+      - "conftest.py"
+      - "Makefile"
+      - "pytest.ini"
+      - "requirements-dev.txt"
```
consideRatio, line 44: Allows us to run this manually, which is very useful to verify if something has broken due to a PR or due to a change over time.

```suggestion
      - "requirements-dev.txt"
  workflow_dispatch:
```

.github/workflows/docker-x86.yml
```diff
@@ -0,0 +1,74 @@
+name: Build and test x86 Docker Images
```
consideRatio, line 1: Regarding referencing this as x86, when using docker buildx, we refer to x86-64 as amd64 I think. I suggest we stick with the amd64 reference for consistency.

consideRatio, line 1: Can you add a descriptive comment in the top to clarify the purpose of having this dedicated workflow file?

The point as I understand it, is to run another workflow that will finish a lot quicker, allowing us to conclude if things work or not far quicker, when we don't expect there to be any platform specific breaking change.

.github/workflows/docker-x86.yml
mathbunnyru: Nice idea!
Maybe it's worth adding to other workflows as well.

.github/workflows/docker-x86.yml
mathbunnyru: Done

mathbunnyru: Yes, you got it right. Done.

mathbunnyru:
Nice, it takes less than 30 minutes, merging this one.

PR #1543, opened 2021-12-01
Update Julia to 1.7.0

tomyun:
A [new stable version](https://julialang.org/blog/2021/11/julia-1.7-highlights/) just came out.


mathbunnyru:
Thanks @tomyun!

Issue #1407, opened 2021-07-16
Improve performance of CI system


consideRatio:
We have crippled our CI systems performance after introducing support for arm64 based images. A key reason for this is that emulation of arm64 images from the amd64 based runners github provide is far worse, besides the fact that we end up building base-notebook and minimal-notebook for arm64 in sequence alongside the other images now.

I'm not fully sure how we should optimize this long run, but under the assumption that we will have high performance self-hosted arm64 based GitHub Action runners that can work in parallel to the amd64 runners. Below is an overview of a very optimized system, where several parts can be done separately.

1. __Nightly builds__
   We have nightly builds with `:nightly-amd64` and `nightly-arm64` tags
2. __amd64 / arm64 in parallel__
   All tests for amd64 and arm64 run in parallel, relying on `nightly-amd64` and `nightly-arm64` caches
3. __Images in parallel where possible__
   All tests for individual images are run in a dedicated job that `needs` its base image job to complete.

   Some images can run in parallel:
   - base
   - minimal
   - scipy | r
   - tensorflow | datascience | pyspark
   - all-spark
4. __Avoid rebuilds when merging__
   Tests finish by updating a github container registry associated with a PR. By doing so, our publishing job on merge to master can opt to use the images as they were built during tests if they are considered fresh enough.
5. __Parallel manifest creation__
   Merge to default branch triggers manifest creation jobs on both amd64 and arm64. If we opt to not optimize using step 4 then this could also build fresh images using nightly cache first.
6. __Combine manifests into one before pushing to official registry__
   Merge to default branch triggers a job that pulls both the amd64 image and arm64 image and defines a combined docker manifest which is then pushed to our official container registry. I think this could be done with something like `docker manifest create <name of combined image> <amd64 only image> <arm64 only image>` but @manics knows more and I lack experience with this.
   
### Standalone performance issue

This standalone issue will go away by using better strategies like above. It isn't so critical to fix either I'd say. But currently, we build minimal-notebook again without using cache during `push-multi` assuming `push-multi` for `base-notebook` has already run. It is because we re-tag jupyter/base-notebook:latest I think.

mathbunnyru:
Related: https://github.com/jupyter/docker-stacks/issues/1203

mathbunnyru:
Added this issue to the milestone.

mathbunnyru:
Another idea is to improve PRs speed - build ARM images only in master or when the commit has some pre-defined string.
[...]

PR #1545, opened 2021-12-10
Try to use old mamba version

mathbunnyru:


base-notebook/Dockerfile
```diff
@@ -81,7 +81,11 @@ WORKDIR /tmp
 # CONDA_MIRROR is a mirror prefix to speed up downloading
 # For example, people from mainland China could set it as
 # https://mirrors.tuna.tsinghua.edu.cn/github-release/conda-forge/miniforge/LatestRelease
-ARG CONDA_MIRROR=https://github.com/conda-forge/miniforge/releases/latest/download
+
+# Using fixed version of mamba, because the latest one has problems with arm under qemu
+# See: https://github.com/jupyter/docker-stacks/issues/1539
+ARG CONDA_MIRROR=https://github.com/conda-forge/miniforge/releases/download/4.10.3-7
```
minrk, line 87: This sets the _initial_ mamba version, but the `conda update --all` in the next step always upgrades mamba to latest.

mathbunnyru: You're right, I will solve this another way.

mathbunnyru:
My approach should work, but we have a much bigger problem.

https://github.com/jupyter/docker-stacks/runs/4493047317?check_suite_focus=true#step:7:4475
```
#5 [1/4] FROM docker.io/jupyter/minimal-notebook@sha256:5d8ba694b92d9fe5802529b7ccf8bad23de6632d0bbaa92fc3fdc8a31cdc9c9c
4476
  #5 resolve docker.io/jupyter/minimal-notebook@sha256:5d8ba694b92d9fe5802529b7ccf8bad23de6632d0bbaa92fc3fdc8a31cdc9c9c done
4477
  #5 sha256:2f66eefcb707429dfa3d1b8d0cdcbf330bb3d2624db2821eb2eb2e7323e0358a 1.48kB / 1.48kB 0.0s done
4478
  #5 sha256:fd44b7e34afa5f14902e0239880773be42ca449f5d1dbab726b4a17cde4f48f4 41.11MB / 235.96MB 0.3s`
```

Because we build so many images, they take too much space and we end up pulling images from dockerhub.
Which means we're effectively building not the code which is at the current branch, but some mix of pas image and the new code.

mathbunnyru:
@consideRatio I will for now completely disable arm images building at all :(
I think approach without multi-arch image will work better for us, it might be a little bit more confusing for the users, but I think it will work better for them as well.

mathbunnyru:
Nice, the only `resolve docker.io` I see are `ubuntu:focal` images.
Let's get mamba pinning back, but this time I will only pin it for arm images.

consideRatio:
> @consideRatio I will for now completely disable arm images building at all :(
I think the approach "build everything with some hacks in one place" doesn't work well for us.

That sounds reasonable. Given the limited maintenance capacity we have, and that this blocks everything if not maintained, cutting away something to save something else seems reasonable.

I think this strategy hasn't been robust enough given that each failure cause a lot of additional work and is hard to resolve due to the monolith CI jobs. I consider a resolution towards having arm again after disabling it would be to put in more work in making the CI system work in a more composable manner (#1407), I think that would make it handle multiple architectures in a more maintenance sustainable way.

---

Are you making some last attempts to get things to work atm?

mathbunnyru:
Yes, let's wait for the latest build - it should run faster (because caching images works) and it gives expected result (because of caching).
It might be the issue with latest mamba, but it might be something else :(

mathbunnyru:
Finally, this seems to work.
Let's merge this and hope it will work during push step.

consideRatio:
@mathbunnyru it worked! You are amazing! Thank you for your thorough work!!! :heart: :tada:

PR #1566, opened 2022-01-11
Implement docs configuration from scratch to make it simple

mathbunnyru:


docs/conf.py
```diff
@@ -20,72 +14,67 @@
...
-html_theme = "alabaster"
+html_theme = 'alabaster'
+
+# Add any paths that contain custom static files (such as style sheets) here,
+# relative to this directory. They are copied after the builtin static files,
+# so a file named "default.css" will overwrite the builtin "default.css".
+html_static_path = ['_static']
+
+# File above was generated using this command:
+# sphinx-quickstart --project "docker-stacks" --author "Project Jupyter" -v "latest" -r "latest" -l en --no-sep --no-makefile --no-batchfile
```
mathbunnyru, line 68: I generated the file above using this command line and only appended settings to it.
This will make it easy to update to newer sphinx versions and keeps this file clean.

.pre-commit-config.yaml
```diff
@@ -16,6 +16,7 @@ repos:
     hooks:
       - id: black
         args: [--target-version=py39]
+        exclude: docs/conf.py
```
mathbunnyru, line 19: This file is 80% generated, we don't want to run linter on it.

consideRatio: If this was a file that wasn't supposed to be manually modified but instead repeatedly re-generated, I'd agree with excluding it. In this case though, I suggest we run black on the generated output once and avoid an exception to this file just because it was generated initially.

I observed that the only differences were the use of double quotes compared to single quotes, so no manual interventions seems relevant to take along with running black.

docs/conf.py
consideRatio: I love when these kinds of notes are added, nice!

You left out though what version of sphinx was used, was it sphinx 4?

mathbunnyru: It was the latest one, 4.3.2.
I will add version here.

docs/conf.py
```diff
@@ -96,91 +84,3 @@
...
-        "navigation.html",
-        "relations.html",
-        "searchbox.html",
-    ]
-}
-
-# -- Options for HTMLHelp output ------------------------------------------
-
-# Output file base name for HTML help builder.
-htmlhelp_basename = "docker-stacksdoc"
```
consideRatio, line 125: Note how our .reathedocs.yaml describes "all" formats to be rendered, but we now configure only the `html` output format, and not latex/pdf/htmlhelp etc. I suggest that as part of this PR, you also update the RTD config file to have `formats: []` to opt out of the additional formats like PDF files to be built.

We've done so across most jupyterhub projects now. Sometimes we got build failures, not about the `html` build failing, but for a `pdf` build failing or similar which wasn't very important for us to debug and fix because few use them. I personally have found the PDF builds to be a distraction, because i could end up at such search result via google instead of a more updated easy to read result on an actual website.

https://github.com/jupyter/docker-stacks/blob/cfc62597b072ed7fbf92870d00704363e0653779/.readthedocs.yaml#L23-L24

mathbunnyru:
@consideRatio thanks for your thorough review.
I think this is ready to go now? :)

consideRatio (approved):

consideRatio:
Wiee nice work keeping this nice and clean @mathbunnyru!

PR #1568, opened 2022-01-12
Do not fix sphinx version

mathbunnyru:
I think this was coming from times when myst-parser installed old sphinx version.


mathbunnyru:
Seems to work for docs, merging :)

consideRatio:
I figure it helped us ensure sphinx 1 was upgraded in a certain RTD env, because sphinx 2+ is still required and unless we had it, RTD builds could fail.

Issue #1551, opened 2021-12-18
Add a multi-architecture stack based on datascience-notebook without Julia


massongit:
**What docker images this feature is applicable to?**

`jupyter/datascience-notebook`

**What changes do you propose?**

Add a multi-architecture stack based on `jupyter/datascience-notebook` without Julia.

**How does this change will affect users?**

Users who do not use Julia will be able to use `jupyter/datascience-notebook` on M1 Mac and latest Docker Desktop for Mac.

massongit:
https://github.com/jupyter/docker-stacks/issues/1549#issuecomment-1019451833
>I think it works now with Docker 4.4.2 🎉

So I close it.

mathbunnyru:
Yes, amd64 image should work fine.

PR #1569, opened 2022-01-12
Add datascience-python-r-notebook

massongit:
fix https://github.com/jupyter/docker-stacks/issues/1551

I add `datascience-python-r-notebook`, which is a multi-architecture stack based on `jupyter/datascience-notebook` without Julia.
If it add, users who do not use Julia will be able to use `jupyter/datascience-notebook` on M1 Mac and latest Docker Desktop for Mac.

### DIFF
--- a/.github/workflows/docker-amd64.yml
+++ b/.github/workflows/docker-amd64.yml
@@ -10,6 +10,7 @@ on:
 
       - "all-spark-notebook/**"
       - "base-notebook/**"
+      - "datascience-python-r-notebook/**"
       - "datascience-notebook/**"
       - "minimal-notebook/**"
       - "pyspark-notebook/**"
@@ -32,6 +33,7 @@ on:
 
       - "all-spark-notebook/**"
       - "base-notebook/**"
+      - "datascience-python-r-notebook/**"
       - "datascience-notebook/**"
       - "minimal-notebook/**"
       - "pyspark-notebook/**"
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,6 +10,7 @@ on:
 
       - "all-spark-notebook/**"
       - "base-notebook/**"
+      - "datascience-python-r-notebook/**"
       - "datascience-notebook/**"
       - "minimal-notebook/**"
       - "pyspark-notebook/**"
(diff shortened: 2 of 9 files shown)


mathbunnyru (changes requested):
Also https://github.com/jupyter/docker-stacks/blob/master/docs/using/selecting.md#image-relationships has to be upgraded.

To be honest, I don't think adding new image is what we should do.
I do understand, that you want the image to work under arm, but the better approach would be to fix our build system to use native arm runners.
It's difficult, but it's worth it.

consideRatio (changes requested):
I'm strongly opposed to adding another image to this repo. If we should do that, I think we should first have clear guidelines on when and why we do decide to add another images. My opposition stems from a care of this projects long term maintainability, not that the addition has no value for an end user.

I've seen several open source projects run into trouble for accepting more and more variations of something when there are infinite amounts of variations possible. I think they all need to say stop after a while as it becomes unsustainable to keep adding new parts.

@romainx @mathbunnyru what do you think?

massongit:
Thank you for your reviews!
I agree that this approach is not good for maintainability.
I would like to cooperate as much as possible to find to install Julia under ARM.

benz0li:
@massongit You may be interested in https://gitlab.b-data.ch/jupyterlab/r/docker-stack, which combines https://github.com/rocker-org/rocker-versioned with https://github.com/jupyter/docker-stacks (± minimal-notebook) including [code-server](https://github.com/coder/code-server) and further customisations.

See also https://demo.jupyter.b-data.ch for a minimal deployment (CPU: 2 cores, RAM: 4 GB) based on https://gitlab.com/b-data/docker/deployments/jupyter serving either [registry.gitlab.b-data.ch/jupyterlab/r/verse:latest-root](https://gitlab.b-data.ch/jupyterlab/r/verse/container_registry) or [registry.gitlab.b-data.ch/jupyterlab/julia/ver:latest-root](https://gitlab.b-data.ch/jupyterlab/julia/ver/container_registry).  
ℹ️ Both [JupyterHub](https://gitlab.b-data.ch/jupyterhub/jupyterhub/container_registry) and JupyterLab images are multi-arch (`linux/amd64`, `linux/arm64/v8`), the latter Debian-based.

Both JupyterLab images have _stock_ Python (currently v3.9.2) installed and focus on Julia and R, respectively. Thus, apart from `jupyter-server-proxy jupyterhub jupyterlab jupyterlab-git notebook nbconvert [radian]` and its dependencies, no further Python packages are installed.

PR #1570, opened 2022-01-12
Spark version incompatibility with Delta

agustingenoud:
Calling the "latest" tag ( "FROM jupyter/pyspark-notebook:latest" ) installs Spark 3.2 version, which is incompatible with delta lake ( https://docs.delta.io/latest/releases.html#-compatibility-with-as ). Installing the last version of Spark 3.1.x resolves the issue. In the jupyter docker repo that would be version 3.1.18

### DIFF
--- a/docs/using/recipes.md
+++ b/docs/using/recipes.md
@@ -530,7 +530,7 @@ By adding the properties to `spark-defaults.conf`, the user no longer needs to e
 ```dockerfile
 FROM jupyter/pyspark-notebook:lab-3.1.18
 
-ARG DELTA_CORE_VERSION="1.0.0"
+ARG DELTA_CORE_VERSION="1.1.0"
 RUN pip install --quiet --no-cache-dir delta-spark==${DELTA_CORE_VERSION} && \
      fix-permissions "${HOME}" && \
      fix-permissions "${CONDA_DIR}"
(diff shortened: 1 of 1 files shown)


docs/using/recipes.md
```diff
@@ -528,7 +528,7 @@ Please note that the [Delta Lake](https://delta.io/) packages are only available
 By adding the properties to `spark-defaults.conf`, the user no longer needs to enable Delta support in each notebook.
 
 ```dockerfile
-FROM jupyter/pyspark-notebook:latest
+FROM jupyter/pyspark-notebook:lab-3.1.18
 
 ARG DELTA_CORE_VERSION="1.0.0"
```
mathbunnyru, line 533: It seems that there is a release newer release of delta lake, which supports newer spark.
https://github.com/delta-io/delta/releases/tag/v1.1.0

So, let's bump version here

mathbunnyru (changes requested):

docs/using/recipes.md
agustingenoud: In that case if the decision is to go with the latest version of both the DELTA_CORE_VERSION should be changed to 1.1.0. 

```suggestion
ARG DELTA_CORE_VERSION="1.1.0"
```

docs/using/recipes.md
```diff
@@ -528,9 +528,9 @@ Please note that the [Delta Lake](https://delta.io/) packages are only available
 By adding the properties to `spark-defaults.conf`, the user no longer needs to enable Delta support in each notebook.
 
 ```dockerfile
-FROM jupyter/pyspark-notebook:latest
+FROM jupyter/pyspark-notebook:lab-3.1.18
```
mathbunnyru, line 531:
```suggestion
FROM jupyter/pyspark-notebook:latest
```

PR #1573, opened 2022-01-18
Add some typing to tests

mathbunnyru:
Main change is to add some typing, but there are some other changes as well.
Sorry they are all packed in one PR, but there are no significant changes here.

### DIFF
--- a/all-spark-notebook/test/test_spark_notebooks.py
+++ b/all-spark-notebook/test/test_spark_notebooks.py
@@ -4,10 +4,12 @@
 import logging
 
 import pytest
-import os
+from pathlib import Path
+
+from conftest import TrackedContainer
 
 LOGGER = logging.getLogger(__name__)
-THIS_DIR = os.path.dirname(os.path.realpath(__file__))
+THIS_DIR = Path(__file__).absolute()
 
 
 @pytest.mark.parametrize(
@@ -15,9 +17,9 @@ THIS_DIR = os.path.dirname(os.path.realpath(__file__))
     # TODO: add local_sparklyr
     ["local_pyspark", "local_spylon", "local_sparkR", "issue_1168"],
 )
-def test_nbconvert(container, test_file):
+def test_nbconvert(container: TrackedContainer, test_file: str) -> None:
     """Check if Spark notebooks can be executed"""
-    host_data_dir = os.path.join(THIS_DIR, "data")
+    host_data_dir = THIS_DIR / "data"
     cont_data_dir = "/home/jovyan/data"
     output_dir = "/tmp"
     timeout_ms = 600
@@ -29,7 +31,7 @@ def test_nbconvert(container, test_file):
         + f"--execute {cont_data_dir}/{test_file}.ipynb"
     )
     c = container.run(
-        volumes={host_data_dir: {"bind": cont_data_dir, "mode": "ro"}},
+        volumes={str(host_data_dir): {"bind": cont_data_dir, "mode": "ro"}},
         tty=True,
         command=["start.sh", "bash", "-c", command],
     )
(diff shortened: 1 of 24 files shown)


mathbunnyru:
Forgot to mention this - with typing, implementing new changes becomes easier (IDEs like VS Code do much much better with typing).
And, I didn't add typing in all possible places, but I think this is better than nothing.

conftest.py
```diff
@@ -74,9 +77,7 @@ def run(self, **kwargs):
         -------
         docker.Container
         """
-        all_kwargs = {}
-        all_kwargs.update(self.kwargs)
-        all_kwargs.update(kwargs)
+        all_kwargs = self.kwargs | kwargs
```
consideRatio, line 80: Oh I had never seen this before, today I learned something new!

minimal-notebook/test/test_nbconvert.py
```diff
@@ -19,17 +21,19 @@
...
+    host_data_dir = THIS_DIR / "data"
     cont_data_dir = "/home/jovyan/data"
     output_dir = "/tmp"
     LOGGER.info(
         f"Test that the example notebook {test_file} can be converted to {output_format} ..."
     )
     command = f"jupyter nbconvert {cont_data_dir}/{test_file}.ipynb --output-dir {output_dir} --to {output_format}"
     c = container.run(
-        volumes={host_data_dir: {"bind": cont_data_dir, "mode": "ro"}},
+        volumes={str(host_data_dir): {"bind": cont_data_dir, "mode": "ro"}},
```
consideRatio, line 36: The same situation I commented on above.

conftest.py
mathbunnyru: Yep, it's Python 3.9 feature.
https://docs.python.org/3/whatsnew/3.9.html#dictionary-merge-update-operators

I've never heard any complaints, that we use latest python version, so I think it's reasonable to continue to use new features sometimes :)

tagging/create_manifests.py
```diff
@@ -11,7 +11,7 @@
 from .manifests import ManifestHeader, ManifestInterface
 
 
-logger = logging.getLogger(__name__)
+LOGGER = logging.getLogger(__name__)
```
consideRatio, line 14: Not sure what I think in general about either small case or upper case. Can you put your finger on why you want to have it upper case rather than lower case?

It is a change off topic to the PR, so I'm leaning towards thinking it would been better to avoided complicating this otherwise very extensive PR.

consideRatio: I observe below that this makes things consistent within the project. Due to that I suggest we include this change and resolve this comment.

consideRatio (approved):
LGTM!! I haven't used these type annotations actively and have little experience overall, but it sounds very reasonable to go for!

minimal-notebook/test/test_nbconvert.py
consideRatio: Nevermind, let's go with this, I see in other situations you have used functions on the returned Path object even though you don't do it here.

all-spark-notebook/test/test_spark_notebooks.py
```diff
@@ -29,7 +31,7 @@ def test_nbconvert(container, test_file):
         + f"--execute {cont_data_dir}/{test_file}.ipynb"
     )
     c = container.run(
-        volumes={host_data_dir: {"bind": cont_data_dir, "mode": "ro"}},
+        volumes={str(host_data_dir): {"bind": cont_data_dir, "mode": "ro"}},
```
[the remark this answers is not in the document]
consideRatio: Resolved like https://github.com/jupyter/docker-stacks/pull/1573#discussion_r786942540

mathbunnyru:
amd64 works fine, so merging this.

PR #1574, opened 2022-01-19
Add version args to base-notebook dockerfile

asdfsx:
To install specific notebook/jupyterhub/jupyterlab into base-notebook image, add version args to dockerfile

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -126,16 +126,19 @@ RUN set -x && \
             fix-permissions "/home/${NB_USER}"; \
     fi;
 
+ARG NOTEBOOK_VERSION=default
+ARG JUPYTERHUB_VERSION=default
+ARG JUPYTERLAB_VERSION=default
+
 # Install Jupyter Notebook, Lab, and Hub
 # Generate a notebook server config
 # Cleanup temporary files
 # Correct permissions
 # Do all this in a single RUN command to avoid duplicating all of the
 # files across image layers when the permissions change
-RUN mamba install --quiet --yes \
-    'notebook' \
-    'jupyterhub' \
-    'jupyterlab' && \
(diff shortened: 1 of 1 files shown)


base-notebook/Dockerfile
```diff
@@ -126,16 +126,19 @@ RUN set -x && \
             fix-permissions "/home/${NB_USER}"; \
     fi;
 
+ARG NOTEBOOK_VERSION=default
+ARG JUPYTERHUB_VERSION=default
+ARG JUPYTERLAB_VERSION=default
```
mathbunnyru, line 131: If we change default values to empty strings, then the check is not needed and we can just write
```
mamba install --quiet --yes notebook="${NOTEBOOK_VERSION}"
```

And if someone specified version, they will get it and if not - they will get default version.

mathbunnyru (changes requested):

consideRatio:
This is a slippery slope, why not allow configuration of another software version as well if we allow these three to be configurable?

I suggest to reject thus PR, the workaround of building from an already built image and installing a specific version doesnt seem to cause much downside to me.

Another option would be to perhaps declare version restrictions in conda before it resolves things, then the user could provide any kind of restriction without us having to have logic hardcoded for individual packages etc.

mathbunnyru:
What @consideRatio tells makes a lot of sense to me.
Both his options seem reasonable and building from an existing image in my opinion is much easier than rebuilding the whole stack with some environment variables set.

asdfsx:
I'm trying to build a base-notebook image with specific version.   
And I found there is a variable to specific PYTHON version, so I add these 3 variables.

Maybe I should just using the existing notebook and reinstall jupyterlab and jupyterhub with other version.

mathbunnyru:
Yes, it would be a better choice.

PR #1575, opened 2022-01-19
Jupyter Notebook deprecation -> default to JupyterLab

### LINKED ISSUE #1217: Jupyter Notebook deprecation
[Jupyter Notebook][1] **is deprecated and it is advised to transition to JupyterLab**.

> Please note that this repository is currently maintained by a skeleton crew of maintainers from the Jupyter community. We encourage users to transition to JupyterLab, where more immediate support can occur.

The recent issue #1205 has shown that keeping both may lead to issues.

The newly preferred backend for JupyterLab 3.0 is [`jupyter-server`][3] instead of [`notebook`][1]: https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html#jupyter-server. 

## Roadmap

1. **Deprecation**: Maintain compatibility with both backends and deprecate Jupyter Notebook. Jupyter Notebook remains the default but a message indicates that it is advised to transition to JupyterLab.
2. **Switch the default to JupyterLab**: Switch to JupyterLab by default. It is still possible to start Jupyter Notebook by positioning a new `JUPYTER_ENABLE_NB` variable. 
3. **Drop Jupyter Notebook**: Move to [`jupyter_server`][3] alone and point people to [`nbclassic`][4] for a classic-like experience. It provides the Notebook UI as a new-server extension and also translates old-server extensions into new-server extensions.  (To be clear, this isn't just something that looks like the Notebook UI, it is literally the same code, running through the compatibility layer.)

Note: Another option could be [retrolab][5] - the look and feel of the Notebook UI but built with Lab components.

## Deprecation

- [x] Done -> PR #1209

Changes
- **With no option (no specific environment variable)**: Jupyter Notebook is launched (no change). However a warning / deprecation message is printed pointing to [the dedicated section][6] in the README.
- **With `JUPYTER_ENABLE_LAB`**: Jupyter Lab is launched (no change).
- Notice added in [README][6]

## Switch the default to JupyterLab

- [x] Done -> PR #1575

Changes
- **With no option (no specific environment variable)**: JupyterLab is launched.
- **With `JUPYTER_ENABLE_LAB`**: JupyterLab is launched. A warning message informs that this option is no more required.
- **With `DOCKER_STACKS_JUPYTER_CMD=notebook`**: Jupyter Notebook is launched. This variable also allows to run other `jupyter` commands
- Documentation is updated to document the new default and the new start option

## Drop Jupyter Notebook support

- [x] Not planned

Changes
- Jupyter Notebook support dropped (along with corresponding environment variables).
- A solution providing a _classic-like_ UI built atop of JupyterLab will be proposed. The solution is not chosen at this time.

[1]: https://github.com/jupyter/notebook
[2]: https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html#v3-0
[3]: https://github.com/jupyter-server/jupyter_server
[4]: https://github.com/jupyterlab/nbclassic
[5]: https://github.com/jtpio/jupyterlab-classic
[6]: https://github.com/jupyter/docker-stacks#jupyter-notebook-deprecation-notice

romainx:
Hello,

This PR intends to close #1217.

The idea is to 
- switch to JupyterLab by default i.e. running the `jupyter lab` command
- permit to run different commands in order to switch back for example to the classic Jupyter Notebook.

Instead of defining a closed option, I think it's more flexible to define a new option `JUPYTER_CMD` that can be used to launch different `jupyter` commands.

| Command     | Backend          | Frontend         |
|-------------|------------------|------------------|
| `lab` (def) | Jupyter Server   | JupyterLab       |
| `notebook`  | Jupyter Notebook | Jupyter Notebook |
| `nbclassic` | Jupyter Server   | Jupyter Notebook |
| `server`    | Jupyter Server   | None?            |
| `retro`*    | Jupyter Server   | RetroLab         |

_*Not installed at this time, but it could be the case in the future or it could be customized by end users._
Thanks 🙏  @krassowski, your [answer on SO](https://stackoverflow.com/a/67804732/4413446) helped me a lot to better understand the Jupyter landscape.

For example to switching back to classic notebook backend and frontend.

```bash
docker run -it --rm -p 8888:8888 -e JUPYTER_CMD=notebook jupyter/base-notebook
# Executing the command: jupyter notebook
```

At this time I've not checked if the value is in a list, if not it will be managed by the `jupyter` command.

```bash
docker run -it --rm -p 8888:8888 -e JUPYTER_CMD=foo jupyter/base-notebook
# Jupyter command `jupyter-foo` not found.
```

## Warning and deprecation

I've left a warning that is intended to tell that `JUPYTER_ENABLE_LAB` is no more used and has no effect.
I'm not sure we have to manage here Jupyter Notebook deprecation since we do not launch it anymore by default.
If the user decides to run it intentionally, it's not necessary to raise a warning it should be done by Jupyter Notebook.

## Todo

- [x] Decide if we want to use this solution -> 👍 
- [x] Decide if we want to perform additional control to check if `JUPYTER_CMD` holds a proper value -> 👎 
- [x] Decide if we want to keep a deprecation notice about the `JUPYTER_ENABLE_LAB` flag -> 👍 but it has to be changed
- [x] Update or remove the deprecation notice
- [x] Update the documentation
- [x] Improve test coverage 

@mathbunnyru and @consideRatio I would like to have your opinion. Once decided I will take care of the remaining tasks.

Many thanks 😄


manics:
It's there any sort of standard variable across the Jupyter ecosystem for defining the Jupyter app/command?

This has kind of come up in JupyterHub before, so it'd be nice if everyone in Jupyter could use a common variable!

romainx:
@manics Agree, however I don't know. I've only seen `JUPYTERHUB_SINGLEUSER_APP` from [here](https://jupyterhub.readthedocs.io/en/stable/reference/config-user-env.html#classic-notebook-ui) but I don't think it can be reused here.

benz0li:
What about `jupyterhub-sigleuser` in https://github.com/jupyter/docker-stacks/blob/d8056125ae1d1ff0b39c27b28437646e59081fe2/base-notebook/start-singleuser.sh#L13

Should this be changed to `jupyter-labhub`?

romainx:
@benz0li good point! I will have a look 👀.

mathbunnyru (commented):
I do like this solution - it's quite clear what's gonna be launched.

One thing to add about this solution - we should probably parametrize more tests based on what we launch, it will make our test suite work in more situations.

benz0li:
> Should this be changed to `jupyter-labhub`?

Does not seem to be necessary:

---

### JupyterHub < v2.0 (JupyterLab ≥ v3.0)

Opens Jupyter Notebook by default.  
ℹ️ Setting `c.Spawner.default_url = '/lab'` in `jupyterhub_config.py` will open JupyterLab by default [when using the `NotebookApp` backend].

--

> By default, the single-user notebook server uses the (old) `NotebookApp` from the [notebook](https://github.com/jupyter/notebook) package.

— [Configuring user environments — JupyterHub 1.4.2 documentation > Switching to Jupyter Server](https://jupyterhub.readthedocs.io/en/1.4.2/reference/config-user-env.html#switching-to-jupyter-server)

ℹ️ Setting `c.Spawner.environment = { 'JUPYTERHUB_SINGLEUSER_APP': 'jupyter_server.serverapp.ServerApp' }` in `jupyterhub_config.py` will use Jupyter Server’s `ServerApp` backend.  
👉 Setting `c.Spawner.default_url = '/lab'` or `c.Spawner.default_url = '/tree'` in `jupyterhub_config.py` is required.

---

### JupyterHub ≥ v2.0 (JupyterLab ≥ v3.0)

> By default the single-user server launches JupyterLab, which is based on [Jupyter Server](https://jupyter-server.readthedocs.io/).

— [Configuring user environments — JupyterHub 2.0.0 documentation > Switching back to classic notebook](https://jupyterhub.readthedocs.io/en/2.0.0/reference/config-user-env.html#switching-back-to-classic-notebook)

ℹ️ Setting `c.Spawner.default_url` in `jupyterhub_config.py` has no effect.

--

The `ServerApp` is the default backend when running JupyterHub ≥ 2.0.

ℹ️ Setting `c.Spawner.environment = { 'JUPYTERHUB_SINGLEUSER_APP': 'notebook.notebookapp.NotebookApp' }` in `jupyterhub_config.py` will use the (old) `NotebookApp` backend.  
👉 Opens Jupyter Notebook by default. Setting `c.Spawner.default_url = '/lab'` in `jupyterhub_config.py` will open JupyterLab by default.

romainx:
@benz0li thank you and I confirm about `jupyterhub-singleuser` vs. `jupyter-labhub`.

As per [the doc](https://jupyterhub.readthedocs.io/en/stable/reference/config-user-env.html#switching-back-to-classic-notebook) the app launched can be controlled through `JUPYTERHUB_SINGLEUSER_APP`.

- Default -> `SingleUserLabApp` is launched
- `JUPYTERHUB_SINGLEUSER_APP=notebook.notebookapp.NotebookApp` -> `SingleUserNotebookApp` is launched

See also [this thread](https://discourse.jupyter.org/t/jupyterhub-singleuser-and-jupyter-server/10942/10).

```bash
# default
docker run -it --rm -p 8888:8888 \
      jupyter/base-notebook start-singleuser.sh \
      | grep SingleUserLabApp
# [I 2022-01-19 16:43:54.025 SingleUserLabApp mixins:615] Starting jupyterhub single-user server version 2.0.2
# NotebookApp
docker run -it --rm -p 8888:8888 \
      -e JUPYTERHUB_SINGLEUSER_APP=notebook.notebookapp.NotebookApp \
      jupyter/base-notebook start-singleuser.sh \
      | grep SingleUserNotebookApp
# [I 2022-01-19 16:44:25.061 SingleUserNotebookApp mixins:615] Starting jupyterhub single-user server version 2.0.2
```

base-notebook/test/test_start_container.py
```diff
@@ -11,17 +11,22 @@
 
 
 @pytest.mark.parametrize(
-    "env,expected_server",
+    "env,expected_server,expected_warning",
```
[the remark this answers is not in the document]
romainx: @mathbunnyru thanks. Tried to improve tests, warnings are now parametrized.

mathbunnyru:
My two cents:

- Decide if we want to use this solution
👍 , it looks really good to me.

- Decide if we want to perform additional control to check if JUPYTER_CMD holds a proper value
👎
I think this is too much.
It will not look good in bash and if docs are gonna be updated, I think a very small amount of people will encounter the problem.
Also, the message `jupyterlub-singleuser: command not found` is quite easy to debug.

- Decide if we want to keep a deprecation notice about the JUPYTER_ENABLE_LAB flag
👍 
I think we should warn if this variable is set at all and it doesn't matter what it is set to.
And this variable will be always ignored.

Also, please, add `Improve tests coverage` as a part of your to-do list.

base-notebook/start-notebook.sh
```diff
@@ -15,10 +19,8 @@ if [[ "${RESTARTABLE}" == "yes" ]]; then
 fi
 
 if [[ -n "${JUPYTER_ENABLE_LAB}" ]]; then
```
mathbunnyru, line 21: This should be change from `-n` to `-v` and if this variable is set at all, then we should warn the user it won't make any difference.

romainx:
@mathbunnyru totally agree with you. I will continue this way.
Just to be sure on the point below. Do you mean that I should add it to the todo list just to check it or do you think there is room for improvement?
> Also, please, add Improve tests coverage as a part of your to-do list.
Thanks !

base-notebook/start-notebook.sh
romainx: Done 👍

base-notebook/test/test_start_container.py
```diff
@@ -6,60 +6,73 @@
...
+import time
 
 from conftest import TrackedContainer
 
 LOGGER = logging.getLogger(__name__)
 
 
 @pytest.mark.parametrize(
-    "env,expected_server,expected_warnings",
+    "env,expected_command,expected_start,expected_warnings",
```
mathbunnyru, line 17: Nice!

docs/using/common.md
```diff
@@ -69,7 +69,7 @@ You do so by passing arguments to the `docker run` command.
   (The `start-notebook.sh` script will `su ${NB_USER}` after adding `${NB_USER}` to sudoers.)
   **You should only enable `sudo` if you trust the user or if the container is running on an isolated host.**
 - `-e GEN_CERT=yes` - Instructs the startup script to generates a self-signed SSL certificate and configure Jupyter Notebook to use it to accept encrypted HTTPS connections.
-- `-e JUPYTER_ENABLE_LAB=yes` - Instructs the startup script to run `jupyter lab` instead of the default `jupyter notebook` command.
+- `-e JUPYTER_CMD=yes` - Instructs the startup script to run `jupyter ${JUPYTER_CMD}` instead of the default `jupyter lab` command.
```
mathbunnyru, line 72: `yes` should be changed to the list of available options or to some subsection of the docs.

I propose to move `Switching back to classic` from recipes to a subsection here because it's not a recipe, but a feature we support ourselves.

mathbunnyru (changes requested):
Please, grep `JUPYTER_ENABLE_LAB` - if there is something left, let's get rid of it.

I will also update the tag `33add21fab64` when this gets merged and built.

mathbunnyru:
And I do actually like the table which you added in your first message: https://github.com/jupyter/docker-stacks/pull/1575#issue-1107980909

I think it's worth to add it to the docs as well.

base-notebook/test/test_start_container.py
```diff
@@ -2,56 +2,77 @@
...
+        (["JUPYTER_CMD=nbclassic"], "jupyter nbclassic", True, []),
+        (
+            ["JUPYTERHUB_API_TOKEN=my_token"],
+            "jupyterhub-singleuser",
+            False,
+            ["WARNING: using start-singleuser.sh"],
+        ),
     ],
 )
 def test_start_notebook(
```
mathbunnyru, line 39: We should probably name this `test_start_jupyter`? :)

mathbunnyru: The name `start-notebook.sh` should also be changed, but let's not touch it in this PR, it will break so many users.

romainx: Ok @mathbunnyru so maybe I should let it with the same name as the file it tests `test_start_notebook`?

docs/using/common.md
romainx: It's done. I've chosen to put it in the _Alternative Commands_ section to avoid to have too much level 2 titles in the page. Let me know if you have a different opinion.

mathbunnyru (approved):
This looks awesome.

Let's wait for @consideRatio and then give it a go.

consideRatio (approved):
Beautiful! Thank you so much for your thorough work to make this change @romainx!

## Naming of environment variables

One thing that I think is a bit problematic is the naming of our environment variables in general. They aren't named to indicate they influence this project's scripts.

I'm considering if we could try to do something about that to help people find their way to this projects documentation to learn more about the environment variable that they otherwise may not be sure what it does.

Overall I'd love to see that this project would prefix its environment variables, but, we would need to deprecate many old variables if we wanted to do that anyhow.

### Example situation

Someone configures their installation of a JupyterHub Helm chart in a k8s cluster to start the servers differently, which works, because they use a jupyter/docker-stacks based image.

```yaml
singleuser:
  extraEnv:
    JUPYTER_CMD: notebook
```

After this, someone copies configuration to a situation when they are not using an image derived from a docker-stacks based image, and voila - they are confused what's going on etc.

### Naming of environment variables - conclusion

I'm okay with merging this as it is. I wanted though to raise the question on what you think about prefixing the new environment variable with `DOCKER_STACKS_` as an initial step towards making our environment variables be prefixed like that in general.

README.md
```diff
@@ -48,7 +48,7 @@ where `hostname` is the name of the computer running docker and `token` is the s
 Docker destroys the container after notebook server exit, but any files written to `~/work` in the container remain intact on the host.
 
 ```bash
-docker run --rm -p 10000:8888 -e JUPYTER_ENABLE_LAB=yes -v "${PWD}":/home/jovyan/work jupyter/datascience-notebook:33add21fab64
+docker run --rm -p 10000:8888 -e -v "${PWD}":/home/jovyan/work jupyter/datascience-notebook:33add21fab64
```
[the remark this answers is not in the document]
romainx: You're right I can change some wordings like
- Jupyter ~Notebook server~ Server
- in a browser loads ~Jupyter Notebook server~  JupyterLab

romainx:
@consideRatio thanks a lot 😄! and this is a very good remark.
If @mathbunnyru is OK with that I think it's better to change it now, it's not a problem for me.
If we stick with this prefix, it's better to do it before releasing a new variable that we will have to deprecate.
So let's go for? 

- `JUPYTER_CMD`-> `DOCKER_STACKS_JUPYTER_CMD` 🤔 not too long?

README.md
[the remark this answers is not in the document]
romainx: Done 👍

docs/index.rst
```diff
@@ -42,7 +42,7 @@ Visiting ``http://<hostname>:10000/lab?token=<token>`` in a browser loads Jupyte
 where ``hostname`` is the name of the computer running docker and ``token`` is the secret token printed in the console.
 Docker destroys the container after notebook server exit, but any files written to ``~/work`` in the container remain intact on the host.::
 
-    docker run --rm -p 10000:8888 -e JUPYTER_ENABLE_LAB=yes -v "${PWD}":/home/jovyan/work jupyter/datascience-notebook:33add21fab64
```
mathbunnyru, line 45: Unfortunately, these docs are duplicated, so you need to change the docs here as well 🙁

mathbunnyru (changes requested):
I'm ok with whatever env variable you choose :)

docs/index.rst
romainx: 😭 I will do it.

romainx: Done!

romainx:
@mathbunnyru and @consideRatio I think I'm done with this PR :real:. 
I've renamed the variable.
I let you merge it!
Thanks for your thoughtful reviews 🙏 

I will just update the issue to reflect what has been done and it will be 👌.

mathbunnyru:
Done. Welcome back, @romainx :)

PR #1578, opened 2022-01-21
Unify docker usage in tests

mathbunnyru:
Benefits of `container.run_and_wait`:
- it's typed and it's type is `str`. I don't like Docker approach, where one function can return different things depending on parameters (but in their case it makes sense).
- warnings and errors are checked, unless explicitly asked not to check them
- return code is always checked
- it's easy to debug broken tests - because not all the tests used `LOGGER.debug`, now they will
- less code :)

### DIFF
--- a/.github/workflows/docker-amd64.yml
+++ b/.github/workflows/docker-amd64.yml
@@ -76,3 +76,13 @@ jobs:
 
       - name: Test Docker Images
         run: make -C main test-all
+
+      - name: Clone Wiki
+        uses: actions/checkout@v2
+        with:
+          repository: ${{github.repository}}.wiki
+          path: wiki
+
+      - name: Run Post-Build Hooks
+        id: hook-all
+        run: make -C main hook-all
--- a/all-spark-notebook/test/test_spark_notebooks.py
+++ b/all-spark-notebook/test/test_spark_notebooks.py
@@ -37,10 +37,9 @@ def test_nbconvert(container: TrackedContainer, test_file: str) -> None:
         tty=True,
         command=["start.sh", "bash", "-c", command],
     )
-    # Spark warning
     warnings = TrackedContainer.get_warnings(logs)
-    assert len(warnings) == 1
-    assert "An illegal reflective access operation has occurred" in warnings[0]
+    # Some Spark warnings
+    assert len(warnings) == 5
 
     expected_file = f"{output_dir}/{test_file}.md"
     assert expected_file in logs, f"Expected file {expected_file} not generated"
--- a/all-spark-notebook/test/test_spark_notebooks.py
+++ b/all-spark-notebook/test/test_spark_notebooks.py
@@ -30,14 +30,11 @@ def test_nbconvert(container: TrackedContainer, test_file: str) -> None:
         + f"--output-dir {output_dir} "
         + f"--execute {cont_data_dir}/{test_file}.ipynb"
     )
-    c = container.run(
+    logs = container.run_and_wait(
+        timeout=timeout_ms / 10 + 10,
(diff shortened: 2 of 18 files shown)


mathbunnyru:
I also plan to unify `run_detached` usages - I think it's gonna be easy in most cases, but this will require another PR.

conftest.py
```diff
@@ -85,6 +89,32 @@ def run(self, **kwargs: typing.Any):
...
+        assert rv == 0 or rv["StatusCode"] == 0
+        return logs
+
+    @staticmethod
+    def get_warnings(logs: str) -> list[str]:
+        return [line for line in logs.split("\n") if line.startswith("WARNING")]
+
+    @staticmethod
+    def get_errors(logs: str) -> list[str]:
+        return [line for line in logs.split("\n") if line.startswith("ERROR")]
```
mathbunnyru, line 116: This is partially not the same as it was, but I think we should only worry about our errors and not the whole log output.

mathbunnyru:
Finally, this seems to work 😂

.github/workflows/docker-amd64.yml
```diff
@@ -76,3 +76,13 @@ jobs:
 
       - name: Test Docker Images
         run: make -C main test-all
+
+      - name: Clone Wiki
```
mathbunnyru, line 80: This is also useful to have in fast builds

all-spark-notebook/test/test_spark_notebooks.py
```diff
@@ -22,22 +22,24 @@ def test_nbconvert(container: TrackedContainer, test_file: str) -> None:
...
     )
-    c = container.run(
+    logs = container.run_and_wait(
+        timeout=60,
+        no_warnings=False,
         volumes={str(host_data_dir): {"bind": cont_data_dir, "mode": "ro"}},
         tty=True,
         command=["start.sh", "bash", "-c", command],
     )
-    rv = c.wait(timeout=timeout_ms / 10 + 10)
```
mathbunnyru, line 38: Dividing by 10 doesn't make sense

conftest.py
romainx: I would make a small change to avoid code duplication.

```python
@staticmethod
def get_errors(logs: str) -> list[str]:
    return _lines_starting_with(logs, "ERROR")

@staticmethod
def get_warnings(logs: str) -> list[str]:
    return _lines_starting_with(logs, "WARNINGS")

@staticmethod
def _lines_starting_with(logs: str, pattern: str) -> list[str]:
    return [line for line in logs.splitlines() if line.startswith(pattern)]
```

romainx (changes requested):
@mathbunnyru great 👍 
I had the same idea because I counted 25 copies of `c.logs(stdout=True).decode("utf-8")` 😓.
So it's a very good Initiative to factorize this code.
I had a different approach to wrap the returned container into our own object in order to provide then helpers methods directly on the object like `container.get_warnings()`. However the code is not done 😄 and this approach with static methods works and solves basically the same problem.
I've just made a small remark on the code.

conftest.py
mathbunnyru: Done as you suggested (but it should be `WARNING`, not `WARNINGS`).

mathbunnyru:
Seems to work for amd64, merging this :)

mathbunnyru:
@romainx if you would like to refactor this even more and create useful abstractions - please, feel free to do it :)

conftest.py
romainx: OK.

romainx:
@mathbunnyru no thank you it's really perfect as it is.

PR #1581, opened 2022-01-23
Fix arm pandoc test

mathbunnyru:


base-notebook/Dockerfile
```diff
@@ -28,13 +28,14 @@ ENV DEBIAN_FRONTEND noninteractive
...
     apt-get install --yes --no-install-recommends \
-    tini \
-    wget \
     ca-certificates \
-    sudo \
-    locales \
     fonts-liberation \
-    run-one && \
+    locales \
+    pandoc \
```
mathbunnyru, line 34: Sorted, and added this line.
Pandoc is not installed by default in arm images.

test/test_packages.py
```diff
@@ -62,7 +62,6 @@
 
 # List of packages that cannot be tested in a standard way
 EXCLUDED_PACKAGES = [
-    "tini",
```
mathbunnyru, line 65: This is system package, it won't be tested at all, so we don't need to exclude it.

mathbunnyru:
`jupyter/base-notebook` will have only one failing test, i hope it's gonna be fixed in the near future.
And, I will try to fix other arm issues as well.

base-notebook/Dockerfile
consideRatio: Can you make inline note about that?

This is knowledge that can be useful to retain like an inline comment i think

mathbunnyru: Done, please, take a look

consideRatio (approved):

PR #1583, opened 2022-01-25
Try ubuntu:devel as base image

mathbunnyru:
**Do not merge.**
I would like to check the issues we're gonna have with the next Ubuntu version.


mathbunnyru:
It's a bit early, but there might be a need for some big change, so it's better to know in advance :)

consideRatio:
So, not merging this, right? You could make it a draft to clarify that.

mathbunnyru:
Wow, this seems to work fine, that's great news :)

PR #1585, opened 2022-01-27
Prototype for multi-arch manifest

### LINKED ISSUE #1401: Tags and manifests are created based on amd64 images, but assumed for the arm64 images
In #1399 we add let this repo publish arm64 based images rather than the amd64 based images we currently have published.

This issue is meant to represent the problem of doing so without also documenting the version of packages installed etc, which may be different for an arm64 based image compared to an amd64 based image. Also, we only publish manifests based on the amd64 images rather than both, or clarify it is only the manifest of the amd64 based image.

romainx:
Hello,

This PR is a proposal to fix #1401.
It has two main points described below.

## 1. Make the manifest tool able to handle different os/architectures

I've made a first beta prototype to make it work on different architecture--there is still a lot of work to do. So the following call will generate a manifest from the image `docker pull --platform linux/arm64 jupyter/base-notebook:b020a0cf3b96`

```bash
python3 -m tagging.create_manifests --short-image-name base-notebook --platform linux/arm64 --owner jupyter --tag b020a0cf3b96 --wiki-path ../wiki

grep "arm" ../wiki/manifests/base-notebook-b020a0cf3b96.md | head
# * OS / Architecture: linux/arm64
# apt/now 2.0.6 arm64 [installed,local]
# base-files/now 11ubuntu5.4 arm64 [installed,local]
# base-passwd/now 3.5.47 arm64 [installed,local]
# bash/now 5.0-6ubuntu1.1 arm64 [installed,local]
# bsdutils/now 1:2.34-0.1ubuntu9.1 arm64 [installed,local]
```

TODO (at least):

- [ ] put architecture in filename (or find another way of doing it)
- [ ] manage the index

## 2. Create the manifest after the push to the registry

By creating the manifest **after the push** to the registry we would be able to access easily (without having to export them) the images of the various architectures.
I guess we will benefit from the `buildx` cache (or at least I'm pretty sure it's possible to do so) so the pull should not be a penalty.

@consideRatio and @mathbunnyru what do you think about this solution?
Once again is only a early prototype only modified to illustrate the proposal.
Thanks.

### DIFF
--- a/tagging/create_manifests.py
+++ b/tagging/create_manifests.py
@@ -54,6 +54,7 @@ def append_build_history_line(
 def create_manifest_file(
     short_image_name: str,
     owner: str,
+    platform: str,
     wiki_path: str,
     manifests: list[ManifestInterface],
     container: Container,
@@ -68,7 +69,7 @@ def create_manifest_file(
     )
 
     markdown_pieces = [
-        ManifestHeader.create_header(short_image_name, owner, BUILD_TIMESTAMP)
+        ManifestHeader.create_header(short_image_name, owner, platform, BUILD_TIMESTAMP)
     ] + [manifest.markdown_piece(container) for manifest in manifests]
     markdown_content = "\n\n".join(markdown_pieces) + "\n"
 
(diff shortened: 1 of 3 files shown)


mathbunnyru:
@romainx I will take a few days to think if this is a good direction to go :)
I haven't forgotten about this PR :)

mathbunnyru:
In general, I think we can try this, I'm not sure how the buildx cache work, but if it's easier to implement the feature this way and it doesn't slow down our already slow CI, then we should try this out.

romainx:
@mathbunnyru 👍  thanks for the feedback. 
I will go further on this and give it a try. It will not be easy to test so maybe I will have to perform a test outside of the project to validate our hypothesis regarding the impact on performances.
I will keep you posted here.

romainx:
Hello @mathbunnyru I will not go further on this according to what you intend to do in https://github.com/jupyter/docker-stacks/issues/1407#issuecomment-1029984033 and our discussion in #1604.
My fear is to loose time doing something that will not be in line with what you want to do. If worflows are split by platform in different runner doing this change is useless.
So, I'm also closing this one.
Best.

Issue #1333, opened 2021-05-25
Better documentation on how NB_UID, NB_GID, CHOWN_HOME work and how to use them properly


mathbunnyru:
We have quite a complicated setup using NB_UID, NB_GID, CHOWN_HOME so it's not obvious at all how to start using our docker images easily and in the correct way.

Some examples:
https://github.com/jupyter/docker-stacks/issues/1330

jeffyjefflabs:
#1229 is related too, I think

mathbunnyru:
Yes, you're right, thanks!

rigzba21:
Hello! I'm interested in contributing to this issue.

mathbunnyru:
@rigzba21 that's great!

I think, we have many issues with these variables.
First of all, you can try to solve the issue https://github.com/jupyter/docker-stacks/issues/1229.
It will definitely give you an understanding why we need these variables and how to use them.

After that, it would be great to search for issues (possibly closed), which also had troubles with these variables.

And then it would be great to understand how to improve our docs to make it easier for newcomers to use our docker stacks.

Sorry, that I don't know the exact steps to do, if someone has a better idea, feel free to post it here.

rigzba21:
Great! I'll start by taking a look at https://github.com/jupyter/docker-stacks/issues/1229. Thank you for the advice for a starting point.

jlphillipsphd:
#1465 is related

masip85:
Hello

I had problems with this long ago:

https://github.com/jupyter/docker-stacks/issues/885

From that issue next one is linked:

https://github.com/jupyter/docker-stacks/issues/1077

mathbunnyru:
Thanks for mentioning these issues @jlphillipsphd @masip85 !

@rigzba21 if you still have a desire to fix this problem, now you have many examples where it's not obvious how our docker images behave.

rigzba21:
Hi all, thank you for the examples, this is helpful!

rigzba21:
Linking https://github.com/jupyter/docker-stacks/issues/1497 here.

mathbunnyru:
We have updated our docs and now we describe UID, GID and how to use them and how to troubleshoot common problems.

Configuration: https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#user-related-configurations
Troubleshooting: https://jupyter-docker-stacks.readthedocs.io/en/latest/using/troubleshooting.html

Issue #1223, opened 2021-01-24
No permission to access /home/jovyan/work folder


doglasfc-docker:
when I use following command:
docker run --rm -p 10000:8888 -e JUPYTER_ENABLE_LAB=yes -v "$PWD":/home/jovyan/work jupyter/datascience-notebook

I don't have the permission to navigate  /home/jovyan/work folder to create any files in jupyter environment? 

(Since I set the the volume to $PWD, I want to create  a file in $PWD location and see it in /home/jovyan/work of Docker image instance.

Sorry, I'm the docker learner.

Bidek56:
Have you searched prior issues before posting this? Please [see](https://github.com/jupyter/docker-stacks/issues/542) which has references to few solutions.
What OS are you using? WSL2 is known to have permission issues.

romainx:
Hello @doglasfc-docker,

Thanks @Bidek56 for this suggestion. In addition, we have not yet taken the time to do it but we would like to create a troubleshooting page and one of the recurring issue / question is this one. This initiative is described in the issue #1071 (I have just added your suggestion @Bidek56) and there is a section called _Permission denied when mounting local folders_ listing several issues reporting the same problem. So please take the time to look at them to find the right solution at your problem.

I'm closing this issue since I'm pretty sure you will find the solution in those past issues. Feel free to post a new comment in this issue if it's not the case.

Thanks for your understanding.
Best.

Issue #1187, opened 2020-11-12
Volume: Permission denied


ogierpaul:
## What docker image you are using?
jupyter/minimal-notebook

## What complete docker command do you run to launch the container (omitting sensitive values)?
````
# Creating a volume with docker create -- volume
docker volume create --name stagingarea
# Start a jupyter minimal notebook container and map the volume (stagingarea) to a container directory (/data/stagingarea)
docker run -d \
  -v stagingarea:/home/jovyan/stagingarea\
  -p 8888:8888 \
  jupyter/minimal-notebook
````

## What steps do you take once the container is running to reproduce the issue?

1. Visit http://localhost:8888 & Login with token
2. Go to /home/jovyan/stagingarea
3. Attempt to create a simple notebook, or text file (Greetings.txt) in the volume

## What do you expect to happen?
- I want to: Create a simple text file and save it in the volume
- In order to: share this text file with other containers

## What actually happens?
- I have this error message: https://cloud.githubusercontent.com/assets/1175888/12765291/4ce76dfe-c9fe-11e5-86e0-b952c8a65777.png
- Similar to this issue: https://github.com/jupyter/docker-stacks/issues/114

## Why I want to do that
- I want to create a staging area so a python program  in jupyter can download data from one source, write it to a docker volume, and another service (like postgres,..) can read data from that volume and ingest it.

### Connecting as root
- It works when I run the jupyter docker image. as root -e GRANT_SUDO=yes \
````
# Creating a volume with docker create -- volume
docker volume create --name stagingarea
# Start a jupyter minimal notebook container and map the volume (stagingarea) to a container directory (/data/stagingarea)
docker run -d \
  -v stagingarea:/home/jovyan/stagingarea\
  -p 8888:8888 \
  -e JUPYTER_TOKEN=letmein \
  -e GRANT_SUDO=yes \
  --user root \
  jupyter/minimal-notebook
````

### Using a bind mount
````
docker run -d \
  -v -v /Users/paulogier/87-docker_parent/usingjupyterwithdocker/Step7_Volumes/stagingarea:/home/jovyan/stagingarea )
  -p 8888:8888 \
  -e JUPYTER_TOKEN=letmein \
  jupyter/minimal-notebook

````

romainx:
Hello @ogierpaul,

You can use one of the following option or combination of options to achieve what you want to do and change the ownership of your mount that is by default owned by `root`.
[...]

Issue #1085, opened 2020-05-18
Conda install problem: Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.


thudxz:
Hi! 

I am using the latest datascience-notebook image and want to install a bioconductor package using the following code:

`
RUN conda install -c bioconda bioconductor-geoquery
`

It returned the following message:

`
Collecting package metadata (current_repodata.json): ...working... done
Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.
`

I googled and found two similar issues [https://github.com/conda/conda/issues/9653](conda/issues/9653) and [https://github.com/conda/conda/issues/9367](conda/issues/9367). However, I can not solve the problem.

romainx:
Hello @thudxz,

We have turned `conda` channel priority to `strict`.

```bash
$ conda config --show | grep priority
# channel_priority: strict

# To see its meaning
$ conda config --describe channel_priority
```

So you can install the package by disabling it since the package you want to install comes from the `bioconda` channel.

```bash
# Install by disabling channel priority at command level
$ conda install --no-channel-priority -c bioconda bioconductor-geoquery

$ conda list | grep bio
# bioconductor-biobase      2.46.0            r36h516909a_0    bioconda
# bioconductor-biocgenerics 0.32.0                    r36_0    bioconda
# bioconductor-geoquery     2.54.0                    r36_0    bioconda
# bioconductor-limma        3.42.0            r36h516909a_0    bioconda
```

## Note

The last version `2.56` is not installed since it requires `r-base 4.0` which is not (yet) installed on `jupyter-datascience`.

```bash
$ conda search -c bioconda --info bioconductor-geoquery=2.56.0

# Loading channels: done
# bioconductor-geoquery 2.56.0 r40_0
# dependencies: 
#   - bioconductor-biobase >=2.48.0,<2.49.0
#   - bioconductor-limma >=3.44.0,<3.45.0
#   - r-base >=4.0,<4.1.0a0
# [...]
```

Please tell us if it's Ok for you?
Best

thudxz:
It worked. Thank you.

romainx:
Great to hear that.
Thank you.

## Note
I will add this case to the troubleshooting doc I intend to write (#1071)

PR #1589, opened 2022-01-31
📝Add troubleshooting page

trallard:
- Adding a troubleshoot page in the docs - using this also to track the cases identified in #1071 
- It should also help with #1333 

## Permission denied when mounting local folders

- [x] Unable to access host directory from within container : getting permission denied #542 
- [x] No permission to access /home/jovyan/work folder #1223
- [x] datascience-notebook needs to chown -R /home/jovyan to uid 1000 as root before starting #1003
- [x] Volume: Permission denied #1187

## Container launching issue

In recent issues (#1035, #1057) the same problem and the same root cause have been identified.
The VS Code plugin is running (silently) in the background and it prevents users to launch their Jupyter Docker Stack.

- [x] Tokens are being rejected #1035
- [x] Token note working #1057

## Package installation
Packages cannot be installed the cause can be the strict conda channel priority.

- [x] Conda install problem: Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve. #1085


mathbunnyru:
I think this is a good direction because we are frequently asked the same questions.

Several things are definitely worth adding in my opinion:
- why we do not add your favorite package :)
- issues we have with arm images and x86 also, related to our build process
- how to give root permissions to the user
- who `jovyan` is :)

I mean, these are the questions I saw quite frequently and it's not obvious how to answer them.
And, of course, it's completely up to you which questions you would like to answer in this PR.

trallard:
Great thanks @mathbunnyru  - I am wondering for the suggestions would it make sense to have a FAQ or keep everything in the same troubleshooting section? 🤔

mathbunnyru:
I think it's not that important, you can choose whatever option you would like to and it's easy to change in the future.

Just a small note: if you delete `test/foo` file, CI will finish much faster (it won't build docker images).

trallard:
Done - 😬 did not realize I had a stray file from when I was working on the examples

mathbunnyru:
> Done - 😬 did not realize I had a stray file from when I was working on the examples

Not done yet :)

trallard:
Right - I have tried to cover all the cases linked to in #1071 

it would be grant if @mathbunnyru and @romainx  could have a look at this 👀

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+       -p 8888:8888 \
+       --user root \
+       -e CHOWN_EXTRA="<container-dir>" \
+       -e CHOWN_EXTRA_OPTS="-R" \
+       jupyter/minimal-notebook
+   ```
+
+   where:
+
+   - `CHOWN_EXTRA=<some-dir>`: will change the ownership and group of the specified container directory (non-recursive by default). You need to provide full paths starting with `/`.
```
mathbunnyru, line 55:
```
CHOWN_EXTRA="<some dir>,<some other dir>"`
```
Let's mention here quickly, that we support several dirs.

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+   # jovyan is now the owner of /home/jovyan/stagingarea
+   $ ls -la ~/stagingarea/
+   -rw-r--r-- 1 jovyan users    0 Feb  1 14:41 kale.txt
+   ```
+
+   ```{admonition} Additional notes
+      - If you are mounting your volume inside the `/home/` directory, you can use the `-e CHOWN_HOME=yes` and `CHOWN_HOME_OPTS="-R"` flags instead of the `-e CHOWN_EXTRA`
+        and `-e CHOWN_EXTRA_OPTS` in the example above.
+      - This solution should work in most cases where you have created a docker volume (i.e. using the [`docker volume create --name <my-volume>`
+        command](https://docs.docker.com/storage/volumes/#create-and-manage-volumes)) and mounted it using the`-v` flag in `docker run`.
```
mathbunnyru, line 76: I don't like this style of line breaks in markdown source code, usually, I prefer to break the lines by sentences or sentence parts - usually, it makes the source code more readable.

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+
+2. **Matching the container's UID/GID with the host's**
+
+   Docker handles mounting host directories differently from mounting volumes, even though the syntax is essentially the same (i.e. `-v`).
+
+   When you initialize a Docker container using the flag `-v`, the host directories are bind-mounted directly into the container.
+   Therefore, the permissions and ownership are copied over and will be **the same** as the ones in your local host
+   (including user ids) which may result in permissions errors when trying to access directories or create/modify files inside.
+
+   Suppose your local user has a `UID` and `GID` of `1234`. To fix the UID discrepancies between your local directories and the container's
```
mathbunnyru, line 87: Let's use different values for uid and gid, because it might happen, I suppose.

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+   directories, you can run the container with an explicit `NB_UID` and `NB_GID` to match the that of the local user:
+
+   ```bash
+   $ docker run -it --rm \
+       --user root \
+       -p 8888:8888 \
+       -e NB_UID=1234 \
+       -e NB_GID=1234 \
+       -v $(PWD)/test:/home/jovyan/work \
```
mathbunnyru, line 96: We need to add quotes for PWD.

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+
+   # you should see an output similar to this
+   Update jovyan's UID:GID to 1234:1234
+   Running as jovyan: bash
+   ```
+
+   where:
+
+   - `NB_IUD` and `NB_GID` should match the local user's UID and GID.
+   - You must use `--user root` to ensure that the `UID` and `GID` are updated at runtime.
```
mathbunnyru, line 107: *must* in bold - I saw you did this in other places, and I like it

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+
+   ````{admonition} Additional notes
+   - The caveat with this approach is that since these changes are applied at runtime, you will need to re-run the same command
+     with the appropriate flags and environment variables if you need to recreate the container (i.e. after removing/destroying it).
+   - If you pass a numeric UID, it must be in the range of 0-2147483647
+   - This approach only updates the UID and GID of the **existing `jovyan` user** instead of creating a new user. From the above example:
+     ```bash
+     $ id
+     uid=1234(jovyan) gid=1234(jovyan) groups=1234(jovyan),100(users)
+   ````
```
mathbunnyru, line 117:
```suggestion
     ```
```

mathbunnyru (changes requested):

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,311 @@
...
+   ````{admonition} Additional notes
+   - The caveat with this approach is that since these changes are applied at runtime, you will need to re-run the same command
+     with the appropriate flags and environment variables if you need to recreate the container (i.e. after removing/destroying it).
+   - If you pass a numeric UID, it must be in the range of 0-2147483647
+   - This approach only updates the UID and GID of the **existing `jovyan` user** instead of creating a new user. From the above example:
+     ```bash
+     $ id
+     uid=1234(jovyan) gid=1234(jovyan) groups=1234(jovyan),100(users)
+     ```
+   ````
```
mathbunnyru, line 118:
```suggestion
   ```{admonition} Additional notes
   - The caveat with this approach is that since these changes are applied at runtime, you will need to re-run the same command
     with the appropriate flags and environment variables if you need to recreate the container (i.e. after removing/destroying it).
   - If you pass a numeric UID, it must be in the range of 0-2147483647
   - This approach only updates the UID and GID of the **existing `jovyan` user** instead of creating a new user. From the above example:
     ```bash
     $ id
     uid=1234(jovyan) gid=1234(jovyan) groups=1234(jovyan),100(users)
     ```
   ```
```

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+   - The caveat with this approach is that since these changes are applied at runtime, you will need to re-run the same command
+     with the appropriate flags and environment variables if you need to recreate the container (i.e. after removing/destroying it).
+   - If you pass a numeric UID, it must be in the range of 0-2147483647
+   - This approach only updates the UID and GID of the **existing `jovyan` user** instead of creating a new user. From the above example:
+     ```bash
+     $ id
+     uid=1234(jovyan) gid=1234(jovyan) groups=1234(jovyan),100(users)
+   ````
+
+## Permission issues after changing the UID/GIU and USER in the container
```
mathbunnyru, line 119: The syntax highlight is broken here :(

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+    docker run -it --rm \
+        -p 8888:8888 \
+        --user root \
+        -e NB_USER=callisto \
+        -e NB_UID=1234 \
+        -e NB_GID=1234 \
+        -e CHOWN_HOME=yes \
+        -e CHOWN_HOME_OPTS="-R" \
+        -w "/home/${NB_USER}" \
+        -v $(PWD)/test:/home/callisto/work \
```
mathbunnyru, line 143: Also, quotes are needed here

mathbunnyru (changes requested):

mathbunnyru:
@trallard please, try to fix the syntax, it's not easy to read the docs with broken markdown highlighting.

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+   - You must use `--user root` to ensure that the `UID` and `GID` are updated at runtime.
+
+   ````{admonition} Additional notes
+   - The caveat with this approach is that since these changes are applied at runtime, you will need to re-run the same command
+     with the appropriate flags and environment variables if you need to recreate the container (i.e. after removing/destroying it).
+   - If you pass a numeric UID, it must be in the range of 0-2147483647
+   - This approach only updates the UID and GID of the **existing `jovyan` user** instead of creating a new user. From the above example:
+     ```bash
+     $ id
+     uid=1234(jovyan) gid=1234(jovyan) groups=1234(jovyan),100(users)
```
mathbunnyru, line 116:
```suggestion
     id
     # uid=1234(jovyan) gid=1234(jovyan) groups=1234(jovyan),100(users)
```

I prefer this syntax.
1. You can copy this command and paste it and it will work (even though you will get some comments).
2. The output is not parsed as bash input and highlighted differently, so it looks better.

I saw some examples below with `$` as well, please, also fix them.

mathbunnyru (changes requested):

mathbunnyru:
@trallard if you don't mind, I can edit this page myself - I think your wording is good, but I will fix the snippets a bit just to make them match the existing style.

trallard:
Fixed the syntax @mathbunnyru - I also think I managed to fixed all the snippets style to match the rest of the docs. 

The only thing - I am not sure of is the line break comment/fix

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+   with the appropriate flags and environment variables if you need to recreate the container (i.e. after removing/destroying it).
+ - If you pass a numeric UID, it **must** be in the range of 0-2147483647
+ - This approach only updates the UID and GID of the **existing `jovyan` user** instead of creating a new user. From the above example:
+   ```bash
+   id
+   # uid=1234(jovyan) gid=5678(jovyan) groups=5678(jovyan),100(users)
+   ```
+````
+
+## Permission issues after changing the UID/GIU and USER in the container
```
mathbunnyru, line 120:
```suggestion
## Permission issues after changing the UID/GID and USER in the container
```

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+   Suppose your local user has a `UID` and `GID` of `1234` and `5678`, respectively. To fix the UID discrepancies between your local directories and the container's
+   directories, you can run the container with an explicit `NB_UID` and `NB_GID` to match the that of the local user:
+
+   ```bash
+   docker run -it --rm \
+       --user root \
+       -p 8888:8888 \
+       -e NB_UID=1234 \
+       -e NB_GID=5678 \
+       -v "$(PWD)"/test:/home/jovyan/work \
```
mathbunnyru, line 96: This way of getting PWD won't work.
It tries to **execute** `PWD` command, which doesn't exist.
The correct way of doing this is to either use pwd command or the environment variable.

We've always used the env variable like this:

```suggestion
       -v "${PWD}"/test:/home/jovyan/work \
```

mathbunnyru (changes requested):

docs/using/troubleshooting.md
mathbunnyru: Other places have this problem as well

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+  ```
+
+  This command will launch the container with a specific user UID and add that user to the `users` group to modify the files in the default `/home` and `/opt/conda` directories.
+  Further avoiding issues when trying to `conda install` additional packages.
+
+- Use `docker inspect <container_id>` and look for the [`Mounts` section](https://docs.docker.com/storage/volumes/#start-a-container-with-a-volume) to verify that the volume was created and mounted accordingly:
+
+  ```json
+  # for example, for a my-vol volume created with
+  # docker volume create --name <my-vol>
```
mathbunnyru, line 225: I don't think comments in json are an official feature of json and this is better to be avoided.

mathbunnyru (changes requested):

docs/using/troubleshooting.md
trallard: you are correct there - fixed

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+- when clicking on the URL displayed on your command line logs, you face a "This site cannot be reached" page on your web browser
+- using the produced token and/or URL results in an "Invalid credentials" error on the Jupyter "Token authentication is enabled" page
+
+  ```bash
+  # example log output from the docker run command
+
+  [...]
+  Or copy and paste one of these URLs:
+    http://3d4cf3809e3f:8888/?token=996426e890f8dc22fa6835a44442b6026cba02ee61fee6a2
+    or http://127.0.0.1:8888/?token=996426e890f8dc22fa6835a44442b6026cba02ee61fee6a2
```
mathbunnyru, line 281: This should be commented as well

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,310 @@
...
+- Pass absolute paths to the `-v` flag:
+
+  ```bash
+  -v "${PWD}"/<my-vol>:/home/jovyan/work
+  ```
+
+  This example uses the syntax `$(PWD)`, which is replaced with the full path to the current directory at runtime. The destination
+  path should also be an absolute path starting with a `/` such as `home/jovyan/work`.
+
+- You might want to consider using the Docker native `--user <UID>` and `--group-add users` flags instead of `-e NB_GID` and `-e NB_UID`:
```
mathbunnyru, line 206:
```suggestion
- You might want to consider using the Docker native `--user <UID>` and `--group-add users` flags instead of `-e NB_UID` and `-e NB_GID`:
```

mathbunnyru (changes requested):

mathbunnyru:
After merging this, let's update our bug report Issue template to explicitly ask users if they have read this page.
https://github.com/jupyter/docker-stacks/blob/master/.github/ISSUE_TEMPLATE/bug_report.md

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,307 @@
...
+
+- `root` is the owner of `/home` or a mounted volume
+- when starting the container, you get an error such as `Failed to change ownership of the home directory.`
+- getting permission denied when trying to `conda install` packages
+
+**Some things to try:**
+
+1. **Ensure the new user has ownership of `/home` and volume mounts**
+
+   For example, say you want to create a user `callisto` with a `GID` and `UID` of `1234`, you will have to add the following flags to the docker run command:
```
mathbunnyru, line 132:
```suggestion
   For example, say you want to create a user `callisto` with a `UID` and `GID` of `1234`, you will have to add the following flags to the docker run command:
```

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,307 @@
...
+   - `-w "/home/${NB_USER}"` sets the working directory to be the new user's home
+
+   ```{admonition} Additional notes
+    In the example above, the `-v` flag is used to mount the local volume onto the new user's `/home` directory.
+
+    However, if you are mounting a volume elsewhere, you also need to use the `-e CHOWN_EXTRA=<some-dir>` flag to avoid any permission
+    issues (see the section [Permission denied when mounting volumes](../using/troubleshooting.html#permission-denied-when-mounting-volumes) in this page).
+   ```
+
+2. **Dynamically assign the user ID and GID**
```
mathbunnyru, line 172:
```suggestion
2. **Dynamically assign the user id (UID) and group id (GID)**
```

mathbunnyru (approved):
Great work @trallard.

Sorry for being too thorough sometimes :)

trallard:
💜   love thorough reviews - I often make silly mistakes so it's great having someone else looking at stuff for me

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,307 @@
+# Troubleshooting Common Problems
+
+When troubleshooting, you may see unexpected behaviors or receive an error message. This section provides advice on
+how to identify and fix some of the most commonly encountered issues.
```
mathbunnyru, line 4:
```suggestion
When troubleshooting, you may see unexpected behaviors or receive an error message.
This section provides advice on how to identify and fix some of the most commonly encountered issues.
```

mathbunnyru: Here is an example of how I like to put line breaks.

mathbunnyru: So, some lines are longer, some are shorter, but in general they are still not too long, but most importantly - I don't cut the sentence in some random part, which in my opinion makes it easier to read.

mathbunnyru, line 4: Please, do not use force pushes.
It makes the review process much worse.
Also, it seems, that you lost the commit which rearranges these lines.

trallard: so sorry 😢  rebased instead of merging master - promise not to do it again

docs/using/troubleshooting.md
```diff
@@ -0,0 +1,307 @@
...
+
+Most of the `docker run` flags used in this document are explained in detail in the [Common Features, Docker Options section](../using/common.html#Docker-Options) of the documentation.
+
+## Permission denied when mounting volumes
+
+If you are running a Docker container while mounting a local volume or host directory using the `-v` flag like so:
+
+```bash
+docker run -it --rm \
+    -p 8888:8888 \
```
mathbunnyru, line 15:
```suggestion
    -p 8888:8888 \
    -v <my-vol>:<container-dir> \
```

You lost this :)

romainx (approved):
Hello @trallard thanks a lot for that. I'm sure it will help a lot our users and also save us some time answering duplicated issues.
You did a very good job, congrats ❤️ 
You have already met our best :detective: @mathbunnyru for the review, so I have nothing to add. @mathbunnyru thank you 🙏 for thorough review.
Let's merge it 🚀 and we will improve it over time.

PR #1592, opened 2022-02-02
Install more r packages for arm

mathbunnyru:
Summary by Erik: In the past we have omitted installing three R packages in our arm64 image specifically. Now though, it seems like arm64 support has improved and we only need to omit one R package.

### DIFF
--- a/r-notebook/Dockerfile
+++ b/r-notebook/Dockerfile
@@ -47,7 +47,7 @@ RUN mamba install --quiet --yes \
     fix-permissions "${CONDA_DIR}" && \
     fix-permissions "/home/${NB_USER}"
 
-# These packages are not easy to install under arm
+# `r-tidymodels` is not easy to install under arm
 # hadolint ignore=SC2039
 RUN set -x && \
     arch=$(uname -m) && \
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
If it works, this looks great to me - how did you realize this @mathbunnyru? Did you observe some changelog somewhere? Got a link if so?

mathbunnyru:
Nope, I had in memory that `r-rmarkdown` had some issues with pandoc.
And I've recently added pandoc to the image.
So I decided to check everything once again and was able to install these two packages.

consideRatio:
> So I decided to check everything once again

How did you go about this? Did you start a build and then when all steps until the step where r packages are getting installed - you started a container based on the completed step just before and then manually tried to installing each package individually?

mathbunnyru:
`Failed to fetch http://ports.ubuntu.com/ubuntu-ports/pool/universe/t/texlive-base/texlive-latex-recommended_2019.20200218-1_all.deb  Unable to connect to ports.ubuntu.com:http: [IP: 91.189.88.142 80]`

Some external issues :(

consideRatio:
> Some external issues :(

Let's restart it - right?

mathbunnyru:
I'm not in a hurry, I would just wait some time, because I might have seen it today already.

mathbunnyru:
Green, merging this.

PR #1593, opened 2022-02-02
Remove mamba pin for aarch64

### LINKED ISSUE #1539: Build is stuck for jupyter/scipy-notebook arm64 image
There is a timeout and it happened several times already:
https://github.com/jupyter/docker-stacks/runs/4292319380?check_suite_focus=true

I think the reason might be in QEMU :(

maresb:
Closes #1539

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -121,18 +121,6 @@ RUN set -x && \
     fix-permissions "${CONDA_DIR}" && \
     fix-permissions "/home/${NB_USER}"
 
-# Using fixed version of mamba in arm, because the latest one has problems with arm under qemu
-# See: https://github.com/jupyter/docker-stacks/issues/1539
-RUN set -x && \
-    arch=$(uname -m) && \
-    if [ "${arch}" == "aarch64" ]; then \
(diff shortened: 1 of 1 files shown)


mathbunnyru:
Let's indeed try this out, thank you!

consideRatio:
Let's hope for the best! Thanks @maresb !!!

maresb:
I'm trying to follow along in the workflow logs, but I'm confused. There are two workflows:
* Build, test, and publish Docker Images
* Build and test amd64 Docker Images

This seems redundant. Is the point that QEMU is slow so that you can save time by testing without, but you want the publishing workflow to be uniform across all platforms?

mathbunnyru:
Yes, slow build times under QEMU is the reason.
Most of the PRs are enough to be tested under x86_64 only, that's why I implemented a separate workflow.
If a feature is similar for arm and x86 or for x86 only, we can skip long tests and merge PRs faster.

maresb:
Green!!! :tada:

consideRatio:
Wieeeee!!!! :heart: :tada: :rocket:

PR #1595, opened 2022-02-02
Envvar comments

maresb:
Supersedes #1519

Does this work?:
Closes #1519

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -182,22 +182,18 @@ if [ "$(id -u)" == 0 ] ; then
         #
         # - We use the `--preserve-env` flag to pass through most environment
         #   variables, but understand that exceptions are caused by the sudoers
-        #   configuration: `env_delete`, `env_check`, and `secure_path`.
+        #   configuration: `env_delete` and `env_check`.
         #
-        # - We use the `--set-home` flag to set the HOME variable appropriatly.
+        # - We use the `--set-home` flag to set the HOME variable appropriately.
         #
-        # - We reduce the `env_delete` list of default variables to be deleted. It
-        #   has higher priority than the `--preserve-env` flag and `env_keep`
-        #   configuration.
+        # - To reduce the default list of variables deleted by sudo, we could have
+        #   used `env_delete` from /etc/sudoers. It has higher priority than the
+        #   `--preserve-env` flag and the `env_keep` configuration.
         #
-        # - We disable the `secure_path` which is set by default in /etc/sudoers as
-        #   it would override the PATH variable.
-        # Note on the purpose of "PATH=${PATH}":
-        #   In case "${cmd[@]}" is "bash", then PATH will be used by this bash shell.
-        #   However, PATH is irrelevant to how the above sudo command resolves the
-        #   path of "${cmd[@]}". Sudo's path resolution is done via the "secure_path"
-        #   variable set above in /etc/sudoers.d/path.
-
+        # - We preserve PATH and PYTHONPATH explicitly. Note however that PATH is
+        #   irrelevant to how the above sudo command resolves the path of `${cmd[@]}`.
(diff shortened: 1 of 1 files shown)


base-notebook/start.sh
```diff
@@ -168,12 +168,32 @@ if [ "$(id -u)" == 0 ] ; then
...
         PYTHONPATH="${PYTHONPATH:-}" \
         "${cmd[@]}"
-        # Note on the purpose of "PATH=${PATH}":
-        #   In case "${cmd[@]}" is "bash", then PATH will be used by this bash shell.
-        #   However, PATH is irrelevant to how the above sudo command resolves the
-        #   path of "${cmd[@]}". Sudo's path resolution is done via the "secure_path"
-        #   variable set above in /etc/sudoers.d/path.
-
+        # Notes on how we ensure that the environment that this container is started
+        # with is preserved (except vars listen in JUPYTER_ENV_VARS_TO_UNSET) when
```
mathbunnyru, line 172:
```suggestion
        # with is preserved (except vars listed in JUPYTER_ENV_VARS_TO_UNSET) when
```

consideRatio: Is `JUPYTER_ENV_VARS_TO_UNSET` a feature we have active?

mathbunnyru: It is implemented, documented, and even tested.
I have no idea though if anyone uses it.

mathbunnyru (approved):

base-notebook/start.sh
mathbunnyru: It is implemented, documented, and even tested.
I have no idea though if anyone uses it

base-notebook/start.sh
```diff
@@ -168,12 +168,32 @@ if [ "$(id -u)" == 0 ] ; then
...
+        # - We use the `--set-home` flag to set the HOME variable appropriately.
+        #
+        # - To reduce the default list of variables deleted by sudo, we could have
+        #   used `env_delete` from /etc/sudoers. It has higher priority than the
+        #   `--preserve-env` flag and the `env_keep` configuration.
+        #
+        # - We preserve PATH and PYTHONPATH explicitly. Note however that PATH is
+        #   irrelevant to how the above sudo command resolves the path of `${cmd[@]}`.
+        #   Sudo uses the "secure_path" variable we modified above in
+        #   /etc/sudoers.d/path. (PATH is useful e.g. when `${cmd[@]}` is `bash`.)
```
consideRatio, line 196: I don't understand the last parenthesis very well, could we rephrase/remove it?

maresb: Hmm... could you help me to rephrase it?

What I'm trying to say is that if you run `base-notebook start.sh some-command`, then `PATH` is likely completely irrelevant. However, if you run `base-notebook start.sh bash`, then you get a Bash shell with the expected `PATH` set.

consideRatio: Hmm... because you assume PATH is irrelevant for a command you start? Maybe... I consider that a quite specific assumption. If you start a jupyter server which is very common, wouldn't the PATH be retained and relevant for the bash terminal sessions run within there for example?

maresb: Yes, good point. Perhaps it's better to say something like:

> The `PATH` will mainly be relevant for any terminals spawned by the `${cmd[@]}`.

mathbunnyru: I would say something like:

> The PATH will be relevant for any processess spawned by ${cmd[@]} (including new terminals).

mathbunnyru (approved):
Nice, I think this is quite clear.

mathbunnyru:
Merging this one - it's only a comment and I think is quite clear now.

consideRatio:
Thank you @maresb and @mathbunnyru!!!

PR #1594, opened 2022-02-02
Use micromamba to bootstrap mamba

maresb:
If Micromamba is as mature as I think, then there is no longer any need to use Miniforge.

In case there are significant bugs in Micromamba, then maybe we find them here.

:eyes:

PR #1598, opened 2022-02-03
Cleanup of #1594

maresb:
Temporary...

To be merged into `maresb:micromamba` after we see whether or not CI passes on #1594


maresb:
Merged back into #1594. This is now obsolete.

Issue #1597, opened 2022-02-03
[DOC] - Broken hyphens in documentation


trallard:
Not related to the images but the documentation - hyphens seem to not be rendering properly. I thought it was as I was working on the docs but the published version also seems broken (see image below):

![https://jupyter-docker-stacks.readthedocs.io/en/latest/ - Home Page ](https://user-images.githubusercontent.com/23552331/152376237-2730364f-a4bb-444e-b706-036fd6ef0ff4.png)

**Checked on:**
- Brave browser
- Chrome 
- Running on MacOS

PR #1600, opened 2022-02-04
Use sphinx_rtd theme and separate docs environment

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1597

Alabaster hasn't received any updates in 3 years.
This theme is also not very active, but it was last updated 5 months ago.

I think it looks better and has more modern style.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,13 +132,16 @@ dev/%: PORT?=8888
 dev/%: ## run a foreground container for a stack
 	docker run -it --rm -p $(PORT):8888 $(DARGS) $(OWNER)/$(notdir $@)
 
-dev-env: ## install libraries required to build docs and run tests
+install-dev-env: ## install libraries required to build images and run tests
 	@pip install -r requirements-dev.txt
 
 
 
 docs: ## build HTML documentation
 	sphinx-build docs/ docs/_build/
+
+install-docs-env: ## install libraries required to build docs
+	@pip install -r requirements-docs.txt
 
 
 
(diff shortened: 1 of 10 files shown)


requirements-dev.txt
```diff
@@ -6,5 +6,6 @@ pre-commit
 pytest
 requests
 sphinx
+sphinx_rtd_theme>=1.0.0
```
mathbunnyru, line 9: In the rtd environment, sphinx_rtd_theme is preinstalled to some old version, which causes troubles.

mathbunnyru:
CI should be a bit faster, also, it's easier to debug pip packages, when only the needed ones are installed.

mathbunnyru:
@trallard please, take a look, I think you will like this :)

trallard:
I was actually going to open an issue to replace the theme for JupyterBook since plenty projects in the Jupyter ecosystem are migrating to this

Makefile
```diff
@@ -132,14 +132,17 @@ dev/%: PORT?=8888
 dev/%: ## run a foreground container for a stack
 	docker run -it --rm -p $(PORT):8888 $(DARGS) $(OWNER)/$(notdir $@)
 
-dev-env: ## install libraries required to build docs and run tests
+install-dev-env: ## install libraries required to build images and run tests
```
trallard, line 135: Nice one! I do prefer having the docs and dev requirements separately

trallard (approved):

consideRatio:
@choldgraf i think you overview the options well, do you have suggestions on a theme to use for a jupyter project in the switch from alabaster?

Issue #1588, opened 2022-01-30
The `help` command doesn't work in notebooks with the R interpreter


davidvandebunte:
**What docker image you are using?**

`jupyter/r-notebook` (and `jupyter/datascience-notebook`)

**What complete docker command do you run to launch the container (omitting sensitive values)?**

```bash
docker run --rm -it --entrypoint /bin/bash jupyter/datascience-notebook
```

**What steps do you take once the container is running to reproduce the issue?**

1. `R` (to start the R interpreter)
2. Any `help` command

In detail:
```
(base) jovyan@dd0b1bd0a300:~$ R

R version 4.1.2 (2021-11-01) -- "Bird Hippie"
...
Type 'q()' to quit R.

> help(demo)
/opt/conda/lib/R/bin/pager: 11: exec: less: not found
Warning message:
error in running command
```

**What do you expect to happen?**

The `help` for the command to display in my terminal.

**What actually happens?**

The command crashes, see above.

My fix for the last year or so has been to install the `less` system package in my own images. Could it be installed upstream?

mathbunnyru:
Yes, I think it's perfectly fine to add `less` to our image to make help work in `R`.

I would suggest adding it to `minimal-notebook` with a comment.
Having a test for this would also be nice, probably.

Could you draft a PR, please?

PR #1601, opened 2022-02-04
Only install system packages which are useful for many people

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1588

### DIFF
--- a/minimal-notebook/Dockerfile
+++ b/minimal-notebook/Dockerfile
@@ -17,7 +17,7 @@ RUN apt-get update --yes && \
     tzdata \
     unzip \
     vim-tiny \
-    # Inkscape shall be installed to be able to convert SVG files
+    # Inkscape is installed to be able to convert SVG files
     inkscape \
     # git-over-ssh
     openssh-client \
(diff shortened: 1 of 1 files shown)


consideRatio:
I appreciate having comments for everything we have included, so that there is clarity for long term maintenance about the purpose/value of all packages.

I cant say if these will be reasonabe to remove or not in practice. I'm okay doing it but would appreciate additional reviewers!

mathbunnyru:
I suppose, if some package is really needed for some crucial feature, we should:
- docs, why this package is needed, and what feature won't work (inline docs like these help a lot)
- test, which tests the feature (not just the package)
- if we delete the package, the users will come with - "hey, this feature doesn't work after the update" and we will add the package back with something from above
- super-useful package, like `wget` or `vim` is also ok

So, in general, I'm not afraid of changes like that :)

consideRatio (approved):
Thanks for your thorough work!! ♥️🎉

mathbunnyru:
Let's merge this, I don't think this will cause us many problems.

PR #1605, opened 2022-02-05
Error and fail start.sh on misconfiguration

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1517


mathbunnyru:
I want to know your opinion and also need to fix the tests.

consideRatio (commented):
Hmmm you make the log function exit on error. I'm not so fond of making its purpose be expanded to that while still being named just `_log`.

I think I'd favor being explicit and including multiple `exit` codes. One could also consider having separate non-zero exit codes for the errors when exiting to allow users recognize an error via the exit code as well.

mathbunnyru:
Thanks for your suggestion, I will fix this.
In general, do you agree with exiting on errors like these?

consideRatio:
Hmmm... I think so, but I'm not that confident about it.

Do we have default values for the environment variables etc that can cause trouble, forcing someone to explicitly configure something to avoid errors? That would be a signal to me that we have been too strict.

mathbunnyru:
I think the defaults are good, but I think there are many cases, where user sets one variable and forgets about another one (like chaning username requires `--user root`).
In cases like this the behavior is not clear.
The user gets a warning, but I beleive no one reads warnings until "it somehow works".

mathbunnyru:
Closing this one, I don't think I'm gonna have enough time in the nearest future to finish this PR.

PR #1607, opened 2022-02-06
Present community stacks as a table

mathbunnyru:
It's the same amount of information, but without duplications like `is a community Jupyter Docker Stack image` or `Try it on`.
I think this is visually much more clear and easy to choose from.

Also refactored GPU flavours, because their descriptions are quite long and it makes sense to have them in a separate table.

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -5,3 +5,4 @@ default: true
 MD013:
   # Number of characters
   line_length: 200
+  tables: false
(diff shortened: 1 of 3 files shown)


mathbunnyru:
Before: https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#community-stacks
After: https://jupyter-docker-stacks--1607.org.readthedocs.build/en/1607/using/selecting.html#community-stacks

consideRatio (approved):
This looks so much better, great job!

PR #1608, opened 2022-02-06
Fail make docs on docs warnings

mathbunnyru:
I think, this is the default behaviour on RTD, I would like it to be the default behaviour when I run `make docs`.
Makes local build process more similar to what happens in the CI.


mathbunnyru:
Will merge this, quite a simple change.

PR #1617, opened 2022-02-10
Fix the recipe "Add a Python 3.x environment"

### LINKED ISSUE #1433: default virtual env in console
I am using `jupyter/base-notebook` with jupyterhub.

I change the python3 version with the recipe : https://jupyter-docker-stacks.readthedocs.io/en/latest/using/recipes.html#add-a-python-3-x-environment

I uncomment the line 
`ENV CONDA_DEFAULT_ENV ${conda_env}`

All works as expected (the notebooks are executed in ${conda_env} ) but the console is still in the env named `base`.
I can activate it manually with `conda activate myEnv` but I was hoping to start directly in the right virtualenv (myEnv).

Do you know a solution for that ?

Regards.

romainx:
Hello,

This fixes #1433.
See [my comment](https://github.com/jupyter/docker-stacks/issues/1433#issuecomment-1034499440) in the issue for explanations.

Best.

### DIFF
--- a/docs/using/recipes.md
+++ b/docs/using/recipes.md
@@ -101,7 +101,7 @@ Ref: <https://github.com/jupyter/docker-stacks/issues/440>
 ## Add a Python 3.x environment
 
 The default version of Python that ships with conda/ubuntu may not be the version you want.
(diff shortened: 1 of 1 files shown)


mathbunnyru:
Thanks. If you could check that `Python 2` version works, it would also be great, but in can be done in another PR.
And I'm not sure we should keep `Python 2` version at all :)

PR #1618, opened 2022-02-10
Unique recipe to install a custom conda env

romainx:
Hello,

@mathbunnyru I've checked as you suggested and this is a follow-on PR for #1617.

After a check, I dot not see a reason to keep two recipes for the same thing. The recipe kept permits to create a custom conda env and to register it as a Jupyter kernel whatever the chosen python version (I've tested it with Python 2.7 and it seems to work correctly). So I've kept only one recipe (the initial Python 3) and made some wording changes and some cleaning.

Let me know if it's OK for you.
Best.

### DIFF
--- a/docs/using/recipes.md
+++ b/docs/using/recipes.md
@@ -71,43 +71,16 @@ RUN mamba install --yes --file /tmp/requirements.txt && \
 
 Ref: [docker-stacks/commit/79169618d571506304934a7b29039085e77db78c](https://github.com/jupyter/docker-stacks/commit/79169618d571506304934a7b29039085e77db78c#r15960081)
 
-## Add a Python 2.x environment
+## Add a custom conda environment and Jupyter kernel
 
-Python 2.x was removed from all images on August 10th, 2017, starting in tag `cc9feab481f7`.
-You can add a Python 2.x environment by defining your Dockerfile inheriting from one of the images like so:
-
-```dockerfile
-# Choose your desired base image
-FROM jupyter/scipy-notebook:latest
-
-# Create a Python 2.x environment using conda, including the ipython kernel
-# and the kernda utility. Add any additional packages you want available for use
-# in a Python 2 notebook to the first line here (e.g., pandas, matplotlib, etc.)
-RUN mamba create --quiet --yes -p "${CONDA_DIR}/envs/python2" python=2.7 ipython ipykernel kernda && \
-    mamba clean --all -f -y
-
-USER root
-
-# Create a global kernelspec in the image and modify it so that it properly activates
-# the python2 conda environment.
-RUN "${CONDA_DIR}/envs/python2/bin/python" -m ipykernel install && \
-    "${CONDA_DIR}/envs/python2/bin/kernda" -o -y /usr/local/share/jupyter/kernels/python2/kernel.json
-
-USER ${NB_UID}
-```
-
-Ref: <https://github.com/jupyter/docker-stacks/issues/440>
-
-## Add a Python 3.x environment
-
-The default version of Python that ships with conda/ubuntu may not be the version you want.
-The instructions to add a conda environment with a different version and make it accessible to Jupyter are very similar to Python 2.x. Still, they are slightly simpler (no need to switch to `root`):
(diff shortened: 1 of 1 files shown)


docs/using/recipes.md
```diff
@@ -123,17 +96,13 @@ RUN mamba create --quiet --yes -p "${CONDA_DIR}/envs/${conda_env}" python=${py_v
...
+# create Python kernel and link it to jupyter
 RUN "${CONDA_DIR}/envs/${conda_env}/bin/python" -m ipykernel install --user --name="${conda_env}" && \
     fix-permissions "${CONDA_DIR}" && \
     fix-permissions "/home/${NB_USER}"
 
 # any additional pip installs can be added by uncommenting the following line
-# RUN "${CONDA_DIR}/envs/${conda_env}/bin/pip" install
-
-# prepend conda environment to path
-ENV PATH "${CONDA_DIR}/envs/${conda_env}/bin:${PATH}"
```
romainx, line 136: Don't figure why it is necessary to prepend the env `PATH`. `PATH` manipulation is handled by the `conda activate`.

mathbunnyru: I might be wrong here, but what happens, if someone creates a new py3 environment with his favourite python version and then runs `python3` command .

I think, if we prepend to PATH, then the favourite python3 will be run.
If we append to `.bashrc`, does it work for the next `RUN` step?
Also, does it work, when we change the `$NB_USER` and actually run the docker using `--user root`?


That's why I'm quite sure prepending to PATH does what expected and I'm not confident about .bashrc.
Could you check this out, please?

mathbunnyru (changes requested):

docs/using/recipes.md
romainx: Hello,

Here my answers.

> I might be wrong here, but what happens, if someone creates a new py3 environment with his favourite python version and then runs `python3` command .

It will run the `python3` executable corresponding to the activated conda environment.

> If we append to .bashrc, does it work for the next RUN step?

No but there is no additional `RUN` step and the previous uses a full path call `"${CONDA_DIR}/envs/${conda_env}/bin/pip"`. So I do not think it's a good reason to keep it.

> Also, does it work, when we change the `$NB_USER` and actually run the docker using `--user root`?
> [...] Could you check this out, please?

Sure, it works as expected with this command.

```
docker run -it --rm \     
    -p 8888:8888 \
    --user root \
    -e NB_USER="romain" \
    -e CHOWN_HOME=yes \
    -w "/home/${NB_USER}" jupyter/custom-notebook
```

IMO you have two option tuning the `PATH` yourself or let `conda` activation script do it for you and also take care to perform other activation actions (source additional variables for example). I do not think it's a good idea to mix these two modes.

> Activating environments is essential to making the software in the environments work well. Activation entails two primary functions: adding entries to PATH for the environment and running any activation scripts that the environment may contain .

You can have a look to [the `conda active` doc](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#activating-an-environment) to make you own opinion.

Best

romainx: Why are you requiring a change before asking the question?

mathbunnyru: Sorry, I should have pressed comment.
Probably should go to sleep :)
Sorry once again ❤️

romainx: Not a problem but I'm investing time here to help and especially on this follow-on PR you asked me to do. I try always to be nice but it's a bit harsh to receive this kind of review.

mathbunnyru (approved):

docs/using/recipes.md
romainx: @mathbunnyru it's Ok and my English is not my greatest strength too (I hope 😄). We have worked together here for a long time now and your investment is impressive. So I hope we will continue to work in a good mood like we do in the past. I think we have sometimes different approaches, a complementary knowledges, but not the same investment now. However, we should try to keep this good collaboration. On my side I will try to do it even if I invest less time here than you do. Working here for free should continue to be pleasant and even fun while doing things right like we always try to do. So I'm going to bed too 💤, have a nice sleep 😄.

Issue #1622, opened 2022-02-12
Add an ability to run tests in parallel


mathbunnyru:
Right now, it's impossible to run tests in parallel.
I know only one reason why it doesn't work (I've quickly checked this running `make test/base-notebook` in two tabs) - we mount container port to host `8888` port (it is hardcoded).
Instead we should choose random port, then it should work fine.

Then we will be able to easily use [pytest-xdist](https://pypi.org/project/pytest-xdist/) and make our tests run in parallel.

PR #1624, opened 2022-02-14
Run tests in parallel

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1622

`time TEST_IMAGE="jupyter/base-notebook" pytest test/ base-notebook/test/ --log-level debug`:
Executed in  145.72 secs

Same with `-n auto`:
Executed in   44.42 secs

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,7 +1,9 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
+from contextlib import closing
 import os
 import logging
+import socket
 from typing import Any, Optional
 
 import docker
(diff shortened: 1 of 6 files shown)


conftest.py
```diff
@@ -108,14 +118,6 @@ def run_and_wait(
         assert rv == 0 or rv["StatusCode"] == 0
         return logs
 
-    def get_host_port(self, container_port: str) -> str:
-        """Returns the host port associated with the tracked container's port."""
-        assert isinstance(self.container, Container)
-        self.container.reload()
```
mathbunnyru, line 114: This doesn't actually work well when running many tests in parallel :=(
I don't really want to dig in the reason, but I think I've found an implementation, which works more reliable.

romainx (approved):
Hello @mathbunnyru,

Very good initiative 👍. I think we can merge it.
Thanks.

PR #1629, opened 2022-02-15
Add inline comments about pytest-xdist

consideRatio:
I ended up needing to google around a bit to learn about the `-n` flag that turned out to come from pytest-xdist, so I figured it was helpful to document it and provide the flag in its long form. This is a followup to https://github.com/jupyter/docker-stacks/pull/1624.


mathbunnyru:
Low-latency review 😂

PR #1631, opened 2022-02-16
[WIP] - Proof of concept - optimising CI build 

trallard:
Related to #1407 

**Notes**:
- I am moving away from `Make` to using [doit](https://pydoit.org/contents.html) some of the advantages of this are:
  - Python-based so we get all the syntactic sugar
  - it allows for much more flexibility when it comes to set tasks, dependencies, and manipulating artefacts/build outputs
  - others projects in the Jupyter ecosystem have started adopting it for automation and to complement their CI workflows and it works pretty well

## Done:
 -  [x] started to adopt the workflow shown in #1407 
   - [x] right-hand side of the workflow is now implemented - see image attached (the side in green is implemented except that the push is not working as we need a PAT - see below)
 -  [x] added cache for the Docker build and for Python dependencies 
 - [x] created  [doit](https://pydoit.org/contents.html) tasks
 - [x] parallel GH jobs for: testing and manifest/tag creation
 - [x] Docker images and manifests are saved as GH artifacts
 - [x] Add types to the `dodo.py` file

## Questions
-   What would you suggest doing after? 
   - Completing the main workflow in #1407  -> pushing all the tags to DockerHub and the manifests to the Wiki? (The advantage is that if that is done now at least we get an optimised workflow from what we currently have for x86 images)
   - Working on the left-hand side of the workflow?
- How long do we want to keep the artefacts for? right now I set a 1-day policy while I was doing some tests but not sure how long would be a reasonable time here 

## TODO:
- Get a PAT (personal access token) so that we can push to GHCR

---

docker-stacks CI workflow schematic

---
Updated by Ayaz:
Moved `Add types to the `dodo.py` file` to `Done` section.


mathbunnyru:
@trallard I merged the PR where tests are run for all dependent images.
So running tests for a specific image now only takes one line of code.
Please, merge master to this branch and fix the code so it uses new way to run the tests.

trallard:
@mathbunnyru could you please have a look at this PR? - I have updated the [top description ](https://github.com/jupyter/docker-stacks/pull/1631#issue-1139915656) to reflect what I have done and what still needs doing

Also, this action run should give you a good idea of how this whole thing is working right now <https://github.com/jupyter/docker-stacks/actions/runs/1899972804>

trallard:
Sure no rush @mathbunnyru 

I use https://www.lucidchart.com/pages/ lucidchart for most of my schematics

.github/workflows/build-test-docker.yml
```diff
@@ -0,0 +1,231 @@
...
+      - "base-notebook/**"
+      - "datascience-notebook/**"
+      - "minimal-notebook/**"
+      - "pyspark-notebook/**"
+      - "r-notebook/**"
+      - "scipy-notebook/**"
+      - "tensorflow-notebook/**"
+
+      - "tagging/**"
+      - "test/**"
```
mathbunnyru, line 21:
```suggestion
      - "tests/**"
```

.github/workflows/build-test-docker.yml
```diff
@@ -0,0 +1,231 @@
...
```
mathbunnyru, line 41:
```suggestion
      - "tests/**"
```

.github/workflows/build-test-docker.yml
```diff
@@ -0,0 +1,231 @@
...
+      - "datascience-notebook/**"
+      - "minimal-notebook/**"
+      - "pyspark-notebook/**"
+      - "r-notebook/**"
+      - "scipy-notebook/**"
+      - "tensorflow-notebook/**"
+
+      - "tagging/**"
+      - "tests/**"
+      - "Makefile"
```
mathbunnyru, line 22:
```suggestion
      - "Makefile"
      - "dodo.py"
```

dodo.py
```diff
@@ -0,0 +1,400 @@
...
+from pathlib import Path
+from subprocess import PIPE
+
+import doit
+from doit import task_params
+from doit.tools import CmdAction
+from tests.images_hierarchy import get_test_dirs
+
+# global doit config
+DOIT_CONFIG = {"verbosity": 2, "default_tasks": ["build_docs"]}
```
mathbunnyru, line 15: I'm not sure we should have `build_docs` as a default task.

mathbunnyru, line 15: I'm not sure we should have build_docs as a default task.

dodo.py
```diff
@@ -0,0 +1,400 @@
...
+
+
+def task_build_docs():
+    """Build Sphinx documentation 📝
+    setting uptodate to False will force the task to run every time
+    """
+
+    return dict(
+        file_dep=[*P.DOCS_MD, *P.DOCS_RST, *P.DOCS_PY],
+        actions=[U.do("sphinx-build", "-W", P.DOCS, "docs/_build/html")],
```
mathbunnyru, line 35: I think you forgot to include some recently added options to the docs build.

mathbunnyru, line 35: We didn't put our files to html subfolder, do you really need to do it?

dodo.py
```diff
@@ -0,0 +1,400 @@
...
+    )
+
+
+# https://pydoit.org/task-creation.html#delayed-task-creation
+@doit.create_after(executed="build_docs")
+def task_docs_check_links():
+    """Checks for any broken links in the Sphinx documentation 🔗
+    only created after the docs are built"""
+    return dict(
+        file_dep=[*P.DOCS.rglob("_build/*.html")],
```
mathbunnyru, line 47: As you put the html files in the html subfolder, this rglob will probably give empty output.

dodo.py
```diff
@@ -0,0 +1,400 @@
...
+            ],
+        )
+
+
+# -----------------------------------------------------------------------------
+# Support classes and methods
+# -----------------------------------------------------------------------------
+
+
+class P:
```
mathbunnyru, line 246: Maybe let's rename this to Paths?

dodo.py
```diff
@@ -0,0 +1,400 @@
...
+# -----------------------------------------------------------------------------
+
+
+def task_build_docs():
+    """Build Sphinx documentation 📝
+    setting uptodate to False will force the task to run every time
+    """
+
+    return dict(
+        file_dep=[*P.DOCS_MD, *P.DOCS_RST, *P.DOCS_PY],
```
mathbunnyru, line 34: Maybe it would be easier to put the whole `docs` directory as a dependency.

requirements-dev.txt
```diff
@@ -8,3 +8,4 @@ pytest
 pytest-xdist
 requests
 tabulate
+doit
```
mathbunnyru, line 11: Let's keep this list sorted :)

dodo.py
```diff
@@ -0,0 +1,400 @@
...
+
+    # git specific - used for tagging
+    SOURCE_DATE_EPOCH = (
+        subprocess.check_output(["git", "log", "-1", "--format=%ct"])
+        .decode("utf-8")
+        .strip()
+    )
+
+    GIT_COMMIT_SHA = (
+        subprocess.check_output(["git", "rev-parse", "--short", "HEAD"])
```
mathbunnyru, line 326: Maybe we could use plumbum in places like this to make our life easier?
An example: https://github.com/jupyter/docker-stacks/blob/master/tagging/git_helper.py

mathbunnyru:
May I also ask you to remove parts of Makefile / workflows that will be redundant after switching to doit?

dodo.py
```diff
@@ -0,0 +1,400 @@
...
+    # ]
+    ALL_IMAGES = [
+        "base-notebook",
+        "minimal-notebook",
+    ]
+
+    AMD64_IMAGES = ["datascience-notebook", "tensorflow-notebook"]
+
+
+class U:
```
mathbunnyru, line 308: Ans this should be probably named Utils?

trallard:
Thanks for the review - ideally all of the Makefile will be redundant after moving over to doit. 

Of course this implies tasks like building docs and creating Dec environments will also be done through doit tasks. 

I was also planning to add a section in the docs on how to execute the tasks locally (i.e for development)

dodo.py
trallard: I will have to look at how these two work together - when I tried to call the tests directly as in the Makefile I kept getting errors since plumbum is used under the hood

mathbunnyru (commented):
I think I have finished reviewing this PR.
Added some things that bother me - I saw some logic that is now duplicated in dodo.py, and I don't think this duplication is good and it would be great to get rid of it.

I do support this work and I think it's great 👍 
Thank you @trallard.

.github/workflows/build-test-docker.yml
```diff
@@ -0,0 +1,232 @@
...
+
+jobs:
+  build-images:
+    name: Build Docker Images
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    outputs:
+      short_sha: ${{ steps.slug.outputs.short_sha }}
+      hash_tag: ${{ steps.slug.outputs.hash_tag }}
```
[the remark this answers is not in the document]
trallard: Nope only need one - forgot to delete the redundant one (will fix this)

.github/workflows/build-test-docker.yml
```diff
@@ -0,0 +1,232 @@
...
+          python -m doit -f main/dodo.py docker_create_manifest
+
+      - name: Save manifests as artifact 💾
+        uses: actions/upload-artifact@v2
+        with:
+          name: docker-images-manifest-${{ needs.build-images.outputs.short_sha }}
+          path: |
+            wiki/manifests/
+          # TODO: decide on a retention policy - also might make the above a smart regex
+          retention-days: 1
```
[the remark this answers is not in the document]
trallard: Right now we do not reuse them - but we'll do later as we move towards the whole new workflow

dodo.py
```diff
@@ -0,0 +1,400 @@
...
+            uptodate=[False],
+            actions=[
+                (U.set_env, [f"{P.OWNER}/{image}"]),
+                (
+                    U.do(
+                        *U.PYTEST_ARGS,
+                        *test_dirs,
+                    )
+                ),
+            ],
```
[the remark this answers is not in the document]
trallard: That was my plan at the beginning - though since [`run_tests.py` uses plumbum ](https://github.com/jupyter/docker-stacks/blob/2e7b2562e0dd5778fbfce67d757b707086118638/tests/run_tests.py?rgh-link-date=2022-03-04T14%3A01%3A38Z#L20) these two (doit task and run_test.py) end up being incompatible. 

To avoid duplication the easier way would be to run tests directly through doit tasks and get rid of `run_tests.py` altogether

.github/workflows/build-test-docker.yml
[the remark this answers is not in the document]
mathbunnyru: Nice 👍

dodo.py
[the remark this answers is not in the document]
mathbunnyru: It would be nice to know the reason for this conflict.
I would like to continue to use plumbum :)

mathbunnyru:
> What would you suggest doing after?
> Completing the main workflow in https://github.com/jupyter/docker-stacks/issues/1407 -> pushing all the tags to DockerHub and the manifests to the Wiki? (The advantage is that if that is done now at least we get an optimised workflow from what we currently have for x86 images)
> Working on the left-hand side of the workflow?

I think you're free to choose whatever you would like to do first :)

> How long do we want to keep the artefacts for? right now I set a 1-day policy while I was doing some tests but not sure how long would be a reasonable time here

These artifacts are not intended (at least now) for external usage.
They are only needed to build and push image (and then we can actually get rid of these artifacts).
On the other way, if something goes wrong, having them available for a couple of days is good for debugging.
Maybe let's keep them for a couple of days?

mathbunnyru:
@trallard for the past couple of days I've made some adjustments to make sure pre-commits do not give errors they shouldn't actually give for mypy.
At the same time, now mypy will correctly trigger for the code which isn't typed and hooks will fail (they didn't before in GitHub Actions), so if you merge master to this branch, you might see some errors.
And I added some more hooks (but I don't think they will trigger for this PR).

Actually, I wanted to ask you - how would you feel if I made a couple of commits to your branch fixing the issues?
I do not want in any way to steal your work, don't get me wrong, please ❤️ 
I might not be able to resolve all of them, because I don't have much experience with dodo / passing info from one job to another one, but I will be able to resolve some simple issues and actually I will gain some experience with dodo, which is nice.
I won't do this without your approval, so, please, tell me what you think 😅

trallard:
Thanks for the heads up @mathbunnyru I have been pretty busy these last weeks so I was planning on finishing this PR early next week. 
In the meantime if you want to push some stuff in the meantime I am happy for you to do so ☺️  teamwork is better than solo work. 

Then I can help wrap this up on Monday

dodo.py
mathbunnyru: Removed html suffix

.github/workflows/build-test-docker.yml
```diff
@@ -0,0 +1,232 @@
...
+        with:
+          path: main
+
+      - name: Cache dependencies 🧠
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-dev.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
```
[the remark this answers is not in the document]
mathbunnyru: Removed all pip caches for now.

dodo.py
mathbunnyru: I see why it's not possible, because file_dep wants files :(

dodo.py
mathbunnyru: Now, I've put the files in the right dir, so this should be fine.

dodo.py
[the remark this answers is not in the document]
mathbunnyru: I think I made it work with run_tests. And now there is no duplication.

Also, I would like to have a way to test only one image locally.
Right now I use run_tests.py for that.

dodo.py
mathbunnyru: I haven't found a way to show the target list by default :(

dodo.py
mathbunnyru: I directly used tagging.git_helper and it works well.

dodo.py
mathbunnyru: I tried to call doit (from doit) with `list` command, but it doesn't work, because the database is locked.
If you don't know how to solve this, just press "Resolve", because this is not a big deal.

mathbunnyru:
I fixed some things here and there.
Hope you will like them.
I will pause for now :)

One more thing to add - currently our doit config doesn't support building/testing only one image.
Please, if there is a way to fix this, it would be great.
Most of the time I'm only interested in building one image and I just do `make build/some-notebook` or test it using similar command.
It would be nice to be able to switch from `make` to `doit`.

dodo.py
```diff
@@ -0,0 +1,408 @@
...
+    @staticmethod
+    def inspect_image(image: str) -> None:
+        """
+        Since we are sharing artifacts across jobs we need to make sure that these are loaded properly.
+        The easiest way to check this is to run `docker inspect` on the image
+        """
+        try:
+            (docker["image", "inspect", image] > "/dev/null")()
+        except plumbum.ProcessExecutionError:
+            print(f"Image not found: {image}")
```
mathbunnyru, line 387: Maybe we should fail, if we don't succeed here?

mathbunnyru:
A bit more thoughts:
- We don't have to actually checkout to `main` in workflow. I think the only place where we might need this is `Create tags and manifests`, where we actually checkout two repos.
- I think it's ok to ask users manually run `python -m pip install -r requirements-dev.txt` (the same for the docs), as a starting point to use the project. This is what I've done in your workflow, so it doesn't depend on Makefile at all.

You asked for `GHCR_PAT` token.
I'm not sure this is actually needed.
https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry
GitHub docs tell to use `GITHUB_TOKEN`.
Maybe this is gonna work in our repo as well?

mathbunnyru:
@trallard how is it going here?
I don't want to put pressure on you, but it would be nice to hear the status of this PR.

trallard:
Hi @mathbunnyru  I am so sorry - work and life have been chaotic recently. I will get to it later today (have some time blocked) 🙏🏽  thanks for the ping and for bringing this to the top of my list

trallard:
update - I ended up super busy - I have now cancelled all my meetings and blocked time for tomorrow I need to get this out of the way ⚡ 🙏🏽

dodo.py
trallard: I had placed the `build_docs` as a placeholder - while I found a good default.  Replaced now

trallard:
Made some progress - added some TODOs so I know what is missing:

- [ ] Check the GHCR push bit
- [ ] Finish cleaning up the Makefile
- [ ] Improve the inspect function to make it more robust
- [ ] Modify `task_docker_build` to allow for individual image build
- [ ] Overall check

dodo.py
```diff
@@ -0,0 +1,484 @@
...
+            Paths.REQS_DOCS,
+        ),
+    ]
+    if Utils.IS_CI:
+        actions.insert(
+            0,
+            Utils.do(*Utils.PIP, "--upgrade", "pip"),
+        )
+
+    return dict(actions=actions)
```
mathbunnyru, line 87: I don't think this is needed.
I think it's easier to tell people that they need to install `requirements.txt` than to tell them that they need to install doit first, then they have to run `doit setup_docs` to install other requirements.
It's what many python-based projects require.
The same goes for `task_setup_dev`

dodo.py
```diff
@@ -0,0 +1,484 @@
...
+
+from tagging.git_helper import GitHelper
+
+docker = plumbum.local["docker"]
+
+# global doit config
+DOIT_CONFIG = {"verbosity": 2, "default_tasks": ["what"]}
+
+
+def task_what() -> dict[str, Any]:
```
mathbunnyru, line 35: Well done 👍

dodo.py
```diff
@@ -0,0 +1,484 @@
...
+
+Command to list all the available tasks in the command line:
+doit list
+
+Get information on a single task:
+doit info <task-name>
+
+Targeted run of an individual task:
+Examples:
+    $ doit docs_build
```
mathbunnyru, line 15: This is really good for anyone who wants to develop our project 👍

mathbunnyru:
@trallard also I wanted to tell you not to worry about test failures related to `nbconvert` and `jinja2`.
They have nothing to do with your PR.
https://github.com/jupyter/docker-stacks/pull/1675

mathbunnyru:
@trallard I think I've fixed the tests, so I merged master to this branch to trigger the workflows here.

.github/workflows/build-test-docker.yml
```diff
@@ -0,0 +1,200 @@
...
+        id: slug
+        run: |
+          echo "::set-output name=short_sha::$(echo ${GITHUB_SHA} | cut -c1-7)"
+
+      - name: Save Docker images as artifact 💾
+        uses: actions/upload-artifact@v2
+        with:
+          name: docker-images-${{ steps.slug.outputs.short_sha }}
+          path: |
+            main/.github/built-docker-images/
```
mathbunnyru, line 111:
```suggestion
            .github/built-docker-images/
```

.github/workflows/sphinx.yml
```diff
@@ -34,21 +36,23 @@ jobs:
...
         with:
-          python-version: "3.9"
+          python-version: 3.9
 
-      - name: Install Dev Dependencies
+      - name: Install Doc Dependencies 📦
         run: |
-          python -m pip install --upgrade pip
-          make install-docs-env
+          python -m pip install doit plumbum
```
mathbunnyru, line 49: I think it would be easier to put these both in `requirements-docs.txt` as well and instead of running `python -m doit setup_docs`, directly run `python -m install -r requirements-docs.txt`.

mathbunnyru:
Also, I gave this a thought: do we actually need the GHCR pushing at all?
In my head, it was only needed to have an intermediate place between two very similar processes for arm and x86 and the final process of actually pushing manifests and the images+tags to docker hub.
But you have used `actions/download-(upload-)artifact` approach to pass this and this looks much better for me 👍 
It is actually better because it works in branches (and so it works in PRs).

Maybe for now we can not use GHCR at all and get away with artifacts?
It might be that I miss a point or don't understand something, please feel free to prove me wrong :)

.github/workflows/sphinx.yml
trallard: agree - was experimenting with this but it makes sense to keep the whole workflow as simple as possible

.github/workflows/sphinx.yml
```diff
@@ -34,21 +36,22 @@ jobs:
...
         with:
-          python-version: "3.9"
+          python-version: 3.9
 
-      - name: Install Dev Dependencies
+      - name: Install Doc Dependencies 📦
         run: |
-          python -m pip install --upgrade pip
-          make install-docs-env
+          python -m pip install - r "requirements-docs.txt"
```
mathbunnyru, line 49:
```suggestion
          python -m pip install -r requirements-docs.txt
```

mathbunnyru:
I don't get it :(
Is doit running **all** `task_*` methods no matter what I actually call?
When I run `python -m doit docs_build` it runs `task_docker_load_images()` even though I didn't add such a dependency :(

schettino72:
> Is doit running all task_* methods no matter what I actually call?

Yes! The body of task creators (`task_*`) are *always executed, and should be used to return Task Metadata only.
Actual task execution (when it is selected) executes a task's `action`.

* Actually there are ways to avoid that...

mathbunnyru:
Thanks, @schettino72. That was not obvious.

What would you suggest?
I want to make sure some tasks are run only in CI.
That's why I just put `assert`s in the beginning of these tasks.
But obviously, it makes it fail during local development, for example.

https://github.com/jupyter/docker-stacks/pull/1631/commits/8f8e5033e7d4277faca7cafe9c14802f48a49b60

schettino72:
```python3
def task_docker_load_images() -> dict[str, Any]:
    """
    Load docker images from previously downloaded artifacts 📥
    Since we are running in a CI environment and within a separate job than
    the one where the images are built, we need to load the images from the
    `CI_IMAGE_TAR`
    """

    def check_image():
        assert Utils.IS_CI
        assert (
            Utils.CI_IMAGE_TAR.exists()
        ), f"not found images archive in: {Utils.CI_IMAGE_TAR}"

    return dict(
        name="load_images",
        doc="Load and inspect Docker images",
        actions=[
            check_image,
            Utils.do("docker", "load", "--input", Utils.CI_IMAGE_TAR),
            # TODO: @trallard to add a more robust inspect
            (Utils.inspect_image, [Utils.get_images()[-1]]),
            Utils.do("docker", "images"),
        ],
    )

mathbunnyru:
Fail of `Build Sphinx Documentation and check links` is ok - somewhy github returned 403 during check_url.

mathbunnyru:
@trallard I merged some simple parts of this PR to master, to make this PR a bit more focused.
Would you like to continue the work here or you don't have plans to finish this?

No pressure here, I would like to know to better understand the current situation ❤️

trallard:
Follow up

>I merged some simple parts of this PR to master
What has already been merged? - do I need to remove anything from this PR then?

mathbunnyru:
I'm closing this PR, because approach with self-hosted runners and many jobs seems to work fine.

PR #1637, opened 2022-02-25
add the COPY_HOME option in start.sh

djangoliv:
These changes make it possible to force the copy of the files of jovyan user to NB_USER 

With kubespawner and user who is not jovyan, volumes mounted in /home/${NB_USER} prevent copying home jovyan files.

With kubespawner and a configuration like bellow:

```
c.KubeSpawner.volume_mounts = [
    {
        'name': 'common',
        'mountPath': '/home/{username}/files',
    },
]
```

Kubespwner create `/home/{username}` to mount the declared volume.
So when we execute start.sh the directory `/home/{username}` already exist and there is no copy of jovyan files.

With this PR, the copy can now be force with the `COPY_HOME` environment variable.

Regards

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -76,6 +76,7 @@ if [ "$(id -u)" == 0 ] ; then
     # - NB_GID: a group id we want our user to belong to
     # - NB_GROUP: the groupname we want for the group
     # - GRANT_SUDO: a boolean ("1" or "yes") to grant the user sudo rights
+    # - COPY_HOME: a boolean ("1" or "yes") to copy the original home files
     # - CHOWN_HOME: a boolean ("1" or "yes") to chown the user's home folder
     # - CHOWN_EXTRA: a comma separated list of paths to chown
     # - CHOWN_HOME_OPTS / CHOWN_EXTRA_OPTS: arguments to the chown commands
@@ -108,9 +109,9 @@ if [ "$(id -u)" == 0 ] ; then
     # directory if it doesn't already exist, and update the current working
     # directory to the new location if needed.
     if [[ "${NB_USER}" != "jovyan" ]]; then
-        if [[ ! -e "/home/${NB_USER}" ]]; then
+        if [[ ! -e "/home/${NB_USER}"  ||  "${COPY_HOME}" == "yes" || "${COPY_HOME}" == "1" ]]; then
             _log "Attempting to copy /home/jovyan to /home/${NB_USER}..."
-            mkdir "/home/${NB_USER}"
+            mkdir -p "/home/${NB_USER}"
             if cp -a /home/jovyan/. "/home/${NB_USER}/"; then
                 _log "Success!"
             else
--- a/docs/using/troubleshooting.md
+++ b/docs/using/troubleshooting.md
@@ -141,6 +141,7 @@ If you have also **created a new user**, you might be experiencing any of the fo
         -e NB_GID=1234 \
         -e CHOWN_HOME=yes \
         -e CHOWN_HOME_OPTS="-R" \
+        -e COPY_HOME="yes" \
         -w "/home/${NB_USER}" \
         -v "${PWD}"/test:/home/callisto/work \
         jupyter/minimal-notebook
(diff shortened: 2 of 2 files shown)


consideRatio:
This feature has been raised for discussion already in https://github.com/jupyterhub/kubespawner/issues/574#issuecomment-1048706438. I wrote the following that I think remains relevant.

>This is a discussion related to jupyter/docker-stacks. My understanding is that you discuss the following feature summarized.
>
>1. You have ensured there is a `/home/{username}/some_dir` ahead of time
>2. As part of the `start.sh` script of jupyter/docker-stacks runs, you have made it switch from a `jovyan` user to `{username}`. As part of this script running, you want to be able to make `start.sh` change its default behavior when the `{username}` home folderexists. You want either to replace the `{username}` folder if it exists, or in some way merge content from `jovyan` home folder to the existing `{username}` folder.
>
>I would overall say that this is a quite uncommon action, and could easily lead to loss of data if misunderstood by someone. Due to that, I'm not so happy about the idea of supporting a feature like that - if I understood it correctly.

---

> Oh... I couldn't transfer this issue to jupyter/docker-stacks as it was outside the organization. Anyhow, that is where this belongs. You could perhaps copy paste things from this to there if you want to raise it for discussion? Please if so, also include my response.

---

I don't think this feature should be implemented as I think it isn't an intuitive enough user experience, and I believe it could lead to confusion and loss of data - besides adding to a long term maintenance complexity.

djangoliv:
@consideRatio Sorry for the noise on kubespawner issues.

It is obviously not up to kubespawner to handle this use case.

But in docker-stacks this feature is already implemented in start.sh . This PR is just a fix for a special case.

I agree that renaming the user and their files may seem dangerous, but keeping the username in jupyterhub is very comfortable.

base-notebook/start.sh
```diff
@@ -108,9 +109,9 @@ if [ "$(id -u)" == 0 ] ; then
     # directory if it doesn't already exist, and update the current working
     # directory to the new location if needed.
     if [[ "${NB_USER}" != "jovyan" ]]; then
-        if [[ ! -e "/home/${NB_USER}" ]]; then
+        if [[ ! -e "/home/ogiorgis"  ||  "${COPY_HOME}" == "yes" || "${COPY_HOME}" == "1" ]]; then
```
mathbunnyru, line 112: `ogiorgis` was probably the name of your user and it should be changed.
And there are too many spaces near `||`.

base-notebook/start.sh
```diff
@@ -76,6 +76,7 @@ if [ "$(id -u)" == 0 ] ; then
     # - NB_GID: a group id we want our user to belong to
     # - NB_GROUP: the groupname we want for the group
     # - GRANT_SUDO: a boolean ("1" or "yes") to grant the user sudo rights
+    # - COPY_HOME: a boolean ("1" or "yes") to copy the otiginal home files
```
mathbunnyru, line 79:
```suggestion
    # - COPY_HOME: a boolean ("1" or "yes") to copy the original home files
```

docs/using/troubleshooting.md
```diff
@@ -159,6 +160,7 @@ If you have also **created a new user**, you might be experiencing any of the fo
 
    - `-e NB_USER=callisto`: will create a new user `callisto` and automatically add it to the `users` group (does not delete jovyan)
    - `-e NB_UID=1234` and `-e NB_GID=1234`: will set the `UID` and `GID` of the new user (`callisto`) to `1234`
+   - `-e COPY_HOME=yes`: ensure that the new user will have the same files than the old user in its `/home` directory
```
mathbunnyru, line 163:
```suggestion
   - `-e COPY_HOME=yes`: ensure that the new user will have the same files as the old user in its `/home` directory
```

mathbunnyru (changes requested):
I'm not an expert on how `start.sh` works in different cases and I trust @consideRatio's opinion on this one.

consideRatio:
I want to avoid mixed signals of both reviewing with code suggestions and saying the feature is problematic to accept, so i'll close this PR for now.

Not because this feature doesn't have value at least to you, but because of what was mentioned above.

PR #1641, opened 2022-02-28
[MAINT] - Updates to GH actions workflows

### LINKED ISSUE #1540: ci maintenance: queue workflows or jobs with new GitHub workflow `concurrency` option?
I think this GitHub workflow thingy can be really useful. It makes two workflows being triggered not run in parallell if they are within the same group. This will make us queue workflows in sequence.

```yaml
concurrency: 
  group: ${{ github.head_ref }}
```

This can also be applied to specific jobs, but not steps within jobs. This makes it even more valuable if we manage to split apart our build workflow into multiple separate jobs and then merges them into a single later. By doing so, we can mark the job that does the push to github and could run into merge conflicts marked to belong to a specific concurrency group to ensure that never happens rather than marking the entire workflow with all steps that could happen cuncurrently.

trallard:
- :construction_worker: [MAINT] - Use concurrency in GH actions: right now I have set it so that new runs on a given workflow cancel any other runs (for the same workflow)
- :construction_worker: Disable schedule for forks - I always get notifications that some workflows failed so this improvement prevents scheduled actions from running on forks

Closes #1540


mathbunnyru (approved):

consideRatio (approved):
:tada: thanks for your work @trallard, I appreciate the inline comments!!

PR #1644, opened 2022-03-05
Refactor package helper and fix tests for aarch64

mathbunnyru:
After this PR tests pass on arm base-notebook image 🎉.
I tested this under m1 mac, should also work fine on arm linux.

### DIFF
--- a/tests/base-notebook/test_packages.py
+++ b/tests/base-notebook/test_packages.py
@@ -47,30 +47,31 @@ LOGGER = logging.getLogger(__name__)
 # Mapping between package and module name
 PACKAGE_MAPPING = {
     # Python
-    "matplotlib-base": "matplotlib",
     "beautifulsoup4": "bs4",
-    "scikit-learn": "sklearn",
-    "scikit-image": "skimage",
-    "spylon-kernel": "spylon_kernel",
+    "matplotlib-base": "matplotlib",
     "pytables": "tables",
+    "scikit-image": "skimage",
+    "scikit-learn": "sklearn",
+    "spylon-kernel": "spylon_kernel",
     # R
     "randomforest": "randomForest",
-    "rsqlite": "DBI",
     "rcurl": "RCurl",
     "rodbc": "RODBC",
+    "rsqlite": "DBI",
 }
 
 # List of packages that cannot be tested in a standard way
 EXCLUDED_PACKAGES = [
-    "python",
-    "hdf5",
+    "bzip2",
+    "ca-certificates",
     "conda-forge::blas[build=openblas]",
+    "hdf5",
+    "mamba[version='<0.18']",
+    "openssl",
     "protobuf",
+    "python",
     "r-irkernel",
     "unixodbc",
-    "bzip2",
-    "openssl",
-    "ca-certificates",
 ]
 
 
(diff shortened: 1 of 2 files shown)


tests/base-notebook/test_packages.py
```diff
@@ -47,30 +47,31 @@
...
 
 # List of packages that cannot be tested in a standard way
 EXCLUDED_PACKAGES = [
-    "python",
-    "hdf5",
+    "bzip2",
+    "ca-certificates",
     "conda-forge::blas[build=openblas]",
+    "hdf5",
+    "mamba[version='<0.18']",
```
mathbunnyru, line 69: I sorted these 2 lists, but also added this line - it's a workaround for testing under arm.
As soon as we remove this workaround for arm, this line can be deleted.

tests/base-notebook/test_packages.py
```diff
@@ -106,53 +107,44 @@ def r_package_predicate(package: str) -> bool:
...
-) -> int:
+def _check_import_package(package_helper: CondaPackageHelper, command: list[str]):
     """Generic function executing a command"""
     LOGGER.debug(f"Trying to import a package with [{command}] ...")
-    rc = package_helper.running_container.exec_run(command)
-    return rc.exit_code  # type: ignore
+    exec_result = package_helper.running_container.exec_run(command)
+    assert (
+        exec_result.exit_code == 0
+    ), f"Import package failed, output: {exec_result.output}"
```
mathbunnyru, line 116: This way we do not lose the output of the test and using exceptions to show the errors is more pythonic.

tests/base-notebook/test_packages.py
```diff
@@ -106,53 +107,44 @@ def r_package_predicate(package: str) -> bool:
...
-        package_helper, ["R", "--slave", "-e", f"library({package})"]
-    )
+    _check_import_package(package_helper, ["R", "--slave", "-e", f"library({package})"])
 
 
 def _check_import_packages(
     package_helper: CondaPackageHelper,
-    filtered_packages: Iterable[str],
-    check_function: Callable[[CondaPackageHelper, str], int],
-    max_failures: int,
```
mathbunnyru, line 136: I think we have never used `max_failures != 0`, so it's easier to remove this.

tests/base-notebook/test_packages.py
```diff
@@ -106,53 +107,44 @@ def r_package_predicate(package: str) -> bool:
...
+    check_function: Callable[[CondaPackageHelper, str], None],
 ) -> None:
     """Test if packages can be imported
 
     Note: using a list of packages instead of a fixture for the list of packages since pytest prevents use of multiple yields
     """
     failures = {}
     LOGGER.info("Testing the import of packages ...")
-    for package in filtered_packages:
+    for package in packages_to_check:
```
mathbunnyru, line 140: This variable is more clear to me.

tests/base-notebook/test_packages.py
consideRatio: > it's a workaround for testing under arm.
As soon as we remove this workaround for arm, this line can be deleted.

That sounds like an excellent inline comment!

```suggestion
    # FIXME: When we remove a workaround for arm regarding mamba, we can 
    #        test installation of mamba as well and remove this exception.
    #        See ???
    "mamba[version='<0.18']",
```

I like to explain a bit that something can be removed in the future, and link out to an issue where it is tracked, so something like the comment above but with a functional issue to understand if its time to remove it.

consideRatio (approved):
This LGTM, I suggested an inline comment but its not crucial.

Thanks for the review comments, it helped!

tests/base-notebook/test_packages.py
mathbunnyru: I agree. Done 👍

PR #1645, opened 2022-03-07
Cleanup .gitignore

mathbunnyru:


.gitignore
```diff
@@ -39,28 +39,125 @@ pip-delete-this-directory.txt
...
+# The content above is copied from                               #
+# https://github.com/github/gitignore/blob/main/Python.gitignore #
+# Please, add the content only below                             #
+
 # Mac OS X
 .DS_Store
 
-dockerspawner
-dockerspawner.tar.gz
-*.orig
```
mathbunnyru, line 63: I'm not sure these are needed.
I haven't seen anything in this repo that created such files.

.gitignore
```diff
@@ -39,28 +39,127 @@ pip-delete-this-directory.txt
...
+#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
+#  and can be added to the global gitignore or merged into this file.  For a more nuclear
+#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
+#.idea/
+
+##################################################################
+#             The content above is copied from                   #
+# https://github.com/github/gitignore/blob/main/Python.gitignore #
+#       Please, add the content only below these lines           #
+##################################################################
```
mathbunnyru, line 158: This is the main idea - just use the general Python gitignore and add only a few files that we explicitly need.

consideRatio: I like such separation! I typically put what we manually add at the top as that will be seen for everyone directly opening the file while whats in the end may not be seen.

consideRatio (approved):

PR #1646, opened 2022-03-07
Update mypy config and make it work well with pre-commit

mathbunnyru:
What I've done:
- Updated mypy config to use `--strict --follow-imports error`. This makes this tool much more efficient in finding errors.
- Fixed all the errors
- Made `pre-commit` mypy hook use the same config
- Made sure the `pre-commit` hook works well and disabled `mypy` for `git commit` stage
- Added some inline comments to explain why and how I did it.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -4,6 +4,9 @@ follow_imports = normal
 strict = False
 no_incremental = True
 
+[mypy-Cython.*]
+ignore_missing_imports = True
+
 [mypy-docker.*]
 ignore_missing_imports = True
 
@@ -20,6 +23,9 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 
 [mypy-pyspark.*]
+ignore_missing_imports = True
+
+[mypy-setuptools.*]
 ignore_missing_imports = True
 
 [mypy-tensorflow.*]
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,6 +1,6 @@
-# MMypy is an optional static type checker for Python that aims to combine
+# Mypy is an optional static type checker for Python that aims to combine
 # the benefits of dynamic (or "duck") typing and static typing.
-# 
+#
 # Documentation: http://www.mypy-lang.org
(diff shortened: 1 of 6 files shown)


mathbunnyru:
Unfortunately, `pre-commit` mypy doesn't work the same way as explicit mypy run.
So, it's better to run `mypy --config-file ./mypy.ini ./` for now.
CC: @trallard

mypy.ini
```diff
@@ -4,6 +4,9 @@ follow_imports = normal
 strict = False
 no_incremental = True
 
+[mypy-Cython.*]
```
consideRatio, line 7: Can you add a comment to the top of this config file saying what we use it for, and what mypy is?

consideRatio:
> Unfortunately, `pre-commit` mypy doesn't work the same way as explicit mypy run. So, it's better to run `mypy --config-file ./mypy.ini ./` for now.

If we cant use if with pre-commit then that is also relevance to clarify in the file as a comment/fixme preferably linking out to something relevant to track if its to be fixed over time - a github issue in this or upstream repo.

consideRatio:
> Unfortunately, `pre-commit` mypy doesn't work the same way as explicit mypy run.
> So, it's better to run `mypy --config-file ./mypy.ini ./` for now.

If it doesn't work the same way, and it's not of importance, I suggest we stick with anything that can be done with pre-commit instead of sticking with an option that requires manual intervention etc. I've not read up on mypy or what it does though.

mypy.ini
mathbunnyru: Thanks!
I added the docs to make it clear.

mathbunnyru:
@consideRatio thank you for very good suggestions.
I think I made it work well for `pre-commit` and documented the workaround.
Please, take a look and tell me what you think.

consideRatio (approved):
Beautiful @mathbunnyru! Thanks for being so thorough with the transition to make this code typed, with all this setup I think this repo is really set to maintain typings!!! :heart: :tada:

PR #1647, opened 2022-03-08
Fix bashate under macos

mathbunnyru:
I used another way to check if env variable is set: https://stackoverflow.com/a/13864829/4881441
It works under macos (I was a bit annoyed by this error).


mathbunnyru:
Closing, it's too painful to deal with bash

PR #1651, opened 2022-03-14
Add some links for pre-commit and add default hooks

mathbunnyru:


.pre-commit-config.yaml
```diff
@@ -1,8 +1,16 @@
...
-  skip: [hadolint-docker]
-
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
 repos:
+  # `pre-commit sample-config` default hooks
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.1.0
+    hooks:
+      - id: trailing-whitespace
```
consideRatio, line 9: Hmmm... This seems to be enforcing that we _do not_ have trailing whitespace. Does that mean that the final new line will be removed? I think having a final new line is very good practice when working with `git` as it makes you able to add a line below without seeing a change in the previous line etc.

I'd be unhappy about seeing the newlines be trimmed away at the end of files.

We use the `end-of-file-fixer` hook in most jupyterhub org repo's to ensure we get a new line at the end of the file. It is an autoformatter as well btw, so it automatically adds it if needed.

mathbunnyru: `trailing-whitespace` just trims the redundant whitespaces on the line.
It doesn't remove final new line.
You can see that my PR passes the tests and doesn't almost change any code in the repository, so it's ok.
It works well with `end-of-file-fixer` as well, which is also added just one line before this.

tensorflow-notebook/Dockerfile
```diff
@@ -15,4 +15,3 @@ RUN mamba install --quiet --yes \
     mamba clean --all -f -y && \
     fix-permissions "${CONDA_DIR}" && \
     fix-permissions "/home/${NB_USER}"
```
mathbunnyru, line 17: Actually, one final newline is still here, otherwise GitHub would display that it didn't find the newline.

.pre-commit-config.yaml
consideRatio: It will be redundant for Python files, and YAML, and I think also Markdown files thanks to black/prettier. Do you think we should retain it for the possible other file format that isn't managed by black and prettier?

I'd prefer to avoid adding it unless we do believe it can have use, otherwise it becomes additional complexity which in my mind almost always has some cost - if not only to rule out it wasn't related to some unknown behaviour observed.

mathbunnyru: I think it should be simple and polished.
We have some other formats already - some text configuration files, ipynb, Makefile, gitignore, shell files.
I have just checked and it worked on `requirements-dev.txt` file and it fixed it as well.
So, it should be easy for developers as well.

I say we give it a try and if it adds complexity, then we remove this.

PR #1655, opened 2022-03-16
Run pre-commit tools: isort, requirements-txt-fixer

mathbunnyru:
isort, requirements-txt-fixer both automatically fix files, so it should be good for developers as well.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,13 @@ repos:
       - id: pyupgrade
         args: [--py39-plus]
 
+  # Automatically sort python imports
+  - repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        args: [--profile, black]
+
   # Autoformat: Python code
   - repo: https://github.com/psf/black
     rev: 22.1.0
@@ -53,9 +60,11 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.1.0
     hooks:
-      - id: trailing-whitespace
-      - id: end-of-file-fixer
       - id: check-added-large-files
+      - id: end-of-file-fixer
+      - id: requirements-txt-fixer
+        args: [requirements-dev.txt, requirements-docs.txt]
+      - id: trailing-whitespace
 
   # Lint: Dockerfile
   - repo: https://github.com/hadolint/hadolint.git
@@ -108,12 +117,13 @@ repos:
       - id: nbstripout
 
   # nbQA provides tools from the Python ecosystem like
-  # pyupgrade, black, and flake8, adjusted for notebooks.
+  # pyupgrade, isort, black, and flake8, adjusted for notebooks.
   - repo: https://github.com/nbQA-dev/nbQA
     rev: 1.3.1
     hooks:
       - id: nbqa-pyupgrade
         args: [--py39-plus]
+      - id: nbqa-isort
       - id: nbqa-black
         args: [--target-version=py39]
       - id: nbqa-flake8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -63,7 +63,6 @@ repos:
       - id: check-added-large-files
       - id: end-of-file-fixer
       - id: requirements-txt-fixer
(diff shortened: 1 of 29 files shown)


mathbunnyru:
I think now it is pretty much everything, that can be checked in this project automatically and can reduce maintenance time.
If someone knows other useful tools to have, please, tell me :)

tests/scipy-notebook/data/cython/setup.py
```diff
@@ -1,4 +1,6 @@
-from setuptools import setup
-from Cython.Build import cythonize
+# These lines are not sorted by isort on purpose
+# see: https://stackoverflow.com/a/53356077/4881441
+from setuptools import setup  # isort:skip
+from Cython.Build import cythonize  # isort:skip
```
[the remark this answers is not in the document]
consideRatio: Thanks for adding a comment on why you `# isort:skip` them with a motivation etc, I love that practice!

.pre-commit-config.yaml
```diff
@@ -53,9 +60,11 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.1.0
     hooks:
-      - id: trailing-whitespace
-      - id: end-of-file-fixer
       - id: check-added-large-files
+      - id: end-of-file-fixer
+      - id: requirements-txt-fixer
+        args: [requirements-dev.txt, requirements-docs.txt]
```
consideRatio, line 66: Did you find you needed to pass these files explicitly? I think what will happen now, is that they are always checked no matter if they are changed or not. which I guess is fine - it is lightning fast anyhow I guess and won't probably cause issues. It does require us to explicitly list what we update. It can of course be kind of nice in a way, but also is something that can break as files are renamed etc.

mathbunnyru: I think we have to pass files explicitly.
https://github.com/pre-commit/pre-commit-hooks/blob/master/pre_commit_hooks/requirements_txt_fixer.py#L135
Actually, if we rename a file, this will break, and we will have to update this, but I guess it's only 2 files and it should be ok.

consideRatio (approved):
These changes look fine to me. I'd appreciate if another maintainer would approve this as well as I know there has been controversy about use of `isort` even though I personally appreciate the output i see below I don't remember why it can be found problematic.

In my mind, as long as we have pre-commit to help us, I'm fine with autoformatters forcing us to align with a systematic practice on how to do things.

.pre-commit-config.yaml
consideRatio: @mathbunnyru hmmm but pre-commit is passing arguments of the changes files by default I think, so it will be passed information. I think requirements-txt-fixer just ignores files it doesn't want to work on.

consideRatio: Example: https://github.com/jupyterhub/jupyterhub/blob/454e356e4dbfa8e291212f5eaff18dc0c14f5e11/.pre-commit-config.yaml#L39-L45

mathbunnyru: This is a valid point, I will do some quick-and-dirty tests :)

mathbunnyru:
There is a thing about sorting python imports:
Actually, it's already used in jupyterhub since Feb 2019.
https://github.com/jupyterhub/jupyterhub/pull/2418

https://github.com/jupyterhub/jupyterhub/pull/2418#discussion_r257451617
The choice between isort vs reorder_python_imports was made because this feature was absent:
https://github.com/PyCQA/isort/issues/679

So, no reason not to use isort now :)

mathbunnyru:
I've just tried reorder_python_imports instead of isort and here is the small part of the diff:
```diff
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
-
-
 import requests
-from conftest import TrackedContainer, find_free_port
+from conftest import find_free_port
+from conftest import TrackedContainer
```

So, in this project the only differences are (at least by default):
- it removes empty lines between shebang/comments and code
- it prefers one import per line

It doesn't make any of the tools better/worse than the other, so I think let's go with isort :)

consideRatio:
+1 for isort, especially because I think you don't have nbqa-reorder-python-imports.

mathbunnyru:
Thank you @trallard ! Merging this :)

PR #1658, opened 2022-03-17
Update sphinx-book-theme to 0.3.0rc2 and remove custom.css

mathbunnyru:


docs/_static/custom.css
```diff
@@ -1,3 +0,0 @@
-body div.sphinxsidebarwrapper p.logo {
```
mathbunnyru, line 1: This was made for alabaster theme, I don't think we need it now.

mathbunnyru: I checked this and it doesn't make difference, as far as I see.

mathbunnyru: Yes, this worked only for alabaster.
https://github.com/bitprophet/alabaster/blob/master/alabaster/static/alabaster.css_t#L141

If someone wants to implement this for the new theme, I would gladly accept the PR, but it's better not to have code which doesn't do what it looks like.

PR #1659, opened 2022-03-17
Do not print git diff on pre-commit fail

mathbunnyru:
@consideRatio
This actually reverts your previous PR:
https://github.com/jupyter/docker-stacks/pull/1307

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ img-rm-dang: ## remove dangling images (tagged None)
 
 
 pre-commit-all: ## run pre-commit hook on all files
-	@pre-commit run --all-files --hook-stage manual || (printf "\n\n\n" && git --no-pager diff --color=always)
+	@pre-commit run --all-files --hook-stage manual
 pre-commit-install: ## set up the git hook scripts
 	@pre-commit --version
 	@pre-commit install
--- a/all-spark-notebook/Dockerfile
+++ b/all-spark-notebook/Dockerfile
@@ -6,6 +6,10 @@ FROM $BASE_CONTAINER
 
 LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 
+# Fix: https://github.com/hadolint/hadolint/wiki/DL4006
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -170,7 +170,7 @@ img-rm-dang: ## remove dangling images (tagged None)
 
 
 pre-commit-all: ## run pre-commit hook on all files
-	@pre-commit run --all-files --hook-stage manual || (printf "\n\n\n" && git --no-pager diff --color=always)
```
mathbunnyru, line 173: The main reason to revert this:
Because of `||` we don't fail the CI when we should.

Another reason:
Actually, all automatic pre-commit fixes will be pushed automatically be pre-commit.ci and the diff can be seen as a commit later.

When manually running pre-commit, I think it's not difficult to type `git diff` if you really want to see the diff.

mathbunnyru:
The failure is expected, actually.
Because this is what I saw in the previous run of pre-commit in GitHub.

I saw:
```
Lint Dockerfiles.........................................................Failed
- hook id: hadolint-docker
- exit code: 1
```

And the build wasn't mark as failed, and I spent some time what might be wrong with pre-commit / hadolint docker.
And the problem was in our Makefile.

I will fix the docker issue here as well.

all-spark-notebook/Dockerfile
```diff
@@ -6,6 +6,10 @@ FROM $BASE_CONTAINER
 
 LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 
+# Fix: https://github.com/hadolint/hadolint/wiki/DL4006
+# Fix: https://github.com/koalaman/shellcheck/wiki/SC3014
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
```
mathbunnyru, line 11: It's easier to add this to every Dockerfile.
This way hadolint and anyone who reads the docker file will know we're using bash with pipefail.

consideRatio (approved):

PR #1661, opened 2022-03-17
Run black pre-commit on python code in docs

mathbunnyru:
It would be nice if people could copy our code from the documentation and paste it into their projects and it would work without any changes.
This also makes the code in the docs follow the same principles as we have in our main projects.
It also makes the review process easier - because the docs are easier to read.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -130,6 +130,10 @@ repos:
     rev: v1.12.1
     hooks:
       - id: blacken-docs
+        # --skip-errors is added to allow us to have python syntax highlighting even if
+        # the python code blocks includes jupyter specific additions such as % or !
+        # See https://github.com/asottile/blacken-docs/issues/127 for an upstream
+        # feature request about this.
         args: [--target-version=py39, --skip-errors]
 
 # Docker hooks do not work in pre-commit.ci
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -130,7 +130,7 @@ repos:
     rev: v1.12.1
     hooks:
       - id: blacken-docs
-        args: [--target-version=py39]
+        args: [--target-version=py39, --skip-errors]
 
 # Docker hooks do not work in pre-commit.ci
 # See: <https://github.com/pre-commit-ci/issues/issues/11>
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -99,7 +99,7 @@ repos:
       - id: markdownlint
         args: ["--fix"]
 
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -99,8 +99,6 @@ repos:
       - id: markdownlint
         args: ["--fix"]
 
-  # Run tools on Jupyter notebooks
```
mathbunnyru, line 102: This is a bit confusing, because there are actually no sections in this file, so better to remove this.

.pre-commit-config.yaml
```diff
@@ -99,8 +99,6 @@ repos:
...
-
   # strip output from Jupyter notebooks
```
mathbunnyru, line 102:
```suggestion
  # Strip output from Jupyter notebooks
```

docs/using/specifics.md
```diff
@@ -143,7 +143,7 @@ sdf_len(sc, 100, repartition = 1) %>%
 Spylon kernel instantiates a `SparkContext` for you in variable `sc` after you configure Spark
 options in a `%%init_spark` magic cell.
 
-```python
+```text
```
consideRatio, line 146: It would be preferable to have `python` remain here even though its not exactly Python syntax highlighting we would need - as this includes jupyter magic `%%` stuff.

Will it error if you go with `python` / did it automatically declare this to be `text` because it found it to be invalid python?

consideRatio (commented):
LGTM but if possible, we should stick with python syntax highlighting even when its a jupyter magic involved.

docs/using/specifics.md
mathbunnyru: Yes, it will error :(
I actually found `--skip-errors` flag which is what we want.
And right now ipython magic is not supported and there is already an upstream issue.
https://github.com/asottile/blacken-docs/issues/127

I think using this flag is the best solution here.

.pre-commit-config.yaml
```diff
@@ -130,7 +130,7 @@ repos:
     rev: v1.12.1
     hooks:
       - id: blacken-docs
-        args: [--target-version=py39]
+        args: [--target-version=py39, --skip-errors]
```
consideRatio, line 133:
```suggestion
        # --skip-errors is added to allow us to have python syntax highlighting even if
        # the python code blocks includes jupyter specific additions such as % or !
        args: [--target-version=py39, --skip-errors]
```

consideRatio (approved):

.pre-commit-config.yaml
mathbunnyru: Nice!

PR #1670, opened 2022-03-30
Simplify binder Dockerfile

mathbunnyru:


binder/Dockerfile
```diff
@@ -14,15 +14,4 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
...
-# Fix permissions on README.ipynb as root
-USER root
-
-RUN fix-permissions README.ipynb
-
-# Switch back to jovyan to avoid accidental container runs as root
-USER ${NB_UID}
-
-WORKDIR "${HOME}"
+COPY --chown=${NB_UID}:${NB_GID} binder/README.ipynb "${HOME}"/README.ipynb
```
mathbunnyru, line 17: This actually makes the same as above but in a simple way.
Also, I don't think `fix-permissions` works on files, `--chown` does.

PR #1673, opened 2022-04-01
Make tests a module and make imports absolute

mathbunnyru:
My motivation is that this way it works well in PyCharm when I open the root of this project.
I'm not a pro in python imports, unfortunately.

### DIFF
--- a/tests/all-spark-notebook/test_spark_notebooks.py
+++ b/tests/all-spark-notebook/test_spark_notebooks.py
@@ -5,7 +5,8 @@ import logging
 from pathlib import Path
 
 import pytest  # type: ignore
-from conftest import TrackedContainer
+
+from tests.conftest import TrackedContainer
 
 LOGGER = logging.getLogger(__name__)
 THIS_DIR = Path(__file__).parent.resolve()
--- a/Makefile
+++ b/Makefile
@@ -207,6 +207,6 @@ run-sudo-shell/%: ## run a bash in interactive mode as root in a stack
 
 test/%: ## run tests against a stack
 	@echo "::group::test/$(OWNER)/$(notdir $@)"
-	tests/run_tests.py --short-image-name "$(notdir $@)" --owner "$(OWNER)"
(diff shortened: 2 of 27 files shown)


consideRatio:
Hmmm I dont understand what this resolves.

> it works well in PyCharm when I open the root of this project.

How does what not work otherwise etc?

tests/all-spark-notebook/test_spark_notebooks.py
```diff
@@ -5,7 +5,8 @@
 from pathlib import Path
 
 import pytest  # type: ignore
-from conftest import TrackedContainer
```
mathbunnyru, line 8: When I open PyCharm at the root of this project, it tries to resolve this import.
Because there is no __init__.py, it tries to import from the root of the repo and fails to find this import.

mathbunnyru:
> I'd appreciate this PR getting reviewed by someone who does

Me too :) Could you please ping someone who knows python imports and packaging model well?

tests/all-spark-notebook/test_spark_notebooks.py
mathbunnyru: I think it worked, because pytest might do something with sys.path.
https://docs.pytest.org/en/6.2.x/pythonpath.html#standalone-test-modules-conftest-py-files

mathbunnyru:
@manics we need help to organize our python files / packages / imports so it is good python code and works well in IDEs and mypy.

manics:
I'm definitely not an expert in pytest 🤣.

I think adding `__init__.py` to tests is OK to support imports though see the caveats:
https://docs.pytest.org/en/6.2.x/goodpractices.html#tests-outside-application-code

Do relative imports work when you add `__init__.py`?

mathbunnyru:
> Do relative imports work when you add __init__.py?

I think they partially work: code runs, but mypy is not happy:
https://github.com/jupyter/docker-stacks/pull/1673#discussion_r840573015

trallard:
I'd be happy to review this tomorrow 👍🏽

mathbunnyru:
I will merge this one, I've used this branch for a while and it seems to work fine.

Issue #1657, opened 2022-03-16
cannot start docker container after changing user name


mayswind:
### What docker image(s) are you using?

base-notebook

### OS system and architecture running docker image

Alpine Linux v3.11/amd64

### What Docker command are you running?

*See the reproducing steps*

### How to Reproduce the problem?

1. Create a Dockerfile, and fill with these following content (change the default user name)
```
FROM jupyter/base-notebook:lab-3.3.0

ENV NEW_NB_USER="jupyter"
ENV NB_USER=$NEW_NB_USER \
    HOME=/home/$NEW_NB_USER

USER root

RUN cd / \
  && usermod -md /home/$NB_USER -l $NB_USER jovyan

WORKDIR $HOME

USER $NB_UID
```
2. Use this Dockerfile to build a new docker image
```
docker build . -t "jupyter-base-notebook-with-new-user-name"
```
3. Start a new docker container with this docker image
```
docker run --net=host jupyter-base-notebook-with-new-user-name
```

### Command output

```bash session
Entered start.sh with args: jupyter lab
```

### Expected behavior

Jupyterlab should be started successfully (like using the image `jupyter/base-notebook:lab-3.2.0`)

### Actual behavior

There are no error message after executing `docker run`, and jupyterlab would no start.

### Anything else?

I have submit a pr (#1649) before, fix the problem this issue metions. In that pr, @consideRatio mentions some questions, I try to explain them.
In the #1546 pr, the `start.sh` script have add some code to check whether there are some problems with current user, group or home folder. The script would get the user id / group id of `jovyan` via `id` command (see the code in the below screenshot)
![image](https://user-images.githubusercontent.com/2211648/158622502-eb62a396-749e-48d1-b017-541b9658eb2a.png)
Then, the following code (see the code in the below screenshot) would check whether the `jovyan` user / group id equals to the current user.
![image](https://user-images.githubusercontent.com/2211648/158627300-6368185e-4fe5-42b1-b91c-6ced9faf4e30.png)
But the `id` command would return a non-zero value when user name does not exist, and then the script would exit. So the following checking code would never execute, and the warn log would not be output too. I think it should ensure that the user / group id of `jovyan` can be obtained. So in #1649 I have add two additional function to call `id` command to avoid the main process geting a non-zero value and ensure the remain code can be executed successfully.

mathbunnyru:
We actually have a way to change the username to the desired one:
[...]

PR #1674, opened 2022-04-01
Fix changing username in Dockerfile

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1657

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -204,9 +204,6 @@ else
         _log "WARNING: container must be started as root to grant sudo permissions!"
     fi
 
-    JOVYAN_UID="$(id -u jovyan 2>/dev/null)"  # The default UID for the jovyan user
-    JOVYAN_GID="$(id -g jovyan 2>/dev/null)"  # The default GID for the jovyan user
-
     # Attempt to ensure the user uid we currently run as has a named entry in
     # the /etc/passwd file, as it avoids software crashing on hard assumptions
     # on such entry. Writing to the /etc/passwd was allowed for the root group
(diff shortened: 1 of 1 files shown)


base-notebook/start.sh
```diff
@@ -204,9 +204,6 @@ else
         _log "WARNING: container must be started as root to grant sudo permissions!"
     fi
 
-    JOVYAN_UID="$(id -u jovyan 2>/dev/null)"  # The default UID for the jovyan user
```
mathbunnyru, line 207: `2>/dev/null` is not required, so I deleted it.

base-notebook/start.sh
```diff
@@ -239,13 +236,13 @@ else
     # A misconfiguration occurs when the user modifies the default values of
     # NB_USER, NB_UID, or NB_GID, but we cannot update those values because we
     # are not root.
-    if [[ "${NB_USER}" != "jovyan" && "${NB_USER}" != "$(id -un)" ]]; then
+    if [[ "${NB_USER}" != "$(id -un)" && "${NB_USER}" != "jovyan" ]]; then
         _log "WARNING: container must be started as root to change the desired user's name with NB_USER=\"${NB_USER}\"!"
     fi
-    if [[ "${NB_UID}" != "${JOVYAN_UID}" && "${NB_UID}" != "$(id -u)" ]]; then
+    if [[ "${NB_UID}" != "$(id -u)" && "${NB_UID}" != "$(id -u jovyan)" ]]; then
```
mathbunnyru, line 242: I changed the order of these two expressions.
Now, it won't execute `id -u jovyan` if the NB_UID is the one the current user has.
And because we don't call this expression, this will work fine when there is no `jovyan` user at all.

mathbunnyru:
To be honest, I don't want to write a test case for this right now, because I'm not sure we promise this will always work.
But, this fix seems to be an easy one, so let's merge it.

base-notebook/start.sh
```diff
@@ -204,9 +204,6 @@ else
         _log "WARNING: container must be started as root to grant sudo permissions!"
     fi
 
-    JOVYAN_UID="$(id -u jovyan 2>/dev/null)"  # The default UID for the jovyan user
-    JOVYAN_GID="$(id -g jovyan 2>/dev/null)"  # The default GID for the jovyan user
```
mathbunnyru, line 208: Oh, I see.
These were saved before changing /etc/passwd :(

mathbunnyru:
Closing this one, I don't think I'm gonna have enough time in the nearest future to finish this PR.

PR #1672, opened 2022-03-31
Remove JUPYTER_ENABLE_LAB variable warning

mathbunnyru:
This warning was there for more than 2 months.
Let's remove it, because I think everyone who wanted to change something, already did it.

PR #1675, opened 2022-04-01
[TMP] Disable tests

mathbunnyru:
Our CI fails now on any change:
https://github.com/jupyter/docker-stacks/pull/1672

What happens:
Because of our complicated build system, when we build minimal image, instead of building it on top of just created base image, we end up pulling an old image.
And this happens for other images as well.
And, we also rebuild during `push`, which is also not great.

This is **a major problem**.
What I want to do - merge this, rebuild images in master 5 times.
This way all our images eventually will be updated as they were build from source code.

After that, testing cab be enabled back.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -93,9 +93,6 @@ jobs:
           # Full logs for CI build
           BUILDKIT_PROGRESS: plain
 
-      - name: Test Docker Images
-        run: make -C main test-all
-
       - name: Clone Wiki
         uses: actions/checkout@v3
         with:


mathbunnyru:
I know this sounds crazy, but I don't see how to fix our broken CI now without some hotfix like this.

mathbunnyru:
The reason why CI fails:

Jinja2 has changed the import locations of `Markup` and `escape`:
`Markup and escape should be imported from MarkupSafe.`:
https://jinja.palletsprojects.com/en/3.1.x/changes/#version-3-1-0

nbconvert already fixed it:
https://github.com/jupyter/nbconvert/pull/1737/files

And the latest version with the fix is available: https://anaconda.org/conda-forge/nbconvert/files

But I think because of how we build images we end up with incompatible packages.

mathbunnyru:
Merging this one, because it affects all the PRs and makes life much more difficult. 
I will manually trigger the workflow several times (each time waiting for the previous one to finish).
After that, I will try to reenable testing.

PR #1679, opened 2022-04-07
Add --quiet to mamba info

mathbunnyru:
This will reduce output we don't need, for example here: https://github.com/jupyter/docker-stacks/wiki/all-spark-notebook-068e3c5ddab0#python-packages


consideRatio (approved):
Nice!

Issue #915, opened 2019-07-30
HEALTHCHECK Instructions?


rwmajor2:
I have a general question.  We base some of our Docker images on the Docker-stacks Dockerfiles.  We are following [CISP Docker Security hardening guidelines](https://www.cisecurity.org/benchmark/docker/) and one of the checklist items is:
```
 Ensure that HEALTHCHECK instructions have been added to container images
	Guidance:  Docker engine periodically checks the running container instances against that instruction to ensure containers are still operational
```
Can anyone provide any suggestions on what this is and how we may implement it in docker-stacks?  According to Docker help documentation, a HEALTHCHECK might be something like:
```
HEALTHCHECK --interval=5m --timeout=3s CMD curl -f http://localhost/ || exit 1
```
Thoughts?

@GrahamDumpleton

GrahamDumpleton:
Can't see what health checks have got to do with security hardening. They are an operations feature to ensure your application is running. Unless they think that your application not running may have been caused by hackers and so qualifies somehow as a security event.  Also be aware that defining health checks in a container image itself only really pertains to Docker's container run time. They aren't used by other container platforms such as Kubernetes. In Kubernetes health checks are defined in the separate deployment configuration of Kubernetes where they more rightly belong. So not sure if it really belongs in the container image itself. It should be really part of how you deploy things in the container platform.

rwmajor2:
Thanks @GrahamDumpleton, that's fair enough.  Thanks for the feedback.

FYI, below is the "rationale" from the CISP guideline:


An important security control is that of availability. Adding the HEALTHCHECK instruction to your container image ensures that the Docker engine periodically checks the running container instances against that instruction to ensure that containers are still operational.
Based on the results of the health check, the Docker engine could terminate containers which are not responding correctly, and instantiate new ones.

romainx:
Hello @rwmajor2 

It's a late answer however I hope it could be useful to someone.

# Standard HTTP probe

As far as I know there is currently no specific "health" end point available on Jupyter, it seems to be confirmed by this issue https://github.com/jupyter/notebook/issues/1857.
Since notebook are protected it's not possible to use a standard `HTTP` health check like.

```dockerfile
# I'm using wget since curl is not available
HEALTHCHECK CMD wget -q --spider http://127.0.0.1:8888 > /dev/null || exit 1
```
This will always return `HTTP 405` and the `wget` command the `8` exit status, meaning
"Server issued an error response". 

```bash
[...]

PR #1687, opened 2022-04-22
Fix HEALTHCHECK command for JupyterHub

yacchin1205:
It seems that the HEALTHCHECK command accesses the wrong endpoint when running the image in JupyterHub.
Since JupyterHub adds a prefix (obtained by `JUPYTERHUB_SERVICE_PREFIX`) to the endpoint, I assume that adding this to the HEALTHCHECK command will allow HEALTHCHECK to work correctly with JupyterHub.

This PR may also fix the following: https://github.com/jupyter/docker-stacks/issues/915#issuecomment-1104230461


mathbunnyru:
@yacchin1205 could you please tell me, if you checked this locally?

yacchin1205:
@mathbunnyru I have checked this locally and have confirmed that the built image shows a status of healthy both by running it via `docker run` and by running it on JupyterHub.

mathbunnyru:
Great, thanks for your contribution!

PR #1689, opened 2022-05-01
Switch to ubuntu:22.04

mathbunnyru:


consideRatio:
Wieee nice everything just works?! Nice!

mathbunnyru:
> Wieee nice everything just works?! Nice!

Yes, it seems to work just fine :)

consideRatio:
I think the value for end users of a switch this early, may be lower than staying a while longer. For the same reason it can make sense to not switch to py310 just after its out. But, how long to wait? Unsure.

Maybe waiting 2 months for py310 and ubuntu for example? Just guessing the time window that is reasonable to wait to find a reasonable balance between using the latest vs avoiding issues because of broken compatellity and software not supporting something modern like ubuntu 2204 and py310

mathbunnyru:
Python 3.10 was released on October 04, 2021 😃 
So, with this one, we don't have to wait any longer if we don't want and there is no reason to (everything works) :)

I'm also in favor of updating ubuntu - we have tags for people, who would like to use old versions.
Indeed, I believe, that if people do not pin any specific version, their build can be broken for many other reasons.

But updating early will give an ability for users who like modern software / modern compilers and so on use it more easily.

benz0li:
FYI: The Rocker Project aims for a 90-day grace period: https://github.com/rocker-org/rocker-versioned2/pull/424#issue-1205603358

consideRatio:
I suggest we wait for 22.04.1 to be released, 4th August, or ~90 days, that is the delay that Ubuntu seem to suggest for their 20.04 LTS users.

> Users of Ubuntu 21.10 will soon be offered an automatic upgrade to 22.04. Users of 20.04 LTS will be offered the automatic upgrade when 22.04.1 LTS is released, which is scheduled for the 4th of August.
>
>_From https://fridge.ubuntu.com/2022/04/21/ubuntu-22-04-lts-jammy-jellyfish-released/_

---

>But updating early will give an ability for users who like modern software / modern compilers and so on use it more easily.

While some people want to be at the bleeding edge, it would be helpful for them to have this upgraded to 22.04.0, but others don't care about the ubuntu version itself and just want things to work reliably, where an update may still be incompatible with some package they rely on. I think what we offer should be inclined towards reliability as this is secondary to the purpose of our images in my mind.

> we have tags for people, who would like to use old versions.

While we have tags, we have a lot of tags and many won't be having ubuntu 20.04 pinned. If they do, they won't get updates for other software during this time - only those not pinning ubuntu 20.04.

With this said, I see there as compromise no matter what we do. And given this, I think we should wait for the first patch release for ubuntu upgrades which would be 22.04.1 expected at 4th Aug.

mathbunnyru:
I get your point, let's wait before 22.04.1 is released.
I will mark this as a draft (so no one merges this PR), but will leave it open.

docs/maintaining/tasks.md
```diff
@@ -18,8 +18,9 @@ To build new images and publish them to the Docker Hub registry, do the followin
 
 ## Updating the Ubuntu Base Image
 
-`minimal-notebook` is based on the Latest LTS Ubuntu docker image.
-Other images are directly or indirectly inherited from `minimal-notebook`.
+`base-notebook` is based on the LTS Ubuntu docker image.
+We wait for the first point release of the new LTS Ubuntu before updating the version.
```
mathbunnyru, line 22: @consideRatio I agree with you, that's why I've already changed the docs to represent how we do it :)

mathbunnyru: Also, the official name for Ubuntu is `point release`, so I've decided to use it here.
https://wiki.ubuntu.com/PointReleaseProcess

mathbunnyru:
There were updates in the DockerHub 2 days ago, so I think we can merge this one.
Also, I checked two arm VMs and they tell `22.04.1` already.

PR #1690, opened 2022-05-01
Make workflows look better

mathbunnyru:


.github/workflows/docker-amd64.yml
```diff
@@ -70,16 +68,16 @@ jobs:
       - name: Install Dev Dependencies
         run: |
           python -m pip install --upgrade pip
-          make -C main install-dev-env
+          python -m pip install -r requirements-dev.txt
```
consideRatio, line 71: You can also safely avoid `python -m` in this context if you want.

PR #1691, opened 2022-05-01
Better arm image for datascience-notebook

mathbunnyru:


mathbunnyru:
`datascience-notebook` is not built in CI (because of some illegal instruction under QEMU), but this build works fine on my M1 Mac, so this image will be better in the future when we start building it for arm.

PR #1696, opened 2022-05-02
Try to fix mamba under aarch64 QEMU by setting ENV G_SLICE=always-malloc

mathbunnyru:


maresb:
I added a few commits at #1697

maresb:
I also have a batch of 9 runs here: https://github.com/maresb/docker-stacks/actions

PR #1697, opened 2022-05-02
Proposed improvements for G_SLICE

maresb:


maresb:
Assuming this `G_SLICE=always-malloc` approach works, I want to tidy things up to make them ready to merge.

My logic here is simply that I don't understand what side-effects `G_SLICE=always-malloc` has, so I'd rather not impose it anywhere downstream, thus avoiding it as an `ENV`, and only using it on `aarch64`. The drawback is that anyone downstream who builds for `aarch64` with QEMU may face breakage, but it seems more likely to me that anyone building downstream is doing so without emulation.

Secondly, I don't think we need it on `base-notebook` since I've never seen the issue arise there.

maresb:
Good news, 9/9 of my runs make it to the "Checkout Wiki Repo" stage: https://github.com/maresb/docker-stacks/actions

So I believe this PR solves the problem.

More good news: it wasn't easy, but I managed to locally reproduce the hanging build on master by running overnight.

Instructions: in `scipy-notebook/Dockerfile`, add

```Dockerfile
FROM docker.io/jupyter/minimal-notebook@sha256:1979b6252a3f0e21d1bdc4654cd7482acc038baf301251a549d5f259d8c3fc33
```

Then in Bash run

```bash
for i in {1..100}; do echo "Run: $i"; sleep 15; docker buildx build  -t build-multi-tmp-cache/scipy-notebook:latest ./scipy-notebook --no-cache --build-arg OWNER=jupyter --platform "linux/arm64"; done
```

Hopefully with this I'll be able to figure out what's going on and get it fixed upstream so that we can remove these ugly workarounds ASAP.

mathbunnyru:
@maresb I squash-merged this to master, to make it easier to revert later.
Thank you!
Hope you (with mamba team, maybe) will be able to find the solution to this.

maresb:
Thanks @mathbunnyru! And I apologize for all the trouble this caused, I was really convinced before that the problem was solved.

Issue #1530, opened 2021-11-15
Latest scipy-notebook:lab-3.1.18 docker image actually contains lab 3.2.0


carlosefr:
The latest docker image for `jupyter/scipy-notebook:lab-3.1.18` does not contain Jupyter Lab 3.1.18, but 3.2.0. This is an issue if one needs to stay on 3.1.x for some reason.

**What docker image you are using?**

`jupyter/scipy-notebook:lab-3.1.18`, specifically `jupyter/scipy-notebook@sha256:4aa1a2cc3126d490c8c6a5505c534623176327ddafdba1efbb4ac52e8dd05e81`

**What complete docker command do you run to launch the container (omitting sensitive values)?**

`docker run --rm -ti jupyter/scipy-notebook:lab-3.1.18 jupyter --version | grep lab`

**What do you expect to happen?**

The Jupyter Lab version should be 3.1.18, as per the docker tag.

**What actually happens?**

The Jupyter Lab version is 3.2.0.

mathbunnyru:
This is the corresponding build: https://github.com/jupyter/docker-stacks/runs/3915506220?check_suite_focus=true

Right now I have no ideas why this could have happened.

mathbunnyru:
I think I know the reason for this.

Unfortunately, because of our complicated build system, we end up compiling images twice.
It's easy to see that pushing takes a lot of time.
Actually, the process looks like this:
1. Build.
2. Tag.
3. Push (but unfortunately, we also build here).

That's why we might end up in a situation like this.
I have checked both minimal and base images with the same tag.
And though they should be ancestors, it's not true.

mathbunnyru:
@consideRatio any ideas, how to fix this?

I do actually have one, but I'm nor sure people will like it.
Do not build real multi-images.
Instead, use `-arm` suffix tag for arm based images.

This way, we will achieve many things simultaneously:
1. Faster build times - we can have two separate jobs, one for arm, another one for amd64.
2. Reducing complexity. Although we will have to add another type of complexity, but I think it will be much easier.
3. Images will be cached during one build process, because they won't take so much space.
4. Test arm images.
5. We can actually have correct tags and manifests for arm images as well.

consideRatio:
If we use a suffix, it should be `arm64` as `arm` often is used to represent arm32.

[...]

Issue #1402, opened 2021-07-15
Tests of arm64 based images are not performed


consideRatio:
Assuming we merge #1399, then we only run tests for the amd64 based image.

My suggestion is that we try to acquire arm64 based self-hosted GitHub runners that can assist this project and other projects in the Jupyter ecosystem. If we have that, then we can run tests and such easily.

manics:
I've been looking at free resources for self-hosted GitHub runners. The most promising looks like Oracle Cloud which has "4 Arm-based Ampere A1 cores and 24 GB of memory usable as one VM or up to 4 VMs" in the "Always Free" service
- https://www.oracle.com/uk/cloud/free/#always-free

You need a credit card to sign-up to prove your identity (and presumably to stop you signing up for multiple free accounts), but they say they won't use it if you go over the free resource quota unless you switch to a paid account:
- https://www.oracle.com/uk/cloud/free/faq.html

GitHub have a "Deploy to Oracle Cloud" button for Arm to make it easy to setup:
- https://www.google.com/search?client=firefox-b-d&q=oracle+github+runner+

There are disadvantages though:
- You're using the same VM every time, not a freshly provisioned one, so you have to be careful of what you globally install or configure
- Untrusted code will be run on the VM if it's used for testing PRs. I think you can use job and runner labels to control which job runs where, but you'd need to be very careful to avoid a security hole

trallard:
Following here and since I already mentioned it in <https://github.com/jupyter/docker-stacks/issues/1407#issuecomment-1029955372> 

We can ask for permanent free access to machines at WorksOnArm/equinix-metal-arm64-cluster - this could help us unblock `arm`-related build/test

mathbunnyru:
@trallard 

> I can help with submitting the proposal - but also know @mathbunnyru offered to help

Please, proceed and ask for arm runner, I don't have much spare time right now.
I will be glad to help if I can ❤️ 

Thank you for your ideas and working on this project 👍

consideRatio:
> If we have that, then we can run tests and such easily.

I've learned a few things that are problematic.

## Challenges with self-hosted arm runners

1. The ARM runners can't use github actions like actions/setup-python or actions/setup-node etc, because they rely on pre-installed Python/Node versions in the runner. They also make checksum checks etc which makes this not possible to workaround by providing such installations for arm.
[...]

PR #1703, opened 2022-05-12
Implement build system using self-hosted aarch64 runners, GitHub `needs` jobs feature and reusable workflows

mathbunnyru:
I've decided to implement our build system from scratch.

Ideas, that came to my mind:
- one job in a workflow only does one essential job, only one platform
- so, we will have lots of jobs, like `build-x86-minimal-notebook`, but we can easily write dependencies between them
- share everything as GitHub artifacts between these jobs
- amd64 and aarch64 are almost independent
- do not even try to merge tags between amd64 and aarch64, let's add the `aarch64-` prefix for all arm builds

Implementation details:
- I rely heavily on reusable workflows. This way there is almost zero code duplication
- The duplication is only needed in the main `docker.yml` workflow. This file can be seen as a simple config file of build dependencies
- No build system is actually needed because we will be relying on the `needs` feature of GitHub
- Heavily rely on well-made GitHub workflows `actions/upload-artifact` and `actions/download-artifact` to pass the image between jobs

TODO: 
- [x] aarch64: self-hosted runners
- [x] self-hosted runners: easy setup docs
- [x] adapt taggers to support aarch64 prefix
- [x] adapt manifest creation for the new build system
- [x] delete `buildx` Makefile parts (or event the Makefile itself)
- [x] add all images to the main workflow
- [x] fix documentation where it mentions `aarch64` or `make` commands
- [x] test, that only fresh images are used - add an empty file to the image and check, that it persists in every image built

Fix: https://github.com/jupyter/docker-stacks/issues/1407
Fix: https://github.com/jupyter/docker-stacks/issues/1530
Fix: https://github.com/jupyter/docker-stacks/issues/1402
Fix: https://github.com/jupyter/docker-stacks/issues/1401
Fix: https://github.com/jupyter/docker-stacks/issues/1203
Supersedes: https://github.com/jupyter/docker-stacks/pull/1631
Fix is no longer needed for: https://github.com/jupyter/docker-stacks/issues/1539

Upsides:
- (all the issues above)
- no need to use Makefile
- no `sudo rm -rf` to maximise space, because we will have a small amount of images in the local cache (hopefully)
- the summary of jobs shows a nice dependency graph
- aarch64 are first-class images now, there are no special hacks for building them (except for native runners and sharing state in VM)
- adding new platforms is theoretically possible now, but we will have to have native runners for them as well (in practice - do not expect new platforms)
- no need for a separate amd64 workflow
- easily adaptable for GitHub native aarch64 runners when they will be available
- no need to manually create sections to make GitHub steps look good - I think it will be easier to jump right into the error if one occurs
- QEMU aarch64 bug fix removal (no longer needed), which will allow faster `mamba` execution on multi-CPU machines

Downsides:
- Actions in self-hosted runners are run in VMs environment: this means worse security and some work to clean up the environment.
- in theory should be extremely fast, because everything that can be run in parallel, will be. The overhead of using a clean environment is small. But `docker save/load` and `uploading/downloading` artifacts is quite slow.
- maintenance self-hosted GitHub runners
- we will have an `aarch64-` prefix for every arm tag
- local experience will be a bit different from the remote one, I don't want to support `docker buildx` and multi-plaform in Makefile
- to add a new core image, more lines of changes will be required. But since we haven't added new images for many years, this is not a problem.


mathbunnyru:
This is just a beginning.

Снимок экрана 2022-05-12 в 16 42 16

mathbunnyru:
Current limitation: https://github.com/docker/buildx/issues/847

mathbunnyru:
@trallard @consideRatio @parente @romainx @maresb this is ready for review.

The only thing it doesn't support is `aarch64`.

The only thing needed to add this support is one of two things:
- Being able to `docker buildx build`, then upload the file, then download it on another job (machine) and build the imaged bases on the pulled one (https://github.com/docker/buildx/issues/847)
- Having self-hosted aarch64 runners

I prefer the second option, as it seems to be much more faster and doesn't create many hacks for this project.
But, I don't have any experience with running my own GitHub runner.
I will give it a try, when I have some inspiration :)

Please, share your ideas about this approach.
I will also publish total build time when the build finishes.

mathbunnyru:
Finally, this workflow runs in 50 minutes.
As we add aarch64 images, this time will not increase (assuming that building aarch64 images takes similar time to build).

Running in the master branch takes some time to push the images, but this time there will be no rebuilds during the push, so this should be OK.

mathbunnyru:
I would also like to personally thank @trallard.
I have learned a few tricks from your PR, like using `needs` for jobs and using docker save/load + actions upload/download.

maresb:
Hey @mathbunnyru this looks very exciting! I'm on the road and sadly don't have time (or likely ability) to help properly review this, but there is one thing which seems strange to me, where if I'm very lucky I might be able to say something helpful. (Otherwise, please forgive the noise.)

>  Being able to docker buildx build, then upload the file, then download it on another job (machine) and build the image based on the pulled one

What is "the file" in this context? Is it a Docker image or an actual file?

In case you want to push/pull images which should be used as cache, then it is possible to coax buildx into doing this, given that you set the (IMHO highly counterintuitive) correct environment variables and CLI flags. If this is the case, I might be able to help you somewhat by comparing with a few things I already have up-and-running.

Otherwise, this looks more elaborate than anything I've ever done with GH Actions, so I'll probably use it as a reference in some future project, so thanks in advance! :smile:

trallard:
I can have a look and review this on Monday/Tuesday 🙏🏽  sorry for the silence recently work has been busy

mathbunnyru:
AFAIK, using GitHub container registry in branches is not possible 😢 
You're right, these images are used for several purposes:
- for a subsequent image, as you told
- to run this image and run tests in it / create manifests

maresb:
I have a recent project using buildx and docker-compose on GitLab, so it's not something I can directly give you, but the essential command seems to be:

```bash
export DOCKER_BUILDKIT=1
docker buildx build --build-arg BUILDKIT_INLINE_CACHE=1 --cache-from from-image ...
```

Setting `BUILDKIT_INLINE_CACHE=1` as a build arg will make the resulting image usable as cache when pulled.

The `--cache-from` argument indicates which pulled images are allowed to be used as cache.

As long as I'm not missing anything, this should allow you to reuse images between different runners and such.

That's a shame about GHCR. Perhaps you could set up a separate DockerHub account? All you need is some generic registry.

mathbunnyru:
Thanks, I will try it out!
And could you tell me, how exactly do you pass this cache between jobs?
Are you using `docker save` or are you using the registry?

> That's a shame about GHCR. Perhaps you could set up a separate DockerHub account? All you need is some generic registry.

I don't think this would work.
I mean, to be able to run in branches, I will have to add the credentials, not as secrets, but to write them in code.
Our images then will not be safe, unfortunately.

I thought about using services in GitHub and run the registry as a service, but I have zero experience with that. And especially, using reuseable workflows, which is a pretty new feature of GitHub.

maresb:
> And could you tell me, how exactly do you pass this cache between jobs?

Using `docker push` and `docker pull`, with the registry as global storage. Does this make sense, at least in theory?

I'm a bit confused about how you'd like to protect the images. (Sorry I don't yet have an understanding of how secret sharing works on GitHub with multiple branches, perhaps this is why I'm confused?)

Ah, maybe I understand now... you effectively want a private registry for each branch? Would it be sufficient to use a single set of credentials, passed as secrets, but build images which have branch-specific tags?

Unfortunately I don't know much about GitHub in particular, so if GitHub isn't populating the credentials in your desired branches, I don't know how to help.

consideRatio:
> First of all, you can't use GitHub secrets in any branch other than the main branch.

I think there may be an exception for container registry access actually. At least you can make PRs read "private" registries, so maybe PRs can also push to them for specific images? To trial this, we need permissions in the jupter org itself, adding images to ghcr.io is not a project level thing but a org level thing.

I'm on vacation, will be back on monday, i can see if someone can help us trial this.

.github/workflows/docker.yml
```diff
@@ -7,6 +7,14 @@ on:
   pull_request:
     paths:
```
trallard, line 8: Maybe for the less familiar with Actions add a comment at the top like:

"Composable and reusable GitHub action workflows" or something similar + the link to the docs section

.github/workflows/docker.yml
```diff
@@ -39,84 +54,181 @@ on:
 
       - "tagging/**"
       - "tests/**"
-      - "Makefile"
       - "requirements-dev.txt"
   workflow_dispatch:
 
+# https://docs.github.com/en/actions/using-jobs/using-concurrency
 concurrency:
-  # only cancel in-progress jobs or runs for the current workflow - matches against branch & tags
```
trallard, line 47: can we keep the comment? - it helps folks to understand what is being done here if they are less familiar with GH actions

docs/contributing/tests.md
```diff
@@ -5,11 +5,7 @@ We greatly appreciate pull requests that extend the automated tests that vet the
 ## How the Tests Work
 
 A [GitHub Action workflow](https://github.com/jupyter/docker-stacks/blob/master/.github/workflows/docker.yml)
-runs the following commands against pull requests submitted to the `jupyter/docker-stacks` repository:
-
-1. `make build-all-multi` - which builds all the Docker images
-2. `make test-all` - which tests the newly created Docker images
-   This `make` command builds and then tests every docker image.
+runs tests against pull requests submitted to the `jupyter/docker-stacks` repository:
```
trallard, line 8:
```suggestion
runs tests against pull requests submitted to the `jupyter/docker-stacks` repository.
```

.github/workflows/docker-build-upload.yml
```diff
@@ -0,0 +1,55 @@
...
+jobs:
+  build-upload:
+    runs-on: ${{ inputs.runsOn }}
+    steps:
+      - name: Checkout Repo ⚡️
+        uses: actions/checkout@v3
+      - name: Create dev environment 📦
+        uses: ./.github/actions/create-dev-env
+
+      - name: Load parent built image to docker 📥
```
trallard, line 32:
```suggestion
      - name: Load parent built image to Docker 📥
```

.github/workflows/docker-build-upload.yml
```diff
@@ -0,0 +1,55 @@
...
+          platform: ${{ inputs.platform }}
+
+      - name: Build image 🛠
+        run: docker build --rm --force-rm --tag jupyter/${{ inputs.image }} ${{ inputs.image }}/
+        env:
+          DOCKER_BUILDKIT: 1
+          # Full logs for CI build
+          BUILDKIT_PROGRESS: plain
+        shell: bash
+      - name: Save image as a file 💾
```
trallard, line 46: To avoid confusion
```suggestion
      - name: Save image as a tar for later use 💾
```

.github/workflows/docker-build-upload.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: docker save jupyter/${{ inputs.image }} -o /tmp/${{ inputs.image }}-${{ inputs.platform }}.tar
+        shell: bash
+
+      - name: Upload artifact 💾
```
trallard, line 50:
```suggestion
      - name: Upload image as artifact 💾
```

.github/workflows/docker-tag-manifest-push.yml
```diff
@@ -0,0 +1,69 @@
+name: Download image artifact from GitHub artifacts, tag it and push to DockerHub
```
trallard, line 1:
```suggestion
name: Download Docker image from GitHub artifacts, tag and push it to DockerHub
```

.github/actions/load-image/action.yml
```diff
@@ -0,0 +1,29 @@
+name: "Load image"
```
trallard, line 1: Keeping consistency 

```suggestion
name: Load Docker image
```

.github/workflows/docker-tag-manifest-push.yml
```diff
@@ -0,0 +1,69 @@
...
+      matrix:
+        image: ${{ fromJson(inputs.images) }}
+
+    steps:
+      - name: Checkout Repo ⚡️
+        uses: actions/checkout@v3
+      - name: Create dev environment 📦
+        uses: ./.github/actions/create-dev-env
+
+      - name: Load image to docker 📥
```
trallard, line 33:
```suggestion
      - name: Load image to Docker 📥
```

.github/workflows/docker.yml
```diff
@@ -39,84 +54,181 @@ on:
...
-        run: make push-all-multi
+  build-amd64-base:
+    uses: ./.github/workflows/docker-build-upload.yml
+    with:
+      parentImage: ""
+      image: base-notebook
+      platform: amd64
+      runsOn: ubuntu-latest
+
+  test-amd64-base:
```
trallard, line 74: Actually in hindsight why is the build and test in separate jobs - so that you can kick - off building the child image as soon as the parent is built?

An alternative to avoid the save + load time accumulation might be to test right after build - then save it as artefact for use in other jobs. 
This might delay some downstream builds but might shave some time on the load task

trallard (commented):
Made some minor changes 

Added some comments on the logic here and there but I think this is a great improvement

trallard:
Another question - how does this then work with #1631 - I suppose a later PR once I manage to wrap this would be to add or integrate doit with all the workflows and build task to have one single way to build things moving forward

.github/workflows/docker.yml
mathbunnyru: Sure, added this info both for reuseable workflows and for composite actions.

.github/workflows/docker.yml
mathbunnyru: Yes, for example we now have a big dependency chain (which is probably the longest one):
base->minimal->scipy->pyspark->all-spark.

If we make testing inside the build step, then we will make this chain run for 5min49s longer (we will be running images).
At the same time the longest run of tests including downloading image is 5min44s.

So, there is practically no difference.

At the same time, if we add a general test, which runs X seconds, in my architecture it will add around X seconds to complete, but when we run inside the build step, it will add around 5X time.
So, current solution is better adapted for future test suite increase.

The times are based on this particular run (I hope I haven't made a mistake):
https://github.com/jupyter/docker-stacks/actions/runs/2321552950


Also, I think, having tests in the same step as a build is OK, and reduces the github graph size, so I might try it after we merge (or reject) this PR.

mathbunnyru:
> Before I forget - @mathbunnyru I love this work and the clarity gained from the parallel jobs is great ✨

Thank you ❤️

> I said I would help with the request - work has been a nightmare but will work on something and tag you all for input/reviews
> This will allow us to have the self-hosted runners - it might be a fun learning experience

Great news! I haven't started working on this actively, so I will wait for you 😂 

> As for using GHCR - I wonder if this will actually make the whole process any faster - ultimately we'd be pushing and pulling multiple images, multiple times - might need to do a comparison on saving as artifact + load time to make a decision

Agreed. I would like to hear from @consideRatio about his experience using GHCR in PRs and branches.

> Another question - how does this then work with #1631 - I suppose a later PR once I manage to wrap this would be to add or integrate doit with all the workflows and build task to have one single way to build things moving forward

To be honest, I think using doit might become an overkill.
I mean, it's a great tool to deal with dependencies.
But if this is merged, then Makefile is basically a transofrmation from `make build/base` to `docker build args...`, there is not much logic there.

@trallard thank you for your review.
I hope I haven't forgotten to answer any of your questions :)

.github/actions/create-dev-env/action.yml
```diff
@@ -0,0 +1,16 @@
+name: "Build environment"
```
mathbunnyru, line 1:
```suggestion
name: Build environment
```

mathbunnyru:
Small detail - even though we use a lot of steps, I see the consistent result in the build times.
It's around 50m each time.
Last time it was 47m 14s.
And the chain which takes the longest to run and is the actually the total time looks like this:
`build-base`->`build-minimal`->`build-base`->`build-scipy`->`build-pyspark`->`build-all-spark`->`test-all-spark`->`tag-push-tensorflow`(these jobs wait for images to be tested)->`wiki-update`.

trallard:
I do not know why I did not think of this when made my review 

But another thing to consider is we generally want to shift left (i.e test as early as possible) so following this I'd prefer having the test within the same job as the build. 
Why?
Imagine the case where we build `base` then continue building images before testing. If we defer the testing and this fails then we will have built images that will more than likely fail the tests too. 
So ideally we want to test and fail as early as possible - especially since the overall time seems unaffected

mathbunnyru:
Ok, let's try it out :)

mathbunnyru:
Used push force just a little bit, to be able to easily revert the commit.

mathbunnyru:
It took approximately the same time in total.
53m 18s

docs/maintaining/aarch64-runner.md
```diff
@@ -0,0 +1,23 @@
...
+
+For `aarch64` images, we have self-hosted VMs, provided by [Oracle OCI](https://www.oracle.com/cloud/).
+
+To setup a new runner:
+
+1. Create an Oracle `VM.Standard.A1.Flex` with _1 OCPU_ and _6 GB_ using `Ubuntu 22.04` image.
+2. Run under `root`:
+
+   ```bash
+   /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/mathbunnyru/docker-stacks/asalikhov/new_build_system/aarch64-runner/setup.sh)"
```
mathbunnyru, line 11: I will change this before merging to master

mathbunnyru:
@trallard @consideRatio @parente @romainx @maresb this is done I believe 😂 

Снимок экрана 2022-07-04 в 21 53 37

Three runners are ready to use and working well.
Sharing data between them using artifacts works as well.
Docs are updated as well.
The description is also updated - please, read it to see, what's going on.

I will see why docs linkcheck fails later (even though I added an exception).

maresb:
This is really, really cool!!!

Now that it's up-and-running, I can see and understand how it works. The execution graph is beautiful!

I managed to quickly skim the entire PR contents. I didn't have time to do a properly thorough or critical review, but my impression is that this is really high-quality work. I feel like I learned a lot by reading through this. Excellent job!!! :+1:

mathbunnyru:
Thank you, @maresb!
I will also remove QEMU G_SLICE bug fix, which doesn't look good and makes aarch64 mamba command slower.

maresb:
Ah, right, since you're not using QEMU anymore. That's great!

But you could do that in a future PR. No reason to hold up this PR if it's passing.

mathbunnyru:
Yes, you're right, I can do it later.
Still, I want to receive at least one thorough review, because this is the biggest change I've ever done in this repo and it took half a year to actually come up with this.
Until I receive this review, I might change a thing or two, while waiting 😂

mathbunnyru:
@manics I forgot to mention you initially.
Please, if you have some time, review this PR :)

mathbunnyru:
I gave a bit more thought and I'm going to reconfigure my runners - I'm going to have two runners with 2 CPU each.
Some things are working faster, when there are several cores - extraction in mamba, tests running, (maybe) docker save.
So, we should see some speed improvement for building aarch64 images.

mathbunnyru:
I changed this option to require approval for all outside collaborators.
This is a simple measure to make self-hosted runners less vulnerable to arbitrary code execution.

Снимок экрана 2022-07-05 в 11 12 00

mathbunnyru:
It was with 3 runners with 1 CPU: [1h 38m 19s](https://github.com/jupyter/docker-stacks/actions/runs/2612384390/usage)
Now it's with 2 runners with 2 CPU: [1h 21m 29s](https://github.com/jupyter/docker-stacks/actions/runs/2614134590/usage)

The best we can get with this scheme is around 1h 11m.
To do this, we will have to have 6 self-hosted aarch64 runners.

This is actually something we might want to have, because in master branch pushing to registry will also take some time and we have only 2 runners, which means we're not doing it the best way we can.

benz0li:
@mathbunnyru Have you thought of running them on a Mac Mini (M1, 2020)?

I am hosting [GitLab Runner](https://gitlab.com/gitlab-org/gitlab-runner)s in a Debian VM on a Mac mini (M1, 2020) to build my `linux/arm64/v8` images.  
ℹ️ Even the first generation Mx machines outperform everything else out there.

mathbunnyru:
> @mathbunnyru Have you thought of running them on a Mac Mini (M1, 2020)?
> 
> I am hosting [GitLab Runner](https://gitlab.com/gitlab-org/gitlab-runner)s in a Debian VM on a Mac mini (M1, 2020) to build my `linux/arm64/v8` images. ℹ️ Even the first generation Mx machines outperform everything else out there.

I thought about running everything on Mac, but there is no free M1 Cloud Mac, as far as I know.
I might be wrong here.
You're using your personal Mac Mini, right?

Currently, 2 Oracle VMs with 2 CPUs each will be free for me.

benz0li:
> I thought about running aarch64 builds on M1 Mac, but there is no free M1 Cloud Mac, as far as I know.
> I might be wrong here.

No, you are not.

> You're using your personal Mac Mini, right?

Yes. It was a $900 investment.

mathbunnyru:
To be honest, I don't think it would be a good idea to use personal equipment in a very popular open-source project.
I mean, it's totally fine for personal projects and I would have a setup as you have for my personal projects (if I needed to).

But if something happens to the person who owns the hardware or he decides to stop providing it as a service or the hardware breaks, it would be quite difficult to make this project work again.
In my case, I added a complete guide to set up a new runner. If I decide to stop using Oracle cloud, then someone has to follow the instructions and it should work just fine. I did nothing to my VMs, except the things I documented. Moreover, I recreated everything from scratch after the last change, so it should be fine.
Another possibility is that Oracle stops providing these runners, but I hope it's not gonna happen and there are other alternatives (Amazon, for example, but not as good in a free tier).

mathbunnyru:
There were some reviews in the past, so I'm going to merge this when this passes.
Now, I have some time to fix with some bugs, if there will be any, in master branch.

mathbunnyru:
I was able to enable building `datascience-notebook` under aarch64.
And here comes the best part:
even in master with full `datascience-notebook` and pushing to DockerHub build time is [1h 23m 22s](https://github.com/jupyter/docker-stacks/actions/runs/2620522730/usage)

trallard:
Wow @mathbunnyru I have been away as I am stretched with work but kudos on this major refactoring. You've done an incredible work 🙂 thank you

Issue #1698, opened 2022-05-06
[BUG] - Healthcheck fails when using self-signed SSL certificate


remisalmon:
### What docker image(s) are you using?

base-notebook

### OS system and architecture running docker image

Amazon Linux 2

### What Docker command are you running?

```
docker run --rm \
           --detach \
           --publish 8888:8888 \
           --name jupyter \
           --volume ${HOME}/work:/home/jovyan/work \
           --env GEN_CERT=yes \
           --env RESTARTABLE=yes \
           jupyter/base-notebook:latest
```

### How to Reproduce the problem?

1. Run the docker command above
2. Run `docker ps`, containers shows as `(unhealthy)`
3. Run `wget -O- --no-verbose --tries=1 --no-check-certificate http://localhost:8888/api`
4. Run `wget -O- --no-verbose --tries=1 --no-check-certificate http://localhost:8888/api`

### Command output

```bash session
Step 2:

CONTAINER ID   IMAGE                                     COMMAND                  CREATED      STATUS                  PORTS                                                                                      NAMES
...   jupyter/base-notebook:latest              "tini -g -- start-no…"   3 days ago   Up 3 days (unhealthy)   0.0.0.0:8888->8888/tcp, :::8888->8888/tcp                                                  jupyter
```

Step 3:
```
No data received.
```

Step 4:
```
ERROR: cannot verify localhost's certificate, issued by ‘/C=XX/ST=XX/L=XX/O=generated/CN=generated’:
  Self-signed certificate encountered.
    ERROR: certificate common name ‘generated’ doesn't match requested host name ‘localhost’.
To connect to localhost insecurely, use `--no-check-certificate'.
```
```


### Expected behavior

Healthchecks should work when using self-signed SSL certificates.

### Actual behavior

When running the base notebook image with `GEN_CERT=yes`, healthchecks are failing in Docker because the check is targeting http://localhost (see https://github.com/jupyter/docker-stacks/blob/60b4e4b298ccccf1da88fb030a8f48778e3d40a7/base-notebook/Dockerfile#L168).

### Anything else?

_No response_

bgalvao:
ah nice, this is what I was looking for 😄 . I think this is related to the issue I'm having and [posted on Discourse](https://discourse.jupyter.org/t/newer-scipy-notebook-version-3-3-4-with-gen-cert-yes-creates-ssl-errors-http-request-does-not-work-with-traefik-reverse-proxy/14148).

[...]

PR #1704, opened 2022-05-13
Fix HEALTHCHECK when GEN_CERT=yes

mathbunnyru:
Fix: https://github.com/jupyter/docker-stacks/issues/1698


mathbunnyru:
I checked this one locally, both using GEN_CERT and without - in both cases the container reports to be healthy.

PR #1711, opened 2022-05-26
Bump Python 3.9 → 3.10

### LINKED ISSUE #1688: [ENH] - Add a jupyter/minimal-notebook:python-3.10.x image

minimal-notebook

### What changes are you proposing?

I would like to see a `jupyter/minimal-notebook:python-3.10.x` image.

They can use a newer Python version

### Anything else?

_No response_

maresb:
## Describe your changes

Bump Python 3.9 → 3.10

## Issue ticket if applicable

Closes #1688 

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -85,7 +85,7 @@ RUN echo "auth requisite pam_deny.so" >> /etc/pam.d/su && \
 USER ${NB_UID}
 
 # Pin python version here, or set it to "default"
-ARG PYTHON_VERSION=3.9
+ARG PYTHON_VERSION=3.10
 
 # Setup work directory for backward-compatibility
 RUN mkdir "/home/${NB_USER}/work" && \
--- a/tests/base-notebook/test_python.py
+++ b/tests/base-notebook/test_python.py
@@ -10,7 +10,7 @@ LOGGER = logging.getLogger(__name__)
 
 
 def test_python_version(
-    container: TrackedContainer, python_next_version: str = "3.10"
+    container: TrackedContainer, python_next_version: str = "3.11"
 ) -> None:
     """Check that python version is lower than the next version"""
     LOGGER.info(f"Checking that python version is lower than {python_next_version}")


mathbunnyru:
I think we have a test, that checks python version.
Probably need to fix it :)

mathbunnyru:
And, because of our current build system, it's technically impossible to fix this test.

maresb:
Ah, ok. I'm finally back from traveling, so hopefully I can manage to wrap my head around #1703. I'd really like to understand all the clever stuff you're doing there! :smile:

mathbunnyru:
Wow, the test seems to pass. It shouldn’t 😂

mathbunnyru:
@maresb I fixed this test in master. Please, merge it and it should fail now :)

maresb:
@mathbunnyru ok, good, now it's failing as expected.

mathbunnyru:
@maresb this is just a joke 😂

trade_offer

maresb:
Ooh, very tempting! :joy: 

Honestly this weekend is unfortunately looking bleak for me accomplishing any fun work. I'm on the road next week, and I have a lot to prepare.

I'll see if I can download it and have a look for when I'm offline...

mathbunnyru:
No pressure here, @maresb :)
Many of us are doing this for fun, and you can take as long as you need to :)

maresb:
Ah, and I should be clear, I have not tested this myself, so there could be some package further down the chain which is incompatible with 3.10. (But since it's been over ½ year, I think most reputable packages should have adapted by now.)

mathbunnyru:
That's a good point. That's why our current CI builds something strange, then tests it, then incorrectly tags it, and done.
At the same time, the `amd64` workflow should catch problems like this, because it builds what expected and tests it well.

maresb:
Oh, that's great to know! I'm glad that `amd64` tests are working properly.

On the other hand, the `arm64` migration on Conda-Forge is not 100%, so there's a significant probability that something on `amd64` is good on Python 3.10 while `arm64` isn't.

bjornjorgensen:
INFO     test_python:test_python.py:16 Checking that python version is lower than 3.10

Is this one of our tests?

bjornjorgensen:
I open https://github.com/jupyter/docker-stacks/pull/1714 where I have changed the test next python version. I did't run on my github, soo I have open it her.

mathbunnyru:
This will work now, but I will have to rebuild many times, waiting each time for the workflow to complete, to actually update all the images.
I will merge when tests pass.

mathbunnyru:
The test we have for python version is checking the exact match of python version, so the tests should pass.

PR #1713, opened 2022-05-27
Try to use ghcr.io

mathbunnyru:
## Describe your changes

This is the same as https://github.com/jupyter/docker-stacks/pull/1703, but uses ghcr.io to pass images between different jobs.


mathbunnyru:
`denied: installation not allowed to Create organization package`

mathbunnyru:
I think it's actually possible to use ghcr.io in branches, but we will need to set some permissions.
https://github.community/t/cant-authenticate-with-actions-token-for-pr-event/170752/24

Example that it works:
https://github.com/jcansdale-test/docker-ghcr-pull-request-tests/pull/1

consideRatio:
@mathbunnyru I have permissions in the jupyter org to "create packages" or define docker images for us. Is there images you want me to define? They would be under `ghcr.io/jupyter/<image name>`.

mathbunnyru:
@consideRatio yes, please, do.

https://github.com/jupyter/docker-stacks/pull/1713/files#diff-380746580c95ded40154f8b5dacf213d5697aa17a0283e52be8377fbea76159cR26

I used `ghcr.io/jupyter/${{ inputs.image }}-${{ inputs.platform }}` name to tag the image while pushing to ghcr.io.
So let's create:

```
base-notebook-amd64
minimal-notebook-amd64
scipy-notebook-amd64
r-notebook-amd64
tensorflow-notebook-amd64
datascience-notebook-amd64
pyspark-notebook-amd64
all-spark-notebook-amd64
```

And the same with `aarch64` suffix, though they won't be used right now.

mathbunnyru:
@consideRatio after you create these images, could you please tell me how security works for ghcr.io images in PR branches?

I mean, what would prevent someone from creating PR, which writes some nasty stuff to the images after they are built by us in `master` branch but before they are pushed from ghcr.io to dockerhub.
Do we need to prevent this by saving hashes and then verifying them or something like that?

consideRatio:
For reference, this is how one goes about creating new packages:
https://github.com/dask/dask-gateway/issues/484#issue-1151853633

For reference, this is the permissions of a secrets.github_token that is possible:
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

---

I think I've been incorrect... I thought since you could grant read permissions to a PR, to read a private image, you could have write permissions also. That doesn't seem to be the case.

If we could use PRs from forks and have permissions to push to a registry, that would at best be something we could use for intermediary processing though, so we would need to have separate image repositories for those with wide permissions and trusted-only permissions.

To conclude, I've messed up thinking we can push to ghcr.io from a workflow triggered by a pull request rather than a push to a branch. What to do?

I think we are forced to use artifacts rather than ghcr.io as intermittent storage when pull requests are involved, rather than ghcr.io's container registry.

mathbunnyru:
Thank you for the explanation. I’m closing this PR then and I will stick to the original PR.

consideRatio:
@mathbunnyru I've created 8x2 images in https://github.com/orgs/jupyter/packages now, and I think I granted the same permissions to all of them.

1. GitHub actions can get write access to them from this repo via permissions.packages=write if the github_token isn't constrainted by it being a PR
2. Members of this repo has access to the packages based on the github repo permissions they have.

As an admin, I think you should be able to declare them to be "public" so that the images can be seen by everyone. For now, they are private until you go to package settings -> change visibility (in the danger zone of settings)

mathbunnyru:
Ok, I will try it out!

.github/workflows/docker.yml
```diff
@@ -65,6 +65,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
+  packages: write
+
```
consideRatio, line 70: To be clear, unless this workflow triggers from a `push` event or `workflow_dispatch` event, etc rather than a `pull_request` event, `write` is not a permissions granted to the `github_token`.

mathbunnyru: I think we can't have what we really need.
https://github.community/t/cant-authenticate-with-actions-token-for-pr-event/170752/31

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

I think we can make it work for PRs made from `jupyter/docker-stacks` repo (not forks), but we can't have it for all the people. Or we will use some bad way with PAT tokens and basically any PR will be able to change our packages.

PR #1717, opened 2022-05-31
Compare major.minor python version to be expected one

mathbunnyru:
## Describe your changes

We pin `major.minor` python version in the base Dockerfile.
So it makes sense to check this version stays the same.


mathbunnyru:
Merging this one, seems to be a simple change.

mathbunnyru:
Merging this one, seems to be a simple change.

PR #1725, opened 2022-06-14
Adding list of groups to `${NB_USER}`

IslamAlam:
## Describe your changes
A useful feature for adding a list of groups to `${NB_USER}` in the case of needing to read multiple volumes with different access rights. For instance, it is handy to add a user to groups given from ActiveDirectory.

### DIFF
--- a/base-notebook/start.sh
+++ b/base-notebook/start.sh
@@ -75,6 +75,7 @@ if [ "$(id -u)" == 0 ] ; then
     # - NB_UID: the desired user id
     # - NB_GID: a group id we want our user to belong to
     # - NB_GROUP: a group name we want for the group
+    # - NB_UNPRIVILEGED_GROUPS: list of groups to add for NB_USER as "104(input),125(render),124(kvm),107(messagebus)"
     # - GRANT_SUDO: a boolean ("1" or "yes") to grant the user sudo rights
     # - CHOWN_HOME: a boolean ("1" or "yes") to chown the user's home folder
     # - CHOWN_EXTRA: a comma separated list of paths to chown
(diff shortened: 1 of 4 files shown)


mathbunnyru:
@IslamAlam the reason for that is because we use not freshly built images to use as a base image, but end up pulling old images from dockerhub.

base-notebook/start.sh
```diff
@@ -104,6 +105,23 @@ if [ "$(id -u)" == 0 ] ; then
         useradd --home "/home/${NB_USER}" --uid "${NB_UID}" --gid "${NB_GID}" --groups 100 --no-log-init "${NB_USER}"
     fi
 
+    # Create additional groups for NB_USER
+    _IFS=${IFS}; IFS=,
+    for gid in ${NB_UNPRIVILEGED_GROUPS?}; do
+        g_id=$(echo "${gid}" | awk -F '[()]' '{print $1}')
+        g_name=$(echo "${gid}" | awk -F '[()]' '{print $2}')
+        # make sure group name, no longer than 32 characters
+        g_name="${g_name:0:32}"
```
mathbunnyru, line 114: I don't think we should manipulate user's input this way.
If they use long group name and it fails when starting the container, it's actually better, than fixing the group name and making them believe everything is fine.

base-notebook/start.sh
```diff
@@ -104,6 +105,23 @@ if [ "$(id -u)" == 0 ] ; then
...
+
+        if ! getent group "${g_id:?}" >/dev/null 2>&1; then
+            _log "create group with: groupadd --force --gid ${g_id:?} ${g_name:-"g_${g_id:?}"} "
+            eval 'groupadd --force --gid "${g_id:?}" "${g_name:-"g_${g_id:?}"}" '
+            _log "Add user to group: usermod -a -G ${g_name:-"g_${g_id:?}"} ${NB_USER}"
+            eval 'usermod -a -G "${g_name:-"g_${g_id:?}"}" ${NB_USER}'
```
mathbunnyru, line 120: Why are we calling `eval` instead of calling these commands directly?

mathbunnyru (changes requested):
To be honest, I'm not sure we should add more stuff to our start scripts.
I mean, they are already over-complicated, not very well tested.

Have you tried using [Startup hooks](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#startup-hooks) for your purposes?

mathbunnyru:
> @IslamAlam the reason for that is because we use not freshly built images to use as a base image, but end up pulling old images from dockerhub.

This is merged in master. Please, merge `master` branch to your branch, to build images correctly.

consideRatio:
I think that this may be stretching the functionality that makes sense for jupyter/docker-stacks to maintain long term beyond what is reasonable.

I think [startup hooks](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#startup-hooks) is the way to go for something more advanced like this.

mathbunnyru:
Closing this PR. Please, if you think this is relevant for many users, fix the issues, reopen PR and let's wait for other people to ask for this feature.

PR #1731, opened 2022-06-29
Update passwd function

irvcaza:
## Describe your changes

The use of `IPython.lib.passwd()` does not match with [Jupyter server documentation](https://jupyter-server.readthedocs.io/en/latest/operators/public-server.html#preparing-a-hashed-password) and may lead to confusion on how to import/use it.

### DIFF
--- a/docs/using/common.md
+++ b/docs/using/common.md
@@ -16,7 +16,7 @@ You can pass [Jupyter server options](https://jupyter-server.readthedocs.io/en/l
 
    ```bash
    docker run -it --rm -p 8888:8888 jupyter/base-notebook \
-       start-notebook.sh --NotebookApp.password='sha1:7cca89c48283:e3c1f9fbc06d1d2aa59555dfd5beed925e30dd2c'
+       start-notebook.sh --NotebookApp.password='argon2:$argon2id$v=19$m=10240,t=10,p=8$JdAN3fe9J45NvK/EPuGCvA$O/tbxglbwRpOFuBNTYrymAEH6370Q2z+eS1eF4GM6Do'
    ```
 
 2. To set the [base URL](https://jupyter-server.readthedocs.io/en/latest/operators/public-server.html#running-the-notebook-with-a-customized-url-prefix) of the notebook server, you can run the following:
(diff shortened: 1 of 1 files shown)


docs/using/common.md
```diff
@@ -10,7 +10,8 @@ This page describes the options supported by the startup script and how to bypas
 
 You can pass [Jupyter server options](https://jupyter-server.readthedocs.io/en/latest/operators/public-server.html) to the `start-notebook.sh` script when launching the container.
 
-1. For example, to secure the Notebook server with a custom password hashed using `IPython.lib.passwd()` instead of the default token,
+1. For example, to secure the Notebook server with a [custom password](https://jupyter-server.readthedocs.io/en/latest/operators/public-server.html#preparing-a-hashed-password)
+   hashed using `notebook.auth.security.passwd()` instead of the default token,
```
consideRatio, line 14: Ah, is it even so that we should now use `jupyter_server` instead here? `notebook` is being phased out in favor of `jupyter_server` and have lots of the same capabilities.

irvcaza: You are right! I was almost sure I got that function from Jupyter-server's docs.

consideRatio: Ah maybe a fix for that projects documentation as well then? :) Thank you for your work in the jupyter ecosystem @irvcaza!!!

consideRatio: Atm you retain the notebook reference, should it be retained or switched to jupytet_server?

irvcaza: Let's change it to `jupyter_server `and do the proper changes to `jupyter_server` documentation.

Thank you for your support @consideRatio !

consideRatio: from jupyter_server.auth import passwd

Should it be with .security or not here?

irvcaza: Great observation! both work (I tested them)

`jupyter_server.auth.security` works because the function is defined in [that file](https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/auth/security.py#L21) 

`jupyter_server.auth` works because (_If I understand corectly_) it is imported in the [`__init__.py`](https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/auth/__init__.py#L4)

I prefer the second one just because it matches jupyter_server's [documentation](https://jupyter-server.readthedocs.io/en/latest/operators/public-server.html#preparing-a-hashed-password)

But if there is a stronger reason to change it to `jupyter_server.auth.security`, I have no problem with that.

mathbunnyru: I think it's a good idea to match the docs 👍

mathbunnyru:
Merging this one 👍

PR #1727, opened 2022-06-16
Upgrading Spark->3.3,Hadoop->3,Scala->2.13,Java->17

Bidek56:
## Describe your changes
Upgrading:
  - Spark -> 3.3
  - Hadoop -> 3.3
  - Scala -> 2.13
  - Java -> 17

Breaking change: [spylon-kernel ](https://github.com/vericast/spylon-kernel) not longer works since the package is no longer maintained. 
For anyone looking for a Scala kernel, I would suggest looking at [Almond](https://github.com/almond-sh/almond)

## Checklist (especially for first-time contributors)

- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests
- [x] I will try not to use force-push to make the review process easier for reviewers
- [x] I have updated the documentation for significant changes

PR #1733, opened 2022-07-02
Upgrades to Spark 3.3.0

rhazegh:
Bumps spark from 3.2.1 to 3.3.0

See https://archive.apache.org/dist/spark/spark-3.3.0/ for `spark_checksum` and `hadoop_version`.


rhazegh:
It seems there is already a pull request for this. https://github.com/jupyter/docker-stacks/pull/1727

PR #1745, opened 2022-07-09
Snowpark notebook

john77eipe:
## Adding snowpark supported dockerfile

Adding new kid on the block - snowpark for snowflake

- [ x] I have performed a self-review of my code
- [ x] If it is a core feature, I have added thorough tests
- [ x] I will try not to use force-push to make the review process easier for reviewers
- [x ] I have updated the documentation for significant changes


mathbunnyru:
Hello, @john77eipe!

I think it would be a maintenance burden for us to have an image for every python package we did not include our images.
I suggest you maintain your own image and have it on top of one of our own images.

PR #1749, opened 2022-07-17
Up docs dependencies and move requirements to docs folder

mathbunnyru:


.readthedocs.yaml
```diff
@@ -27,4 +27,4 @@ sphinx:
 # Optionally declare the Python requirements required to build your docs
 python:
   install:
-    - requirements: requirements-docs.txt
+    - requirements: docs/requirements.txt
```
mathbunnyru, line 30: Docs folder is self-contained, so it's easier to store this file there.

PR #1770, opened 2022-08-10
Run github pipelines in ubuntu 22.04

mathbunnyru:
## Describe your changes

Let's update and pin Ubuntu versions for GitHub runners as well.

https://github.blog/changelog/2022-08-09-github-actions-ubuntu-22-04-is-now-generally-available-on-github-hosted-runners/

### DIFF
--- a/.github/workflows/docker-wiki-update.yml
+++ b/.github/workflows/docker-wiki-update.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   tag-push:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     steps:
       - name: Checkout Repo ⚡️
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -79,7 +79,7 @@ jobs:
       parentImage: ""
       image: base-notebook
       platform: amd64
-      runsOn: ubuntu-latest
+      runsOn: ubuntu-22.04
 
   aarch64-minimal:
     needs: [aarch64-base]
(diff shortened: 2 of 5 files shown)


bjornjorgensen:
-1 for this.

"Otherwise, our recommendation is to use ubuntu-latest, which currently utilizes Ubuntu 20.04 but will begin running on Ubuntu 22.04 in the near future. This will ensure your workflows are always using a recent OS and removes the need to constantly update workflow files with image versions."

Lets use `ubuntu-latest`

mathbunnyru:
Well, my motivation is that arm runners have ubuntu 22.04, that's why to make everything in sync, it makes sense to me to use the same Ubuntu version on GitHub runners as well.

bjornjorgensen:
ok, but we have it as github recommends. So I don't find any reasons to change that. 

We will get 22.04 when github recommends that.

mathbunnyru:
I'm ok keeping it `ubuntu-latest`.

bjornjorgensen:
[GitHub Actions: Larger Runners using Ubuntu-latest label will now use Ubuntu-22.04](https://github.blog/changelog/2022-12-01-github-actions-larger-runners-using-ubuntu-latest-label-will-now-use-ubuntu-22-04/)

mathbunnyru:
> @mathbunnyru
> [GitHub Actions: Larger Runners using Ubuntu-latest label will now use Ubuntu-22.04](https://github.blog/changelog/2022-12-01-github-actions-larger-runners-using-ubuntu-latest-label-will-now-use-ubuntu-22-04/)

Got it, thanks.

PR #1776, opened 2022-08-22
Improve performance of fix-permissions script

peterisr:
## Describe your changes

Using the `-exec command {} +` (instead of `;`)  action of `find`, we can improve the performance of fix-permissions script significantly. The difference between the two is that the former (`+`) invokes a single process and passes all the files to it as arguments, whereas the latter (`;`) invokes one process per matched file resulting in thousands of `chmod` runs per build.

My benchmarks show about a 30 second improvement in steps that involve fix-permissions for the `build-all` target of `make`. More importantly for me, user defined Docker images, which use the fix-permissions script, will also show perf improvements.

## Issue ticket if applicable

n/a

### DIFF
--- a/base-notebook/fix-permissions
+++ b/base-notebook/fix-permissions
@@ -23,13 +23,13 @@ for d in "$@"; do
             -group "${NB_GID}" \
             -a -perm -g+rwX \
         \) \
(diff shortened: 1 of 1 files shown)


mathbunnyru:
Thank you, merging this one 👍

Issue #1781, opened 2022-09-06
[ENH] - Exposing the Spark UI port


galimba39:
### What docker image(s) is this feature applicable to?

pyspark-notebook

### What changes are you proposing?

It would be useful for pyspark/spark users to expose the SparkUI port 4040 (please refer to this [https://spark.apache.org/docs/latest/monitoring.html](url)).

### How does this affect the user?

- Improve the overall Pyspark experience as a user

bjornjorgensen:
https://jupyter-docker-stacks.readthedocs.io/en/latest/using/specifics.html#apache-spark

galimba39:
I am not sure I have understood correctly your answer @bjornjorgensen but I'll try to expand my reasoning below.
I was thinking to expose it to make the interaction with the Docker Desktop more confortable.
Something like this

![image](https://user-images.githubusercontent.com/42516847/188617533-5807a86b-c501-44b4-8f9d-5c9e11f0ca49.png)

comes in handy when you plan to add different volumes (one for data, one for code, one for spark local storage etc) and you prefer to use UI to handle both ports and external volumes.

mathbunnyru:
@galimba39 what does this window currently show for `pyspark-notebook` image?

mathbunnyru:
I guess we just have to add `EXPOSE 4040` to the spark images to make sure docker knows which ports we are exposing.

bjornjorgensen:
`docker run -d -p 8888:8888 -p 4040:4040 -p 4041:4041 jupyter/pyspark-notebook` work and it works with docker-compose. I don't think its a good options to expose 4040 as default for everyone. Its a page without password and everyone can stop running spark seasons thru the web page.

galimba39:
@mathbunnyru it currently shows port 8888 only. Just adding EXPOSE 4040 return the image in my previous comment.

@bjornjorgensen 
I'm not an expert in Docker but if I have correctly understood from [this](https://docs.docker.com/config/containers/container-networking/#published-ports), only published ports can be reached from outside and exposed ports shouldn't raise a risk until published.

bjornjorgensen:
hmm.. have been reading [this one..](https://www.cloudbees.com/blog/docker-expose-port-what-it-means-and-what-it-doesnt-mean)

The strange thing her is that I build spark master daily and start it with a docker-compose where I expose - "4040:4040"

But in the docker file I don't have 4040 only 
# Should match the service 
[...]

PR #1783, opened 2022-09-08
Add EXPOSE

bjornjorgensen:
## Describe your changes
Add `EXPOSE` 4040
## Issue ticket if applicable

[ENH] - Exposing the Spark UI port #1781


mathbunnyru:
> shall we add `EXPOSE` to all others dockerfiles?

I don't think so (at least for pyspark). This is the only image where we install pyspark.

mathbunnyru (approved):

bjornjorgensen:
Ok. how about all-spark-notebook then?
https://github.com/jupyter/docker-stacks/blob/main/all-spark-notebook/Dockerfile

mathbunnyru:
> Ok. how about all-spark-notebook then?
> https://github.com/jupyter/docker-stacks/blob/main/all-spark-notebook/Dockerfile

We don't need it there.
Exposed ports are inherited and `all-spark` is based on `pyspark`.

I've checked that these ports are inherited by using `docker image inspect` on both `base` and `minimal` images and port `8888/tcp` is exposed in both these images (even though we explicitly declare it only in the base image).

Issue #1737, opened 2022-07-05
[ENH] - Create multi-platform images


mathbunnyru:
### What docker image(s) is this feature applicable to?

all-spark-notebook, base-notebook, datascience-notebook, minimal-notebook, pyspark-notebook, r-notebook, scipy-notebook, tensorflow-notebook

### What changes are you proposing?

It would be great to merge `amd64` image with `aarch64` image with the same tag into a single multi-platform image

### How does this affect the user?

It would allow users not to specify `aarch64-` prefix.
Also, it will make `make build/<somestack>-notebook` work as expected, because it will always use correct parent image.

### Anything else?

Implementation detail - this should probably be done in a separate reusable workflow, after pushing all images to DockerHub.
It would take a look which tags were pushed for `amd64` and which for `aarch64` and it would merge these tags.

nathancarter:
I would love to see this.  Just to give an example of why it's important:

Just got an M1 Mac and tried to re-use a VS Code devcontainer I have that's based on jupyter/datascience-notebook.  A build script in one of my projects, which used to run just fine, now runs out of memory, because the AMD devcontainer is being run on an ARM chip and thus has to go through some semi-supported emulation business that I don't fully understand, except to note that out-of-memory errors are documented as something to be expected.  I have to physically go to my PC to run that build script...kind of the opposite of the point of Docker.  While I guess I could add the `aarch64-` prefix to the `FROM` line of my Dockerfile, that will make the project work on my Mac but not my PC.

In other words, things got worse when I bought a better computer.  And there's no workaround.  I'm not complaining; I'm just trying to clarify the significance of @mathbunnyru's issue.  Thanks for opening it!

mathbunnyru:
Thank you for sharing your use case.

I think right now you can use your have Dockerfile loke `FROM $BASE_CONTAINER` and have this `BASE_CONTAINER` set to different things for Mac and PC, if it's possible.
I know it's not ideal, but if it works, it's not that bad.

msemelman:
Not sure if it completely fits current building setup, but maybe buildx could be used to simplify this:

https://docs.docker.com/build/buildx/

mathbunnyru:
> Not sure if it completely fits current building setup, but maybe buildx could be used to simplify this:
> 
> https://docs.docker.com/build/buildx/

Unfortunately, `docker buildx` didn't work for us and created many issues and they were leading to bugs and very unpleasant things.
So far, current solution only has this issue.
[...]

PR #1789, opened 2022-09-20
Merge tags and create multi-platform images

mathbunnyru:
## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/1737


.github/workflows/docker-merge-tags.yml
```diff
@@ -0,0 +1,57 @@
...
+          path: /tmp/tags/
+
+      - name: Login to Docker Hub 🔐
+        if: github.ref == 'refs/heads/main' || github.event_name == 'schedule'
+        uses: docker/login-action@49ed152c8eca782a232dede0303416e8f356c37b # dependabot updates to latest release
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Merge tags for the images
```
mathbunnyru, line 54: Add icon

README.md
```diff
@@ -115,6 +115,6 @@ This change is tracked in the issue [#1217](https://github.com/jupyter/docker-st
 
 ## CPU Architectures
 
-- We publish containers for both `amd64` (`x86_64`) and `aarch64` platforms, except for `tensorflow-notebook`, which only supports `amd64` for now
+- We publish containers for both `x86_64` and `aarch64` platforms, except for `tensorflow-notebook`, which only supports `x86_64` for now
 - We do not create multi-platform images
```
mathbunnyru, line 119: Fix the docs

mathbunnyru:
There is a 10-15 minutes improvement in branches, but the main branch might be slow because `docker manifest` will push many times.
I will squash merge this and then, if it doesn't work, I will disable merging tags and fix it later.

mathbunnyru:
It actually works and only add 2-3 minutes to the CI.
But this process seems to be stuck for two images.
I will cancel and try to investigate what's happening.

mathbunnyru:
This is how it looks.
Cool thing is that I was able to make it work properly even in cases, where different versions of software are installed in x86_64 and aarch64 images.

Screenshot 2022-09-21 at 11 50 09

mathbunnyru:
> It actually works and only adds 2-3 minutes to the CI.
> But this process seems to be stuck for two images.
> I will cancel and try to investigate what's happening.

It seems to work now, but I expect to see failures sometimes (unfortunately, docker has some issues).
And thee build time in main branch is [1h 13m 24s](https://github.com/jupyter/docker-stacks/actions/runs/3096055828/usage).
It's faster, than it was, because I'm running more stuff in parallel and use aarch64 runners only to build/test/create tags and for everything else I use native Github x86_64 runners.

Overall, I think our build system now works 100% as intended and I'm happy about it.

PR #1793, opened 2022-09-29
Build images with ubuntu-20.04 and python-3.7

mathbunnyru:
## Describe your changes

I think that our build system works well and doesn't have any known bugs and it finally supports both `x86_64` and `aarch64` and multi-arch images for these platforms properly.
I also think, some users still want to use old images (not all people can switch to the new ubuntu or python version quickly).

So, I had an idea to rebuild the images in this order:
- First, ubuntu-20.04 with python 3.7, 3.8, 3.9, 3.10
- Then, ubuntu-22.04 with python 3.7, 3.8, 3.9, 3.10 as well

After this, people will be able to choose the old version(s) if they want to.
And tags will work correctly for these images as well.
*This doesn't mean, that we will support these versions in the future.*

After all these rebuilds, the latest versions will be the same as they currently are.
But in the process, they will change and people who don't use fixed tags might get an unexpected downgrade.
I expect these rebuilds to be done in less than 2 days.

Recently, I also introduced a new tag `python-major.minor` (without patch). It makes sense because usually, people update the patch versions easily.

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -1,9 +1,9 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-# Ubuntu 22.04 (jammy)
-# https://hub.docker.com/_/ubuntu/tags?page=1&name=jammy
-ARG ROOT_CONTAINER=ubuntu:22.04
+# Ubuntu 20.04 (focal)
+# https://hub.docker.com/_/ubuntu/tags?page=1&name=focal
+ARG ROOT_CONTAINER=ubuntu:20.04
 
 FROM $ROOT_CONTAINER
 
@@ -85,7 +85,7 @@ RUN echo "auth requisite pam_deny.so" >> /etc/pam.d/su && \
 USER ${NB_UID}
 
 # Pin python version here, or set it to "default"
-ARG PYTHON_VERSION=3.10
+ARG PYTHON_VERSION=3.7
 
 # Setup work directory for backward-compatibility
 RUN mkdir "/home/${NB_USER}/work" && \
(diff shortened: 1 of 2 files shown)


mathbunnyru:
Please, tell me what you think about this @consideRatio

mathbunnyru:
Even the oldest version is building fine. That's great because we won't probably have to change any code/image.

consideRatio:
My understanding is that you are proposing a one off build with older versions of ubuntu and python by temporarily adjusting the CI sytem.

I'm okay with it but not clearly :+1: or :-1:. My main concern is that we will hint that we build py3.7 etc images regularly by publishing a fresh tag of py37 etc now. This can cause people to think "oh nice they publish these versions regularly as well!" and then show up asking why we don't actually do that etc.

I'm :+1: on the judgement call you make!

mathbunnyru:
Yes, I propose to do it only once. I will try do it relatively quickly. Then I will also add the information about the built images to the docs (I guess a table with 7 entries will do it well).

If we had github arm runners, I guess we would be able to always build absolutely everything.
Because of overall parallelism, we would be able to build it in the same amount of time.
But it doesn't seem to be possible now with only 2 arm self-hosted runners.
(Also, we should be careful with tags in this case, but that's not difficult, I suppose).

mathbunnyru:
Thanks @consideRatio!
Yes, everything seems to work really well.
I had to restart things once as far as I remember because of network issues (and I don't think we can and should do anything about it).

There is also a cool thing about structuring our build system in many jobs: if anything fails, we can rerun only failed jobs and it saves so much time (usually network related fail happens during push to dockerhub, so on the last stages). And even if we restart, we get consistent results (no duplication of anything or missing something). I like this a lot as a maintainer 😂

Issue #1796, opened 2022-10-03
[ENH] - Enable clipboard


adrienpacifico:
### What docker image(s) is this feature applicable to?

datascience-notebook

### What changes are you proposing?

This code does not work with docker images:

```python
import pandas as pd
df = pd.read_clipboard()
```
It throws a `PyperclipException`.

I do not know if it is possible, but it would be really nice to have this working with docker images.

### How does this affect the user?

- The use will be able to use any library that needs to access to the paperclip.

mathbunnyru:
Hi @adrienpacifico!
I guess I found the solution here: https://github.com/asweigart/pyperclip/issues/161#issuecomment-1112942799
I haven't checked it out, but please try it.

I also think this issue is not related to our docker images in particular and you should use the thread with the message I mentioned (that's why I'm closing this issue).

adrienpacifico:
Hi @mathbunnyru!

I am a bit sad that you closed the issue, as I think that being able to copy/paste at least for local jupyterlab docker image could be a nice option to have! 

It could be: 
1. Documented on how to proceed on https://jupyter-docker-stacks.readthedocs.io/
2. Maybe have xclip installed in minimal-notebook (not in base).

If anyone interested, I made it work on a debian based linux with the following method: 

1. Start jupyterlab container with `docker run --network host -v $(pwd):/home/jovyan/work -e GRANT_SUDO=yes --user root -e DISPLAY -it -v /tmp/.X11-unix:/tmp/.X11-unix jupyter/base-notebook`

2. copy paste the link from your terminal in your browser to access to Jupyter.
example:
```bash
Or copy and paste one of these URLs:
        http://pop-os:8889/?token=c370a0043b9bf8e2e10c39a501fe96f4fbe52da9f58bbfd0
```
3. From jupyter/jupyterlab install xclip with `!sudo apt-get update` if you are in a notebook cell (or remove the `!` if you use jupyterlab's terminal). 
4. From jupyter/jupyterlab install xclip with `!sudo apt-get install xclip --yes` if you are in a notebook cell (or remove the `!` if you use jupyterlab's terminal).

4. Install pandas if needed based on the docker image you are using
5. ```python
import pandas as pd
pd.read_clipboard()
```

And it works 🎉

It would be interesting to me to know if there exist a solution like that for macos (or windows).

mathbunnyru:
Great solution @adrienpacifico!
I guess you're right and I was a bit too fast to close this issue.

[...]

PR #1799, opened 2022-10-05
Install xclip to support clipboard functionality

mathbunnyru:
## Issue ticket if applicable

To deal with: https://github.com/jupyter/docker-stacks/issues/1796
I will merge this if the image size increase is small and I will add a recipe to enable clipboard functionality.


mathbunnyru:
The size of the image stayed pretty much the same.
I will update after I merge this change (so it is easier to test).

Issue #1804, opened 2022-10-12
[ENH] - Add openpyxl to scipy-notebook to support .xlsx Excel files


dietmar:
### What docker image(s) is this feature applicable to?

scipy-notebook

### What changes are you proposing?

Add the [openpyxl](https://openpyxl.readthedocs.io) package to the image.

Currently, the pandas function `read_excel` can only read old-style `.xls` files but not new-style `.xlsx` files, because the `scipy-notebook` image does contain the package [xlrd](https://xlrd.readthedocs.io), but not [openpyxl](https://openpyxl.readthedocs.io).
Since pandas 1.3.0, `read_excel` uses `xlrd` for old-style `.xls` files only, for new-style `.xlsx` files `openpyxl` is required, see the following links:

* What's new in pandas 1.3.0: https://pandas.pydata.org/pandas-docs/version/1.3.0/whatsnew/v1.3.0.html
* Pandas issue for this topic: https://github.com/pandas-dev/pandas/issues/28547
* Pandas documentation for `read_excel` (under "engine"): https://pandas.pydata.org/docs/reference/api/pandas.read_excel.html

### How does this affect the user?

"New-style" `.xlsx` files are IMO **very** common (and more common than `.xls` these days), so including the capability to work with such files per default makes a lot of sense.

The `scipy-notebook` image seems the right place for this, because this is where all the Python dependencies are added (such as `xlrd`). Initially I came across this issue using the `datascience-notebook` image, but that is based off `scipy-notebook`.

### Anything else?

I'll be working on a pull request myself right away.

PR #1805, opened 2022-10-12
add openpyxl to scipy-notebook (#1804)

dietmar:
## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/1804

### DIFF
--- a/docs/using/selecting.md
+++ b/docs/using/selecting.md
@@ -107,6 +107,7 @@ It contains:
   [matplotlib-base](https://matplotlib.org/),
   [numba](https://numba.pydata.org/),
   [numexpr](https://github.com/pydata/numexpr),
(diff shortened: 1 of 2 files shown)


mathbunnyru:
We also describe what packages we include here, so this will have to be updated as well.
https://github.com/jupyter/docker-stacks/blob/main/docs/using/selecting.md#jupyterscipy-notebook

Let's see how much this increases the image size. If the change is small, the I guess we can include this.

mathbunnyru:
I don't see any size increase, so @dietmar please update the docs and I will merge this.

dietmar:
> I don't see any size increase, so @dietmar please update the docs and I will merge this.

Docs updated. Thanks for merging @mathbunnyru!

PR #1815, opened 2022-10-26
Introduce policy to update python version

mathbunnyru:

### DIFF
--- a/docs/maintaining/tasks.md
+++ b/docs/maintaining/tasks.md
@@ -16,6 +16,14 @@ To build new images and publish them to the Docker Hub registry, do the followin
 4. Try to avoid merging another PR to main branch until all pending builds are complete.
    This way, you will know which commit might have broken the build and also have correct tags for moving tags (like `python` version).
 
+## Updating Python version
+
+When a new `Python` version is released, we wait for two things:
+
+- all the dependencies to be available (as wheels or in `conda-forge`).
+- the first `python` patch release for this version.
+  This allows to avoid many bugs, which can happen in a major release.
+
 ## Updating the Ubuntu Base Image
 
 `base-notebook` is based on the LTS Ubuntu docker image.


mathbunnyru:
@bjornjorgensen please take a look

bjornjorgensen (approved):

bjornjorgensen:
When a new major `Python` version

add major?

mathbunnyru:
> When a new major `Python` version
> 
> add major?

I'm not sure what is the correct name for Python versioning (because I think `3` is a major version).

benz0li:
> I'm not sure what is the correct name for Python versioning

I am confident Python uses [Semantic Versioning](https://semver.org): `MAJOR.MINOR.PATCH`

manics:
Python doesn't use semantic versioning. For example, these breaking changes were made in 3.10:
https://docs.python.org/3/whatsnew/3.10.html#removed

mathbunnyru:
Thanks @manics. I guess I will keep this PR as it is then.
Could you please share your opinion on this policy?

manics:
From a technical perspective, it sounds fine to me.

From a user perspective- that depends on your goals.
https://jupyter-docker-stacks.readthedocs.io/en/latest/#using-old-images
Do you see this project as leading the way, and encouraging users to use a particular version of Python? Or do you want to be more responsive to users, which means gathering feedback from e.g. https://discourse.jupyter.org/ to gauge the general feeling on which Python should be supported and when a switch should be made?

benz0li:
> Python doesn't use semantic versioning.

Agreed: https://peps.python.org/pep-0440/

mathbunnyru:
> From a technical perspective, it sounds fine to me.
> 
> From a user perspective- that depends on your goals. https://jupyter-docker-stacks.readthedocs.io/en/latest/#using-old-images Do you see this project as leading the way, and encouraging users to use a particular version of Python? Or do you want to be more responsive to users, which means gathering feedback from e.g. https://discourse.jupyter.org/ to gauge the general feeling on which Python should be supported and when a switch should be made?

I would like to find a good balance between switching too early and switching too late.
With a similar Ubuntu policy, I didn't see any negative feedback, which is nice.

I guess, this policy will also make it clear for our users when we will support the new python version and they won't actually have to wait for a long time after a release (around 2 months I suppose).
At the same time, we have never deleted old images, so users, who want to stick with them, can do it.

PR #1821, opened 2022-11-01
Show tags for third-party github actions

mathbunnyru:

### DIFF
--- a/.github/workflows/docker-merge-tags.yml
+++ b/.github/workflows/docker-merge-tags.yml
@@ -61,7 +61,7 @@ jobs:
 
       - name: Login to Docker Hub 🔐
         if: github.ref == 'refs/heads/main' || github.event_name == 'schedule'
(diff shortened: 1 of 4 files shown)


mathbunnyru:
@consideRatio @manics @minrk there is a brand new feature, which allows to specify tag in comments and these comments will be updated by `dependabot` as well!
This makes it both secure (when using tags) and human-readable.

https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/

mathbunnyru:
I merged this one, created PR so I could mention you and let you know about this nice feature.

consideRatio:
Nice improvement, thanks for sharing @mathbunnyru!

mathbunnyru:
Today you willl see two more improvements like this. Hope you don't mind. I think they will be useful in other repos as well and I don't know a better way to share these cool things.

PR #1835, opened 2022-11-23
generate-config prompts if a config exists, so remove first

OKaluza:
## Describe your changes

When I'm building with an ubuntu:20.04 base, the base-notebook build fails on `jupyter notebook --generate-config`, it appears there is an existing file and it prompts whether to overwrite it which halts the build. 

Simply removing the existing file before re-generating the config fixed our issue.
Using the "-f" flag prevents prompting or breaking things if the file doesn't already exist.

- [ ] ~~If it is a core feature, I have added thorough tests~~
- [ ] ~~I have updated the documentation for significant changes~~

### DIFF
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -39,6 +39,7 @@ RUN mamba install --quiet --yes \
     'notebook' \
     'jupyterhub' \
     'jupyterlab' && \
+    rm -f "/home/${NB_USER}/.jupyter/jupyter_notebook_config.py" && \
     jupyter notebook --generate-config && \
     mamba clean --all -f -y && \
     npm cache clean --force && \


mathbunnyru:
I can't reproduce this problem. Could you please tell me how to do it?

My diff:
```diff
-ARG ROOT_CONTAINER=ubuntu:22.04
+ARG ROOT_CONTAINER=ubuntu:20.04
```

My commands:
```
make build/docker-stacks-foundation
make build/base-notebook
```

They work just fine. I'm using x86_64 Linux Ubuntu 22.04 host machine.

OKaluza:
I can no longer duplicate it either, no idea what's going on it was happening every time yesterday but I can remove the fix and it is no longer needed, I can only guess I must have used some other base image somewhere by accident, sorry for the noise

mathbunnyru:
No worries, I will close this PR then.

Issue #1851, opened 2022-12-27
[BUG] - scipy-notebook has old jupyter_server version < 2.0 - Jupyter Server Terminals fails to start


EnigmaCurry:
### What docker image(s) are you using?

scipy-notebook

### OS system and architecture running docker image

Ubuntu/amd64

### What Docker command are you running?

```
IMAGE=jupyter/scipy-notebook; docker pull ${IMAGE} ; docker run --rm -it ${IMAGE}
```

### How to Reproduce the problem?

 * docker pull jupyter/scipy-notebook 
 * docker run --rm -it jupyter/scipy-notebook

### Command output

```bash session
Using default tag: latest
latest: Pulling from jupyter/scipy-notebook
Digest: sha256:d04c1e21489cc8c5ee6fe03c5748cf77872ccd6bbaa1cd50cd54ac424b68378a
Status: Image is up to date for jupyter/scipy-notebook:latest
docker.io/jupyter/scipy-notebook:latest
Entered start.sh with args: jupyter lab
Executing the command: jupyter lab
[W 2022-12-27 22:59:05.372 ServerApp] jupyter_server_terminals | error adding extension (enabled: True): Jupyter Server Terminals requires Jupyter Server 2.0+
    Traceback (most recent call last):
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 320, in add_extension
        extpkg = ExtensionPackage(name=extension_name, enabled=enabled)
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 166, in __init__
        super().__init__(*args, **kwargs)
      File "/opt/conda/lib/python3.10/site-packages/traitlets/traitlets.py", line 1360, in __init__
        value = self._traits[key]._cross_validate(self, getattr(self, key))
      File "/opt/conda/lib/python3.10/site-packages/traitlets/traitlets.py", line 746, in _cross_validate
        value = obj._trait_validators[self.name](obj, proposal)
      File "/opt/conda/lib/python3.10/site-packages/traitlets/traitlets.py", line 1232, in __call__
        return self.func(*args, **kwargs)
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 175, in _validate_name
        self._module, self._metadata = get_metadata(name)
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/utils.py", line 55, in get_metadata
[...]

PR #1853, opened 2022-12-29
Pin jupyter_server>=2.0.0

mathbunnyru:
## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/1851

### DIFF
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -40,7 +40,7 @@ RUN mamba install --quiet --yes \
     'ipympl'\
     'ipywidgets' \
     # Temporary fix for: https://github.com/jupyter/docker-stacks/issues/1851
-    'jupyter_server==2.0.0' \
+    'jupyter_server>=2.0.0' \
     'matplotlib-base' \
     'numba' \
     'numexpr' \
--- a/scipy-notebook/Dockerfile
+++ b/scipy-notebook/Dockerfile
@@ -39,6 +39,8 @@ RUN mamba install --quiet --yes \
     'h5py' \
     'ipympl'\
     'ipywidgets' \
+    # Temporary fix for: https://github.com/jupyter/docker-stacks/issues/1851
+    'jupyter_server==2.0.0' \
     'matplotlib-base' \
     'numba' \
     'numexpr' \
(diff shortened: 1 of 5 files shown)


mathbunnyru:
This is a hotfix.
It would be nice to add the test to make sure the issue won't happen again (I guess we can use healthcheck test).
Then we'll have to add this hotfix to all the affected images.
And I guess this issue will be gone in a few weeks and we will be able to remove the fix (at least I hope so).

consideRatio:
Is this a way to force install a modern version, or to force 2.0.0 but not 2.0.1 etc? I think 2.0.6 is out.

You can do >=2 if you want to force v2 to be installed instead of v1

scipy-notebook/Dockerfile
```diff
@@ -39,6 +39,8 @@ RUN mamba install --quiet --yes \
     'h5py' \
     'ipympl'\
     'ipywidgets' \
+    # Temporary fix for: https://github.com/jupyter/docker-stacks/issues/1851
+    'jupyter_server==2.0.0' \
```
consideRatio, line 43:
```suggestion
    'jupyter_server>=2.0.0' \
```

consideRatio (approved):

bjornjorgensen:
It this a problem only with scipy-notebook?

tests/base-notebook/test_packages.py
```diff
@@ -69,6 +69,7 @@
     "ca-certificates",
     "conda-forge::blas[build=openblas]",
     "hdf5",
+    "jupyter_server[version='>=2.0.0']",  # Temporary fix for: https://github.com/jupyter/docker-stacks/issues/1851
```
bjornjorgensen, line 72: It looks like this only test if the import of this package works or not. Try to remove `[version='>=2.0.0']`

mathbunnyru: No, this won't work, because our parsing is a bit broken.
I fixed the test (not in the beautiful way, but it works).

mathbunnyru:
> It this a problem only with scipy-notebook?

As I told above, I will have to add healthcheck test to know if other images are affected.
I did it, so we'll see the results in 1h30 minutes.

jabbera:
I'm unsure if you folks care but this breaks voila as they have a dependency on < 2.

mathbunnyru:
> I'm unsure if you folks care but this breaks voila as they have a dependency on < 2.

Unfortunately, some packages will break and that's (kind of) expected behaviour.
It is generally not recommended to use the latest version (and it is even better to use sha/date tag), if you don't want anything to break.

Issue #1856, opened 2023-01-06
[BUG] Log WARNING: no write access to /home/jovyan


simonglet:
### What docker image(s) are you using?

datascience-notebook, scipy-notebook

### OS system and architecture running docker image

MacOS 10.13.6, High Sierra: 17.7.0 Darwin Kernel Version 17.7.0: Fri Oct 30 13:34:27 PDT 2020; root:xnu-4570.71.82.8~1/RELEASE_X86_64 x86_64

### What Docker command are you running?

docker run -p8890:8888 -v mylocaldir:/home/jovyan/work jupyter/scipy-notebook:b807da51d7f2

### How to Reproduce the problem?

Run the command and these lines will be logged:

$ docker run -p8890:8888 -v mylocaldir:/home/jovyan/work jupyter/scipy-notebook:latest
Entered start.sh with args: jupyter lab
WARNING: no write access to /home/jovyan. Try starting the container with group 'users' (100), e.g. using "--group-add=users".

if you let the container launch, open Jupyter, and double-click on the Python3 Notebook icon to get the following error message in a popup alert:
Unexpected error while saving file: Untitled.ipynb [Errno 2] No such file or directory: '/home/jovyan/.~Untitled.ipynb' -> '/home/jovyan/Untitled.ipynb'

### Expected behavior

Being able to mount a local dir in the container.

### Actual behavior

Cannot see the directory in Jupyter
Cannot create a file in the / dir

mathbunnyru:
Please, take a look here:
https://jupyter-docker-stacks.readthedocs.io/en/latest/using/troubleshooting.html#permission-denied-when-mounting-volumes

I think the instruction provided there should help.

simonglet:
Thank you for the reference.

By opening a bash session in the running container, I confirmed that the /home/jovyan directory is not visible from Jupyter. It is there, I can mount in it a writable directory without the extra CHOWN arguments.

So now it seems to be a Jupyter configuration issue. Just in case, how can I fix that?

Thanks

mathbunnyru:
Please check permissions/ownership of your dir in the host machine and inside the container?
Also, please provide as many details as you can.

simonglet:
Thank you for the follow-up.

Trying to run the Jupyter container on my Mac (MacBook Pro, early 2011, Intel i5) now stuck at macOS High Sierra (10.13.6) with Docker 19.03.13 was a bad idea.

[...]

PR #1857, opened 2023-01-07
Improve issue template

mathbunnyru:

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -15,7 +15,6 @@ body:
         Please answer the following sections to help us troubleshoot the problem.
 
   - type: dropdown
-    id: docker_images
     attributes:
       label: What docker image(s) are you using?
       description: Select as many images as applicable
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -15,7 +15,6 @@ body:
         Please answer the following sections to help us troubleshoot the problem.
 
   - type: dropdown
-    id: docker_images
```
mathbunnyru, line 18: I don't think this is useful at all.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -109,3 +108,11 @@ body:
         Tip: You can attach images or log files by clicking this area to highlight it and then dragging files in.
     validations:
       required: false
+
+  - type: checkboxes
+    attributes:
+      label: Latest Docker version
+      description: You should try to use latest Docker version
+      options:
+        - label: I've updated my Docker version to the latest available and the issue still persists
```
mathbunnyru, line 117: We've seen several people, who were using old Docker versions and they were getting some errors. I think this should make them update the Docker version before creating an issue.

mathbunnyru: See: https://github.com/jupyter/docker-stacks/issues/1856

Issue #1709, opened 2022-05-21
[BUG] - Container fails health check when using non-empty "base_url"


zbig-t:
### What docker image(s) are you using?

scipy-notebook

### OS system and architecture running docker image

VMware Photon OS/Linux/x86_64

### What Docker command are you running?

```
docker run -d --name jupyter-lab-1 \
	--restart unless-stopped \
	-p 8888:8888 \
	--user 1026 --group-add users \
	-v /mnt/docker-shared/jupyter-lab:/home/jovyan/work \
	-v /mnt/conf/jupyter-lab/.jupyter:/home/jovyan/.jupyter \
	jupyter/scipy-notebook:latest \
        start-notebook.sh --NotebookApp.base_url=/jupyter/
```

### How to Reproduce the problem?

Start a jupyter container with a non-empty `base_url` - either via `--NotebookApp.base_url` parameter in a `docker run` invocation or a `c.ServerApp.base_url` parameter in a bind-mounted `jupyter_server_config.py` file.

Do `docker ps`

### Command output

```bash session
docker ps
CONTAINER ID   IMAGE                                    COMMAND                  CREATED          STATUS                      PORTS                                                                                            NAMES
27e0b443b186   jupyter/scipy-notebook:latest            "tini -g -- start-no…"   20 minutes ago   Up 20 minutes (unhealthy)   0.0.0.0:8888->8888/tcp, :::8888->8888/tcp                                                        jupyter-lab-1
```

### Expected behavior

`docker ps` showing a healthy jupyterlab container.

### Actual behavior

`docker ps` showing status "unhealthy" when container is configured with custom `base_url`, unless `-e JUPYTERHUB_SERVICE_PREFIX=custom_prefix` is added explicitly during `docker run`.

mathbunnyru:
Hello!
Unfortunately, we can't provide healthcheck, which works in all cases (it will be to much parsing of config files).
If you really need to have healthcheck, use `--health-cmd CUSTOM_COMMAND`, where you add this base_url to our healthcheck.
If you don't want it at all, use `--no-healthcheck`

zbig-t:
Hello

Thank you for your response and clarification, @mathbunnyru. This is not a problem and I kind of suspected that rectifying this programmatically could require adding complexity.

antofthy:
It would be better if this was mentioned in the docker-stacks guide.

mathbunnyru:
> It would be better if this was mentioned in the docker-stacks guide.

[...]

PR #1859, opened 2023-01-19
Parameterize healthcheck by internal port

### LINKED ISSUE #1817: [BUG] - Health Check fails if you change the port jupyter runs on
### What docker image(s) are you using?

minimal-notebook

### OS system and architecture running docker image

RHEL7  docker swarm

### What Docker command are you running?

Not really relevant, but I need to run it in a docker swarm, with a generalise 'ingress service'.

For this I needed to change internal port jupyter runs on needs to be changes for intergation into a 'ingress proxy'
To change the port I made a slight modification the docker image to set the internal port it runs on (see below)

The problem is the docker container  dies unexpectedly after running for 46 seconds.
During that time the service is visible within the conatiner, but not external to the container.
This is because the built-in heathcheck never succeeds, and eventually kills the container with little logged reporting.
(see below)

### How to Reproduce the problem?

Dockerfile, to set port

```dockerfile
FROM "jupyter/minimal-notebook:latest"
# Update Jupyter configuration to set port
RUN set -eux; \
    sed -i 's/port = 8888/port = 8080/' /etc/jupyter/jupyter_notebook_config.py ;\
    sed -i 's/port = 8888/port = 8080/' /etc/jupyter/jupyter_server_config.py ;\
    :;
```
You can also change the port in other ways such as... 

Creating a `~joyvan/.jupyter/jupyter_server_config.py` file (which can also set a password)

or setting a  JUPYTER_PORT environment variable (IF the setting in `/etc/jupyter` configs are removed)

### Command output

When you build and then run the modified docker image, `docker ps`  reports
`Up 9 seconds (health: starting)   8888/tcp`  despite the fact that jupyter is now running on port 8080

46 seconds after starting the container dies with a unhelpful (Signal 15)

Log output...
```
[I 2022-10-28 05:20:00.393 ServerApp] Jupyter Server 1.21.0 is running at:
[I 2022-10-28 05:20:00.393 ServerApp] http://jupyter_service:8080/lab?token=929eaaf2e60f8a947761cb1f2741d745fd46dde62f6fef7c
or http://127.0.0.1:8080/lab?token=929eaaf2e60f8a947761cb1f2741d745fd46dde62f6fef7c
[I 2022-10-28 05:20:00.393 ServerApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[C 2022-10-28 05:20:00.397 ServerApp] 

To access the server, open this file in a browser:
file:///home/jovyan/.local/share/jupyter/runtime/jpserver-8-open.html
Or copy and paste one of these URLs:
http://jupyter_service:8080/lab?token=929eaaf2e60f8a947761cb1f2741d745fd46dde62f6fef7c
or http://127.0.0.1:8080/lab?token=929eaaf2e60f8a947761cb1f2741d745fd46dde62f6fef7c
Entered start.sh with args: jupyter lab
Executing the command: jupyter lab
[C 2022-10-28 05:20:46.261 ServerApp] received signal 15, stopping
[I 2022-10-28 05:20:46.262 ServerApp] Shutting down 2 extensions
[I 2022-10-28 05:20:46.263 ServerApp] Shutting down 0 terminals
```

### Expected behavior

Changing the internal port should not take days of work to track down, it should be straight forward and documented.

The healthcheck should also be properly documented in jupyter-stacks documentation.
This will make it more 'swarm friendly' as well as allow others to integrate it better when port 8888 is NOT available.

Yes you can map the port when doing a 'docker run', but that is NOT always possible.

### Actual behavior

Internal Port changing is undocumented in stacks

Heathcheck kills the container without notice (signal 15 hardly makes it clear) when port is different.

Days of work lost trying to figure out what should be a straight forward and simple task.

### Anything else?

There is an existing environment variable  "JUPYTER_PORT" that defines the default port.
But any such setting is currently overridden by the  configuration files in `/etc/jupyter`

This may be usable to set healthcheck, especially if the config file default is removed, or allows the env var to override.

in Dockerfile....
```
HEALTHCHECK  --interval=15s --timeout=3s --start-period=5s --retries=3 \
    CMD wget -O- --no-verbose --tries=1 --no-check-certificate \
    http${GEN_CERT:+s}://localhost:${JUPYTER_PORT:-8888}${JUPYTERHUB_SERVICE_PREFIX:-/}api || exit 1
```

That Environment variable also needs to be documented in the jupyter-stacks documentation, with the health check.

imiro:
## Describe your changes

Opening PR intended to close #1817 

- [x] Let's remove port setting in config.
- [x] Add JUPYTER_PORT with default value 8888 to the root file
- [x] Add a test which changes the internal port
- [x] Fix the health check command to use this env variable
- [x] Add a health check test (both using the default port and not using it).
- [x] Document health check behaviour (including https://github.com/jupyter/docker-stacks/issues/1709)

## Issue ticket if applicable

Fix #1817 

### DIFF
--- a/tests/base-notebook/test_container_options.py
+++ b/tests/base-notebook/test_container_options.py
@@ -86,8 +86,7 @@ def test_unsigned_ssl(
         None,
         {"DOCKER_STACKS_JUPYTER_CMD": "notebook"},
         {"JUPYTER_PORT": 8171},
-        {"JUPYTER_PORT": 8117,
-         "DOCKER_STACKS_JUPYTER_CMD": "notebook"},
+        {"JUPYTER_PORT": 8117, "DOCKER_STACKS_JUPYTER_CMD": "notebook"},
     ],
 )
 def test_custom_internal_port(
@@ -96,7 +95,9 @@ def test_custom_internal_port(
     """Container should be accessible from the host
     when using custom internal port"""
     host_port = find_free_port()
-    internal_port = env["JUPYTER_PORT"] if (env and ("JUPYTER_PORT" in env.keys())) else 8888
+    internal_port = (
+        env["JUPYTER_PORT"] if (env and ("JUPYTER_PORT" in env.keys())) else 8888
+    )
     running_container = container.run_detached(
         command=["start-notebook.sh", "--NotebookApp.token=''"],
         environment=env,
--- a/tests/base-notebook/test_container_options.py
+++ b/tests/base-notebook/test_container_options.py
@@ -80,11 +80,13 @@ def test_unsigned_ssl(
     assert not warnings
 
 
-def test_custom_internal_port(container: TrackedContainer, http_client: requests.Session) -> None:
+def test_custom_internal_port(
+    container: TrackedContainer, http_client: requests.Session
+) -> None:
     """Container should be accessible from the host
     when using custom internal port"""
     host_port = find_free_port()
-    internal_port=8117
+    internal_port = 8117
     running_container = container.run_detached(
         command=["start-notebook.sh", "--NotebookApp.token=''"],
         environment={"JUPYTER_PORT", internal_port},
--- a/tests/base-notebook/test_container_options.py
+++ b/tests/base-notebook/test_container_options.py
@@ -83,21 +83,21 @@ def test_unsigned_ssl(
 @pytest.mark.parametrize(
     "env",
     [
-        None,
+        {},
         {"DOCKER_STACKS_JUPYTER_CMD": "notebook"},
         {"JUPYTER_PORT": 8171},
         {"JUPYTER_PORT": 8117, "DOCKER_STACKS_JUPYTER_CMD": "notebook"},
     ],
 )
 def test_custom_internal_port(
-    container: TrackedContainer, http_client: requests.Session
+    container: TrackedContainer,
+    http_client: requests.Session,
+    env: dict[str, str],
 ) -> None:
     """Container should be accessible from the host
     when using custom internal port"""
     host_port = find_free_port()
-    internal_port = (
-        env["JUPYTER_PORT"] if (env and ("JUPYTER_PORT" in env.keys())) else 8888
-    )
+    internal_port = env.get("JUPYTER_PORT", 8888)
     running_container = container.run_detached(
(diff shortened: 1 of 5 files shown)


mathbunnyru (changes requested):
Please:
- add the documentation
- add the test changing the internal port and checking that the host binded port is working. You can add this test here: https://github.com/jupyter/docker-stacks/blob/main/tests/base-notebook/test_container_options.py

mathbunnyru (changes requested):
Thank you for the documentation update. I fixed some wording. Please, add a test, as I asked.

imiro:
@mathbunnyru just added a test following your directions. Please help check, I'm not sure if the assertions I used are ideal.

tests/base-notebook/test_container_options.py
```diff
@@ -78,3 +78,24 @@ def test_unsigned_ssl(
     assert "ERROR" not in logs
     warnings = TrackedContainer.get_warnings(logs)
     assert not warnings
+
+
+def test_custom_internal_port(
```
mathbunnyru, line 83: Let's add `pytest.mark.parametrize` almost the same way as it is done in test_health.

mathbunnyru (changes requested):
I think this is almost good to go. Only test parametrisation is left to be done.

mathbunnyru:
> @mathbunnyru just added a test following your directions. Please help check, I'm not sure if the assertions I used are ideal.

I guess you just copied them from similar test, and it looks reasonable to be 👍

tests/base-notebook/test_container_options.py
imiro: Added with [e0fda89](https://github.com/jupyter/docker-stacks/pull/1859/commits/e0fda8994849369c61869a7e4851006b36e9e1ab)

mathbunnyru:
I highly suggest you install pre-commit hooks to make it easier for you (`pre-commit install`).
And also to test locally (`make build/scipy-notebook` to build and `make test/scipy-notebook` to test it, all upstream images have to be be built before, or they will be pulled from DockerHub).

mathbunnyru (approved):

imiro:
Yep noted, totally missed that before, sorry. Thanks for the suggestions. Will do when I work on future PR, thanks for merging this one 👍

Issue #1862, opened 2023-01-24
[BUG] - Jupyter server launches on port 80 instead of 8888 when launched via JupyterHub


twalcari:
### What docker image(s) are you using?

minimal-notebook

### Host OS system and architecture running docker image

Ubuntu 22.04 - amd64

### What Docker command are you running?

Full report of `docker inspect` of the docker container that was created: 
[docker-inspect.txt](https://github.com/jupyter/docker-stacks/files/10492034/docker-inspect.txt)

Command:

```
 start-notebook.sh --notebook-dir=/ --SingleUserNotebookApp.default_url='/lab/tree/project_ghent'
```

Environment:

```
GRANT_SUDO = yes
JPY_API_TOKEN = 0c1f95e5d1384faeaa9b945bf21b8e3b
JUPYTERHUB_ACTIVITY_URL = http://193.191.148.147:19998/hub/api/users/twalcari@ilabt.imec.be/activity
JUPYTERHUB_ADMIN_ACCESS = 1
JUPYTERHUB_API_TOKEN = 0c1f95e5d1384faeaa9b945bf21b8e3b
JUPYTERHUB_API_URL = http://193.191.148.147:19998/hub/api
JUPYTERHUB_BASE_URL = /
JUPYTERHUB_CLIENT_ID = jupyterhub-user-twalcari%40ilabt.imec.be
JUPYTERHUB_HOST =
JUPYTERHUB_OAUTH_ACCESS_SCOPES = ["access:servers!server=twalcari@ilabt.imec.be/", "access:servers!user=twalcari@ilabt.imec.be"]
JUPYTERHUB_OAUTH_CALLBACK_URL = /user/twalcari@ilabt.imec.be/oauth_callback
JUPYTERHUB_OAUTH_CLIENT_ALLOWED_SCOPES = []
JUPYTERHUB_OAUTH_SCOPES = ["access:servers!server=twalcari@ilabt.imec.be/", "access:servers!user=twalcari@ilabt.imec.be"]
JUPYTERHUB_ROOT_DIR = /
JUPYTERHUB_SERVER_NAME =
JUPYTERHUB_SERVICE_PREFIX = /user/twalcari@ilabt.imec.be/
JUPYTERHUB_SERVICE_URL = http://127.0.0.1:0/user/twalcari@ilabt.imec.be/
JUPYTERHUB_USER = twalcari@ilabt.imec.be
LANG = en_US.UTF-8
NB_GID = GPULab Dark Magic
[...]

PR #1863, opened 2023-01-25
Try to fix port for jupyterhub

mathbunnyru:
## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/1862

### DIFF
--- a/base-notebook/jupyter_server_config.py
+++ b/base-notebook/jupyter_server_config.py
@@ -9,6 +9,9 @@ from jupyter_core.paths import jupyter_data_dir
 
 c = get_config()  # noqa: F821
 c.ServerApp.ip = "0.0.0.0"
+# Setting port in the config to make JupyterHub singleuser use the provided port
+# https://github.com/jupyter/docker-stacks/issues/1862
+c.ServerApp.port = os.environ["JUPYTER_PORT"]
 c.ServerApp.open_browser = False
 
 # to output both image/svg+xml and application/pdf plot formats in the notebook file
--- a/base-notebook/jupyter_server_config.py
+++ b/base-notebook/jupyter_server_config.py
@@ -11,7 +11,7 @@ c = get_config()  # noqa: F821
 c.ServerApp.ip = "0.0.0.0"
 # Setting port in the config to make JupyterHub singleuser use the provided port
 # https://github.com/jupyter/docker-stacks/issues/1862
-c.ServerApp.port = os.environ["JUPYTER_PORT"]
+c.ServerApp.port = int(os.environ["JUPYTER_PORT"])
 c.ServerApp.open_browser = False
 
 # to output both image/svg+xml and application/pdf plot formats in the notebook file


mathbunnyru:
@twalcari could you please check this change fixes it for you?

mathbunnyru:
@consideRatio I would like to point out that running our docker images under jupyterhub are not tested at all in this repo.
Do you know what would be the best way to test it (maybe you have some minimal jupyterhub example) and in which repo should we test it?

twalcari:
Hi, I've been able to narrow it down this morning to the `JUPYTERHUB_SERVICE_URL`.

My custom spawner did not implement a value for `port`, which caused it to default to 0. This resulted in the following `JUPYTERHUB_SERVICE_URL`: `http://127.0.0.1:0/user/twalcari@ilabt.imec.be/`

After adding the following lines to my spawner, the JUPYTERHUB_SERVICE_URL is now `http://127.0.0.1:8888/user/twalcari@ilabt.imec.be/` and the problem is gone.

```python
class GPULabSpawner(Spawner):
    ...

    @default("port")
    def _default_port(self):
        return 8888

```

I would suggest failing the startup of the docker container if such an invalid URL is received with an informative error would be better than this strange behavior of the port changing to 80 😅

mathbunnyru:
I guess we should still use the port set in JUPYTER_PORT, shouldn't we?
I don't completely understand what behaviour we should implement (and users expect).

twalcari:
> @consideRatio I would like to point out that running our docker images under jupyterhub are not tested at all in this repo. Do you know what would be the best way to test it (maybe you have some minimal jupyterhub example) and in which repo should we test it?

I was a bit stuck wrt to testing this yesterday because the startup script hangs if it is unable to reach the JupyterHub service. After analyzing the traffic, I determined that the Jupyter process in the docker container does not use the result of the HTTP-requests it does to the hub.

My solution was to launch a small webserver which just reacts with 200 replies to all requests. I used the example from: https://gist.github.com/bradmontgomery/2219997

twalcari:
> I guess we should still use the port set in JUPYTER_PORT, shouldn't we? I don't completely understand what behaviour we should implement (and users expect).

After a bit of searching, it seems like the answer is in `notebook/notebookapp.py`, cfr. https://github.com/jupyter/notebook/pull/5221

I don't think that the fix you committed earlier is needed because of this. It also won't solve the bug uncovered in this issue as it is related to a faulty(?) `JUPYTERHUB_SERVICE_URL`

manics:
My guess for testing JupyterHub (untried, but I might give it a go if I have time) is probably to copy a test from DockerSpawner:
- conftest: https://github.com/jupyterhub/dockerspawner/blob/8503af69161a3a543cc613f93ce7951ad30a1912/tests/conftest.py#L59-L84
- start/stop test https://github.com/jupyterhub/dockerspawner/blob/8503af69161a3a543cc613f93ce7951ad30a1912/tests/test_dockerspawner.py#L42-L61

Longer term https://github.com/jupyterhub/pytest-jupyterhub may avoid the need to copy conftest, you'd still need a spawner implementation though

mathbunnyru:
Ok, I guess I will close this PR and thee issue (because there is an upstream issue).

PR #1868, opened 2023-01-29
Healthcheck url from json

imiro:
## Describe your changes

This PR tries to implement the idea discussed on #1817. Obtain the URL where the jupyter server runs on from a json file located on `/home/$NB_USER/.local/share/jupyter/runtime` directory and use it for healthcheck. Parsing of the json file is done with minimal python currently written as a one-liner command. 

Would appreciate any feedback!

## Issue ticket if applicable

Intended as a solution for https://github.com/jupyter/docker-stacks/issues/1709

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
 
   # Automatically sort python imports
   - repo: https://github.com/pycqa/isort
-    rev: 5.11.4
+    rev: 5.12.0
     hooks:
       - id: isort
         args: [--profile, black]
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -56,7 +56,7 @@ CMD ["start-notebook.sh"]
 # Copy local files as late as possible to avoid cache busting
 COPY start-notebook.sh start-singleuser.sh /usr/local/bin/
 # Currently need to have both jupyter_notebook_config and jupyter_server_config to support classic and lab
-COPY jupyter_server_config.py /etc/jupyter/
+COPY jupyter_server_config.py docker_healthcheck.py /etc/jupyter/
 
 # Fix permissions on /etc/jupyter as root
 USER root
@@ -70,10 +70,7 @@ RUN sed -re "s/c.ServerApp/c.NotebookApp/g" \
 # This healtcheck works well for `lab`, `notebook`, `nbclassic`, `server` and `retro` jupyter commands
 # https://github.com/jupyter/docker-stacks/issues/915#issuecomment-1068528799
 HEALTHCHECK  --interval=5s --timeout=3s --start-period=5s --retries=3 \
-    CMD bash -c 'set -o pipefail; \
-    (cat /home/"${NB_USER}"/.local/share/jupyter/runtime/*.json | \
-    python -c '\''import sys, json; print(json.loads(sys.stdin.read())["url"]+"api");'\'' | \
-    wget -i - -O- --no-verbose --tries=1 --no-check-certificate) || exit 1'
+    CMD python /etc/jupyter/docker_healthcheck.py || exit 1
 
 # Switch back to jovyan to avoid accidental container runs as root
 USER ${NB_UID}
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -70,9 +70,10 @@ RUN sed -re "s/c.ServerApp/c.NotebookApp/g" \
 # This healtcheck works well for `lab`, `notebook`, `nbclassic`, `server` and `retro` jupyter commands
 # https://github.com/jupyter/docker-stacks/issues/915#issuecomment-1068528799
 HEALTHCHECK  --interval=5s --timeout=3s --start-period=5s --retries=3 \
-    CMD cat /home/"${NB_USER}"/.local/share/jupyter/runtime/*.json | \
-    python -c 'import sys, json; print(json.loads(sys.stdin.read())["url"]+"api");' | \
-    wget -i - -O- --no-verbose --tries=1 --no-check-certificate || exit 1
+    CMD bash -c 'set -o pipefail; \
+    (cat /home/"${NB_USER}"/.local/share/jupyter/runtime/*.json | \
+    python -c '\''import sys, json; print(json.loads(sys.stdin.read())["url"]+"api");'\'' | \
+    wget -i - -O- --no-verbose --tries=1 --no-check-certificate) || exit 1'
 
 # Switch back to jovyan to avoid accidental container runs as root
 USER ${NB_UID}
--- a/base-notebook/Dockerfile
+++ b/base-notebook/Dockerfile
@@ -70,7 +70,7 @@ RUN sed -re "s/c.ServerApp/c.NotebookApp/g" \
 # This healtcheck works well for `lab`, `notebook`, `nbclassic`, `server` and `retro` jupyter commands
 # https://github.com/jupyter/docker-stacks/issues/915#issuecomment-1068528799
 HEALTHCHECK  --interval=5s --timeout=3s --start-period=5s --retries=3 \
-    CMD cat /home/${NB_USER}/.local/share/jupyter/runtime/*.json | \
+    CMD cat /home/"${NB_USER}"/.local/share/jupyter/runtime/*.json | \
     python -c 'import sys, json; print(json.loads(sys.stdin.read())["url"]+"api");' | \
     wget -i - -O- --no-verbose --tries=1 --no-check-certificate || exit 1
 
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -21,7 +21,7 @@ repos:
 
   # Automatically sort python imports
   - repo: https://github.com/pycqa/isort
-    rev: 5.11.4
+    rev: 5.12.0
```
mathbunnyru, line 24: Already merged this change to main branch :)

base-notebook/Dockerfile
```diff
@@ -70,8 +70,9 @@ RUN sed -re "s/c.ServerApp/c.NotebookApp/g" \
 # This healtcheck works well for `lab`, `notebook`, `nbclassic`, `server` and `retro` jupyter commands
 # https://github.com/jupyter/docker-stacks/issues/915#issuecomment-1068528799
 HEALTHCHECK  --interval=5s --timeout=3s --start-period=5s --retries=3 \
-    CMD wget -O- --no-verbose --tries=1 --no-check-certificate \
-    http${GEN_CERT:+s}://localhost:${JUPYTER_PORT}${JUPYTERHUB_SERVICE_PREFIX:-/}api || exit 1
+    CMD cat /home/${NB_USER}/.local/share/jupyter/runtime/*.json | \
```
mathbunnyru, line 73:
```suggestion
    CMD cat /home/"${NB_USER}"/.local/share/jupyter/runtime/*.json | \
```

tests/base-notebook/test_healthcheck.py
```diff
@@ -13,22 +13,48 @@
...
+        ),
+        (
+            ["NB_USER=testuser", "CHOWN_HOME=1", "JUPYTER_PORT=8123"],
+            ["start-notebook.sh", "--NotebookApp.base_url=/test"],
+            "/home/testuser",
+        ),
     ],
 )
-def test_health(container: TrackedContainer, env: Optional[list[str]]) -> None:
+def test_health(
```
mathbunnyru, line 47: Wow, I like the number of tests you have 👍 

I have two questions though:
1. Why do you need to change workdir?
2. We do `time.sleep(15)`. I think when we have this many tests it's better to return early (for example have a loop for 15 seconds maximum, which checks each 0.1 second, whether the status is healthy).

mathbunnyru (changes requested):

tests/base-notebook/test_healthcheck.py
imiro: 1. I included it just so it resembles the example on https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#user-related-configurations more closely. It's not really needed actually, I tested without changing the workdir as well and it doesn't cause any problems. Should we not change it to simplify the test?
2. That's a good point. I can try to implement it. When testing locally it didn't feel like it took that long though, I guess pytest runs those tests in parallel

base-notebook/Dockerfile
imiro: Is this to handle the case where "NB_USER" contains whitespace? I'm wondering if we need to quote the entire path `"/home/${NB_USER}/.local/share/jupyter/runtime/*.json"` instead..?

.pre-commit-config.yaml
imiro: Thanks! I'll merge the latest update from the main branch here

tests/base-notebook/test_healthcheck.py
mathbunnyru: 1. I think if it doesn't change behaviour in any visible way, it's better to remove it.

base-notebook/Dockerfile
mathbunnyru: > Is this to handle the case where "NB_USER" contains whitespace?

Yes, though it is not well tested 😂 
Abd we try to keep this style where we use variables with possible space, so it should probably work fine.

> I'm wondering if we need to quote the entire path "/home/${NB_USER}/.local/share/jupyter/runtime/*.json" instead..?

No, if you quote the entire string, `cat` is gonna think you need one specific file (and it won't find it).
So, it should be done the way I wrote it.

mathbunnyru:
@imiro hi! any chance you could take a look at my suggestions?

mathbunnyru:
@imiro are you still interested in this PR?

imiro:
HI @mathbunnyru sorry for getting this stalled, been struggling to find the time to work on this. I've pushed some updates just now

base-notebook/Dockerfile
imiro: Actually I tried to test the current implementation without quoting NB_USER and setting NB_USER to have space and it somehow passed the test 🤔 I committed this anyway though

imiro:
If they look good I'll merge the latest commit from main here, and update the documentation as well

mathbunnyru (commented):
LGTM!

Please, merge main and update docs as you wanted 👍

imiro:
image
I was testing locally just now and I got this error. Don't remember seeing this before, it seems that `/home/testu` is not created event though I started the container with `-e NB_USER=testu` ?
This also reveals a problem with the test because its exit status remains 0 .. so the status is "healthy"

mathbunnyru:
@imiro thank you!

I think I know several problems with current solution:

1. Using bash is a bad idea. I suggest you rewrite the HEALTHCHECK command to run a simple python script. This way, if the file doesn't exist or something else, we will have an apparent problem, which won't be skipped during bash settings (like pipefail and so on).

2. When you use `-e NB_USER=testu`, you also have to use `user=root`, otherwise you won't be able to change username.

base-notebook/Dockerfile
```diff
@@ -70,8 +70,10 @@ RUN sed -re "s/c.ServerApp/c.NotebookApp/g" \
 # This healtcheck works well for `lab`, `notebook`, `nbclassic`, `server` and `retro` jupyter commands
 # https://github.com/jupyter/docker-stacks/issues/915#issuecomment-1068528799
 HEALTHCHECK  --interval=5s --timeout=3s --start-period=5s --retries=3 \
-    CMD wget -O- --no-verbose --tries=1 --no-check-certificate \
-    http${GEN_CERT:+s}://localhost:${JUPYTER_PORT}${JUPYTERHUB_SERVICE_PREFIX:-/}api || exit 1
+    CMD bash -c 'set -o pipefail; \
```
mathbunnyru, line 73: This is getting out of hand! 😂 

Let's write a simple python script, please 😄

imiro: yeah I know =)) adding a simple bash option seems to resolve the issue, so I thought I'll give it a shot lol. alright, I'll try to work on the python script

imiro: Anyway, when I run `make build/base-notebook` followed by `make test/base-notebook`, the test should run on the image I have locally, right?

imiro: Actually, never mind. I got strange behavior, apparently because I forgot to rebuild the image locally

mathbunnyru: > Anyway, when I run `make build/base-notebook` followed by `make test/base-notebook`, the test should run on the image I have locally, right?

Yes, but you need to build all parent images as well - in this case, if you want to have the latest version, you should first build `make build/docker-stacks-foundation`.

mathbunnyru:
@imiro this works as expected now, right?

mathbunnyru (approved):

imiro:
@mathbunnyru Just added a couple more things. I was halfway there sorry.

So another observation is that NB_USER values with spaces aren't valid. I modified the test a bit for this. Please review, thanks!

tests/base-notebook/test_healthcheck.py
```diff
@@ -13,23 +13,106 @@
...
+            return
+
     assert get_health(running_container) == "healthy"
+
+
+@pytest.mark.parametrize(
+    "env,cmd,user",
+    [
+        (["NB_USER=testuser", "CHOWN_HOME=1"], None, None),
+        (["NB_USER='test user'", "CHOWN_HOME=1"], None, "root"),
```
mathbunnyru, line 77: This should actually work.
I think it's better to fix users with spaces after merging this PR.

imiro: Why do you think it should work? I think it's a limitation of Unix in general, you just can't have a username that contain spaces.

imiro: I'm not really sure though. That's just my impression when reading the error message.

imiro: Sorry you're right. At least based on https://serverfault.com/a/73101 and useradd's help, it should be possible to allow space, by giving the `--badnames` param to the `useradd` command. But it's an entirely different issue I guess, as you said. I'll leave it to you whether to address this issue and actually allow "bad username". 

Should we remove these tests, then?

mathbunnyru: You're right.

```
# useradd "crazy user"
useradd: invalid user name 'crazy user'
```

I will delete these tests then (they don't make much sense I suppose).

imiro:
Thanks!

mathbunnyru:
You're welcome. I also modernised a script (just a little bit) and removed the irrelevant information from the docs.
Thank you for the contribution 👍

PR #1879, opened 2023-02-17
Install r-tidymodels on aarch64 images

mathbunnyru:


mathbunnyru:
After some effort, I was able to install `r-tidymodels` on `linux-aarch64`.
So, this eliminates the last difference between `x86_64` and `aarch64` images (except we don't provide `tensorflow-notebook` image at all for `aarch64`).

mathbunnyru:
And, it works 🔥

PR #1882, opened 2023-03-03
Update docker_healthcheck.py

finalstate:
This does not (always) work. Sometimes, other JSON files appear in the directory, like this kernel-xxx file

jpserver-6.json
jpserver-6-open.html
kernel-374f4977-29fc-43e0-8a48-231f6980fdab.json

changing the glob expression to something like "jpserver*.json" instead of simply "*.json" should solve the problem (at least in my case, I checked with a quick patch on a live container, the healthcheck immediately turned 'healthy' again...)

### DIFF
--- a/base-notebook/docker_healthcheck.py
+++ b/base-notebook/docker_healthcheck.py
@@ -11,7 +11,7 @@ import requests
 # As this is a healthcheck, it should succeed or raise an exception on error
 
 runtime_dir = Path("/home/") / os.environ["NB_USER"] / ".local/share/jupyter/runtime/"
-json_file = next(runtime_dir.glob("*.json"))
+json_file = next(runtime_dir.glob("jpserver*.json"))
 
 url = json.loads(json_file.read_bytes())["url"]
 url = url + "api"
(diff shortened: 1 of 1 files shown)


base-notebook/docker_healthcheck.py
```diff
@@ -11,7 +11,7 @@
 # As this is a healthcheck, it should succeed or raise an exception on error
 
 runtime_dir = Path("/home/") / os.environ["NB_USER"] / ".local/share/jupyter/runtime/"
-json_file = next(runtime_dir.glob("*.json"))
+json_file = next(runtime_dir.glob("jpserver*.json"))
```
mathbunnyru, line 14:
```suggestion
json_file = next(runtime_dir.glob("jpserver-*.json"))
```

Let's add `-` to be even more specific.

mathbunnyru:
Thank you for this PR @finalstate.
I will merge it when tests pass.

mathbunnyru:
When we use `-e DOCKER_STACKS_JUPYTER_CMD=notebook`, the file is called `nbserver-7.json`.

finalstate:
> When we use `-e DOCKER_STACKS_JUPYTER_CMD=notebook`, the file is called `nbserver-7.json`.

'*server-*.json' ? this starts to become messy :-) Sorry, I am completely new to this project, so I am not of great help...

mathbunnyru:
It's gonna be `*server-*.json`.
This should work fine, I hope :)

It's nice that @imiro added so many test cases, so we can spot bugs like this.

PR #1887, opened 2023-03-08
Remove --quiet flag when not needed

mathbunnyru:
## Describe your changes

Removing `--quiet` flag from most commands.
It was added only because of Travis CI logs size, which we no longer use.
This flag makes it more difficult to analyze the logs when something (usually dependency solver) goes wrong.


mathbunnyru:
This flag was introduced here: https://github.com/jupyter/docker-stacks/pull/134#issuecomment-193065938

Issue #1464, opened 2021-09-14
JupyterLab Git Extension


kslader8:
**What docker images this feature is applicable to?**
jupyter/scipy-notebook

**What changes do you propose?**
installing jupyterlab-git 
(mamba install jupyterlab-git)

**How does this change will affect users?**
It allows users to more easily collaborate (the main issue research teams I work on have with jupyterhub). I'm actually surprised that jupyterlab-git hasn't been absorbed into the default running of jupyter lab. It's a very wide value add and git is already installed in the minimal notebook.

The only downside is it isn't used by jupyter notebooks, but it's probably a good time to start incentivizing jupyter lab over jupyter notebooks.

mathbunnyru:
@kslader8 could you please draft a PR?
And could you check how the size of the image changes after adding this package?

We try to minimize the image size, especially minimal-notebook(it makes sense to add this to minimal notebook, because git is also installed there as you said).

mathbunnyru:
The author doesn't seem to be interested, so closing this issue.

romeokienzler:
@mathbunnyru Can you please re-open?

Created PR #1895 

New size of image: 3.82 GB

costa:
I'm sorry to report that as of today, this extension does not seem to work (I'm not sure any other extensions work because I need just this one).

I run JupyuterLab "locally" all-fresh with:
```docker-compose.yml
version: '2.1'
services:
  jupy:
    image: jupyter/scipy-notebook
    command: start-notebook.sh --IdentityProvider.token=''
```

If anyone can verify this and prove me wrong/right — or provide any other assistance, it will be greatly appreciated.

Thank you.

mathbunnyru:
@costa unfortunately, I don't think this extension was adapted for JupyterLab 4.x, which is now used in our images.
The upstream issue:
https://github.com/jupyterlab/jupyterlab-git/issues/1245
If you can dedicate your time and help fixing the issue in the upstream project, that would be great.

If you want to use, you can either downgrade jupyterlab in an inherited image or use `jupyter/scipy-notebook:lab-3.6.3` (this image is not up-to-date though).

costa:
@mathbunnyru Thank you, subscribed to the upstream issue (I hope the Jupyter team will be much more efficient fixing it than myself could be), `lab-3.6.3` seems to work for me.

PR #1895, opened 2023-04-25
add jupyterlab-git extension

romeokienzler:
## Describe your changes
add jupyterlab-git extension

New size of the image: 3.82GB

## Issue ticket if applicable

 Fix: #1464 


scipy-notebook/Dockerfile
```diff
@@ -57,7 +57,8 @@ RUN mamba install --yes \
     'statsmodels' \
     'sympy' \
     'widgetsnbextension'\
-    'xlrd' && \
+    'xlrd'\
+    'jupyterlab-git' && \
```
mathbunnyru, line 61: Let's keep this list sorted.

mathbunnyru (changes requested):

mathbunnyru (changes requested):
You will also need to update the docs here: https://github.com/jupyter/docker-stacks/blob/main/docs/using/selecting.md#jupyterscipy-notebook

scipy-notebook/Dockerfile
romeokienzler: @mathbunnyru pushed

romeokienzler:
@mathbunnyru added documentation

mathbunnyru:
@romeokienzler great 👍 
Please, do not forget to add an exception around here, because `jupyterlab-git` is not importable: https://github.com/jupyter/docker-stacks/blob/main/tests/base-notebook/test_packages.py#L72
This will fix CI tests.
You can also get familiar with how our tests work [here](https://jupyter-docker-stacks.readthedocs.io/en/latest/contributing/tests.html)

mathbunnyru (changes requested):
Waiting for tests fix.

romeokienzler:
Thanks @mathbunnyru, done

mathbunnyru (approved):
As this seems to be really popular package for jupyterlab, I think it makes sense to add it, so I'm ok with this change.

PR #1896, opened 2023-04-26
feat(ci): enable `pip` caching in CI

SauravMaheshkar:
Changes made by this PR can be summarized as follows:

- Set the `cache` and `cache-dependency-path` argument to enable caching to speed up CI times.

### DIFF
--- a/.github/actions/create-dev-env/action.yml
+++ b/.github/actions/create-dev-env/action.yml
@@ -13,14 +13,10 @@ runs:
     # actions/setup-python doesn't support Linux aarch64 runners
     # See: https://github.com/actions/setup-python/issues/108
     # python3 is manually preinstalled in the aarch64 VM self-hosted runner
-    - name: Set Up Python 🐍
+    - name: Set Up Python 🐍 and install dev dependencies 📦
       uses: actions/setup-python@v4
       with:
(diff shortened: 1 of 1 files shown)


.github/actions/create-dev-env/action.yml
```diff
@@ -13,14 +13,10 @@ runs:
...
         python-version: 3.x
+        cache: "pip"
+        cache-dependency-path: "**/requirements*.txt"
       if: ${{ inputs.platform == 'x86_64' }}
-
-    - name: Install Dev Dependencies 📦
-      run: |
-        pip install --upgrade pip
-        pip install --upgrade -r requirements-dev.txt
-      shell: bash
```
mathbunnyru, line 26: We have our own aarch64 Github Actions runners.
This won't be executed there, so, whenever someone changes requirements.txt, we will have to manually go to our runners and install packages.
So, unfortunately, I don't think we can get rid of this and that's why we can' use caching.

mathbunnyru (changes requested):

.github/actions/create-dev-env/action.yml
SauravMaheshkar: Ahh makes sense.

Issue #1901, opened 2023-05-23
[BUG] - NodeJS >= 18.0 requires jupyter lab build command


mrkn:
### What docker image(s) are you using?

base-notebook

### Host OS system and architecture running docker image

Ubuntu 22.04 / x86_64

### What Docker command are you running?

`docker run --rm -it jupyter/base-notebook:x86_64-b378681adad0 jupyter lab build`

### How to Reproduce the problem?

1. Run the command below on the host OS system

```
docker run --rm -it jupyter/base-notebook:x86_64-b378681adad0 jupyter lab build
```

### Command output

```bash session
[LabBuildApp] JupyterLab 4.0.0
[LabBuildApp] Building in /opt/conda/share/jupyter/lab
[LabBuildApp] ERROR | Build failed.
Troubleshooting: If the build failed due to an out-of-memory error, you
may be able to fix it by disabling the `dev_build` and/or `minimize` options.

If you are building via the `jupyter lab build` command, you can disable
these options like so:

jupyter lab build --dev-build=False --minimize=False

You can also disable these options for all JupyterLab builds by adding these
lines to a Jupyter config file named `jupyter_config.py`:

c.LabBuildApp.minimize = False
c.LabBuildApp.dev_build = False

If you don't already have a `jupyter_config.py` file, you can create one by
adding a blank file of that name to any of the Jupyter config directories.
The config directories can be listed by running:

jupyter --paths

Explanation:

- `dev-build`: This option controls whether a `dev` or a more streamlined
`production` build is used. This option will default to `False` (i.e., the
`production` build) for most users. However, if you have any labextensions
installed from local files, this option will instead default to `True`.
Explicitly setting `dev-build` to `False` will ensure that the `production`
build is used in all circumstances.

- `minimize`: This option controls whether your JS bundle is minified
during the Webpack build, which helps to improve JupyterLab's overall
performance. However, the minifier plugin used by Webpack is very memory
intensive, so turning it off may help the build finish successfully in
low-memory environments.

/opt/conda/lib/python3.10/site-packages/jupyterlab/debuglog.py:56: UserWarning: An error occurred.
  warnings.warn("An error occurred.")
/opt/conda/lib/python3.10/site-packages/jupyterlab/debuglog.py:57: UserWarning: ValueError: Please install nodejs >=18.0.0 before continuing. nodejs may be installed using conda or directly from the nodejs website.
  warnings.warn(msg[-1].strip())
[...]

PR #1902, opened 2023-05-23
Install nodejs >= 18.0

mrkn:
## Describe your changes

NodeJS >= 18.0 is installed in base-notebook image because `jupyter lab build` command requires it.

## Issue ticket if applicable

Fix: #1901 


mathbunnyru (changes requested):
Please, add an exception like this to the packages import test:
https://github.com/jupyter/docker-stacks/blob/main/tests/base-notebook/test_packages.py#L72

PR #1914, opened 2023-06-03
Do not pin nodejs version

mathbunnyru:


mathbunnyru:
It seems nodejs 18 is getting installed even without pin, so we can remove this. Hope it won't change.

Issue #1924, opened 2023-06-21
[BUG] - Pin pandas version for spark images


bjornjorgensen:
### What docker image(s) are you using?

pyspark-notebook

### Host OS system and architecture running docker image

NA

### What Docker command are you running?

We need to pin `pandas` version to 1.5.3 because spark's `pandas API on spark` does not work 100 % with pandas version 2. 

https://issues.apache.org/jira/browse/SPARK-43291?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel 

### How to Reproduce the problem?

NA

### Actual behavior

NA

### Latest Docker version

- [X] I've updated my Docker version to the latest available, and the issue still persists

PR #1925, opened 2023-06-23
Pin `pandas` to version 1.5.3 to all spark images

bjornjorgensen:
## Describe your changes
This will install `pandas` version 1.5.3 for all spark images.
Spark have a module called `pandas API on spark`, this module uses `pandas` for many tasks.
Apache spark have decided not to upgrade `pandas API on spark` to version 2.0 before Apache spark 4.0 
https://lists.apache.org/thread/jbxrpf81flr6rk1poj9mqjyl8p34ro6w

## Issue ticket if applicable
Fix: #1924


bjornjorgensen:
@mathbunnyru do you understand this? :)

tests/all-spark-notebook/test_spark_notebooks.py
```diff
@@ -38,3 +38,21 @@ def test_nbconvert(container: TrackedContainer, test_file: str) -> None:
 
     expected_file = f"{output_dir}/{test_file}.md"
     assert expected_file in logs, f"Expected file {expected_file} not generated"
+
+
+def test_pandas_version(container: TrackedContainer) -> None:
```
mathbunnyru, line 43: We don't need to repeat tests. If we have a test for some image, I made it so this test will be run for all inherited images as well.

pyspark-notebook/Dockerfile
```diff
@@ -64,9 +64,10 @@ RUN fix-permissions "/etc/ipython/"
 
 USER ${NB_UID}
 
-# Install pyarrow
+# Install pyarrow and pin pandas to version 1.5.3 see #1924
 RUN mamba install --yes \
-    'pyarrow' && \
+    'pyarrow' \
+    'pandas=1.5.3' && \
```
mathbunnyru, line 70: I think a better way would be to install >=1.5.3 and <2.0.0, and then actually pin it (so it won't be updated in the `all-spark-notebook`).

tests/pyspark-notebook/test_spark.py
```diff
@@ -16,3 +16,21 @@ def test_spark_shell(container: TrackedContainer) -> None:
...
+        "-c",
+        "python -c 'import pandas; print(pandas.__version__)'",
+    ]
+    logs = container.run_and_wait(
+        timeout=60,
+        tty=True,
+        command=command,
+    )
+
+    assert "1.5.3" in logs, "pandas version is not 1.5.3"
```
mathbunnyru, line 36: I actually meant trying to run this inside docker image:

```
pdf = pd.DataFrame([("1", "2"), ("0", "3"), ("2", "0"), ("1", "1")], columns=["a", "b"])
psdf = ps.from_pandas(pdf)
self.assert_eq(pdf.cov(), psdf.cov())
```

I'm not sure it will work though.

tests/pyspark-notebook/test_spark.py
```diff
@@ -16,3 +16,21 @@ def test_spark_shell(container: TrackedContainer) -> None:
...
+
+
+def test_pandas_version(container: TrackedContainer) -> None:
+    """Checking if pandas is installed and its version is 1.5.3
+    This is only needed for Spark before 4.0 see #1924"""
+    command = [
+        "start.sh",
+        "bash",
+        "-c",
+        "python -c 'import pandas; print(pandas.__version__)'",
```
mathbunnyru, line 28: Actually, we have a simple way of running python scripts inside a container.
Take a look here: https://jupyter-docker-stacks.readthedocs.io/en/latest/contributing/tests.html#unit-tests

tests/pyspark-notebook/test_spark.py
bjornjorgensen: it is enough to test for the pandas version

mathbunnyru: Ok.

pyspark-notebook/Dockerfile
```diff
@@ -64,9 +64,10 @@ RUN fix-permissions "/etc/ipython/"
 
 USER ${NB_UID}
 
-# Install pyarrow
+# Install pyarrow and pin pandas to version 1.5.3 see #1924
```
mathbunnyru, line 67:
```suggestion
# Install pyarrow
# Temporarily pin pandas to version 1.5.3, see: https://github.com/jupyter/docker-stacks/issues/1924 
```

pyspark-notebook/Dockerfile
```diff
@@ -64,9 +64,10 @@ RUN fix-permissions "/etc/ipython/"
...
 RUN mamba install --yes \
-    'pyarrow' && \
+    'pyarrow' \
+    'pandas>=1.5.3 and <2.0.0' && \
```
mathbunnyru, line 70:
```suggestion
    'pandas>=1.5.3,<2.0.0' \
    'pyarrow' && \
```

tests/pyspark-notebook/test_spark.py
```diff
@@ -16,3 +16,20 @@ def test_spark_shell(container: TrackedContainer) -> None:
...
+        tty=True,
+        command=[
+            "start.sh",
+            "python",
+            "-c",
+            "import pandas; print(pandas.__version__)",
+        ],
+    )
+
+    assert "1.5.3" in logs, "pandas version is not 1.5.3"
```
mathbunnyru, line 35: Instead of all this code, just add a file [here](https://github.com/jupyter/docker-stacks/tree/main/tests/pyspark-notebook/units) with the following content:

```

import pandas
assert pandas.__version__ == "1.5.3"
```

mathbunnyru: This is much easier and works well.

mathbunnyru (changes requested):
I added a few suggestions, please, take a look.

mathbunnyru (approved):

bjornjorgensen:
I think that we will see Apache Spark 4.0.0 in jan. or feb. 2024 and then we can remove this.

mathbunnyru:
> I think that we will see Apache Spark 4.0.0 in jan. or feb. 2024 and then we can remove this.

Great, thanks.

bjornjorgensen:
The error is for cache?

mathbunnyru:
> The error is for cache?

No, we need to add a new exception here: https://github.com/jupyter/docker-stacks/blob/main/tests/base-notebook/test_packages.py#L67

bjornjorgensen:
hmm.. ok, so it just to add pandas to that list?

tests/base-notebook/test_packages.py
```diff
@@ -75,6 +75,7 @@
     "python",
     "r-irkernel",
     "unixodbc",
+    "pandas",
```
mathbunnyru, line 78: You need to actually add `"pandas[version='>"` to this list.
And please, keep it sorted.

tests/base-notebook/test_packages.py
```diff
@@ -71,6 +71,7 @@
     "hdf5",
     "jupyterlab-git",
     "openssl",
+    "pandas[version='>1.5.3,<2.0.0",
```
mathbunnyru, line 74: As far as I remember, it should be exactly as I wrote it in a comment above.

bjornjorgensen: "pandas[version='>" strange but I will give it a try :)

mathbunnyru: Yep, I think the version parsing was written quite a while ago and some format has changed, so it would be nice to update that script, but that doesn't give much trouble.

mathbunnyru:
@bjornjorgensen thank you!
I finally merged this :)
The image should update in approximately 2 hours.

PR #1928, opened 2023-06-28
Remove pre-commit github action

yuvipanda:
## Describe your changes

pre-commit.ci already runs these and automatically fixes these, so the GitHub action is redundant!


.github/workflows/pre-commit.yml
```diff
@@ -1,33 +0,0 @@
...
-        with:
-          python-version: 3.x
-
-      - name: Install pre-commit 📦
-        run: |
-          pip install --upgrade pip
-          pip install --upgrade pre-commit
-
-      - name: Run pre-commit hooks ✅
-        run: pre-commit run --all-files --hook-stage manual
```
consideRatio, line 33: Only the "manual" hook stage runs, i recall this was because pre-commit.ci couldn't run all hooks at least historically.

The pre-commit config should help us clarify the situation in more detail

/ From mobile

mathbunnyru: @consideRatio is right :)

The explanation is here:
https://github.com/jupyter/docker-stacks/blob/main/.pre-commit-config.yaml#L44-L51

```
      # Unfortunately, `pre-commit` only runs on changed files
      # This doesn't work well with `mypy --follow-imports error`
      # See: https://github.com/pre-commit/mirrors-mypy/issues/34#issuecomment-1062160321
      #
      # To work around this we run `mypy` only in manual mode
      # So it won't run as part of `git commit` command
      # But it will still be run as part of `pre-commit` workflow and give expected results
      stages: [manual]
```

yuvipanda: Aaah sorry for not looking through that in more detail!!!!

PR #1929, opened 2023-06-29
Install Pluto.jl and jupyter-pluto-proxy

yuvipanda:
## Describe your changes

[Pluto.jl](https://plutojl.org/) is an alternative reactive notebook frontend focused specifically on Julia. I think shipping this by default in the julia-enabled images helps serve the Julia community better, particularly when used with JupyterHub.

For context, I am working with the Julia users of the [Jupyter Meets the Earth](https://jupytearth.org/) project, and trying to understand how to best serve their needs on a JupyterHub. We currently maintain a massive image that 'has everything', but I'm trying to instead work upstream wherever possible so everyone working in these subfields can benefit. Meeting Julia users where they are at seems a useful path forward here.


yuvipanda:
If this looks acceptable, I'll write a specific test for this too.

docs/using/selecting.md
```diff
@@ -118,6 +118,7 @@ It contains:
 - Everything in `jupyter/minimal-notebook` and its ancestor images
 - The [Julia Programming Language](https://julialang.org/)
 - [IJulia](https://github.com/JuliaLang/IJulia.jl) to support Julia code in Jupyter notebook
+- [Pluto.jl](https://plutojl.org/) reactive Julia notebook interface, made accessible with [jupyter-pluto-proxy](https://github.com/yuvipanda/jupyter-pluto-proxy)
```
mathbunnyru, line 121: This has to be added to 2 places, unfortunately.

tests/base-notebook/test_packages.py
```diff
@@ -93,7 +93,7 @@ def packages(package_helper: CondaPackageHelper) -> dict[str, set[str]]:
 
 def get_package_import_name(package: str) -> str:
     """Perform a mapping between the python package name and the name used for the import"""
-    return PACKAGE_MAPPING.get(package, package)
+    return PACKAGE_MAPPING.get(package, package).replace("-", "_")
```
mathbunnyru, line 96: Let's just add one more package to PACKAGE_MAPPING?

mathbunnyru (commented):
To be honest, I know nothing about Julia or Julia's infrastructure.
So, @yuvipanda, if you know someone from Julia (+Jupyter) community, who can take a look here and approve this change, this would be awesome.
As I understand - do we need to run something special to run Pluto?
If yes, maybe we should probably create a recipe with how to do that.
https://jupyter-docker-stacks.readthedocs.io/en/latest/contributing/recipes.html

tests/base-notebook/test_packages.py
yuvipanda: Done!

docs/using/selecting.md
yuvipanda: Done!

yuvipanda:
@mathbunnyru you don't need anything special to use it. You can see it in the JupyterLab launcher!

image

So I don't think a specific recipe is needed.

yuvipanda:
I'm going to add a simple test to make sure this works.

yuvipanda:
ok, I added a test to make sure this starts! I've also asked some julia users to try comment on this, just as social proof :) 

Anything else you'd like me to do, @mathbunnyru?

mathbunnyru (approved):
Thank you for adding the test.
Let's use the DRY principle and move the test code to a function in `tests/julia-tests/pluto_proxy.py` folder and in the `julia-notebook` and `datascience-notebook` folders add a test which will be calling this function.

JordiBolibar:
Thanks a lot for all the efforts on building a Julia-specific image for the hub! 

I've been heavily using the "massive" image from Jupyter Meets the Earth the last 2.5 years, and I can say that adding Pluto.jl would be fantastic. Having Julia + Pluto.jl + VSCode + Jupyter notebooks with CPUs is pretty much everything we could ask for 🙂

benz0li:
@JordiBolibar There is also the [JupyterLab Julia docker stack](https://github.com/b-data/jupyterlab-julia-docker-stack) and the [Data Science Development Containers](https://github.com/b-data/data-science-devcontainers).  
ℹ️ (GPU accelerated) Multi-arch (`linux/amd64`, `linux/arm64/v8`) container images

mathbunnyru:
> Thanks a lot for all the efforts on building a Julia-specific image for the hub!
> 
> I've been heavily using the "massive" image from Jupyter Meets the Earth the last 2.5 years, and I can say that adding Pluto.jl would be fantastic. Having Julia + Pluto.jl + VSCode + Jupyter notebooks with CPUs is pretty much everything we could ask for 🙂

Great 👍
It always feels great to make users of our images happy 🙂

mathbunnyru:
I think I'll just merge this and then remove code duplication in the `main` branch.
I want to put all the images to the `images` folder and this is blocking me a little bit.

yuvipanda:
Thank you so much, @mathbunnyru! And thanks for chiming in, @JordiBolibar!

PR #1930, opened 2023-07-01
Refactor handling of artifacts

mathbunnyru:


mathbunnyru:
I hit a low disk space issue once again. I think it's better to save all the thing in one folder and then to delete it before running a pipeline.
I also noticed some big temporary files were created in `/tmp` (probably during `docker build` or `docker save`), I will try to deal with this issue later.

PR #1938, opened 2023-07-07
Fix image tagging in hook/%

twalcari:
## Describe your changes

The `hook/%` target refers to the script `tagging.tag_image` which does not exist anymore. This PR updates Makefile to use `tagging.write_tags_file` and `tagging.apply_tags` instead (identical to what is being done in Github Actions).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,8 @@ linkcheck-docs: ## check broken links
 
 
 hook/%: ## run post-build hooks for an image
-	python3 -m tagging.tag_image --short-image-name "$(notdir $@)" --owner "$(OWNER)" && \
+	python3 -m tagging.write_tags_file --short-image-name "$(notdir $@)" --tags-dir /tmp/jupyter/tags/ --owner "$(OWNER)" && \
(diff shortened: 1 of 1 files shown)


twalcari:
Open question: should we move executing `tagging.apply_tags` to the `push/%` target instead of the `hook/%` target? That way the Makefile stays more in sync with how Github Actions workflow.

mathbunnyru:
> The hook/% target refers to the script tagging.tag_image which does not exist anymore. This PR updates Makefile to use tagging.write_tags_file and tagging.apply_tags instead (identical to what is being done in Github Actions).

Thank you!
Makefile is not used by this project at all, to be honest, but I keep it for backward compatibility, if some forks use it.
It can also be used during local development (and sometimes I use it).

> Open question: should we move executing `tagging.apply_tags` to the `push/%` target instead of the `hook/%` target? That way the Makefile stays more in sync with how Github Actions workflow.

I think let's keep `tagging.apply_tags` in `hook/%`, because this way we can easily run this script locally (`push/%` can't be used at all, to be honest, but some forks might still use it though).

Issue #1324, opened 2021-05-22
Verify contributed recipes


mathbunnyru:
Contributed recipes are not checked in CI, were mostly added a long time ago and so they might be out-of-date.
It would be nice if someone checked them, updated packages versions used there and verified that they still work.

consideRatio:
Perhaps we should include a date for when they were added to the list, and allow that date to be updated whenever it is known to "be functional" (vague, but it's fine I think).

mathbunnyru:
Yes, something like `Last verified` near each recipe should be fine.

mathbunnyru:
An example: https://github.com/jupyter/docker-stacks/pull/1617

mathbunnyru:
One more example: https://github.com/jupyter/docker-stacks/issues/1951

mathbunnyru:
I had an idea on how to fix this:

1. Express all recipes as Dockerfiles - done in several commits today.
2. Create a separate file for each Dockerfile in the docs.
   They will be called `feature.dockerfile`
   MyST: https://myst-parser.readthedocs.io/en/latest/syntax/code_and_apis.html#including-code-from-files
   Sphinx: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude
3. Enable pre-commit hooks on these Dockerfile - default hadolint doesn't work well with `.dockerfile`s.
4. The most interesting part - create a separate GitHub Workflow, which will build these recipes. So, our recipes will always be buildable.
5. Mark outdated recipes as outdated. If someone want to fix them, they will easily use my workflow to verify the recipe is working fine.

After these steps I will consider this issue closed and even to be future-proof (because we'll have automatic tests).

mathbunnyru:
@consideRatio I think you're going to like this 🙂

mathbunnyru:
Done.
We currently have 8 working (and automatically tested) Dockerfiles.
And 13 untested and possibly broken ones.
They are marked as warnings in documentation, so people will be aware there might be some problems.
I will be happy to merge fixes for these recipes.

consideRatio:
> @consideRatio I think you're going to like this 🙂

Yes I love it!!! Amazing work as always @mathbunnyru!!!

mathbunnyru:
@consideRatio a bit more improvements here:
GitHub allows to generate a matrix for execution, that's why I don't need to hardcode list of docker images to build and just build them all: https://github.com/jupyter/docker-stacks/pull/1973
[...]

PR #1955, opened 2023-07-28
Build recipes in GitHub Actions

mathbunnyru:
## Issue ticket if applicable

Work on: https://github.com/jupyter/docker-stacks/issues/1324

### DIFF
--- a/.github/workflows/contributed-recipes.yml
+++ b/.github/workflows/contributed-recipes.yml
@@ -36,7 +36,7 @@ jobs:
 
     strategy:
       matrix:
-        docker:
+        dockerfile:
           [
             dask_jupyterlab.dockerfile,
             jupyterhub_version.dockerfile,
--- a/.github/workflows/contributed-recipes.yml
+++ b/.github/workflows/contributed-recipes.yml
@@ -1,0 +1,47 @@
+name: Test contributed recipes
+
+on:
+  schedule:
+    # Images are rebuilt on Monday, so we're testing recipes each Tuesday
+    # Weekly, at 03:00 on Tuesday UTC time
+    - cron: "0 3 * * 2"
+  pull_request:
+    paths:
+      - ".github/workflows/contribured-recipes.yml"
+      - "docs/using/recipe_code/"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/contribured-recipes.yml"
+      - "docs/using/recipe_code/"
+
+jobs:
+  test-recipes:
+    runs-on: ubuntu-latest
+    if: github.repository == 'jupyter/docker-stacks'
+
+    steps:
+      - name: Checkout Repo ⚡️
+        uses: actions/checkout@v3
+
+      - name: Build recipe 🛠
+        run: docker build --rm --force-rm --tag my-custom-image -f ./${{ matrix.dockerfile }} ./
+        env:
+          DOCKER_BUILDKIT: 1
+          # Full logs for CI build
+          BUILDKIT_PROGRESS: plain
+        working-directory: docs/using/recipe_code
+        shell: bash
+
+    strategy:
+      matrix:
+        docker:
+          [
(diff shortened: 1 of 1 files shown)


.github/workflows/contributed-recipes.yml
```diff
@@ -0,0 +1,47 @@
...
+  pull_request:
+    paths:
+      - ".github/workflows/contributed-recipes.yml"
+      - "docs/using/recipe_code/"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/contributed-recipes.yml"
+      - "docs/using/recipe_code/"
```
consideRatio, line 17: Perfect! I think this makes perfect sense, nice work!

mathbunnyru: @consideRatio

I'm not sure if other projects test their docs, but I really like that this one does, because I had (relatively) many complaints that some recipe was not working.
The coolest thing about this approach is that I can even insert tests (if I want to) to the dockerfiles and not to include them in docs.

https://myst-parser.readthedocs.io/en/latest/syntax/code_and_apis.html#including-code-from-files
There is a really useful option:
```
:end-before: end example
```
So, tests can be any usual RUN/COPY or other docker steps and they will be tested by my workflow, but no one will see them in the docs.

consideRatio: Oh thanks for sharing, great idea of using that MyST feature (which I learned about now)!

> many complaints that some recipe was not working.

Yes! I've grown a bit cautious about documentation additions because of such maintenance needs arising. Your work with this is whats required to go hand in hand with growning amounts of recipies I figure!

I think your idea on how to handle this is very promising, I wonder if it can be combined with a badge or note saying when it was last tested and if it was successful at that time or not. For that I think one would need to use some REST API first:
- List runs of a workflow: https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-workflow
- List jobs in a workflow run: https://docs.github.com/en/rest/actions/workflow-jobs?apiVersion=2022-11-28#list-jobs-for-a-workflow-run

Hmmm, but that API needs to be done during doc build, which would freeze the info into the docs built HTML and not update over time if it isn't rebuild. So, maybe stashing away the status to be dynamically read from the docs HTML, for example as part of running test jobs or as a job that `needs` the test job (run multiple times) updating some JSON blob to be read from the built documentation directly.

If so, then one could use a badge from shields.io (more about that both at https://shields.io/badges/dynamic-json-badge and https://shields.io/badges/endpoint-badge). As an example of a shields.io based Dynamic JSON badge, see this badge showing the latest version of the jupyterhub Helm chart based on [a json blob](https://hub.jupyter.org/helm-chart/info.json) we update elsewhere:

[Latest pre-release of the Helm chart](https://jupyterhub.github.io/helm-chart#development-releases-jupyterhub)

```markdown
[![Latest pre-release of the Helm chart](https://img.shields.io/badge/dynamic/json.svg?label=pre&url=https://hub.jupyter.org/helm-chart/info.json&query=$.jupyterhub.pre&colorB=orange&logo=helm)](https://jupyterhub.github.io/helm-chart#development-releases-jupyterhub)
```

mathbunnyru: I will think about this, I don't have the answer to this currently.

> Hmmm, but that API needs to be done during doc build, which would freeze the info into the docs built HTML and not update over time if it isn't rebuild.

I might have a solution to this problem, but it's not for everyone, I suppose.
Use scheduled builds.
Yes, you will have more builds (so, more CPU spent), more failing builds and will need to add a few lines of code to GitHub Workflows. But at the same time, you will catch bugs earlier, if they happen - and they are much easier to track down, when you know it started occurring only in the last week, not since the previous build.
I suppose it's easy to add a badge with the date of build.
And it's easy to add a badge that the GitHub workflow was last successful or not.

Issue #1962, opened 2023-08-09
[BUG] - Healthcheck fails when using proxy


MatthiasJobst:
### What docker image(s) are you using?

base-notebook

### Host OS system and architecture running docker image

Windows 11 as host and linux/amd64 for docker

### What Docker command are you running?

docker compose up with the following dockerfile:

```Dockerfile
version: '3.4'

services:
  datamining:
    container_name: xxxx
    image: xxxx
    build:
      context: .
      dockerfile: ./Dockerfile
    ports:
      - "8888:8888"
    volumes:
      - xxxx:/home/jovyan/work
    environment:
      - DOCKER_STACKS_JUPYTER_CMD=lab
    restart: on-failure
```

### How to Reproduce the problem?

Precondition is that the machine has to operate in a corporate environment using the companies proxy.
Start the container as above.
Check the state of the container with ```docker container ls```
The container is marked as unhealthy.

### Command output

```bash session
abcdefghijk    "tini -g -- start-no…"   x hours ago     Up x hours (unhealthy)        0.0.0.0:8888->8888/tcp                             xxxx
```

### Expected behavior

```abcdedfghi   abcdefghijk    "tini -g -- start-no…"   x hours ago     Up x hours (healthy)        0.0.0.0:8888->8888/tcp                             xxxx```

### Actual behavior

After investigating the issue the problem is that docker_healthcheck.py does not run successfully giving the following error message:
```
Traceback (most recent call last):
  File "/opt/conda/lib/python3.11/site-packages/urllib3/connectionpool.py", line 790, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/urllib3/connectionpool.py", line 536, in _make_request
    response = conn.getresponse()
               ^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/urllib3/connection.py", line 461, in getresponse
    httplib_response = super().getresponse()
                       ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/http/client.py", line 1378, in getresponse
    response.begin()
  File "/opt/conda/lib/python3.11/http/client.py", line 318, in begin
    version, status, reason = self._read_status()
                              ^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/http/client.py", line 287, in _read_status
[...]

PR #1964, opened 2023-08-12
Healthcheck gives correct result even behind proxy

MatthiasJobst:
## Changes
For environments that use proxies, mostly corporate environments, the health check did not work. Proxies are not necessary for the health check so the proxies are disabled.
The health check runs in the docker container and makes a request to a specific address. If that request is routed through a proxy it leaves the docker container. This can cause two types of problems:
- The proxy does not know the ip address and thus the request never terminates
- Outside of the container the ports can change and the request will not reach the proper port.

## Issue ticket

Closed [1962](https://github.com/jupyter/docker-stacks/issues/1962)

- [-] I have updated the documentation for significant changes

### DIFF
--- a/base-notebook/docker_healthcheck.py
+++ b/base-notebook/docker_healthcheck.py
@@ -17,8 +17,8 @@ url = json.loads(json_file.read_bytes())["url"]
 url = url + "api"
 
 proxies = {
-  "http": None,
-  "https": None,
+    "http": None,
+    "https": None,
 }
 
 r = requests.get(url, proxies=proxies, verify=False)  # request without SSL verification
--- a/base-notebook/docker_healthcheck.py
+++ b/base-notebook/docker_healthcheck.py
@@ -16,6 +16,11 @@ json_file = next(runtime_dir.glob("*server-*.json"))
 url = json.loads(json_file.read_bytes())["url"]
 url = url + "api"
 
-r = requests.get(url, verify=False)  # request without SSL verification
+proxies = {
+  "http": None,
+  "https": None,
+}
+
+r = requests.get(url, proxies=proxies, verify=False)  # request without SSL verification
 r.raise_for_status()
 print(r.content)
--- a/base-notebook/docker_healthcheck.py
+++ b/base-notebook/docker_healthcheck.py
@@ -17,8 +17,8 @@ url = json.loads(json_file.read_bytes())["url"]
 url = url + "api"
 
 proxies = {
-    "http": None,
-    "https": None,
+    "http": "",
+    "https": "",
 }
 
 r = requests.get(url, proxies=proxies, verify=False)  # request without SSL verification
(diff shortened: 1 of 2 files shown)


base-notebook/docker_healthcheck.py
```diff
@@ -16,6 +16,11 @@
 url = json.loads(json_file.read_bytes())["url"]
 url = url + "api"
 
-r = requests.get(url, verify=False)  # request without SSL verification
+proxies = {
+    "http": None,
+    "https": None,
+}
```
mathbunnyru, line 22:
```suggestion
proxies = {
    "http": "",
    "https": "",
}
```

This should solve mypy issue.
There seems to be some difference  though.
https://stackoverflow.com/a/72529216/4881441

mathbunnyru (approved):
Overall, LGTM and thank you for adding a test 👍

MatthiasJobst:
The tests that fail do not have anything to do with the test I added. Interestingly they seem to fail without the changes too? 
Maybe some deprecation issue with notebook?

mathbunnyru:
> The tests that fail do not have anything to do with the test I added. Interestingly they seem to fail without the changes too?
> Maybe some deprecation issue with notebook?

Let's hope it was a spurious fail. I restarted failed GitHub jobs.

mathbunnyru:
I guess the reason is that for some reason `mamba` decides to downgrade `jupyterlab` and `notebook`:
```
#6 394.0   Change:
#6 394.0 ────────────────────────────────────────────────────────────────────────────────────────
#6 394.0 
#6 394.0   - _openmp_mutex                  4.5  2_gnu                    conda-forge      24kB
#6 394.0   + _openmp_mutex                  4.5  2_kmp_llvm               conda-forge       6kB
#6 394.0 
#6 394.0   Downgrade:
#6 394.0 ────────────────────────────────────────────────────────────────────────────────────────
#6 394.0 
#6 394.0   - notebook                     7.0.2  pyhd8ed1ab_0             conda-forge       3MB
#6 394.0   + notebook                     6.5.4  pyha770c72_0             conda-forge     306kB
#6 394.0   - jupyterlab                   4.0.4  pyhd8ed1ab_0             conda-forge       6MB
#6 394.0   + jupyterlab                   3.6.5  pyhd8ed1ab_0             conda-forge       5MB
```

I also see a warning because of `jupyter_lsp` not being able to run.
Probably jupyter app doesn't start properly after downgrade and we get an error.
It's definitely your change and I'll need to fix it in the `main` branch.
I will take a look in a few days in more details.

mathbunnyru:
@MatthiasJobst I fixed (actually, mitigated) the issue in the main branch and merged it here.
Hopefully, tests will pass and then I'll merge this PR.

MatthiasJobst:
Thanks. That was quick and solved the issue.

Issue #1969, opened 2023-08-17
[ENH] - /home/jovyan/work is confusing (documentation)


4kbyte:
### What docker image(s) is this feature applicable to?

scipy-notebook

### What change(s) are you proposing?

User Guide documentation suggests mounting a local directory (`$PWD`, etc.) to `/home/jovyan/work` to persist notebooks. An excellent suggestion, let's keep our data around.

But at no point in _Quick Start_, _Selecting an Image_, _Running a Container_, or _Common Features_ does the documentation instruct you that by default, the notebook will save to `/home/jovyan`.

### How does this affect the user?

The user will discover that their data didn't persist only upon running a new container.

Further, there's no immediately available troubleshooting topic or search query that will illuminate that you didn't click "work" in the left sidebar of jupyter-server. The natural inclination is to click the large, friendly Python logo to get a Python notebook, since after all, that's why your here.

But that notebook ends up in `/home/jovyan`.

### Anything else?

I was running:

```
docker run --rm --name=jupyter \
  -p 8888:8888 -v $(pwd):/home/jovyan/work \
  -e RESTARTABLE=yes \
  jupyter/scipy-notebook:python-3.11.4 "$@"
```

I'm using jupyter for instructional purposes. My resolution is to mount `$PWD` to `/home/jovyan` (without the `work` folder) and accept that I end up with extra files on the host from jovyan's `$HOME` (which I presume could be an issue when selecting another image at a later date). Not a problem, since I'm primarily concerned with making sure I don't lose any .ipynb files.

PR #1971, opened 2023-08-18
Add info about root_dir in docs

mathbunnyru:
## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/1969

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ system when the container exits, but any changes made to the `~/work` directory 
 By default, [jupyter's root_dir](https://jupyter-server.readthedocs.io/en/latest/other/full-config.html) is `/home/jovyan`.
 So, new notebooks will be saved there, unless you change the directory in the file browser.
 
-To change the default behaviour, you will need to specify `ServerApp.root_dir` by adding this line to previous command: `start-notebook.sh --ServerApp.root_dir=/home/jovyan/work`.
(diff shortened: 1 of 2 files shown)


mathbunnyru:
@4kbyte please, take a look her.

mathbunnyru:
> @mathbunnyru tyvm! This clarifies the exact issue in #1969 and I wouldn't have known about the `--ServerApp.root_dir` option without your note.
> 
> Could we add this to the user guide as well? We could insert it as Example 3 (and renumber examples after) so that it clarifies and responds to the last sentence of Example 2, "Any other changes made in the container will be lost."
> 
> > Example #
> > This command pulls the `jupyter/scipy-notebook` image tagged with a specific version of Python from Docker Hub if it is not already present on the local host. It then starts a container running Jupyter Server with its root directory set as the current directory on the host machine. New files and changes within Jupyter are then saved regardless of which directory is selected in the file browser.
> > ```
> > docker run --rm --name=jupyter \
> >   -p 8888:8888 -v "${PWD}":/home/jovyan/work \
> >   jupyter/scipy-notebook:python-3.11.4 \
> >     start-notebook.sh --ServerApp.root_dir=/home/jovyan/work
> > ```

Note that README.md is part of the docs as well.
You can see how it's gonna look like as well: https://jupyter-docker-stacks--1971.org.readthedocs.build/en/1971/#quick-start

mathbunnyru:
Ok, I added this to running.md as well.

PR #1972, opened 2023-08-18
Move all images to images dir

mathbunnyru:
## Describe your changes

I think this is a cleaner approach, especially if we will add new images.


mathbunnyru:
There are 2 downsides:
1. External links will need to be fixed
2. Git history is gonna be more difficult to see

I think we should still move forward with this.

@consideRatio what do you think?

consideRatio:
I think this is fine when the files are moved like they are in this PR, making git understand they are moved as compared to a delete old file + create new file.

Thanks for your care for these kinds of details @mathbunnyru!!

PR #1973, opened 2023-08-18
Build all contributed recipes automatically

mathbunnyru:


mathbunnyru:
This automatic generation is nice - no need to change the workflow code when adding new recipe dockerfile.

Next thing to do - build recipes under aarch64 and using `generate_matrix.py` it is easier to create a list of recipes, which are not ready for aarch64.

Issue #1903, opened 2023-05-24
[ENH] - include conda env activation for kernel with custom environment


creative-resort:
### What docker image(s) is this feature applicable to?

base-notebook

### What changes are you proposing?

I'd like to find out, how to best integrate the activation of a custom environment upon kernel launch.

# Problem, that needs this improvement:
When the kernel in the custom environment is started from JupyterLab, the conda environment is unchanged the "base" environment, where Jupyter is installed and run from, because the .bashrc with its conda environment activation never runs in that context. This means, that the PATH environment variable is also the one of the "base" environment, even if the python executable is from the custom env … resulting in modules, that are not purely Python, but properly installed in the custom env, to fail loading.

If I do a conda custom env activation e.g. via shell script, that I call via the kernelspec, the PATH is correct and loading those modules, that previously failed (e.g. plotly kaleido), works.

# Questions:
1. How should **changes to the kernelspec**, that `ipykernel install` creates, be implemented optimally, in the Dockerfile?
2. Is providing and calling a `kernel-launch.sh` script a good idea, or should just the standard command be augmented with `conda activate customenv` , instead?
```
"argv": [
    "~/.jupyter/lab/launch-kernel.sh",
    "custom-env",
    "{connection_file}"
 ],
```

This is the reference to the custom env kernel creation:
https://github.com/jupyter/docker-stacks/blob/b378681adad0506b3613cde9d8a35c3e246dfe71/docs/using/recipes.md?plain=1#L88-L102

### How does this affect the user?

The user would be able to make use of conda modules, that use the full spectrum of conda environments, right away, without having to wonder, why additional binaries can't be found and without manually manipulating the PATH environment variable or absolute paths.

bribroder:
Sounds like you could use nb_conda_kernels
https://github.com/Anaconda-Platform/nb_conda_kernels

mathbunnyru:
@creative-resort sorry for the really late answer.
I might have a few solutions here (they are untested though).

As you told, we have a recipe for using custom environment.
https://github.com/jupyter/docker-stacks/blob/main/docs/using/recipe_code/custom_environment.dockerfile

It has `RUN echo "conda activate ${env_name}" >> "${HOME}/.bashrc"`.
Unfortunately, `.bashrc` is not run by jupyter, that's why environment is not changed.

1. First way of solving this would be to add `-l` to the `start-notebook.sh` script. You can easily change the first line of this script in your inherited image.
[...]

PR #1975, opened 2023-08-20
Update the custom env recipe to correctly activate it by default

mathbunnyru:
## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/1903

### DIFF
--- a/docs/using/recipe_code/custom_environment.dockerfile
+++ b/docs/using/recipe_code/custom_environment.dockerfile
@@ -28,8 +28,9 @@ RUN "${CONDA_DIR}/envs/${env_name}/bin/python" -m ipykernel install --user --nam
 RUN "${CONDA_DIR}/envs/${env_name}/bin/pip" install --no-cache-dir \
     'flake8'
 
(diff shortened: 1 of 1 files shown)


mathbunnyru:
@ creative-resort please, check, if this works for you.

docs/using/recipe_code/custom_environment.dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM jupyter/base-notebook
 
 # name your environment and choose the python version
-ARG env_name=python38
-ARG py_ver=3.8
+ARG env_name=python310
+ARG py_ver=3.10
```
mathbunnyru, line 5: 3.8 also works fine.
I updated these lines to make sure 3.10 also works.
And this recipe will stay up-to-date a bit longer (till 3.10 reaches eol).

mathbunnyru:
I've checked it on my machine and it seems to work well.
@creative-resort please, tell me if you encounter any issues.
I will merge this PR (and close the issue), because I will need this updated recipe to fix another issue as well.

Issue #1959, opened 2023-08-04
[ENH] - Document correct way to persist conda packages


laundmo:
### What docker image(s) is this feature applicable to?

base-notebook

### What change(s) are you proposing?

Documenting a "blessed" way of persisting installed python packages, such that they are not removed after the container is recreated.

Ideally this should be a way which doesn't require persisting the packages automatically installed by the container

### How does this affect the user?

As seen in
- #1836

its not currently obvious how to correctly do this. Finding this issue is also not easy.

a documentation section for this would help users achieve this quickly without a lot of searching, and allow maintainers to choose one way of persisting python packages (entire conda env or just relevant folders? docker host mount or volume?)

bjornjorgensen:
I use a `docker-compose.yml` file 
```
version: '3.7'
services:
  my-service:
    image: jupyter/base-notebook:latest
    volumes:
      - myvolume:/opt/conda
volumes:
  myvolume:

```

mathbunnyru:
@laundmo if you want to persist packages (mamba/conda/pip/apt), then you should create an inherited image and install them only once while building your Dockerfile.
An example for `mamba` and `pip` is available here: https://jupyter-docker-stacks.readthedocs.io/en/latest/using/recipes.html#using-mamba-install-recommended-or-pip-install-in-a-child-docker-image

If you want to persist user files like python scripts, notebooks, text files and so on (created by you), then you should use a mount a volume. This volume might be just a folder on your host machine, or a docker volume (as in the example above).

mathbunnyru:
> and allow maintainers to choose one way of persisting python packages (entire conda env or just relevant folders? docker host mount or volume?)

I don't think one way will work for everyone though.
As I said, there are **2 different types of things we want to persist**, and they should be treated differently.
The first type (packages) is relatively straightforward (use an inherited image).

Persisting user files is more complicated.
1. Some users run our containers on local machines and they manually execute `docker run` and pass `-v` to mount a host's folder.
2. Others do the same but in docker-compose.
3. Some people prefer using docker volumes, which is also fine.
4. At the same time, people use our images as singleuser images for JupyterHub and then they need to set a config there.

The first three methods are common for all Docker ecosystem, the 4th one is documented in JupyterHub or JupyterHub's spawners, that's why we don't document it here.

Hope this helps.

laundmo:
I'm not asking how to do it. i know i can make a inherited image if there should be more packages from the get go, and i know you can mount in various ways.

[...]

Issue #1478, opened 2021-10-12
Populate home dir if it is empty


benz0li:
**What docker images this feature is applicable to?**

`jupyter/base-notebook`

**What changes do you propose?**

Populate home dir if it is empty.

**How does this change will affect users?**

If a user _mounts_ to `/home/{raw_username}` it will be populated from `/home/jovyan` on initial startup.  
ℹ️ `NB_USER` set to `{raw_username}`; the _unescaped_ `username` (e.g. from JupyterHub).

maresb:
Which are the files here that you care about? Is your home directory totally empty, like missing `.bashrc`? Or is it the `work` folder? I'm trying to understand what test cases we should write.

benz0li:
Yes, the home directory is entirely empty if _bind mounted_ to `/home/{raw_username}`. As it exists, `! -e "/home/${NB_USER}"` is `TRUE`, thus nothing is copied from `/home/jovyan/` to `/home/${NB_USER}`.  
ℹ️ `NB_USER` set to `{raw_username}`

It's the whole content of `/home/jovyan/` I care about.

benz0li:
For jupyters [base-notebook](https://hub.docker.com/r/jupyter/base-notebook) this would takes care about

```
-rw-rw-r-- 1 jovyan users  220 Feb 25  2020 .bash_logout
-rw-rw-r-- 1 jovyan users 3823 Nov  8 05:07 .bashrc
drwsrwsr-x 2 jovyan users 4096 Nov  8 05:09 .cache
drwsrwsr-x 1 jovyan users 4096 Nov  8 05:08 .conda
drwsrws--- 3 jovyan users 4096 Nov  8 05:09 .config
drwsrws--- 2 jovyan users 4096 Nov  8 05:09 .jupyter
-rw-rw-r-- 1 jovyan users  807 Feb 25  2020 .profile
-rw-rw-r-- 1 jovyan users  227 Nov  8 05:07 .wget-hsts
drwsrwsr-x 2 jovyan users 4096 Nov  8 05:07 work
```

For my custom [jupyterlab image](https://gitlab.b-data.ch/jupyterlab/r/r-ver/container_registry) this would takes care about

```
-rw-r--r--  1 jovyan users  220 Aug  4 20:25 .bash_logout
-rw-r--r--  1 jovyan users 3768 Nov  5 10:50 .bashrc
drwxr-xr-x  3 jovyan users 4096 Nov  5 10:50 .local
[...]

PR #1977, opened 2023-08-20
Add info how to persist user data

mathbunnyru:
## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/1959

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ and want to launch a single Jupyter Application in a container.
 
 The [User Guide on ReadTheDocs](https://jupyter-docker-stacks.readthedocs.io/en/latest/) describes additional uses and features in detail.
 
-**Example 1:**
+### Example 1
 
 This command pulls the `jupyter/scipy-notebook` image tagged `2023-08-19` from Docker Hub if it is not already present on the local host.
 It then starts a container running a Jupyter Server with the JupyterLab frontend and exposes the container's internal port `8888` to port `10000` of the host machine:
@@ -43,7 +43,7 @@ where:
 
 The container remains intact for restart after the Server exits.
 
-**Example 2:**
+### Example 2
 
 This command pulls the `jupyter/datascience-notebook` image tagged `2023-08-19` from Docker Hub if it is not already present on the local host.
 It then starts an _ephemeral_ container running a Jupyter Server with the JupyterLab frontend and exposes the server on host port 10000.
--- a/docs/using/faq.md
+++ b/docs/using/faq.md
@@ -12,7 +12,7 @@ There are 2 types of data, which you might want to persist.
    ```{note}
    If you install a package inside a running container (for example you run `pip install <package>` in a terminal),
    it won't be preserved when you next run your image.
-   To make it work, install this package in your inherited image and rerun `docker build`.
+   To make it work, install this package in your inherited image and rerun `docker build` command.
    ```
 
 2. If you want to persist user-data (files created by you, like `python` scripts, notebooks, text files and so on),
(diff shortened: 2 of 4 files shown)


mathbunnyru:
@laundmo @bjornjorgensen please, take a look here.

README.md
```diff
@@ -24,7 +24,7 @@ and want to launch a single Jupyter Application in a container.
 
 The [User Guide on ReadTheDocs](https://jupyter-docker-stacks.readthedocs.io/en/latest/) describes additional uses and features in detail.
 
-**Example 1:**
+### Example 1
```
mathbunnyru, line 27: This makes it possible to address it in other parts of docs.

laundmo:
It's good to have this documented, but it doesnt address the core issue of how to persist packages installed by a user. it only documents how to persist packages installed by the admin of the container host.

mathbunnyru:
> but it doesnt address the core issue of how to persist packages installed by a user

Could you please explain, what you mean by this?
Which packages and installed how exactly?

laundmo:
a user of jupyter lab, who only sees the webinterface, may run `!pip install package` to install a package. These wont be persisted with the methods documented here.

mathbunnyru:
Ok, I added a note that a package installed inside a running container will not be preserved.
Please, take a look.

laundmo:
the core issue is that users cant interact with the docker image - by design, and because my users dont have the knowhow. knowing that it wont be persisted is good, but it shouldnt close #1959

mathbunnyru:
> the core issue is that users cant interact with the docker image - by design

Why can't they? They can, but as it always works with docker images, some changes are lost after a container exists, and this is indeed by design.

> and because my users dont have the knowhow. 

That's why we're adding these docs.

> knowing that it wont be persisted is good, but it shouldnt close https://github.com/jupyter/docker-stacks/issues/1959

Could you make a **suggestion** what to add here?
You told this is a documentation issue and I believe I added all the neccessary information to the docs.

laundmo:
> Why can't they? They can, but as it always works with docker images

sorry, this was unclear: they cant access the docker host or start/stop/rebuild images.

> You told this is a documentation issue

initially i thought it was. maybe its actually a feature request.

mathbunnyru:
> > Why can't they? They can, but as it always works with docker images
> 
> sorry, this was unclear: they cant access the docker host or start/stop/rebuild images.
> 
> > You told this is a documentation issue
> 
> initially i thought it was. maybe its actually a feature request.

Ok, then please create a new issue with a feature request, which describes what you want users to be able to do (and we will discuss if it's possible).
Also, if you want to be the whole state saved, then you can use `docker commit` (you need to have access to the host machine), but this is not the best practice: https://stackoverflow.com/a/44480870/4881441

For now, I'm merging this issue as I believe there is nothing more to document on this issue.

benz0li:
@laundmo You might be interested in my/b-data's [JupyterLab](https://github.com/b-data/jupyterlab-r-docker-stack) [docker](https://github.com/b-data/jupyterlab-python-docker-stack) [stacks](https://github.com/b-data/jupyterlab-julia-docker-stack).

Those images

1. allow to mount[^1] the whole user directory and thus preserve all user data.
2. use Python (no Conda/Mamba) and install packages in the home directory.

[^1]: Either volume or bind mount

Demo environment: https://demo.jupyter.b-data.ch

Cross references:

1. https://github.com/jupyter/docker-stacks/issues/1478#issuecomment-1062670304
2. https://github.com/jupyter/docker-stacks/issues/1969

PR #1979, opened 2023-08-24
Implement run-hooks as a separate script

mathbunnyru:
## Issue ticket if applicable

Work on: https://github.com/jupyter/docker-stacks/issues/1478

### DIFF
--- a/docs/using/common.md
+++ b/docs/using/common.md
@@ -86,7 +86,7 @@ You do so by passing arguments to the `docker run` command.
 
   ```{note}
   `NB_UMASK` when set only applies to the Jupyter process itself -
-  you cannot use it to set a `umask` for additional files created during run-hooks.
+  you cannot use it to set a `umask` for additional files created during `run-hooks.sh`.
   For example, via `pip` or `conda`.
   If you need to set a `umask` for these, you **must** set the `umask` value for each command.
   ```
@@ -135,7 +135,7 @@ or executables (`chmod +x`) to be run to the paths below:
 - `/usr/local/bin/before-notebook.d/` - handled **after** all the standard options noted above are applied
   and ran right before the Server launches
 
-See the `run-hooks` function in the [`jupyter/docker-stacks-foundation start.sh`](https://github.com/jupyter/docker-stacks/blob/main/images/docker-stacks-foundation/start.sh)
+See the `run-hooks.sh` script [here](https://github.com/jupyter/docker-stacks/blob/main/images/docker-stacks-foundation/run-hooks.sh) and how it's used in the [`start.sh`](https://github.com/jupyter/docker-stacks/blob/main/images/docker-stacks-foundation/start.sh)
 script for execution details.
 
 ## SSL Certificates
--- a/docs/using/selecting.md
+++ b/docs/using/selecting.md
@@ -36,6 +36,7 @@ It contains:
   with ownership over the `/home/jovyan` and `/opt/conda` paths
 - `tini` as the container entry point
 - A `start.sh` script as the default command - useful for running alternative commands in the container as applications are added (e.g. `ipython`, `jupyter kernelgateway`, `jupyter lab`)
+- A `run-hooks.sh` script, which can source/run files in a given directory
 - Options for a passwordless sudo
 - Common system libraries like `bzip2`, `ca-certificates`, `locales`
 - `wget` to download external files
--- a/images/docker-stacks-foundation/Dockerfile
+++ b/images/docker-stacks-foundation/Dockerfile
@@ -127,7 +127,7 @@ ENTRYPOINT ["tini", "-g", "--"]
 CMD ["start.sh"]
 
 # Copy local files as late as possible to avoid cache busting
-COPY start.sh /usr/local/bin/
+COPY run-hooks.sh start.sh /usr/local/bin/
 
 USER root
 
--- a/images/docker-stacks-foundation/run-hooks.sh
+++ b/images/docker-stacks-foundation/run-hooks.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+# The run-hooks.sh script looks for *.sh scripts to source
+# and executable files to run within a passed directory
+
+if [ "$#" -ne 1 ]; then
+    echo "Should pass exactly one directory"
+    return 1
+fi
+
+if [[ ! -d "${1}" ]] ; then
+    echo "Directory ${1} doesn't exist"
+    return 1
+fi
+
+echo "Running hooks in: ${1} as uid: $(id -u) gid: $(id -g)"
+for f in "${1}/"*; do
+    # Hadling a case when the directory is empty
+    [ -e "${f}" ] || continue
+    case "${f}" in
+        *.sh)
+            echo "Sourcing shell script: ${f}"
+            # shellcheck disable=SC1090
+            source "${f}"
+            ;;
+        *)
+            if [ -x "${f}" ] ; then
+                echo "Running executable: ${f}"
+                "${f}"
+            else
+                echo "Ignoring non-executable: ${f}"
+            fi
+            ;;
+    esac
+done
+echo "Done running hooks in: ${1}"
--- a/images/docker-stacks-foundation/run-hooks.sh
+++ b/images/docker-stacks-foundation/run-hooks.sh
@@ -11,7 +11,7 @@ if [ "$#" -ne 1 ]; then
 fi
 
 if [[ ! -d "${1}" ]] ; then
-    echo "Directory ${1} doesn't exist"
+    echo "Directory ${1} doesn't exist or is not a directory"
     return 1
 fi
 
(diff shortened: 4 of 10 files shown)


mathbunnyru:
@benz0li take a look, please.

mathbunnyru:
@benz0li thanks 👍

I think it would be nice if your images somehow were able to update scripts like this, so we have fewer differences and these differences are easier to find. I don't know if it's easy for you or not though.

benz0li:
Very few modifications:

```diff
diff --git a/base/scripts/usr/local/bin/start.sh.orig b/base/scripts/usr/local/bin/start.sh
index d8b97bd..eb62fc7 100755
--- a/base/scripts/usr/local/bin/start.sh.orig
+++ b/base/scripts/usr/local/bin/start.sh
@@ -53,6 +53,8 @@ if [ "$(id -u)" == 0 ] ; then
     # - CHOWN_HOME: a boolean ("1" or "yes") to chown the user's home folder
     # - CHOWN_EXTRA: a comma separated list of paths to chown
     # - CHOWN_HOME_OPTS / CHOWN_EXTRA_OPTS: arguments to the chown commands
+    # - CHMOD_HOME: a boolean ("1" or "yes") to chmod the user's home folder
+    # - CHMOD_HOME_MODE: mode argument to the chmod command
 
     # Refit the jovyan user to the desired the user (NB_USER)
     if id jovyan &> /dev/null ; then
@@ -75,7 +77,7 @@ if [ "$(id -u)" == 0 ] ; then
         fi
         # Recreate the desired user as we want it
         userdel "${NB_USER}"
-        useradd --no-log-init --home "/home/${NB_USER}" --uid "${NB_UID}" --gid "${NB_GID}" --groups 100 "${NB_USER}"
+        useradd --no-log-init --home "/home/${NB_USER}" --shell "$(which zsh)" --uid "${NB_UID}" --gid "${NB_GID}" --groups 100 "${NB_USER}"
     fi
 
     # Move or symlink the jovyan home directory to the desired users home
@@ -97,12 +99,22 @@ if [ "$(id -u)" == 0 ] ; then
                     exit 1
                 fi
             fi
+        # The home directory could be bind mounted. Populate it if it is empty
+        elif [[ "$(ls -A "/home/${NB_USER}" 2> /dev/null)" == "" ]]; then
+            _log "Populating home dir /home/${NB_USER}..."
+            if cp -a /home/jovyan/. "/home/${NB_USER}/"; then
+                _log "Success!"
+            else
+                _log "Failed to copy data from /home/jovyan to /home/${NB_USER}!"
+                exit 1
+            fi
         fi
         # Ensure the current working directory is updated to the new path
         if [[ "${PWD}/" == "/home/jovyan/"* ]]; then
             new_wd="/home/${NB_USER}/${PWD:13}"
             _log "Changing working directory to ${new_wd}"
             cd "${new_wd}"
+            export CODE_WORKDIR=/home/${NB_USER}/projects
         fi
     fi
 
@@ -120,6 +132,10 @@ if [ "$(id -u)" == 0 ] ; then
             chown ${CHOWN_EXTRA_OPTS} "${NB_UID}:${NB_GID}" "${extra_dir}"
         done
     fi
+    # Optionally change the mode of the user's home folder
+    if [[ "${CHMOD_HOME}" == "1" || "${CHMOD_HOME}" == "yes" ]]; then
+        chmod "${CHMOD_HOME_MODE:-755}" "/home/${NB_USER}"
+    fi
 
     # Update potentially outdated environment variables since image build
     export XDG_CACHE_HOME="/home/${NB_USER}/.cache"
@@ -140,6 +156,7 @@ if [ "$(id -u)" == 0 ] ; then
     unset_explicit_env_vars
     _log "Running as ${NB_USER}:" "${cmd[@]}"
     exec sudo --preserve-env --set-home --user "${NB_USER}" \
+        LD_LIBRARY_PATH="${LD_LIBRARY_PATH:-}" \
         PATH="${PATH}" \
         PYTHONPATH="${PYTHONPATH:-}" \
         "${cmd[@]}"
@@ -196,7 +213,7 @@ else
             # We cannot use "sed --in-place" since sed tries to create a temp file in
             # /etc/ and we may not have write access. Apply sed on our own temp file:
             sed --expression="s/^jovyan:/nayvoj:/" /etc/passwd > /tmp/passwd
-            echo "${NB_USER}:x:$(id -u):$(id -g):,,,:/home/jovyan:/bin/bash" >> /tmp/passwd
+            echo "${NB_USER}:x:$(id -u):$(id -g):,,,:/home/jovyan:$(which zsh)" >> /tmp/passwd
             cat /tmp/passwd > /etc/passwd
             rm /tmp/passwd
 
```

benz0li:
I think relevant is only the part after `The home directory could be bind mounted. Populate it if it is empty`.  
ℹ️ If the directory is not empty, the script fails.

I have no intention of syncing any files or folders with a pre-populated directory between the host and container.

benz0li:
And the [`populate.sh`](https://github.com/b-data/jupyterlab-python-docker-stack/blob/e0295b86406246873f73d5bc5763866729d9c9d6/base/scripts/usr/local/bin/start-notebook.d/populate.sh) script is required at `/usr/local/bin/start-notebook.d` in case someone *bind mounts* to `/home/jovyan`.

benz0li:
P.S.: I use a similar approach for my [Data Science Dev Containers](https://github.com/b-data/data-science-devcontainers)[^1]: https://github.com/b-data/data-science-devcontainers/blob/main/.devcontainer/scripts/usr/local/bin/onCreateCommand.sh

[^1]: Which are based on my [Language](https://github.com/b-data/julia-docker-stack) [Docker](https://github.com/b-data/python-docker-stack) [Stacks](https://github.com/b-data/r-docker-stack).

mathbunnyru:
I think this PR is ready to go, so I will merge it.

@benz0li I have a few questions though if you don't mind:
1. `Populate it if it is empty` is a bit different from what I would like to do (copy from backup without rewrites).
   Is it a matter of personal choice or do you see some issues with the approach I would like to implement?
2. Why do you need the duplication of the restore step - both in `start.sh` from `/home/jovyan/` and in `populate.sh` from backup-dir?

And two more comments on things I will. change in this project:
1. Thank you for reminding me about specifying `--shell` for useradd. I recently changed `useradd` to use long options and noticed this `useradd` lacks specifying shell (the one in Dockerfile does it correctly). 
2. I will also create a PR with keeping LD_LIBRARY_PATH - it definitely makes sense, thanks! I think people might install some software on top of our image which requires changing LD_LIBRARY_PATH and then if they use another username, it won't work.

benz0li:
> 1. `Populate it if it is empty` is a bit different from what I would like to do (copy from backup without rewrites).
>    Is it a matter of personal choice or do you see some issues with the approach I would like to implement?

This is more of a personal choice. My point of view: Only the original content [from the image] ensures that the container works as intended.

> 2. Why do you need the duplication of the restore step - both in `start.sh` from `/home/jovyan/` and in `populate.sh` from backup-dir? I mean, why can't you always restore for the backup-dir?

1. **Populating `/home/jovyan/` in `populate.sh` with the `start-notebook.d` hook**
    * Use case: `docker run -it --rm -p 8888:8888 -v "${PWD}/<empty-dir>":/home/jovyan jupyter/base-notebook`  
      This ensures that `/home/jovyan/` has the original content before https://github.com/jupyter/docker-stacks/blob/2aa9e0fe363ed6a9b6f4a4777558862e77c63362/images/docker-stacks-foundation/start.sh#L41-L46 i.e. anything else.
2. **Populating `/home/jovyan/` in `start.sh`**
    * Use case: Treat the (empty) bind mounted home directory as a special case of the non-existent one.  
      https://github.com/jupyter/docker-stacks/blob/2aa9e0fe363ed6a9b6f4a4777558862e77c63362/images/docker-stacks-foundation/start.sh#L85 +
      ```bash
      elif [[ "$(ls -A "/home/${NB_USER}" 2> /dev/null)" == "" ]]; then
      ```

ℹ️ This way `start.sh` treats a *bind mount* similar to how Docker treats a *volume*:

> If you start a container which creates a new volume, and the container has files or directories in the directory to be mounted such as `/app/`, Docker copies the directory's contents into the volume. The container then mounts and uses the volume, and other containers which use the volume also have access to the pre-populated content.

– [Volumes | Docker Docs > [...] > Populate a volume using a container](https://docs.docker.com/storage/volumes/#populate-a-volume-using-a-container)

👉 This happens only at first run and only if the volume is empty.

benz0li:
@mathbunnyru `elif [[ "$(ls -A "/home/${NB_USER}" 2> /dev/null)" == "" ]]; then` also ensures that the home directory remains untouched on subsequent runs.

benz0li:
Very simple, entirely sane and the outcome easily predictable.

I might extend to

```bash
# The home directory could be bind mounted. Populate it if it is empty
elif [[ "$(ls -A "/home/${NB_USER}" 2> /dev/null)" == "" ]]; then
    _log "Populating home dir /home/${NB_USER}..."
    if cp -a /home/jovyan/. "/home/${NB_USER}/"; then
        _log "Success!"
    else
        _log "ERROR: Failed to copy data from /home/jovyan to /home/${NB_USER}!"
        _log "       The home directory must be empty at the first run!"
        exit 1
    fi
fi
```

i.e. improving the error message on failure.

benz0li:
Cross reference: https://github.com/b-data/jupyterlab-python-docker-stack/issues/1

PR #1980, opened 2023-08-24
Specify /bin/bash shell in useradd

mathbunnyru:


mathbunnyru:
We specify `/bin/bash` shell for `jovyan` in the Dockerfile.
It definitely makes sense to do the same in `start.sh`.

The line to check the problem:
`docker run -it --rm --user root -e NB_UID=120 jupyter/docker-stacks-foundation start.sh cat /etc/passwd | grep jovyan`

mathbunnyru:
Credits to @benz0li in this PR: https://github.com/jupyter/docker-stacks/pull/1979

PR #1994, opened 2023-09-14
Add zip to the `minimal-notebook` image

slesaad:
## Describe your changes
The `minimal-notebook` image comes preinstalled with useful utilities, `unzip` being one of them. But it's missing its counterpart - `zip`.

This PR adds the `zip` package to the image.


mathbunnyru:
I think it was on purpose - I think unzip is more useful in the case of our docker stacks as some libraries/source code might be zipped and you have to unzip them.
If you need zip in your image, I suggest you just install it on top of our images.

slesaad:
Hmm.. I did think it was weird to not have `zip` when `unzip` existed. But I guess it makes sense that `unzip` is used more. If you think this is unnecessary, I'll close it.

Issue #1532, opened 2021-11-15
Improve logging in start.sh


mathbunnyru:
Possible future improvements (just a note for someone, who will want to improve our docker stacks):

1. Test `JUPYTER_DOCKER_STACKS_QUIET` env variable
2. Always add log_level like this: `_log "INFO" "my_message"
2. Based on log level color can be added
3. Exit on fatal errors (or maybe just errors)
4. Add timestamps, like INFO[2021-08-23 11:00:00.000000] my_message
5. Stop using bash, haha 👍

Based on comment: https://github.com/jupyter/docker-stacks/pull/1526#pullrequestreview-804768133

yuvipanda:
@mathbunnyru I think this should still stay open for a while?

mathbunnyru:
Sure, reopened.

mathbunnyru:
@goelakash I updated list of todos

> Based on the log level color can be added

This could be nice for warnings and errors, I think

> Exit on fatal errors (or maybe just errors)

We actually already do `exit 1` each time after `log_error`, maybe let's add one line to a helper function, and remove it from start script(s)?

> Stop using bash, haha 👍

This is unrealistic, so I would call it a day and close the issue if 2 remaining items are done

goelakash:
@mathbunnyru I really like that suggestion to add color - I haven't experimented with it before. Let me try it out today.

mathbunnyru:
@goelakash I think only exiting on fatal errors is left, maybe let's do it and we'll happily close the issue?

goelakash:
Sure, I'll take a look when i get some time soon.

Issue #1528, opened 2021-11-12
Move environment setup from `start.sh` to ENTRYPOINT instead of CMD


manics:
**What docker images this feature is applicable to?**

- `jupyter/base-notebook`
- Notebooks that use startup hooks to configure the environment

**What changes do you propose?**

Split `start-notebook.sh` or `start.sh` into a script that does the environment setup in an ENTRYPOINT, and a script that does the actual notebook startup in CMD.

Originally suggested in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2138#issuecomment-828291826

**How does this change will affect users?**

`start-notebook.sh` calls `start.sh` which handles a lot of setup in the Jupyter environment, including:
- Running setup hooks such as for setting environment variables
  https://github.com/jupyter/docker-stacks/blob/8dfdbfd3a3370ec138a7b85ae42422de8da3ca1b/base-notebook/start.sh#L44
- Customising user names and IDs when started as root
  https://github.com/jupyter/docker-stacks/blob/8dfdbfd3a3370ec138a7b85ae42422de8da3ca1b/base-notebook/start.sh#L46-L51

Since `start-notebook.sh` is set as the CMD if someone passes any arguments when running the Docker container all this setup is ignored. For example
`docker run -e NB_UID=12345 -u 0 jupyter/base-notebook jupyter-lab --debug`
should change the UID from the default `1000` to `12345`, but since the startup scripts aren't run this leads to an error (can't be run as root). Instead you must run
`docker run -e NB_UID=12345 -u 0 jupyter/base-notebook start.sh jupyter-lab --debug`

A concrete example of where this is a problem for users is in the pyspark notebook- it isn't obvious to a user that the pyspark environment is setup by a startup script rather than being baked into the Dockerfile.
- https://github.com/jupyter/docker-stacks/blob/8dfdbfd3a3370ec138a7b85ae42422de8da3ca1b/pyspark-notebook/Dockerfile#L48
- https://discourse.jupyter.org/t/pyspark-library-is-missing-from-jupyter-pyspark-notebook-when-running-with-jupyterhub-zero-to-jupyterhub-k8s/8450

Note we're working around this in JupyterHub 2.0 and Z2JH 2.0 with a breaking change: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2449
Instead of specifying `jupyterhub-singleuser` as the CMD when running the image we'll use the image's default CMD, but I think this change is still generally helpful.

[...]

PR #2006, opened 2023-10-17
Migrate start-notebook.sh to bash

yuvipanda:
Based on

> Stop using bash, haha 👍

from https://github.com/jupyter/docker-stacks/issues/1532.

If there's more apetite for this, I'll try to migrate `start.sh` and `start-singleuser.sh` as well - I think they should all be merged together. We can remove the `.sh` suffixes for accuracy, and keep symlinks in so old config still works. Since the shebang is what is used to launch the correct interpreter, the `.sh` doesn't matter.

Will help fix https://github.com/jupyter/docker-stacks/issues/1532, as I believe all those things are going to be easier to do from python than bash


yuvipanda:
I'll take a look at the test failures if this kinda migration is acceptable! I'm sure I can sort those out if needed.

mathbunnyru:
I like this PR a lot, just a few thoughts about the implementation:
1. Let's add the `.py` extension to the added start-like scripts. It doesn't change much, but some utils (maybe pre-commit hooks) and editors definitely work better (at least not worse).
2. Instead of symlinking to keep compatibility, I propose having simple shell files, which first echo a warning that calling is file is deprecated and then simply pass all the params to the corresponding `.py` file. This way we will be able to first replace a warning with a hard error in a few months and then completely delete the `.sh` files.
3. Let's also document the warning about calling old `.sh` start scripts.
4. I really hope you will have the courage to deal with `start.sh` when we merge this PR 🙂 

Overall, great PR, thank you!

images/base-notebook/start-notebook
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+import os
+import shlex
+import sys
+
+# If we are in a JupyterHub, we pass on to `start-singleuser` instead so it does the right thing
+if "JUPYTERHUB_API_TOKEN" not in os.environ:
```
mathbunnyru, line 9:
```suggestion
if "JUPYTERHUB_API_TOKEN" in os.environ:
```

images/base-notebook/start-notebook
```diff
@@ -0,0 +1,38 @@
...
+    print(
```
mathbunnyru, line 10: You can change this here to `logging.warning`, or in a separate PR (as you wish).

yuvipanda: @mathbunnyru yeah, I think that can be handled separately so this becomes more of a 'straight port'.

mathbunnyru:
I updated your branch (to make sure we test latest versions of everything).

yuvipanda:
Glad to hear you like it, @mathbunnyru! I've addressed all the points you mentioned and I think this is ready for review now.

I do personally prefer leaving out the `.py` suffix, but am alright leaving it in. I find that most tools are ok looking at the shebang for detection in those cases, but I'm sure that's not 100% of them.

I'm definitely interested in tackling `start.sh` as well! I actually wrote https://github.com/yuvipanda/jupyterhub-roothooks/ a while ago that provides some part of that functionality, so there's gonna be stuff for me to steal from there. I also spent today picking up a plan to replicate as exactly as possible bash's `source` semantics in python and I think I've a path there.

My goal would be to do this in such a way that all the existing tests pass, so no new functionality.

yuvipanda:
@mathbunnyru hah, we clashed as I was just about to rebase :)

docs/using/selecting.md
```diff
@@ -56,10 +56,17 @@ It contains:
...
-- A `start-singleuser.sh` script useful for launching containers in JupyterHub
+- A `start-notebook.py` script as the default command
+- A `start-singleuser.py` script useful for launching containers in JupyterHub
 - Options for a self-signed HTTPS certificate
 
+```{warning}
+`jupyter/base-notebook` also contains `start-notebook.sh` and `start-singleuser.sh` files to maintain backwards
+compatibility. External config that explicitly refers to those files should instead update to refer to
+`start-notebook.py` and `start-singleuser.py` instead. The shim `.sh` files will be removed at some future
+date.
```
mathbunnyru, line 67:
```suggestion
`jupyter/base-notebook` also contains `start-notebook.sh` and `start-singleuser.sh` files to maintain backwards compatibility.
External config that explicitly refers to those files should instead
update to refer to `start-notebook.py` and `start-singleuser.py`.
The shim `.sh` files will be removed at some future date.
```

In most of the mardown files here, we try to fit the sentence on one line, and it it's too long, I try to divide on several lines logically (makes it easier to read the source markdown file).

images/base-notebook/start-notebook.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+import os
+import shlex
+import sys
+
+# If we are in a JupyterHub, we pass on to `start-singleuser.py` instead so it does the right thing
```
mathbunnyru, line 8: I appreciate it that you added comments to this file! I'm sure it will help future contributors to understand this file faster.

manics:
> I'm definitely interested in tackling start.sh as well! I actually wrote https://github.com/yuvipanda/jupyterhub-roothooks/ a while ago that provides some part of that functionality, so there's gonna be stuff for me to steal from there. I also spent today picking up a plan to replicate as exactly as possible bash's source semantics in python and I think I've a path there.

If you're modifying the behaviour of the start scripts https://github.com/jupyter/docker-stacks/issues/1528 might fit in too.

mathbunnyru (approved):

images/base-notebook/start-notebook.py
yuvipanda: @mathbunnyru you're welcome :)

yuvipanda:
Yay, thank you for the quick review and merge, @mathbunnyru. And for the long set of tests that made this easy

PR #2007, opened 2023-10-17
[FAST_BUILD] Migrate `runhooks` to python

yuvipanda:
This can't be merged without also migrating `start.sh`, but opening this primarily to
show the implementation of `source`. I think that is the biggest barrier to migrating `start.sh`
to python, and this approach solves it quite neatly - am pretty proud of it :D 

You can test it by creating a folder like this:
```
test
test/01-set-variable.sh
test/02-get-variable.sh
```

And the contents of the files are:

```
$ cat test/01-set-variable.sh 
export HELLO=WORLD
```

```
$ cat test/02-get-variable.sh 
echo "What is the HELLO to? It is to $HELLO"
```

And then run:

```
$ ./run-hooks.py test
Running hooks in: test as 501 gid: 20
Sourcing shell script: test/01-set-variable.sh
Sourcing shell script: test/02-get-variable.sh
What is the HELLO to? It is to WORLD
Done running hooks in: test
```

you can see that `01-set-variable` exports a variable, which is then read by `02-get-variable`!

We can turn this into a unit test later

This is built on top of #2006, see https://github.com/yuvipanda/docker-stacks/compare/start-notebook-python...yuvipanda:docker-stacks:runhooks-py for just the run-hooks change


yuvipanda:
@mathbunnyru take a look at this for the `source` implementation :) To me this was the part of `start.sh` that needed solving, and I think this does a pretty good job. What do you think of this approach?

mathbunnyru:
That's a nice trick, and I love it!
Fortunately, I added testing for `run-hooks` only a couple months ago, so as longs as the tests are ok, we should be fine 🙂 
I have a few ideas and will write them, when we merge previous PR and this one will be easier to read/comment/change.

mathbunnyru:
@yuvipanda I merge the previous PR, please rebase/merge/create new PR (whichever you choose and makes git diff nice and small is ok for me).

yuvipanda:
@mathbunnyru I've merged that in here.

Since `start.sh` uses source to call `run-hooks`, we will have to migrate start and run-hooks together. This PR migrates run-hooks, but of course start has more going on. I'll try work on it later this week, but next week is more likely.

images/docker-stacks-foundation/run-hooks.py
```diff
@@ -0,0 +1,113 @@
...
+        bash_file.write(
+            dedent(
+                f"""
+                #!/bin/bash
+                source {path}
+                {sys.executable} {py_file.name}
+                """
+            ).encode()
+        )
+        bash_file.flush()
```
mathbunnyru, line 68: I prefer having this as actual files, added to the repo.
Having them inlined here, in my opinion, doesn't make anything easier.
Of course, we will have to pass some params to these tiny scripts, but that's ok.

images/docker-stacks-foundation/run-hooks.py
```diff
@@ -0,0 +1,113 @@
...
+        run = subprocess.run(["/bin/bash", bash_file.name])
+
+        if run.returncode != 0:
+            print(
+                f"{path} has failed with return code {run.returncode}, continuing execution"
+            )
+            return
+
+        env_vars = json.load(env_vars_file)
+        os.environ.update(env_vars)
```
mathbunnyru, line 79: What if sourced bash file unsets some variable?

mathbunnyru (changes requested):

images/docker-stacks-foundation/run-hooks.py
yuvipanda: @mathbunnyru yeah that makes sense. I'll iterate with them inline but move them out by the end.

images/docker-stacks-foundation/run-hooks.py
yuvipanda: @mathbunnyru right, so we'd have to diff them right?

images/docker-stacks-foundation/run-hooks.py
```diff
@@ -0,0 +1,113 @@
...
+    print(f"Directory {hooks_directory} does not exist")
+    sys.exit(1)
+
+if not hooks_directory.is_dir():
+    print(f"{hooks_directory} is not a directory")
+    sys.exit(1)
+
+print(f"Running hooks in: {hooks_directory} as {os.getuid()} gid: {os.getgid()}")
+
+for f in hooks_directory.iterdir():
```
mathbunnyru, line 98: As far as I understand, the order is random here.
Filename expansion in Bash is sorting alphabetically.

yuvipanda:
@manics right, that won't work. In those cases, the shell scripts that want to use the upstream functions must source them directly. For that subset of use cases, this is not a backwards compatible change. I'm personally ok with that though, as I think setting vars is a far more common use case.

mathbunnyru:
I agree with @yuvipanda on @manics issue.

Generally speaking, we haven't promised the order in which shell scripts are sourced/called (the current implementation is in order and tests use this though).
I think, if you have `a.sh` and `b.sh`, it's not that difficult to move the function to some common shell file and source it from both `a.sh` and `b.sh`. And, to be honest, I think there are not so many people using this.

mathbunnyru:
@yuvipanda also, when you will be porting `start.sh`, take a look at [`plumbum`](https://plumbum.readthedocs.io/en/latest/).
We already use it in this repo for testing and tagging, so it won't increase the number of technologies used in this repo.

This package is not added in the images, but I suppose it's tiny and without lots of additional dependencies.
And it makes calling executable files easier and more pretty-looking in Python.

images/docker-stacks-foundation/run-hooks.py
yuvipanda: `sorted` (hah!)

images/docker-stacks-foundation/run-hooks.py
yuvipanda: Handled and tested! 

If I add `unset HOME` to the first file, and `echo "Home is: $HOME"` from the second file, I get:

`Home is: `

images/docker-stacks-foundation/run-hooks.py
```diff
@@ -0,0 +1,113 @@
...
+
+if len(sys.argv) != 2:
+    print("Should pass exactly one directory")
+    sys.exit(1)
+
+hooks_directory = PosixPath(sys.argv[1])
+
+if not hooks_directory.exists():
+    print(f"Directory {hooks_directory} does not exist")
+    sys.exit(1)
```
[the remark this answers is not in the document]
yuvipanda: I would personally prefer to keep the `if` as `assert` produces a traceback that's slightly ugly:

```
  File "/Users/yuvipanda/code/docker-stacks/images/docker-stacks-foundation/t.py", line 1, in <module>
    assert False, "Some message"
AssertionError: Some message
```

images/docker-stacks-foundation/run-hooks.py
mathbunnyru: Nice!

images/docker-stacks-foundation/run-hooks.sh
```diff
@@ -2,45 +2,4 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
```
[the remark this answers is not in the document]
yuvipanda: Done

images/docker-stacks-foundation/run-hooks.py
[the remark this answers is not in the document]
mathbunnyru: Ok, as this is something users might see, let's make it more beautiful for them and more ugly for us 😆

yuvipanda:
I'm pacing myself on this one, and spent another 15 minutes on it today addressing some of the review feedback :) `start.sh` is the big one and to start next, but will take it slow.

images/docker-stacks-foundation/run-hooks.sh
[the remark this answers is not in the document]
mathbunnyru: I suppose you forgot to push your changes.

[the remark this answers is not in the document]
yuvipanda: @mathbunnyru verily did. fixed now.

mathbunnyru:
Added a test for unset: https://github.com/jupyter/docker-stacks/pull/2022

mathbunnyru:
Added a test for changing variable value: https://github.com/jupyter/docker-stacks/pull/2023

images/docker-stacks-foundation/run-hooks.sh
```diff
@@ -1,46 +1,6 @@
 #!/bin/bash
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
+# echo "WARNING: Use run-hooks.py instead"
```
mathbunnyru, line 4: Unfortunately, we have to comment this line, otherwise, tests fail, because they do not expect warnings.

mathbunnyru:
@yuvipanda I added some tests for run-hooks and rewrote the test file a little bit, to make adding new tests easier.
I believe these tests will make merging this PR easier (when you are done with start.sh).
I only added tests that should work after this PR as well (so, no functions), but now we check `unset` works and double `export` with value change also works fine.

mathbunnyru:
@yuvipanda do you have plans to continue working on this? 🙂

mathbunnyru:
Closing this PR, as there is no activity here

PR #2008, opened 2023-10-19
Add an ability to specify registry when using docker images

mathbunnyru:
## Describe your changes

Work on: https://github.com/jupyterhub/team-compass/issues/688


mathbunnyru:
This doesn't change the registry to quay yet, but rather gives an ability to specify the registry.
I would like to merge this change separately - if we ever decide to change registry again, it will be simple.
After merging this, changing registry should only require only a few lines of code change.

minrk:
Amazing, great work!

PR #2012, opened 2023-10-20
Build all workflows in my fork

mathbunnyru:
## Describe your changes

This will allow me to first merge potentially breaking changes to my fork, check if they work, and only then merge in the upstream.
I can easily push-force, revert, and do whatever I want in my fork without affecting anyone.
I will have to set up Docker Hub (and Quay.io), but it's worth the trouble.

### DIFF
--- a/.github/workflows/contributed-recipes.yml
+++ b/.github/workflows/contributed-recipes.yml
@@ -35,7 +35,7 @@ jobs:
   test-recipes:
     runs-on: ${{ matrix.runsOn }}
     needs: generate-matrix
-    if: github.repository_owner == 'jupyter' || github.repository_owner == 'mathbunnyru'
+    if: github.repository_owner == 'jupyter'
 
     steps:
       - name: Checkout Repo ⚡️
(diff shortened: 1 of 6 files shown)


mathbunnyru:
One funny thing - this PR will make me create proper secrets in my fork as well, and VS Code extension will finally stop complaining I don't have these secrets 😆

mathbunnyru:
Done, created my Docker Hub repos and added a token to my fork secrets.

mathbunnyru:
I will merge this PR manually to my fork main branch.
It should work. If it doesn't, then I will fix the PR until it works.
Then I will make everyone believe I didn't screw up and force-push my fork if needed 😆 

@consideRatio @yuvipanda @minrk how do you feel about all this? 🙂

mathbunnyru:
Overviews are updated successfully.
https://github.com/mathbunnyru/docker-stacks/actions/runs/6589653550

mathbunnyru:
Seems to work, so I will merge this, because otherwise it's not easy to check switching to `quay.io`.
I will also enable running `wiki-update`, I guess I forgot to change a line there.

mathbunnyru:
> I will also enable running `wiki-update`, I guess I forgot to change a line there.

It's not easy, because some jobs are skipped, and in this case if another job specifies that it need this step, it will also be skipped. There is `if always()`, but I think I will keep it as is right now.

PR #2014, opened 2023-10-21
[FAST_BUILD] Introduce fast build

mathbunnyru:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -67,7 +67,7 @@ jobs:
       image: docker-stacks-foundation
       platform: aarch64
       runsOn: ARM64
-    if: ${{ github.repository_owner == 'jupyter' }}
+    if: github.repository_owner == 'jupyter'
 
   x86_64-foundation:
     uses: ./.github/workflows/docker-build-test-upload.yml
@@ -85,7 +85,7 @@ jobs:
       platform: aarch64
       runsOn: ARM64
     needs: [aarch64-foundation]
-    if: ${{ github.repository_owner == 'jupyter' }}
+    if: github.repository_owner == 'jupyter'
 
   x86_64-base:
     uses: ./.github/workflows/docker-build-test-upload.yml
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -101,8 +101,8 @@ jobs:
     steps:
       - name: Output PR title
         run: |
-          echo ${{ github.event.pull_request.title }}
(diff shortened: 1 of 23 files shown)


mathbunnyru:
This approach allows us to test the most crucial changes in the CI in a fast way.
We need to build 2 images because we really want to check the upload/download part because it might be broken as well.

As I'm not sure this is the best approach, and it adds some noise to the CI graph and the workflow file is bigger, I would like to hear some ideas, maybe there is a better approach.

.github/workflows/docker.yml
```diff
@@ -175,6 +183,7 @@ jobs:
       platform: aarch64
       runsOn: ARM64
     needs: [aarch64-scipy]
+    if: github.repository_owner == 'jupyter' && ${{ !contains(github.event.head_commit.message, '[FAST_BUILD]') }}
```
mathbunnyru, line 186: This line was completely missing (it's a bug).
It worked, because if some job is skipped (because of if:), and another job needs it, then this job will be skipped as well.

It means essentially we can only specify if: in some of the jobs, but explicit is better than implicit.

mathbunnyru:
If it's a PR, then everything depends on whether the PR title has a `[FAST_BUILD]` substring.
During the push to the main branch a `github.event.pull_request.title` property is empty (I hope), so we will always run the full build (which totally makes sense).

mathbunnyru:
I don't understand what's going on 😢

.github/workflows/docker.yml
mathbunnyru: Solved in main

mathbunnyru:
Finally got it working. Today I learned GitHub env variables are strings, not some kind of GitHub magic, so, the "false" string evaluates to true.
What's cool (I suppose it's secure enough, but who knows), is that GitHub evaluates the expression after replacing env variables with values.
https://github.com/jupyter/docker-stacks/actions/runs/6712091950/job/18241119289

mathbunnyru:
@yuvipanda you might find this feature useful - it allows to significantly reduce build time in PRs.
You need to add `[FAST_BUILD]` to the PR name and then only 2 root images will be built (and tested).
I think most of CI/infrastructure/start-script PRs will benefit from this change (and I wish I had thought of this feature a few years ago).

Please, review, if you have some time 🙂

mathbunnyru:
I am merging this. I want to introduce new tests for run-hooks before we merge https://github.com/jupyter/docker-stacks/pull/2007.

PR #2024, opened 2023-11-06
Add ml-notebook

twalcari:
## Describe your changes

This is a follow-up PR on #1936 , as described by @mathbunnyru here: https://github.com/jupyter/docker-stacks/pull/1936#issuecomment-1783594478

One open question: wouldn't it make more sense to call this image "machine-learning-notebook"? I'm not sure what the lore behind the use of the term 'deep-learning' in this project is, but it hurts discoverability imho.


mathbunnyru:
> One open question: wouldn't it make more sense to call this image "machine-learning-notebook"? I'm not sure what the lore behind the use of the term 'deep-learning' in this project is, but it hurts discoverability imho.

@mathbunnyru, @consideRatio, @yuvipanda, @manics let's think about this and let's vote for this PR as well.

consideRatio:
Deep learning indicates large neural networks, its a part of the machine learning field. I think by adding tensorflow and pytorch to an image, it becomes an machine learning (ml) focused image rather than a deep-learning focused ml image. There are loads of libraries out, where perhaps Keras is a expected library for deep learning? Is it part pf the tf image atm?

👍 For ml-notebook or machine-learning-notebook, because i think specializing in the field of ml beyond the core ml functionality may be too specific for jupyter/docker-stacks

mathbunnyru:
Thanks @consideRatio!

I agree, let's rename to `ml-notebook`, keep it short, while everyone understands the meaning.
@twalcari thank you for the suggestion and the work you've put here 👍

yuvipanda:
In [pangeo docker-stacks](https://github.com/pangeo-data/pangeo-docker-images/), there's ml-notebook (which we wish we had named tensorflow-notebook) and pytorch-notebook. pytorch-notebook is newer than ml-notebook, primarily driven by users using pytorch rather than tensorflow. Crucially, pytorch-notebook does *not* inherit from ml-notebook - the idea is that you're using tensorflow *or* pytorch, but not both.

Based on conversations in https://github.com/jupyter/docker-stacks/pull/1936, if the goal is to have just *one* image here, I would say calling this `ml-notebook` makes sense. 

However, something to look out for is docker image size explosion. tensorflow is half a gig compressed, while pytorch is almost 670MB compressed! If those two are not used together often, having two separate images - `tensorflow-notebook` and `pytorch-notebook` is probably better for our CI as well as end users. This is the choice made in the pangeo docker-stacks.

My personal vote is to keep these as two different images, as I've seen a 10+ GB Docker image once and I don't wanna see things like that again ;)

yuvipanda:
Size estimates from looking at x86 wheels on PyPI - [pytorch](https://pypi.org/project/torch/#files) and [tensorflow](https://pypi.org/project/tensorflow/#files)

twalcari:
+1 for the remarks of @yuvipanda . Users never mix the two ML ecosystems, and besides the Docker image size I'm also a bit concerned about the maintainability due to risk of conflicting dependency versions down the road.

tests/deep-learning-notebook/units/unit_pytorch.py
```diff
@@ -0,0 +1,5 @@
+# Copyright (c) Jupyter Development Team.
```
mathbunnyru, line 1: We need to put this to `ml-notebook` dir

mathbunnyru (approved):
LGTM. Let's wait for vote results 👍

manics:
`ml-notebook` or `machine-learning-notebook` sound good to me too.

The Dockerfile seems to be missing from this PR though.

twalcari:
Those files must have gotten mixed up while doing all sorts of git wizardry 😅. Restored and fixed now.

yuvipanda:
@mathbunnyru I appreciate the voting structure and being part of it!

For https://github.com/jupyter/docker-stacks/pull/2024#issuecomment-1794320554 and https://github.com/jupyter/docker-stacks/pull/2024#issuecomment-1794390263, I would very much prefer that tensorflow and pytorch are not mixed in one image. So my vote would be to pursue https://github.com/jupyter/docker-stacks/pull/1936 instead, and have a separate tensorflow and pytorch image. I believe that will be far easier to maintain long term, and is a better user experience too.

mathbunnyru (approved):

mathbunnyru:
Ok, our votes look like this:
- have `ml-notebook`: @mathbunnyru @manics @consideRatio 
- have separate `pytorch-notebook`: @yuvipanda @twalcari 

I included the author of PRs for this voting, it seems to be fair and also eliminates the problem of an equal number of votes.
Let's have a deadline in two weeks: 21 November.
If someone changes their mind (or maybe I didn't understand someone), then please let us know.
On the day of the deadline, the more popular option will be merged.

Both PRs are in a really good state (really good work @twalcari 👍). I've also fixed a few things in both to accommodate the latest changes I made, so there should be no problems merging either of them.

yuvipanda:
Regardless of how this pans out, I really appreciate the process, @mathbunnyru :)

I agree on giving the PR author 1 vote as well to odd out the evens.

consideRatio:
I'm sorry for not being very responsive @mathbunnyru!

My input above was mostly on naming a combined image to ml-notebook or machine-learning-notebook over deep-learning-notebook.

If pytorch is added one way or another and the maintenance toil around that would be embraced, i think i'd lean towards embracing a new image - but I'm not sure.

I have installed both tensorflow and pytorch in images in the past, and these ml libraries habe been beasts compared to other libraries with regards to size and installation requirements.

If both are installed in the same image, there is risk of dependency mismatches making things hard to install (and image size issues), while if they are installed in different images there is one more image to maintain.

I'm open to both strategies if pytorch is to be supported by this project, not confident about what will the better compromise - but i lean towards separate images as that is better for end users likely due to image size reasons.

consideRatio:
I've taken vacation from work to focus on getting things done at home, and is overall not able to invest much time to help atm :/

mathbunnyru:
@consideRatio thank you! I moved your vote to a second option.

yuvipanda:
I also just went and checked the packages installed by [pytorch-notebook](https://github.com/pangeo-data/pangeo-docker-images/blob/master/pytorch-notebook/environment.yml) vs [ml-notebook](https://github.com/pangeo-data/pangeo-docker-images/blob/master/ml-notebook/environment.yml) - there's no overlap other than a jupyterlab extension showing GPU stats. This is primarily to show that they are indeed two separate ecosystems, rather than overlapping parts of one.

mathbunnyru:
@twalcari please update https://github.com/jupyter/docker-stacks/blob/main/.github/workflows/registry-move.yml in both PRs.
The deadline is soon and I will merge one of PRs.

mathbunnyru:
Comparison of build times (this is not perfectly accurate, but still we can see what takes time):
`main`: [1h 33m 17s](https://github.com/jupyter/docker-stacks/actions/runs/6943658525/usage)
`pytorch-notebook`: [1h 37m 52s](https://github.com/jupyter/docker-stacks/actions/runs/6944822077/usage)

We now should build many images in parallel, while we only have 2 aarch64 runners, so there is an increased build time.
We can try another configuration (now we have 2 VMs with 2 CPUs each, we can try 4 VMs with 1 CPU each for slightly bigger cost).

I think we can decrease our build time by around 40 minutes in a perfect-case scenario.
Why 40 minutes: in `pytorch-notebook` branch there is `[x86_64-images-tag-push](https://github.com/jupyter/docker-stacks/actions/runs/6944822077/job/18894863458?pr=1936#step:1:1)` started at `14:47:27 GMT`, while `[aarch64-images-tag-push](https://github.com/jupyter/docker-stacks/actions/runs/6944822077/job/18896689730?pr=1936#step:1:1)` started at `15:30:39 GMT`. This practically means we were waiting for aarch64 runners are the bottleneck.

mathbunnyru:
> @twalcari please update https://github.com/jupyter/docker-stacks/blob/main/.github/workflows/registry-move.yml (it's a new file I recently added) in both PRs. The deadline is soon and I will merge one of PRs.

@twalcari waiting for this one

twalcari:
Thank you for the ping. I was on holidays last week and had missed your message in the sea of GitHub mails that was waiting for me.
I've added ml-notebook / pytorch-notebook to `registry-move.yml` in their respective PR branch.

mathbunnyru:
So, the vote happened here, I'm merging `pytorch-notebook` image.

mathbunnyru:
Closing in favour of: https://github.com/jupyter/docker-stacks/pull/1936

Issue #2037, opened 2023-11-21
Setting NB_USER for running the container should also set XDG_CACHE_HOME to that users home


HasseJohansen:
### What docker image(s) are you using?

datascience-notebook

### Host OS system

Ubuntu 20.04

### Host architecture

x86_64

### What Docker command are you running?

I don't use docker but apptainer running in a HPC cluster

It runs the image as read-only (I cannot set it read/write as that would disable GPU passthrough)

### How to Reproduce the problem?

Run the container mounting your home inside the container and use the NB_USER env variable to set it to your own user

Observe that XDG_CACHE_HOME still points to /home/jovyan/.cache/ (which is a little unexpected when you have set the user)

This makes it impossible to use conda to ex. create an environment in your writeable home..because it will try to write to the cache directory from the readonly container image

### Expected behavior

XDG_CACHE_HOME is set correctly when NB_USER is set

### Actual behavior

XDG_CACHE_HOME is always /home/jovyan/.cache/

### Anything else?

I have got around it by passing --env XDG_CACHE_HOME=${HOME}/.cache myself when starting the container with apptainer

### Latest Docker version

- [X] I've updated my Docker version to the latest available, and the issue persists

benz0li:
This *should* be set here (and preserved): https://github.com/jupyter/docker-stacks/blob/main/images/docker-stacks-foundation/start.sh#L125

@HasseJohansen Do you build the image yourself? Because this project does not provide GPU accelerated images.

HasseJohansen:
Hi @benz0li and thank you for the swift reply

I did notice the line in the start script setting XDG_CACHE_HOME, but somehow it doesn't work for me. I need to pass something like this when running with apptainer --env ${HOME}/cache

I actually retested without passing the above env and had the XDG_CACHE_HOME set to /home/jovyan/.cache which when using apptainer is readonly by default

I will try to reproduce with docker

I am not rebuilding the image. We get it through the singularity-hpc project (https://singularityhub.github.io/singularity-hpc/) which are using official images, but adding a GitHub hosted "recipe" for each upstream repo(in this instance Jupyter docker stacks) to build an env module(https://modules.readthedocs.io/en/latest/)

It will also convert the docker image to the .sif image format used by apptainer

As apptainer is working somewhat transparently on the host

We can do something like `module load jupyter/datascience-notebook` and then ex. run start-notebook.sh which is a script exposed by the env module

[...]

PR #2038, opened 2023-11-22
Remove environment variable XDG_CACHE_HOME

benz0li:
## Describe your changes

Remove environment variable XDG_CACHE_HOME

## Issue ticket if applicable

Fixes https://github.com/jupyter/docker-stacks/issues/2037

### DIFF
--- a/images/docker-stacks-foundation/start.sh
+++ b/images/docker-stacks-foundation/start.sh
@@ -121,9 +121,6 @@ if [ "$(id -u)" == 0 ]; then
         done
     fi
 
-    # Update potentially outdated environment variables since the image build
-    export XDG_CACHE_HOME="/home/${NB_USER}/.cache"
-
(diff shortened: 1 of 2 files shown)


benz0li:
Ping @mathbunnyru

mathbunnyru:
> Ping @mathbunnyru

Yep, already launched the workflows 🙂

mathbunnyru:
I also searched the specification:
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

`$XDG_CACHE_HOME defines the base directory relative to which user-specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.`

I think it is quite reasonable not to set this variable - makes the code simpler and more correct (at least now user-renaming should work better without additional hacks).

Issue #2036, opened 2023-11-20
Running jupyterlab as root instead of jovyan when running in rootless docker mode


anil-resero:
### What docker image(s) is this feature applicable to?

datascience-notebook

### What change(s) are you proposing?

When running the container in rootless docker mode, running Jupyterlab as Jovyan or any other user creates files with subuid and subgid. As a result, deleting these files or updating them requires sudo permissions. If we are able to run jupyterlab in the docker container as root, in the rootless docker environment, the UID/GID of the root in the container will map to the UID/GID of the host user.

### How does this affect the user?

The host user can access the files created by the Jupyterlab without any permission issues as the UID/GID for the host user will map to UID/GID (0/0) of the root user in the container (in rootless docker environment)

benz0li:
@anil-resero You may be interested in my/b-data's [\[CUDA-enabled\] JupyterLab Python docker stack](https://github.com/b-data/jupyterlab-python-docker-stack#using-podman-rootless-mode-3115), which supports running in *rootless mode*.

anil-resero:
@benz0li thank you for the pull request. I tried the new image `quay.io/jupyter/datascience-notebook:2023-11-22` and it worked as expected.

benz0li:
> @benz0li thank you for the pull request. I tried the new image `quay.io/jupyter/datascience-notebook:2023-11-22` and it worked as expected.

@anil-resero PR https://github.com/jupyter/docker-stacks/pull/2039 is not merged yet. One can already start the container with the triplet `-e NB_USER=root` `-e NB_UID=0` `-e NB_GID=0`.

But the crucial part is, that the home directory of `root` is modified to `/home/root` in `/etc/passwd`. Otherwise, not everything will work as expected.

anil-resero:
> > @benz0li thank you for the pull request. I tried the new image `quay.io/jupyter/datascience-notebook:2023-11-22` and it worked as expected.
> 
> @anil-resero PR #2039 is not merged yet. One can already start the container with the triplet `-e NB_USER=root` `-e NB_UID=0` `-e NB_GID=0`.
> 
> But the crucial part is, that the home directory of `root` is modified to `/home/root` in `/etc/passwd`. Otherwise, not everything will work as expected.

@benz0li thanks for the clarification. The simple notebooks I tried worked fine even without changing the home directory. I customized the image as shown below:
```
# Start from a core stack version
FROM quay.io/jupyter/datascience-notebook:2023-11-22

USER root
[...]

PR #2039, opened 2023-11-22
[FAST_BUILD] Add support for Docker/Podman in rootless mode

benz0li:
## Describe your changes

Add support for Docker/Podman in rootless mode.

Enables the container to run as `root` user, e.g.

```bash
podman run -p 8888:8888 -u root -e NB_USER=root -e NB_UID=0 -e NB_GID=0 quay.io/jupyter/scipy-notebook start-notebook.sh --allow-root
```

FYI @anil-resero

## Issue ticket if applicable

Fixes https://github.com/jupyter/docker-stacks/issues/2036


benz0li:
`NB_UID=0` and `NB_GID=0` are mandatory so that the UID/GID is correctly mapped to the host's user.

Due to the additional `cp` option `--no-preserve=ownership`, a `CHOWN_HOME` is not required.

mathbunnyru:
Added `[FAST_BUILD]` to the PR name, to make the build significantly faster.
More info about it: https://github.com/jupyter/docker-stacks/blob/main/.github/workflows/docker.yml#L3-L6

manics:
> IMHO we do not want to use (and touch) the original /root directory.

What's your thinking behind not touching `/root`, but modifying the `root` user? What happens if you create multiple users with the same UID?

```
root@8db194d8497c:/# useradd -m test -u0 --non-unique
root@8db194d8497c:/# su - test
# pwd
/home/test
# id
uid=0(root) gid=1000(test) groups=1000(test)
```

benz0li:
The triplet `-e NB_USER=root -e NB_UID=0 -e NB_GID=0` is strictly required for this use case.

Otherwise you run into

https://github.com/jupyter/docker-stacks/blob/baa110d3f9855f18d8e5db317f48a9591c06cc6b/images/docker-stacks-foundation/start.sh#L70

followed by

https://github.com/jupyter/docker-stacks/blob/baa110d3f9855f18d8e5db317f48a9591c06cc6b/images/docker-stacks-foundation/start.sh#L77 which is supposed to fail.

👉 You cannot remove the user the container is running with (`-u root`).

mathbunnyru:
@manics may I ask you to review this, please?

manics:
It looks like this solves one instance of the problem of running a rootless container without dealing with different subuids.

I'm still not convinced changing root's home directory to `/home/root` is the right approach though. It's true that `/root` may be pre-populated, but so is but since this is the container it's unlikely that there's anything important in there. There's already a well known distinction between root's homedir and standard homedirs, so given that there's going to be confusion regardless of which option is chosen I think sticking with the Linux default is clearer.

Having said that there's actually two different issues being discussed here:
- How should `NB_USER=root NB_UID=0` behave?
- How should `NB_USER=jovyan NB_UID=0` (or any other `NB_USER`) behave? For example, one way to solve the rootless problem is to allow `jovyan` to have `NB_UID=0` which requires `useradd --non-unique`. This isn't perfect since a terminal still shows the username as `root` but `USER=jovyan` and `HOME=/home/jovyan` are correct.

I think it'd be nice to solve the more general problem for rootless running and allow `NB_UID=0` for any `NB_USER`. The specific decision on whether root's home directory should be changed for `NB_USER=root` is something I'm happy for @mathbunnyru to decide since it's a matter of opinion.

mathbunnyru:
Merging this one. I will document the current behaviour in a separate PR.

Issue #2015, opened 2023-10-23
Pluto Notebook problem


andreasskeidsvoll:
### What docker image(s) are you using?

datascience-notebook

### Host OS system

Rocky Linux 8.6

### Host architecture

x86_64

### What Docker command are you running?

Used as singleuser image in customized [zero-to-jupyterhub-k8s](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/tree/b407b541b1ee2a0f699e90699b61c2299c460547) Helm chart.

### How to Reproduce the problem?

Problem occurs when I click the Pluto Notebook link in the JupyterLab Launcher. This does not occur in the 2023-09-11 image, as suggested [in this comment](https://github.com/jupyter/docker-stacks/issues/1992#issuecomment-1715850047)
.

### Actual behavior

The Pluto Notebook interface doesn't work properly, and it seems to be failing when loading different components:
![image](https://github.com/jupyter/docker-stacks/assets/12532460/753f193a-1d8b-4f92-8c0c-832e02561989)

I also get `500 : Internal Server Error` when trying to create new Pluto notebooks, The behavior matches the one seen in [this issue](https://github.com/jupyter/docker-stacks/issues/1992).
 
When I click the Pluto Notebook link in the Launcher, I also get variants of the following log message:
```
[341] signal (4.2): Illegal instruction
in expression starting at none:1
isabsent at /opt/julia/packages/URIs/7Tg5r/src/URIs.jl:239 [inlined]
formaturi at /opt/julia/packages/URIs/7Tg5r/src/URIs.jl:250
jfptr_formaturi_8986 at /opt/julia/compiled/v1.9/Pluto/OJqMt_idFnG.so (unknown line)
_jl_invoke at /cache/build/default-amdci5-5/julialang/julia-release-1-dot-9/src/gf.c:2758 [inlined]
ijl_apply_generic at /cache/build/default-amdci5-5/julialang/julia-release-1-dot-9/src/gf.c:2940
jl_apply at /cache/build/default-amdci5-5/julialang/julia-release-1-dot-9/src/julia.h:1880 [inlined]
do_apply at /cache/build/default-amdci5-5/julialang/julia-release-1-dot-9/src/builtins.c:730
uristring at /opt/julia/packages/URIs/7Tg5r/src/URIs.jl:262
uristring at /opt/julia/packages/URIs/7Tg5r/src/URIs.jl:264 [inlined]
[...]

PR #2044, opened 2023-11-28
Specify multiple architectures for Julia to precompile to

yuvipanda:
## Describe your changes

For amd64 (x86_64), we should specify what specific targets the precompilation should be done for.  If we don't specify it, it's *only* done for the target of the host doing the compilation. When the container runs on a host that's still x86_64, but a *different* generation of CPU than what the build host was, the precompilation is useless and Julia takes a long long time to start up. This specific multitarget comes from
https://docs.julialang.org/en/v1/devdocs/sysimg/#Specifying-multiple-system-image-targets, and is the same set of options that the official Julia x86_64 build is compiled with.  If the architecture the container runs on is different, precompilation may still have to be re-done on first startup - but this *should* catch most of the issues.

h/t to
https://discourse.julialang.org/t/is-it-possible-to-make-precompilation-portable-for-docker-images-built-with-a-different-cpu/95913 which helped point me towards `JULIA_CPU_TARGET`.

## Issue ticket if applicable

Fixes https://github.com/jupyter/docker-stacks/issues/2015 for more information

### DIFF
--- a/images/minimal-notebook/setup-scripts/setup-julia-packages.bash
+++ b/images/minimal-notebook/setup-scripts/setup-julia-packages.bash
@@ -20,11 +20,11 @@ set -exuo pipefail
 if [ "$(uname -m)" == "x86_64" ]; then
     # See https://github.com/JuliaCI/julia-buildkite/blob/70bde73f6cb17d4381b62236fc2d96b1c7acbba7/utilities/build_envs.sh#L24
     # for an explanation of these options
-    JULIA_CPU_TARGET="generic;sandybridge,-xsaveopt,clone_all;haswell,-rdrnd,base(1)"
+    export JULIA_CPU_TARGET="generic;sandybridge,-xsaveopt,clone_all;haswell,-rdrnd,base(1)"
 elif [ "$(uname -m)" == "aarch64" ]; then
     # See https://github.com/JuliaCI/julia-buildkite/blob/70bde73f6cb17d4381b62236fc2d96b1c7acbba7/utilities/build_envs.sh#L54
     # for an explanation of these options
-    JULIA_CPU_TARGET="generic;cortex-a57;thunderx2t99;carmel"
+    export JULIA_CPU_TARGET="generic;cortex-a57;thunderx2t99;carmel"
 fi
 
 # Install base Julia packages
--- a/images/minimal-notebook/setup-scripts/setup-julia-packages.bash
+++ b/images/minimal-notebook/setup-scripts/setup-julia-packages.bash
@@ -5,6 +5,20 @@ set -exuo pipefail
 # - The JULIA_PKGDIR environment variable is set
 # - Julia is already set up, with the setup-julia.bash command
 
+
+# For amd64 (x86_64), we should specify what specific targets the precompilation should be done for.
+# If we don't specify it, it's *only* done for the target of the host doing the compilation.
+# When the container runs on a host that's still x86_64, but a *different* generation of CPU
+# than what the build host was, the precompilation is useless and Julia takes a long long time
+# to start up. This specific multitarget comes from https://docs.julialang.org/en/v1/devdocs/sysimg/#Specifying-multiple-system-image-targets,
+# and is the same set of options that the official Julia x86_64 build is compiled with.
+# If the architecture the container runs on is different, precompilation may still have to be
+# re-done on first startup - but this *should* catch most of the issues.
+# See https://github.com/jupyter/docker-stacks/issues/2015 for more information
+if [ $(uname -m)== "x86_64" ]; then
+    export JULIA_CPU_TARGET="generic;sandybridge,-xsaveopt,clone_all;haswell,-rdrnd,base(1)"
+fi
+
 # Install base Julia packages
 julia -e '
 import Pkg;
--- a/images/minimal-notebook/setup-scripts/setup-julia-packages.bash
+++ b/images/minimal-notebook/setup-scripts/setup-julia-packages.bash
@@ -15,7 +15,7 @@ set -exuo pipefail
 # If the architecture the container runs on is different, precompilation may still have to be
 # re-done on first startup - but this *should* catch most of the issues.
 # See https://github.com/jupyter/docker-stacks/issues/2015 for more information
-if [ $(uname -m)== "x86_64" ]; then
+if [ "$(uname -m)" == "x86_64" ]; then
     export JULIA_CPU_TARGET="generic;sandybridge,-xsaveopt,clone_all;haswell,-rdrnd,base(1)"
 fi
 
--- a/images/minimal-notebook/setup-scripts/setup-julia-packages.bash
+++ b/images/minimal-notebook/setup-scripts/setup-julia-packages.bash
@@ -6,17 +6,25 @@ set -exuo pipefail
 # - Julia is already set up, with the setup-julia.bash command
 
 
-# For amd64 (x86_64), we should specify what specific targets the precompilation should be done for.
-# If we don't specify it, it's *only* done for the target of the host doing the compilation.
-# When the container runs on a host that's still x86_64, but a *different* generation of CPU
-# than what the build host was, the precompilation is useless and Julia takes a long long time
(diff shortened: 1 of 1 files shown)


mathbunnyru:
> I tested this locally and seemed ok. I think the way out here is to build this out on GitHub actions again, push and then we can test on our local x86s's :)

There is a better way.
All the images are available as GitHub artifacts on the summary page.
This is a cool feature of our upload/download architecture - many things can be checked before merging to main.
https://github.com/jupyter/docker-stacks/actions/runs/7023919029?pr=2044

They are only available for 3 days though, but normally it should be enough to download and test the image.

mathbunnyru (approved):

mathbunnyru:
@yuvipanda waiting for you to check the image in the PR on your x86_64 machine.
If everything works fine, then I'll merge.

mathbunnyru:
Do you maybe know how to fix this when running aarch64 images on Apple Silicon Macs?
I mean, it would be great to have faster Julia startup in more cases.
If it's more difficult, then don't worry and let's leave it as is.

I think the important file to look at is this one: https://github.com/JuliaLang/julia/blob/master/base/cpuid.jl
I also noticed there are more Instruction Set Architectures in this file for `x86_64`, for example `skylake`.
Should we include them in this PR as well? (I assume the documentation might be outdated, but I also might be wrong)

mathbunnyru:
Yep, the ISAs mentioned in the docs were added 5 years ago, so the documentation is outdated.
https://github.com/JuliaLang/julia/commit/d5d01ef75428b440976d01fce947550e51715f44

I suggest we resolve this issue for the modern CPUs as well, and also add a link for a `cpuid` file to make it easier to update our script in the future, how do you feel about this?

mathbunnyru:
I also have an idea in the opposite direction - stop trying to deal with this on the build machine at all.

We want `Pkg.precompile()` be run on a user machine.
So let's run it on a user machine.

To do this, we can use (wait for it) `run-hooks` capabilities, we need to add a file which runs this command in the background.

Pros:
- we don't need to have a mess with CPU architectures
- most likely to work for all the CPUs and unlikely to break
- so we don't need to do anything for new Julia versions / new CPUs
- also should work for aarch64 (and even Apple Silicon Macs)

Cons:
- we should definitely run this at background, as this takes some time and we can't wait for this process to finish, so users won't know if Julia is fully ready or not.
- we have to educate users about this
- need to be a bit careful in cases where a user is renamed (but maybe it's gonna be easy)

My opinion on this - if we can fix this issue in a sustainable way and for modern CPUs as well (I'm okay with manual updates of JULIA_CPU_TARGET, but we need to have a clear understanding how and when to update it), let's do it (users will be more happy. If we can't - then it's better to be fixed during runtime.

benz0li:
> I suggest we resolve this issue for the modern CPUs as well, and also add a link for a `cpuid` file to make it easier to update our script in the future, how do you feel about this?

@mathbunnyru These are the targets settings for all the architectures: https://github.com/JuliaCI/julia-buildkite/blob/70bde73f6cb17d4381b62236fc2d96b1c7acbba7/utilities/build_envs.sh#L20-L76

Original post: https://github.com/docker-library/julia/issues/79#issuecomment-1544493063

---

https://github.com/JuliaCI/julia-buildkite/blob/main/utilities/build_envs.sh

mathbunnyru:
> > I suggest we resolve this issue for the modern CPUs as well, and also add a link for a `cpuid` file to make it easier to update our script in the future, how do you feel about this?
> 
> @mathbunnyru These are the targets settings for all the architectures: https://github.com/JuliaCI/julia-buildkite/blob/70bde73f6cb17d4381b62236fc2d96b1c7acbba7/utilities/build_envs.sh#L20-L76
> 
> Original post: [docker-library/julia#79 (comment)](https://github.com/docker-library/julia/issues/79#issuecomment-1544493063)
> 
> File: https://github.com/JuliaCI/julia-buildkite/blob/main/utilities/build_envs.sh

Great, thanks!
Let's add a link to this file, and set JULIA_CPU_TARGETS for aarch64 as well and this should be good to go.

benz0li:
> That's a great find, thanks @benz0li!

@yuvipanda Likewise. https://github.com/jupyter/docker-stacks/issues/2015#issuecomment-1829232453 ff is exemplary in finding out the cause.

> [...] I'd like to stick to what's in the julia-buildkite repo (which is what produces the official binaries, I think?).

Your assumption is correct.

yuvipanda:
Ok, testing on aarch64:

```
❯ docker run -it -p 8888:8888 quay.io/jupyter/julia-notebook:latest /bin/bash
(base) jovyan@0fd491481b74:~$ time julia -e 'import Pluto'

real    0m0.863s
user    0m0.749s
sys     0m0.192s
(base) jovyan@0fd491481b74:~$ ls /opt/julia/compiled/v1.9/Pluto/
OJqMt_j7gqT.ji  OJqMt_j7gqT.so
```

success!

And fine on x86_64 too!

```
➜ docker run -it -p 8888:8888 quay.io/jupyter/julia-notebook:latest /bin/bash
(base) jovyan@f092b817b671:~$ uname -m
x86_64
(base) jovyan@f092b817b671:~$ time julia -e 'import Pluto'

real    0m9.723s
user    0m8.765s
sys     0m1.203s
(base) jovyan@f092b817b671:~$ ls /opt/julia/
artifacts/       conda/           logs/            pluto_notebooks/ registries/      
compiled/        environments/    packages/        prefs/           scratchspaces/   
(base) jovyan@f092b817b671:~$ ls /opt/julia/compiled/v1.9/Pluto/
OJqMt_kStVT.ji  OJqMt_kStVT.so
```

yay!

I think this means this is ready to merge, @mathbunnyru.

mathbunnyru:
> That's a great find, thanks @benz0li!
> 
> While newer architectures do exist in `cpuid`, I want to stick to whatever the official Julia build is built with. Primarily because we don't know what other options are needed - both sandybridge and haswell have options about what is included and what is missing, and we don't know what that would be for newer architectures. So if this passes the test of me downloading the built image and testing it locally, I'd like to stick to what's in the julia-buildkite repo (which is what produces the official binaries, I think?). How does that sound, @mathbunnyru?

Sounds good for me.

yuvipanda:
The compliment in https://github.com/jupyter/docker-stacks/pull/2044#issuecomment-1832692858 made my day brighter, thanks @benz0li :)

mathbunnyru:
> Ok, testing on aarch64:
> 
> ```
> ❯ docker run -it -p 8888:8888 quay.io/jupyter/julia-notebook:latest /bin/bash
> (base) jovyan@0fd491481b74:~$ time julia -e 'import Pluto'
> 
> real    0m0.863s
> user    0m0.749s
> sys     0m0.192s
> (base) jovyan@0fd491481b74:~$ ls /opt/julia/compiled/v1.9/Pluto/
> OJqMt_j7gqT.ji  OJqMt_j7gqT.so
> ```
> 
> success!
> 
> And fine on x86_64 too!
> 
> ```
> ➜ docker run -it -p 8888:8888 quay.io/jupyter/julia-notebook:latest /bin/bash
> (base) jovyan@f092b817b671:~$ uname -m
> x86_64
> (base) jovyan@f092b817b671:~$ time julia -e 'import Pluto'
> 
> real    0m9.723s
> user    0m8.765s
> sys     0m1.203s
> (base) jovyan@f092b817b671:~$ ls /opt/julia/compiled/v1.9/Pluto/
> OJqMt_kStVT.ji  OJqMt_kStVT.so
> ```
> 
> yay!
> 
> I think this means this is ready to merge, @mathbunnyru.

I also tested on my M2 MacBook Air, and it only takes `0m1.359s` (it takes around 40 seconds in the current images).

Amazing work both @yuvipanda and @benz0li, thank you!
Happy to merge this.

yuvipanda:
Thanks for the review and merge, @mathbunnyru. In addition, thanks for setting up the build process in such a way I could actually test the built images without having to merge these!

yuvipanda:
I also wrote a quick blog post about this so this information is easier to find for the next soul that goes searching: https://words.yuvi.in/post/pre-compiling-julia-docker/

PR #2046, opened 2023-12-01
Automatically install latest julia version

mathbunnyru:
## Describe your changes

- Automatically determine the latest stable (but non-LTS) version of Julia
- Julia 1.9.4 is gonna be installed after this PR


mathbunnyru:
@yuvipanda @benz0li please take a look.
This should get rid of PRs which only update Julia version.

mathbunnyru:
`After this operation, 1,043 kB of additional disk space will be used.`
`jq` is lightweight

yuvipanda:
This looks good to me, but what do you think about rewriting this script in python instead? That would allow us to not use jq. I love jq too, but I always find complex bash gets out of hand fast.

mathbunnyru:
> This looks good to me, but what do you think about rewriting this script in python instead? That would allow us to not use jq. I love jq too, but I always find complex bash gets out of hand fast.

Thanks, I will rewrite this script in Python. I might need to use `plumbum` package though, it makes my life much easier when I need to call some utilities from Python.

mathbunnyru:
@yuvipanda done. If you like the Python implementation, I will get rid of `jq` and `setup-julia.sh`.
Kept them for now, so you can compare.

images/minimal-notebook/setup-scripts/setup-julia.py
```diff
@@ -0,0 +1,75 @@
...
+import plumbum
+import requests
+
+curl = plumbum.local["curl"]
+ln = plumbum.local["ln"]
+chown = plumbum.local["chown"]
+fix_permissions = plumbum.local["fix-permissions"]
+
+
+def unify_aarch64(platform: str) -> str:
```
yuvipanda, line 23: Add a docstring?

images/minimal-notebook/setup-scripts/setup-julia.py
```diff
@@ -0,0 +1,75 @@
...
+    return {
+        "aarch64": "aarch64",
+        "arm64": "aarch64",  # To support local building on aarch64 Macs
+        "x86_64": "x86_64",
+    }[platform]
+
+
+def get_latest_julia_url() -> tuple[str, str]:
+    # Get the last stable version of Julia
```
yuvipanda, line 32: Turn comment into a docstring?

yuvipanda (commented):
Yay for rewriting shell scripts in python :) I think that's the way to go, especially if we find ourselves parsing JSON.

Can you add even a simple docstring to all the functions? I find that helps tremendously to both clarity when reading as well as in keeping them well scoped during modification.

I understand the desire to use `plumbum`, but I find it pretty difficult to read and understand, especially because it's overriding array syntax for command execution. One of the big reasons for me for writing things in python over bash is that it makes it much easier for newer contributors to onboard and make contributions, and I think plumbum sort of turns python into a hybrid shell situation that makes it a little difficult. So I do have a strong preference for just using `subprocess`, *especially* when there's no piping or complex shell-like things involved. Here, we're *only* using it to call subprocesses and nothing else (as far as I can tell). So what do you think about dropping plumbum and just using `subprocess`, with the line being 'are we only calling a process, getting its output (optional)? If so, use `subprocess`, else can use `plumbum`'.

mathbunnyru:
@yuvipanda done, please, take a look.

images/minimal-notebook/setup-scripts/setup_julia.py
```diff
@@ -0,0 +1,82 @@
...
+import os
+import platform
+import shutil
+import subprocess
+from pathlib import Path
+
+import requests
+
+
+def unify_aarch64(platform: str) -> str:
```
yuvipanda, line 18: One missing docstring

images/minimal-notebook/setup-scripts/setup_julia.py
```diff
@@ -0,0 +1,82 @@
...
+    Make Julia aware of conda libraries
+    Creates a directory for Julia user libraries
+    """
+    # Link Julia installed version to /usr/local/bin, so julia launches it
+    subprocess.check_call(
+        ["ln", "-fs", f"/opt/julia-{julia_version}/bin/julia", "/usr/local/bin/julia"]
+    )
+
+    # Tell Julia where conda libraries are
+    Path("/etc/julia").mkdir()
```
yuvipanda, line 67: This will fail if this directory already exists, which I think is ok - we don't need to pass `exists_ok=True` here.

yuvipanda (approved):
yay this is awesome! There's one missing docstring, but otherwise this looks good to go.

Pleasure working with you, @mathbunnyru

images/minimal-notebook/setup-scripts/setup_julia.py
mathbunnyru: Thanks, fixed it. I copy-pasted this function from another place, so fixed it there as well.
This is the copy-paste I can live with 😆

images/minimal-notebook/setup-scripts/setup_julia.py
mathbunnyru: It's better to fail and know about it, because I don't expect this dir to exist.

Issue #2043, opened 2023-11-24
Conda environment not fully set in Jupyter


nthiery:
### What docker image(s) are you using?

minimal-notebook

### Host OS system

CentOS 

### Host architecture

x86_64

### What Docker command are you running?

sudo docker run -p8888:8888 --rm docker.io/jupyter/minimal-notebook:latest

(or standard run from JupyterHub)

### How to Reproduce the problem?

Start a Python or C++ kernel, e.g. from a notebook and run the following
command:
```
!env | grep CONDA 
```

### Command output

```bash session
CONDA_DIR=/opt/conda
```

### Expected behavior

See all the usual environment variables that are set when a conda
environment is set; e.g.:
```
CONDA_EXE=/opt/conda/bin/conda
CONDA_PREFIX=/opt/conda
CONDA_PROMPT_MODIFIER=(base) 
_CE_CONDA=
CONDA_SHLVL=1
CONDA_DIR=/opt/conda
CONDA_PYTHON_EXE=/opt/conda/bin/python
CONDA_DEFAULT_ENV=base
```

### Actual behavior

If a barebone shell is launched from Jupyter (e.g. through shebang calls in a notebook), the  conda environment variables are not set.

## Suggestion

Launch Jupyter itself within an activated conda environment. Maybe by using an entry
point such as `conda run jupyter ...`.

For now I am using as workaround:
```
%bash --login
...
```
but this is tricky to find for end users.

## Use case

In a C++ course, we include calls to compilation commands in the course
narrative to explain how to compile; students also include them
in the support of their presentation to run the compilation and execution
of their programs. These compilation commands often need a fully
configured conda environment.

### Latest Docker version

- [X] I've updated my Docker version to the latest available, and the issue persists

mathbunnyru:
This was the solution which was implemented the last time:
https://github.com/jupyter/docker-stacks/pull/1166/files

`echo 'eval "$(command conda shell.bash hook 2> /dev/null)"' >> /etc/skel/.bashrc`
This command is executed before `NB_USER`, meaning it's `.bashrc` will have this line in the shell.
But it only works for terminals, not when opening Jupyter Notebook.

I'm currently thinking about a different solution:
Add a similar line to a file in `/usr/local/bin/before-notebook.d`.
This will source the file right before running the jupyter command, so should work for all the subprocesses.
There is one downside, in case `--user root`, this will run under root, but it mostly sets env variables, so we should be fine.

[...]

Issue #1792, opened 2022-09-29
New conda environment issue with another username


masip85:
### What docker image(s) are you using?

scipy-notebook

### OS system and architecture running docker image

aarch64

### What Docker command are you running?

docker run -it --name myproject --rm \
 -p 8888:8888 \
    --user root \
    -e NB_USER=vicente \
    -e NB_UID="$(id -u)" \
    -e NB_GID="$(id -g)"  \
    -e CHOWN_HOME=yes \
    -e CHOWN_HOME_OPTS="-R" \
    -w "/home/${NB_USER}" \
    -v "${PWD}"/test:/home/vicente/work \
    jupyter/myproject-notebook

### How to Reproduce the problem?

1. Create Dockerfile with conda environment following doc:
https://jupyter-docker-stacks.readthedocs.io/en/latest/using/recipes.html#add-a-custom-conda-environment-and-jupyter-kernel
adding at the end the commented line:
`RUN echo "conda activate ${conda_env}" >> "${HOME}/.bashrc"`

2. Execute docker run with new NB_USER.

### Expected behavior

I want my user has the conda environments in the list of kernels from the beginning (as is documented, but for any other user)

### Actual behavior

```
# create Python kernel and link it to jupyter
RUN "${CONDA_DIR}/envs/${conda_env}/bin/python" -m ipykernel install --user --name="${conda_env}" && \
    fix-permissions "${CONDA_DIR}" && \
    fix-permissions "/home/${NB_USER}"
```
and 

`RUN echo "conda activate ${conda_env}" >> "${HOME}/.bashrc" `,

 in its context is executed only for jovyan home. So I guess is only being executed when jovyan starts a session. When my notebook starts with my user, if I open the terminal I only get default conda environment at the jupyter kernelspec list. 
 
 the And I dont want to use jovyan.

### Anything else?

I can't use  install without _--user_ for install kernel it system wide:
`RUN "${CONDA_DIR}/envs/${conda_env}/bin/python" -m ipykernel install --name="${conda_env}" && \
    fix-permissions "${CONDA_DIR}" && \
    fix-permissions "/home/${NB_USER}"`
`[Errno 13] Permission denied: '/usr/local/share/jupyter'`

And in that case, I would have to hardcode this in dockerfile,
`RUN echo "conda activate ${conda_env}" >> "/home/vicente/.bashrc"`

which is not very pleasant. If you could solve the first issue, I'd like to get a nice way for doing the second one.

[...]

PR #2047, opened 2023-12-02
Fix conda hook to work in both terminal and Jupyter Notebook

mathbunnyru:
## Describe your changes

This makes conda env activate correctly for the whole process, not just the terminals.
I think it is reasonable and desirable to have the same environment inside a terminal and a Jupyter Notebook.

## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/2043
Fix: https://github.com/jupyter/docker-stacks/issues/1792


mathbunnyru:
I tested it this way:

```
make build/docker-stacks-foundation
make build/base-notebook
docker run -it --rm -p 8888:8888 quay.io/jupyter/base-notebook
```

In both the terminal and inside Jupyter notebook the result is the same:
```
env | grep CONDA
CONDA_EXE=/opt/conda/bin/conda
CONDA_PREFIX=/opt/conda
CONDA_PROMPT_MODIFIER=(base) 
_CE_CONDA=
CONDA_SHLVL=1
CONDA_DIR=/opt/conda
CONDA_PYTHON_EXE=/opt/conda/bin/python
CONDA_DEFAULT_ENV=base
```

mathbunnyru:
But this actually creates another problem:
`eval "$(conda shell.bash hook)"` sets PS1, and creates inline bash functions to make `conda activate` and `conda deactivate` work.

But people open new Jupyter Terminals in the Web UI, meaning these inline bash functions are lost.
PS1 is also lost, because `.bashrc` and some other files in Linux contain setting PS1.
It also means in the terminal `conda activate` and `conda deactivate` won't work at all in a Jupyter Terminal.
It's not nice at all.

This practically means that we will have to `eval` twice:
1. in a sourced magical file (which this PR added) - it will activate `base` environment, set proper env variables for the process
2. In the `.bashrc` - this will fix PS1 and add an ability to use `conda (de)activate` in Jupyter Terminals

mathbunnyru:
I have no idea what's going on with the test here.
Moreover, it works funder under aarch64, so I can't even test it easily 😢

mathbunnyru:
Also tested the recipe: `docker build -t test . -f custom_environment.dockerfile`, seems to work well.
PS1 now is also correct.

mathbunnyru:
> @mathbunnyru I think the issue can be considered resolved as soon as
> 
> 1. the conda environment is activated for both JupyterLab (Terminal) and Jupyter Notebooks.
> 2. env variables are identical for IPython (started from JupyterLab Terminal) and Jupyter Notebooks.
>    
>    * This might not be the case for Terminal related env variables, though.
> 
> P.S.: I have added an order of precedence (i.e. numbering) for startup hooks: [b-data/jupyterlab-python-docker-stack@82da8b4](https://github.com/b-data/jupyterlab-python-docker-stack/commit/82da8b437d361987e8915ae92d6e341e045bbd77) ℹ️ Useful, if someone builds on the Jupyter Docker Stacks and wants to add additional startup hooks.

As a developer, I would like to agree - it seems to work, easy to implement and to change for a custom environment.
From a user perspective, I would like to be able to do `conda (de)activate` easily in terminal.
And having correct PS1 would be nice as well.

mathbunnyru:
> Is it really a good idea that the custom environment is activated by default?

I think I can comment this code with some information.

mathbunnyru:
> > `docker run -it --rm -p 8888:8888 --user root -e NB_USER=vicente test` works for me with a new recipe, so I think we can close the issue you mention.
> 
> Please check `import os; os.environ` and `import sys; sys.version` for both Notebook/Console 'Python 3 (ipykernel)' and 'python310'.
> 
> Is this expected?

I checked with commented activation (as it is now in the PR).

`os.environ` results are the same (except JPY_SESSION_NAME).
I think Jupyter Lab/Console doesn't change conda when it is launched, so the variables are all the same

`sys.version` - `3.11.6` in default Notebook/Console and `3.10.13` in `python310`.
Totally fine

```
from typing import Self
```
This snippet works for default Notebook/Console and doesn't work for py310 versions (and it shouldn't work).

I also checked that python3.11 packages are not leaking somehow to `python310` - also works fine, `alembic` is available in default Notebook/Console and is not available in `python310`.

benz0li:
> Totally fine

All fine; Agreed.

> I think Jupyter Lab/Console doesn't change conda when it is launched, so the variables are all the same

Yes. And the Jupyter Docker Stacks project is not supposed to fix IPython.

---

But I am quite sure people will complain that the custom (conda) environment is not activated when running the Notebook/Console with its kernel.

mathbunnyru:
Nice, thanks for your help ❤️
I appreciate it.

I'll merge this when the tests pass.

PR #2049, opened 2023-12-05
Modify the custom Python kernel

benz0li:
- to activate the custom environment
- for the respective Jupyter Notebook and Jupyter Console

## Issue ticket if applicable

Properly fix https://github.com/jupyter/docker-stacks/pull/2047.

Prevent https://github.com/jupyter/docker-stacks/pull/2047#issuecomment-1839343594

## Checklist (especially for first-time contributors)

- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests
- [x] I will try not to use force-push to make the review process easier for reviewers
- [x] I have updated the documentation for significant changes

PR #2050, opened 2023-12-05
Use custom conda environment in Jupyter Notebook when user wants to use it

mathbunnyru:
## Describe your changes

Based on: https://github.com/jupyter/docker-stacks/pull/2049
All the credit goes to @benz0li


mathbunnyru:
I checked the image and `/home/${NB_USER}/.local/share/jupyter/kernels/python310/kernel.json` looks nice.

@benz0li please review 🙂

mathbunnyru:
As I said in the PR title all the credit goes to you.
I think it's a really nice change and our custom environment recipe is now really close to perfect 🙂 
Thank you!

If we will need to add more env variables, it will be easier with Python implementation and the code is much more readable in general.

docs/using/recipe_code/custom_environment.dockerfile
```diff
@@ -28,12 +28,37 @@ RUN "${CONDA_DIR}/envs/${env_name}/bin/python" -m ipykernel install --user --nam
...
+# This changes the custom Python kernel so that the custom environment will
+# be activated for the respective Jupyter Notebook and Jupyter Console
+# hadolint ignore=DL3059
+RUN python <<HEREDOC
+import json
+import os
+from pathlib import Path
+
+env_name = "${env_name}"
+CONDA_DIR = os.environ["CONDA_DIR"]
```
mathbunnyru, line 40: Removed CONDA_DIR hardcore here, we usually use `${CONDA_DIR}` value in our docker files.

yuvipanda: Thoughts on turning this into a file that can be called, rather than HEREDOC? Especially given python is indent sentisitive, I think HEREDOC can be difficult to modify

benz0li:
ℹ️ The default kernel has `indent=1`:

```bash
cat /opt/conda/share/jupyter/kernels/python3/kernel.json
```

```json
{
 "argv": [
  "/opt/conda/bin/python",
  "-m",
  "ipykernel_launcher",
  "-f",
  "{connection_file}"
 ],
 "display_name": "Python 3 (ipykernel)",
 "language": "python",
 "metadata": {
  "debugger": true
 }
}
```

benz0li:
@mathbunnyru See my comments. Thank you.

docs/using/recipe_code/custom_environment.dockerfile
mathbunnyru: I thought about it, but then I thought - but we would like to make our recipe easily-copyable (and then it should run).
But the I thought - let's add it as `/opt/setup-scripts/activate_notebook_custom_env.py`
People, who don't use it, won't notice, but people using custom env will still be happy (and their Dockerfiles) will be shorter - so thank you for suggestion, done.

mathbunnyru:
> ℹ️ The default kernel has `indent=1`:
> 
> ```shell
> cat /opt/conda/share/jupyter/kernels/python3/kernel.json
> ```
> 
>  ```json
> {
>  "argv": [
>   "/opt/conda/bin/python",
>   "-m",
>   "ipykernel_launcher",
>   "-f",
>   "{connection_file}"
>  ],
>  "display_name": "Python 3 (ipykernel)",
>  "language": "python",
>  "metadata": {
>   "debugger": true
>  }
> }
> ```

Thank you. Fixed indent.

mathbunnyru:
The build failure comes from https://github.com/mamba-org/mamba/pull/3036.

PR #2055, opened 2023-12-08
[FAST_BUILD] Clean conda env

mathbunnyru:
## Describe your changes

Introduce the env variable which environment to activate by default.
This makes the custom env recipe smaller.
Also eliminates the problem of double activation in the case of a custom environment (if you call `bash` a few times, you will get a `base, custom, base, custom, base, custom...` chain).

### DIFF
--- a/docs/using/recipe_code/custom_environment.dockerfile
+++ b/docs/using/recipe_code/custom_environment.dockerfile
@@ -33,16 +33,8 @@ RUN "${CONDA_DIR}/envs/${env_name}/bin/pip" install --no-cache-dir \
 # hadolint ignore=DL3059
 RUN /opt/setup-scripts/activate_notebook_custom_env.py "${env_name}"
 
-# Comment the line above and uncomment the section below instead to activate the custom environment by default
-# Note: uncommenting this section makes "${env_name}" default both for Jupyter Notebook and Terminals
+# Comment the line above and uncomment the line below instead
+# to activate the "${env_name}" custom environment default
+# both for Jupyter Notebook and Terminals
 # More information here: https://github.com/jupyter/docker-stacks/pull/2047
-# USER root
-# RUN \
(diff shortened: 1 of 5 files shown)


docs/using/recipe_code/custom_environment.dockerfile
```diff
@@ -33,16 +33,8 @@ RUN "${CONDA_DIR}/envs/${env_name}/bin/pip" install --no-cache-dir \
...
-# RUN \
-#     # This changes a startup hook, which will activate the custom environment for the process
-#     echo conda activate "${env_name}" >> /usr/local/bin/before-notebook.d/10activate-conda-env.sh && \
-#     # This makes the custom environment default in Jupyter Terminals for all users which might be created later
-#     echo conda activate "${env_name}" >> /etc/skel/.bashrc && \
-#     # This makes the custom environment default in Jupyter Terminals for already existing NB_USER
-#     echo conda activate "${env_name}" >> "/home/${NB_USER}/.bashrc"
-
-USER ${NB_UID}
+ENV DOCKER_STACKS_CONDA_ENV "${env_name}}"
```
yuvipanda, line 40: An extra `}`?

mathbunnyru: Thanks, missed it.

mathbunnyru:
With this PR we will also be close to not hardcoding PATH.
This should also help custom envs - they will be able not to accidentally consume binaries from base environment.
I don't think I will do it, but if someone wants to, I can explain what to do.

mathbunnyru:
@yuvipanda @benz0li @consideRatio please tell me, what you think about this.

mathbunnyru:
I think this creates additional complexity to most users, when they don't use custom environments.
And I haven't seen any complaints / issues with the current implementation of custom environments, so let's close this PR.

PR #2061, opened 2023-12-14
Update actions/download-artifact and actions/upload-artifact to v4

mathbunnyru:

### DIFF
--- a/.github/actions/download-manifests/action.yml
+++ b/.github/actions/download-manifests/action.yml
@@ -20,258 +20,258 @@ runs:
   using: composite
   steps:
     - name: Download artifact 📥
-      uses: actions/download-artifact@v3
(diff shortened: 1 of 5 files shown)


mathbunnyru:
I had to update this manually, otherwise, dependabot would create separate PR for each action and one for all workflows. And downloading artifacts that were created from `action/upload-artifact@v3` and below are not supported by `action/download-artifact@v4`.

mathbunnyru:
x86_64 base upload went down from 1m23s to 10s
aarch64 base upload went down from 1m39s to 15s

Finally, we get good performance.
Should significantly decrease 
Obviously, the devs of these actions didn't like my PR: https://github.com/jupyter/docker-stacks/pull/2059 (this PR only made sense because of slow uploads).

PR #1932, opened 2023-07-03
Compress images to improve download/upload speed

mathbunnyru:
## Checklist (especially for first-time contributors)

- [ ] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests
- [ ] I will try not to use force-push to make the review process easier for reviewers
- [ ] I have updated the documentation for significant changes

PR #2062, opened 2023-12-14
Revert "Compress images to improve download/upload speed (#1932)"

mathbunnyru:
This reverts commit 2df2eeca933121d6153d9dd299c6d879a52f184f.

Depends on: https://github.com/jupyter/docker-stacks/pull/2066
I would like to check if we still have speed benefits from compression. If not, let's remove it.


mathbunnyru:
Now we have fast artifacts upload, it might be faster not to compress the image.

mathbunnyru:
It seems that compressing still improves the speed, so let's keep it.

Issue #1909, opened 2023-05-28
CI improvements and speedups


mathbunnyru:
There are a few things that might be improved / changed and everyone is welcome to implement these ideas:

1. Get rid of upload/download artifact and use a registry server. This would mean much less uploads (most layers will be already uploaded). It might be a good idea to use [service containers](https://docs.github.com/en/actions/using-containerized-services/about-service-containers) here.
2. Execute several steps inside one runner. For example, `docker-stacks-foundation`, `base-notebook` and `minimal-notebook` can be built together, and while one image is uploading as artefact, another can already start building. This doesn't look very clean though, as it will introduce complexity.
3. Use [Autoscaling with self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/autoscaling-with-self-hosted-runners). This will allow to create aarch64 runners automatically and only when needed. It will save resources and my money 😆

To be honest, I'm pretty happy with the way CI works currently, so I don't think these are essential to implement, but I might give these ideas a try.

mathbunnyru:
One more idea - to compress/decompress images when downloading as GitHub artifacts.
https://github.com/jupyter/docker-stacks/pull/1932

mathbunnyru:
Two more ideas:
1. Try to zip artifacts before upload (`upload-artifact` is so slow): https://github.com/actions/toolkit/issues/1533#issuecomment-1768315428
   This might not work though
2.  Run tests in parallel with creating image file and uploading - this should be easy to implement as a background process maybe.

PR #2066, opened 2023-12-21
Switch to v4 actions/{upload,download} and start using `pattern:` feature

mathbunnyru:
## Describe your changes

v4 is super fast
Waiting for https://github.com/actions/download-artifact/issues/249 to be fixed though

## Issue ticket if applicable

Close: https://github.com/jupyter/docker-stacks/issues/1909
I don't think these speedups will be that relevant if download/upload actions will become as fast as they should be.


mathbunnyru:
It works! 🎉 
https://github.com/jupyter/docker-stacks/actions/runs/7291094963/job/19869591566

I will also delete the step to display the structure of files, because it is already displayed good enough in the download action.

PR #2074, opened 2024-01-04
[FAST_BUILD] Fix Docker healthcheck when using custom runtime dirs

### LINKED ISSUE #2073: [BUG] Healthcheck fails when using a custom runtime dir
### What docker image(s) are you using?

scipy-notebook (but applies to all images based on the `base-notebook` image)

### Host OS system

RHEL 8.0

### Host architecture

x86_64

### What Docker command are you running?

The following command DOES work as expected (default runtime dir):
```
docker run --rm -p 8888:8888 --name jupyter quay.io/jupyter/scipy-notebook:2023-12-25 start-notebook.sh
```

The following command does NOT work as expected (customized runtime dir):
```
docker run --rm -p 8888:8888 --name jupyter -e JUPYTER_RUNTIME_DIR=/home/jovyan/custom-runtime quay.io/jupyter/scipy-notebook:2023-12-25 start-notebook.sh
```

### How to Reproduce the problem?

1. Start the Jupyter container using the commands above.
2. In another terminal, run the healtcheck script: `docker exec jupyter /etc/jupyter/docker_healthcheck.py`
3. Observe the healthcheck script failing due to server state JSON file(s) not being found.

### Command output

```bash session
$ docker run --rm -p 8888:8888 --name jupyter quay.io/jupyter/scipy-notebook:2023-12-25 start-notebook.sh
$ docker exec jupyter /etc/jupyter/docker_healthcheck.py
b'{"version": "2.12.1"}'

$ docker run --rm -p 8888:8888 --name jupyter -e JUPYTER_RUNTIME_DIR=/home/jovyan/custom-runtime quay.io/jupyter/scipy-notebook:2023-12-25 start-notebook.sh
$ docker exec jupyter /etc/jupyter/docker_healthcheck.py
  File "/etc/jupyter/docker_healthcheck.py", line 14, in <module>
    json_file = next(runtime_dir.glob("*server-*.json"))
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
StopIteration
```

### Expected behavior

Healthcheck script to not fail, e.g. to display `b'{"version": "2.12.1"}'`, even with a customized runtime dir.

### Actual behavior

The healthcheck script fails because it cannot find server state JSON files in the hard-coded default runtime dir.

### Anything else?

The problem is that the `/etc/jupyter/docker_healthcheck.py` healtcheck script hard-codes the default runtime directory to search for server JSON state files as below:
https://github.com/jupyter/docker-stacks/blob/fcb20a914ed20e44a96053caf43eef6e12fb4c04/images/base-notebook/docker_healthcheck.py#L13

When this directory is customized for example via `JUPYTER_RUNTIME_DIR`, then the healthcheck script does not work.

The actual problem is when deploying Jupyter containers as services.
The Jupyter images have a default healthcheck configured as below:
https://github.com/jupyter/docker-stacks/blob/fcb20a914ed20e44a96053caf43eef6e12fb4c04/images/base-notebook/Dockerfile#L66-L70

When the healthcheck fails due to a custom runtime dir, the service is restarted continuously.

I think the healthcheck script should use the output of `jupyter --runtime-dir` which respects customizations:
```
$ docker run --rm -e JUPYTER_RUNTIME_DIR=/home/jovyan/custom-runtime quay.io/jupyter/scipy-notebook:2023-12-25 jupyter --runtime-dir
/home/jovyan/custom-runtime
```

If you agree with the above, I can send a PR with this fix.

### Latest Docker version

- [X] I've updated my Docker version to the latest available, and the issue persists

hhromic:
## Describe your changes

This PR fixes the Docker healthcheck script in `base-notebook` when using custom runtime dirs.

## Issue ticket if applicable

Fix #2073 

- [X] I have performed a self-review of my code
- [X] If it is a core feature, I have added thorough tests
- [X] I have updated the documentation for significant changes

### DIFF
--- a/images/base-notebook/docker_healthcheck.py
+++ b/images/base-notebook/docker_healthcheck.py
@@ -2,8 +2,8 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import json
-from pathlib import Path
 import warnings
+from pathlib import Path
 
 import requests
 
--- a/images/base-notebook/docker_healthcheck.py
+++ b/images/base-notebook/docker_healthcheck.py
@@ -3,24 +3,23 @@
 # Distributed under the terms of the Modified BSD License.
 import json
 import os
-import warnings
+import subprocess
 from pathlib import Path
 
 import requests
 
-# Import Jupyter paths ignoring deprecation warning that does not affect this script
-with warnings.catch_warnings():
-    warnings.filterwarnings("ignore", category=DeprecationWarning)
-    from jupyter_core import paths
-
-# Set HOME according to the user in NB_USER
-# This is necessary for paths.jupyter_runtime_dir() to work correctly
-os.environ["HOME"] = str(Path("/home") / os.environ["NB_USER"])
-
 # Several operations below deliberately don't check for possible errors
 # As this is a healthcheck, it should succeed or raise an exception on error
 
-runtime_dir = Path(paths.jupyter_runtime_dir())
+result = subprocess.run(
+    ["jupyter", "--runtime-dir"],
+    check=True,
+    capture_output=True,
+    text=True,
+    env=dict(os.environ) | {"HOME": str(Path("/home") / os.environ["NB_USER"])},
+)
+runtime_dir = Path(result.stdout.rstrip())
+
 json_file = next(runtime_dir.glob("*server-*.json"))
 
 url = json.loads(json_file.read_bytes())["url"]
--- a/images/base-notebook/docker_healthcheck.py
+++ b/images/base-notebook/docker_healthcheck.py
@@ -11,12 +11,16 @@ import requests
 # Several operations below deliberately don't check for possible errors
 # As this is a healthcheck, it should succeed or raise an exception on error
 
+# Docker runs healtchecks using an exec
+# It uses the default user configured when running the image: root for the case of a custom NB_USER or jovyan for the case of the default image user.
+# We manually change HOME to make `jupyter --runtime-dir` report a correct path
+# More information: <https://github.com/jupyter/docker-stacks/pull/2074#issuecomment-1879778409>
 result = subprocess.run(
     ["jupyter", "--runtime-dir"],
     check=True,
     capture_output=True,
     text=True,
-    env=dict(os.environ) | {"HOME": str(Path("/home") / os.environ["NB_USER"])},
+    env=dict(os.environ) | {"HOME": "/home" + os.environ["NB_USER"]},
 )
 runtime_dir = Path(result.stdout.rstrip())
 
--- a/images/base-notebook/docker_healthcheck.py
+++ b/images/base-notebook/docker_healthcheck.py
@@ -20,7 +20,7 @@ result = subprocess.run(
     check=True,
     capture_output=True,
     text=True,
-    env=dict(os.environ) | {"HOME": "/home" + os.environ["NB_USER"]},
+    env=dict(os.environ) | {"HOME": "/home/" + os.environ["NB_USER"]},
 )
 runtime_dir = Path(result.stdout.rstrip())
 
--- a/images/base-notebook/docker_healthcheck.py
+++ b/images/base-notebook/docker_healthcheck.py
@@ -2,6 +2,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import json
+import os
 import warnings
 from pathlib import Path
 
@@ -11,6 +12,10 @@ import requests
 with warnings.catch_warnings():
     warnings.filterwarnings("ignore", category=DeprecationWarning)
     from jupyter_core import paths
+
+# Set HOME according to the user in NB_USER
+# This is necessary for paths.jupyter_runtime_dir() to work correctly
+os.environ["HOME"] = str(Path("/home") / os.environ["NB_USER"])
 
 # Several operations below deliberately don't check for possible errors
 # As this is a healthcheck, it should succeed or raise an exception on error
--- a/images/base-notebook/docker_healthcheck.py
+++ b/images/base-notebook/docker_healthcheck.py
@@ -2,15 +2,20 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import json
-import os
 from pathlib import Path
+import warnings
 
 import requests
+
+# Import Jupyter paths ignoring deprecation warning that does not affect this script
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", category=DeprecationWarning)
+    from jupyter_core import paths
 
 # Several operations below deliberately don't check for possible errors
 # As this is a healthcheck, it should succeed or raise an exception on error
 
-runtime_dir = Path("/home/") / os.environ["NB_USER"] / ".local/share/jupyter/runtime/"
+runtime_dir = Path(paths.jupyter_runtime_dir())
 json_file = next(runtime_dir.glob("*server-*.json"))
 
 url = json.loads(json_file.read_bytes())["url"]
--- a/mypy.ini
+++ b/mypy.ini
@@ -24,9 +24,6 @@ ignore_missing_imports = True
 [mypy-docker.*]
 ignore_missing_imports = True
 
-[mypy-jupyter_core.*]
-ignore_missing_imports = True
-
 [mypy-matplotlib.*]
 ignore_missing_imports = True
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -24,6 +24,9 @@ ignore_missing_imports = True
 [mypy-docker.*]
 ignore_missing_imports = True
 
+[mypy-jupyter_core.*]
+ignore_missing_imports = True
+
 [mypy-matplotlib.*]
 ignore_missing_imports = True
 
--- a/tests/base-notebook/test_healthcheck.py
+++ b/tests/base-notebook/test_healthcheck.py
@@ -42,6 +42,7 @@ LOGGER = logging.getLogger(__name__)
             ["start-notebook.py", "--ServerApp.base_url=/test"],
             "root",
         ),
+        (["JUPYTER_RUNTIME_DIR=/jupyter-runtime"], ["start-notebook.sh"], None),
     ],
 )
 def test_health(
--- a/tests/base-notebook/test_healthcheck.py
+++ b/tests/base-notebook/test_healthcheck.py
@@ -43,6 +43,15 @@ LOGGER = logging.getLogger(__name__)
             "root",
         ),
         (["JUPYTER_RUNTIME_DIR=/tmp/jupyter-runtime"], ["start-notebook.sh"], None),
+        (
+            [
+                "NB_USER=testuser",
+                "CHOWN_HOME=1",
+                "JUPYTER_RUNTIME_DIR=/tmp/jupyter-runtime",
+            ],
+            ["start-notebook.sh"],
+            "root",
+        ),
     ],
 )
 def test_health(
--- a/tests/base-notebook/test_healthcheck.py
+++ b/tests/base-notebook/test_healthcheck.py
@@ -42,7 +42,7 @@ LOGGER = logging.getLogger(__name__)
             ["start-notebook.py", "--ServerApp.base_url=/test"],
             "root",
         ),
-        (["JUPYTER_RUNTIME_DIR=/jupyter-runtime"], ["start-notebook.sh"], None),
+        (["JUPYTER_RUNTIME_DIR=/tmp/jupyter-runtime"], ["start-notebook.sh"], None),
     ],
 )
 def test_health(


images/base-notebook/docker_healthcheck.py
```diff
@@ -2,15 +2,20 @@
...
 import json
-import os
+import warnings
 from pathlib import Path
 
 import requests
 
+# Import Jupyter paths ignoring deprecation warning that does not affect this script
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", category=DeprecationWarning)
```
mathbunnyru, line 12: I think DeprecationWarning is just a warning (so it doesn't stop execution). I would prefer seeing this warning and not ignoring it.

hhromic: > (so it doesn't stop execution)

That's correct, the warning does not affect the actual functionality of the healthcheck script.

The reason I decided to silence the warning is that it might get confusing for users.
With the warning visible, the output of the healtcheck script looks like this:
```
$ docker exec jupyter /etc/jupyter/docker_healthcheck.py
/etc/jupyter/docker_healthcheck.py:13: DeprecationWarning: Jupyter is migrating its paths to use standard platformdirs
given by the platformdirs library.  To remove this warning and
see the appropriate new directories, set the environment variable
`JUPYTER_PLATFORM_DIRS=1` and then run `jupyter --paths`.
The use of platformdirs will be the default in `jupyter_core` v6
  from jupyter_core import paths
b'{"version": "2.12.2"}'
```

Users will not normally run this script manually, but in case the container is unhealthy, they will see this output from Docker when reporting the state of the healthcheck failure.

Note that the output of the `jupyter --runtime-dir` command does not emit any deprecation warnings because the warning only appears when `jupyter_core.paths` is imported from a non "internal" module ([see source here](https://github.com/jupyter/jupyter_core/blob/9ea909732fda6b7a674a8b7aa8233602e0b6af52/jupyter_core/utils/__init__.py#L76-L90)).

I thought of preserving the same UX for the Docker healthcheck script while directly importing the `jupyter_core.paths` module and not having to externally call the `jupyter --runtime-dir` command as a subprocess.

Let me know if you still would prefer the warning to be emitted.

hhromic:
@mathbunnyru I'm getting the following pre-commit mypy error:
```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

images/base-notebook/docker_healthcheck.py:13: error: Cannot find implementation or library stub for module named "jupyter_core"  [import-not-found]
images/base-notebook/docker_healthcheck.py:13: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 66 source files)
```

This is expected because the `jupyter_core` module does not exist in the testing environment.
Do you have any suggestions on how to handle this situation?
Can we instruct mypy to ignore this particular import and its attributes?

hhromic:
@mathbunnyru I realized that the following tests are failing with this PR:
```
FAILED tests/base-notebook/test_healthcheck.py::test_health[env13-None-root] - AssertionError: assert 'unhealthy' == 'healthy'
FAILED tests/base-notebook/test_healthcheck.py::test_health[env15-cmd15-root] - AssertionError: assert 'unhealthy' == 'healthy'
FAILED tests/base-notebook/test_healthcheck.py::test_health[env14-cmd14-root] - AssertionError: assert 'unhealthy' == 'healthy'
FAILED tests/base-notebook/test_healthcheck.py::test_not_healthy[env1-cmd1-None] - RuntimeError: Container should not be healthy for these testcases.
FAILED tests/base-notebook/test_healthcheck.py::test_not_healthy[env0-None-None] - RuntimeError: Container should not be healthy for these testcases.
FAILED tests/base-notebook/test_healthcheck.py::test_health_proxy[env1-cmd1-root] - AssertionError: assert 'unhealthy' == 'healthy'
FAILED tests/base-notebook/test_healthcheck.py::test_not_healthy[env2-cmd2-None] - RuntimeError: Container should not be healthy for these testcases.
```
The reason is that all of them set a custom `NB_USER` which is not taken into consideration by `jupyter --runtime-dir`:
```
$ docker run --rm -e NB_USER=testuser -e CHOWN_HOME=1 quay.io/jupyter/base-notebook:latest jupyter --runtime-dir
/home/jovyan/.local/share/jupyter/runtime
```
I need to investigate this case a bit more. In the meantime I'll set the PR to Draft.

mathbunnyru:
> @mathbunnyru I'm getting the following pre-commit mypy error:
> 
> ```
> mypy.....................................................................Failed
> - hook id: mypy
> - exit code: 1
> 
> images/base-notebook/docker_healthcheck.py:13: error: Cannot find implementation or library stub for module named "jupyter_core"  [import-not-found]
> images/base-notebook/docker_healthcheck.py:13: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
> Found 1 error in 1 file (checked 66 source files)
> ```
> 
> This is expected because the `jupyter_core` module does not exist in the testing environment. Do you have any suggestions on how to handle this situation? Can we instruct mypy to ignore this particular import and its attributes?
> 
> EDIT: Oh, I just realized that I can just add `jupyter_core` to be ignored in `mypy.ini`. I will do that for now but let me know of what else you would prefer.

First of all, you need to add this module here: https://github.com/jupyter/docker-stacks/blob/main/.pre-commit-config.yaml#L42
This way this module will be available during mypy execution (which happens in an isolated environment).
Then there are 3 cases:
1. If this module already has typing information, then we don't need anything changed in `mypy.ini`.
2. Some modules have a separate `types-<package_name>` package which adds typing information (in such a case, you can add it to `.pre-commit-config.yaml` as well
3. The module doesn't provide typing information in any way - then it still should be installed during mypy check and then your fix becomes valid.

mathbunnyru:
> @mathbunnyru I realized that the following tests are failing with this PR:
> 
> ```
> FAILED tests/base-notebook/test_healthcheck.py::test_health[env13-None-root] - AssertionError: assert 'unhealthy' == 'healthy'
> FAILED tests/base-notebook/test_healthcheck.py::test_health[env15-cmd15-root] - AssertionError: assert 'unhealthy' == 'healthy'
> FAILED tests/base-notebook/test_healthcheck.py::test_health[env14-cmd14-root] - AssertionError: assert 'unhealthy' == 'healthy'
> FAILED tests/base-notebook/test_healthcheck.py::test_not_healthy[env1-cmd1-None] - RuntimeError: Container should not be healthy for these testcases.
> FAILED tests/base-notebook/test_healthcheck.py::test_not_healthy[env0-None-None] - RuntimeError: Container should not be healthy for these testcases.
> FAILED tests/base-notebook/test_healthcheck.py::test_health_proxy[env1-cmd1-root] - AssertionError: assert 'unhealthy' == 'healthy'
> FAILED tests/base-notebook/test_healthcheck.py::test_not_healthy[env2-cmd2-None] - RuntimeError: Container should not be healthy for these testcases.
> ```
> 
> The reason is that all of them set a custom `NB_USER` which is not taken into consideration by `jupyter --runtime-dir`:
> 
> ```
> $ docker run --rm -e NB_USER=testuser -e CHOWN_HOME=1 quay.io/jupyter/base-notebook:latest jupyter --runtime-dir
> /home/jovyan/.local/share/jupyter/runtime
> ```
> 
> I need to investigate this case a bit more. In the meantime I'll set the PR to Draft.

Be careful here.
1. You forgot to include `start.sh`, which actually handles most of the options, including user-renaming.
2. We also need to include `--user root` to actually change the user name

So, the correct line looks like this `docker run --rm -e NB_USER=testuser -e CHOWN_HOME=1 --user root quay.io/jupyter/base-notebook:latest start.sh jupyter --runtime-dir`
And this actually produces the correct result: `/home/testuser/.local/share/jupyter/runtime`

mathbunnyru:
My assumption - health check is run by using `docker exec`, not `docker run`. And `docker exec` doesn't know anything about our user-renaming (it probably runs from `root` or from `jovyan` user).

Our current script seems to work, because it seems that we have the correct environment variable for `NB_USER` during running a healthcheck.
But in your implementation, the script is probably run from another user, so it fails.

Could you please verify this is correct?

images/base-notebook/docker_healthcheck.py
mathbunnyru: Let's call a subprocess with `jupyter --runtime-dir`, so we don't have to suppress warnings.
This has one more benefit for you - you don't have to deal with `mypy` at all 🙂

hhromic:
@mathbunnyru I figured out what was the problem now :)

Docker runs healtchecks using an exec, yes. It uses the default user configured when running the image: `root` for the case of a custom `NB_USER` or `jovyan` for the case of the default image user.

The problem is as follows.

The `jupyter_runtime_dir()` function uses the `HOME` env variable to resolve `~` (the default base dir) for computing the default paths (e.g. when not customizing `JUPYTER_RUNTIME_DIR`). However, the image hard-codes `HOME` to `/home/jovyan` regardless if `NB_USER` has been customized and therefore the default paths were always returned as `/home/jovyan/.local/share/jupyter/runtime`.

I now fixed this by ensuring that `HOME` is set on-the-fly according to `NB_USER` before calling `jupyter_runtime_dir()` in the healthcheck script. Threfore it always returns the correct default for the atcive `NB_USER` usre.

I also added an extra test case to test this situation as well just in case.
All tests are passing now, let me know what do you think!

mathbunnyru:
So, I was right here: https://github.com/jupyter/docker-stacks/pull/2074#issuecomment-1879755740? 🙂

hhromic:
> So, I was right here: [#2074 (comment)](https://github.com/jupyter/docker-stacks/pull/2074#issuecomment-1879755740)? 🙂

If you meant computing home from `NB_USER`, yes :)
See the comment I just wrote above with a detailed explanation.

hhromic:
> First of all, you need to add this module here: https://github.com/jupyter/docker-stacks/blob/main/.pre-commit-config.yaml#L42 This way this module will be available during mypy execution (which happens in an isolated environment). Then there are 3 cases:
> 
> 1. If this module already has typing information, then we don't need anything changed in `mypy.ini`.
> 2. Some modules have a separate `types-<package_name>` package which adds typing information (in such a case, you can add it to `.pre-commit-config.yaml` as well
> 3. The module doesn't provide typing information in any way - then it still should be installed during mypy check and then your fix becomes valid.

I assume `jupyter_core` is a heavy dependency for such a small piece of code in the healthcheck script (only).
But I can add it there (thanks for the hint!). I'll take a look now.

images/base-notebook/docker_healthcheck.py
```diff
@@ -3,14 +3,24 @@
...
+    from jupyter_core import paths
+
+# Set HOME according to the user in NB_USER
+# This is necessary for paths.jupyter_runtime_dir() to work correctly
+os.environ["HOME"] = str(Path("/home") / os.environ["NB_USER"])
+
 # Several operations below deliberately don't check for possible errors
 # As this is a healthcheck, it should succeed or raise an exception on error
```
mathbunnyru, line 21: Let's change these 2 sections

images/base-notebook/docker_healthcheck.py
hhromic: I just saw this comment now.
Yes I also contemplated that option (calling `jupyter --runtime-dir` as a command) but I thought it is a bit overkill considering it would be a Python -> Shell -> Python chain of processes for a single line of Python code :)

Can you confirm you really prefer to not supress the warning so I proceed with that?

mathbunnyru: I don't thinks it's an overkill, if it makes code look simpler, unlikely to break and doesn't produce warnings.
Yes, I confirm I prefer subprocess in this case 🙂

images/base-notebook/docker_healthcheck.py
hhromic: Apologies, the Github review history became a bit confusing to read for me now :)
What change exactly did you mean?

mathbunnyru: I prefer keeping "Several operations below deliberately don't check..." at the top of this file, because it describes the script design.

hhromic: Oh you meant re-ordering them. Sure!

mathbunnyru: Sorry, I meant swap. It's been a long day 😆

hhromic:
@mathbunnyru replaced using `jupyter_core.paths` with calling `jupyter --runtime-dir` in a subprocess now :)
All tests are passing, let me know what do you think!

mathbunnyru (approved):
I added a comment and removed usage of Path - it creates more noise here.

hhromic:
> I added a comment and removed usage of Path - it creates more noise here.

Sounds good to me!
I used Path objects just to be consistent with the rest of the code, but you are right that in the end it is just string manipulation in that part and not any path validation/joining 👍.

Cheers and thanks for the interesting/productive review conversation! 🍻

mathbunnyru:
Thank you for your patience and thorough issue and a great PR @hhromic.
Good luck using our images 🙂

PR #2075, opened 2024-01-05
Automatically install latest spark version

mathbunnyru:
## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/2063
Close: https://github.com/jupyter/docker-stacks/issues/1937

## Checklist (especially for first-time contributors)

- [ ] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests
- [ ] I will try not to use force-push to make the review process easier for reviewers
- [ ] I have updated the documentation for significant changes

PR #2076, opened 2024-01-06
Fix `max(stable_versions)` on Julia version finding

jmuchovej:
## Describe your changes

Since the keys are semantic version strings, that means that `"1.9.4" > "1.10.0"`, which we know isn't true. 🙂

I just added some code to convert the string to tuples, find the max, then convert back to a string.

I first noticed this on the `2024-01-05` build of the `datascience-notebook`, since Julia 1.10.0 was released ~2 weeks ago: https://github.com/JuliaLang/julia/releases/tag/v1.10.0.

### DIFF
--- a/images/minimal-notebook/setup-scripts/setup_julia.py
+++ b/images/minimal-notebook/setup-scripts/setup_julia.py
@@ -37,8 +37,7 @@ def get_latest_julia_url() -> tuple[str, str]:
     ).json()
     stable_versions = {k: v for k, v in versions.items() if v["stable"]}
     latest_stable_version = max(
-        stable_versions,
-        key=lambda ver: [int(sub_ver) for sub_ver in ver.split(".")]
+        stable_versions, key=lambda ver: [int(sub_ver) for sub_ver in ver.split(".")]
     )
     latest_version_files = stable_versions[latest_stable_version]["files"]
     triplet = unify_aarch64(platform.machine()) + "-linux-gnu"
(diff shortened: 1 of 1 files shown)


images/minimal-notebook/setup-scripts/setup_julia.py
```diff
@@ -36,7 +36,10 @@ def get_latest_julia_url() -> tuple[str, str]:
         "https://julialang-s3.julialang.org/bin/versions.json"
     ).json()
     stable_versions = {k: v for k, v in versions.items() if v["stable"]}
-    latest_version_files = stable_versions[max(stable_versions)]["files"]
```
mathbunnyru, line 39: Thanks!
I think it is better to add comparator: `key=lambda ver: [int(sub_ver) for sub_ver in ver.split(".")]` as sort argument.

Funny fact: I actually thought about this the second time I was writing such a script, https://github.com/jupyter/docker-stacks/pull/2075/files

mathbunnyru (changes requested):

images/minimal-notebook/setup-scripts/setup_julia.py
```diff
@@ -36,7 +36,12 @@ def get_latest_julia_url() -> tuple[str, str]:
...
+    latest_version_files = stable_versions[
+        max(
+            stable_versions,
+            key=lambda ver: [int(sub_ver) for sub_ver in ver.split(".")],
+        )
+    ]
```
mathbunnyru, line 44:
```suggestion
    ]["files"]
```

mathbunnyru:
I checked the logs `INFO:tagging.docker_runner:Command result: julia version 1.10.0`.
Seems to work, thank you, merging this one.

jmuchovej:
> I rearranged the code just a little bit (sometimes it's good to have additional variables), hope you don't mind.

All good! That's partially why I took the approach I did at first. Strong preference towards sensible variables that try to avoid chaining lots of things together. 🙂

Glad to see this got merged! Thanks for your quick work on, too. 😄

PR #2081, opened 2024-01-16
Do not bloat spark image with ENV variables

mathbunnyru:

### DIFF
--- a/images/pyspark-notebook/Dockerfile
+++ b/images/pyspark-notebook/Dockerfile
@@ -34,12 +34,6 @@ ARG scala_version
 # But it seems to be slower, that's why we use the recommended site for download
 ARG spark_download_url="https://dlcdn.apache.org/spark/"
 
-# Configure Spark
-ENV SPARK_VERSION="${spark_version}" \
-    HADOOP_VERSION="${hadoop_version}" \
-    SCALA_VERSION="${scala_version}" \
-    SPARK_DOWNLOAD_URL="${spark_download_url}"
-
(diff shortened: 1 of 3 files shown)


tagging/taggers.py
```diff
@@ -136,12 +124,6 @@ def tag_value(container: Container) -> str:
         return "spark-" + version_line.split(" ")[-1]
 
 
-class HadoopVersionTagger(TaggerInterface):
-    @staticmethod
-    def tag_value(container: Container) -> str:
-        return "hadoop-" + _get_env_variable(container, "HADOOP_VERSION")
```
mathbunnyru, line 142: This is pretty useless, because it just says "3". And it custom builds people obviously know their Hadoop version, so they are not interested in tagging.

I also think that getting versions from env is not the best solution, programs should be able to tell their versions themselves.

images/pyspark-notebook/Dockerfile
```diff
@@ -34,20 +34,18 @@ ARG scala_version
 # But it seems to be slower, that's why we use the recommended site for download
 ARG spark_download_url="https://dlcdn.apache.org/spark/"
 
-# Configure Spark
-ENV SPARK_VERSION="${spark_version}" \
-    HADOOP_VERSION="${hadoop_version}" \
-    SCALA_VERSION="${scala_version}" \
-    SPARK_DOWNLOAD_URL="${spark_download_url}"
```
mathbunnyru, line 41: These env variable names are not something common, it's something we made up, so I don't think we should set them in the images.

PR #2087, opened 2024-01-21
Make `start.sh` the entrypoint

manics:
Move `start.sh` from CMD to ENTRYPOINT

This ensures the setup steps performed in start.sh will always be run unless the user overrides the entrypoint. For example, someone can pass `jupyterhub-singleuser` (the default with KubeSpawner) and the image should "just work" as intuitively expected: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2138#issuecomment-828291826

There's a check to prevent re-execution of the script for anyone who continues to pass it as the first CMD.

Closes https://github.com/jupyter/docker-stacks/issues/1528

### DIFF
--- a/images/docker-stacks-foundation/start.sh
+++ b/images/docker-stacks-foundation/start.sh
@@ -34,8 +34,8 @@ else
     cmd=( "$@" )
 fi
 
-# Backwards compatibility: start.sh is executed by default in ENTRYPOINT so
-# should no longer be specified in CMD
+# Backwards compatibility: `start.sh` is executed by default in ENTRYPOINT
+# so it should no longer be specified in CMD
 if [ "${_START_SH_EXECUTED}" = "1" ]; then
     _log "WARNING: start.sh is the default ENTRYPOINT, do not include it in CMD"
     _log "Executing the command:" "${cmd[@]}"
--- a/images/docker-stacks-foundation/start.sh
+++ b/images/docker-stacks-foundation/start.sh
@@ -34,6 +34,17 @@ else
     cmd=( "$@" )
 fi
 
+# Backwards compatibility: start.sh is executed by default in ENTRYPOINT so
+# should no longer be specified in CMD
+if [ "${_START_SH_EXECUTED}" = "1" ]; then
+    _log "WARNING: start.sh is the default ENTRYPOINT, do not include it in CMD"
+    _log "Executing the command:" "${cmd[@]}"
+    exec "${cmd[@]}"
+else
+    export _START_SH_EXECUTED=1
+fi
+
+
 # NOTE: This hook will run as the user the container was started with!
 # shellcheck disable=SC1091
 source /usr/local/bin/run-hooks.sh /usr/local/bin/start-notebook.d
--- a/tests/docker-stacks-foundation/test_python_version.py
+++ b/tests/docker-stacks-foundation/test_python_version.py
@@ -17,13 +17,7 @@ def test_python_version(container: TrackedContainer) -> None:
         tty=True,
         command=["python", "--version"],
     )
-    python = None
-    for line in logs.splitlines():
-        if line.startswith("Python "):
-            python = line
-            break
-
-    assert python is not None
+    python = next(line for line in logs.splitlines() if line.startswith("Python "))
     full_version = python.split()[1]
     major_minor_version = full_version[: full_version.rfind(".")]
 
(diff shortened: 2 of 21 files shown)


images/docker-stacks-foundation/start.sh
```diff
@@ -34,6 +34,17 @@ else
     cmd=( "$@" )
 fi
 
+# Backwards compatibility: start.sh is executed by default in ENTRYPOINT so
+# should no longer be specified in CMD
```
mathbunnyru, line 38:
```suggestion
# Backwards compatibility: `start.sh` is executed by default in ENTRYPOINT
# so it should no longer be specified in CMD
```

tests/docker-stacks-foundation/test_user_options.py
```diff
@@ -284,7 +282,22 @@ def test_secure_path(container: TrackedContainer, tmp_path: pathlib.Path) -> Non
...
         user="root",
         volumes={p: {"bind": "/usr/bin/python", "mode": "ro"}},
-        command=["start.sh", "python", "--version"],
+        command=["python", "--version"],
     )
     assert "Wrong python" not in logs
     assert "Python" in logs
+
+
+def test_startsh_multiple_exec(container: TrackedContainer) -> None:
```
mathbunnyru, line 291: Thank you for adding a test 👍

mathbunnyru: Fixed it as well 🙂

mathbunnyru:
Just to be sure - you searched for all the occurrences of `start.sh` and either removed them (most likely in tests) or rephrased them (mostly in docs)?

manics:
I missed one mention, fixed in the last commit.

tests/docker-stacks-foundation/test_python_version.py
```diff
@@ -17,8 +17,9 @@ def test_python_version(container: TrackedContainer) -> None:
         tty=True,
         command=["python", "--version"],
     )
-    assert logs.startswith("Python ")
-    full_version = logs.split()[1]
+    python = logs.splitlines()[7]
```
mathbunnyru, line 20: I don't really like hardcoding line number. Any log line added or removed from `start.sh` will break this test for no reason.

To solve this, for example, we can choose the line which starts with `Python ` instead.

The same thing is true for the test below.

mathbunnyru (changes requested):

manics:
I've worked out what's causing the test to fail. Previously start-notebook.py ran start.sh ...., now it's the other way around, so now the output is
```
Entered start.sh with args: start-notebook.py
```
and start-notebook.py runs `jupyter lab`, but doesn't log it

tests/docker-stacks-foundation/test_python_version.py
```diff
@@ -17,8 +17,8 @@ def test_python_version(container: TrackedContainer) -> None:
         tty=True,
         command=["python", "--version"],
     )
-    assert logs.startswith("Python ")
-    full_version = logs.split()[1]
+    python = next(line for line in logs.splitlines() if line.startswith("Python "))
```
mathbunnyru, line 20: I think I simplified this line so it stays quite readable and is still robust to logging changes in start.sh

tests/docker-stacks-foundation/test_python_version.py
```diff
@@ -31,4 +31,4 @@ def test_python_pinned_version(container: TrackedContainer) -> None:
         tty=True,
         command=["cat", "/opt/conda/conda-meta/pinned"],
     )
-    assert logs.startswith(f"python {EXPECTED_PYTHON_VERSION}.*")
+    assert f"python {EXPECTED_PYTHON_VERSION}.*" in logs
```
mathbunnyru, line 34: Let's have the simplest check possible here, should be just fine for our purposes.

mathbunnyru (approved):
Overall, great job, @manics, thank you :heart:

I will merge this when tests pass.

yuvipanda:
NICE!

mathbunnyru:
@manics @minrk @yuvipanda I merged this 2 days ago, and it seems to work well 👍
I think, there might be some workarounds in the jupyterhub projects, which might no longer be needed, but I am not sure.

@manics, thank you for this PR, it's great.

minrk:
@mathbunnyru awesome! I think we've mostly removed the workarounds in jupyterhub for more consistent behavior with images not from docker-stacks, so I think this will more directly affect _users_ with a better experience and less configuration required. There are likely docs that should be updated.

consideRatio:
Wieee this is turning out very helpful right now, thanks to this we can still get the benefits from `start.sh` while overriding CMD for all images to ensure a vulnerable package is upgraded. Super happy to see this already proposed and merged!!

Issue #1557, opened 2021-12-24
Small Change to Makefile to facilitate GPU Variants of Current Images


sdwalker62:
Hello everyone! First, I just want to say that at work we almost exclusively use Jupyter as our data science platform and appreciate everything that the community and the Jupyter team has done to create such a wonderful tool!

I have searched through the current list of issues and pull requests open in this repository and have not found and that match this request so I am starting a new one. Apologies if I have missed any.  Our work necessitates the use of cuda-enabled containers as we primarily train deep learning models. We have made a slight modification to the makefile that allows us to re-use most of the infrastructure already in place to build the various levels of notebooks. 

---

I am aware of the fantastic work being done at https://github.com/iot-salzburg/gpu-jupyter/ and suggest anyone who needs gpu-enabled containers to check it out as a first stop. This solution will work for many, and it contains some additional features which are nice to have, but it is slightly different than what we prepose. 

The current docker-stacks starts with the base-notebook which is built from an Ubuntu 20.04 layer. Our preposal is to build a separate set of notebooks based on the nvidia/cuda:11.3.0-cudnn8-runtime-ubuntu20.04 image (we chose this for compatibility with Pytorch). This is similar to what the iot-salzburg repository does in their images but maintains the minimalism found in some of the base notebooks for those who are looking for a solid foundation for their own custom images. With the change to the makefiles it would be easy to build two sets of images: one for the standard notebooks as they exist now, and another for the cuda-enabled variants. 

To address the selection criteria found in the documention for new features: 

* The majority of the data science community working on neural networks uses GPUs to train their models and being able to pull down mirrors of the docker-stacks containers with cuda backed in would be tremendously useful. 
* One of the most beautiful things about the docker-stacks project is the ability to base custom images on any of the notebook images depending on the requirements of the user. Building two sets of containers would be inline with the current design philosophy of the project. 
* The complexity of this change is low. We have a simply python script that replaces lines in the base-notebook and a new section in the Makefile that is minimal and could be hooked into the current build process. 
* The build times are identical to the current build times as only the base layers changes. None of the current images will change, only new tags added. 
* Sustainment of the change should not be difficult at all as the nvidia/cuda containers are regularly updated by Nvidia. Any test that need to be written we can do to support the pull request. 

I would be happy to answer any questions about this proposal!

mathbunnyru:
@sdwalker62 first of all, thank you for your suggestion!
I'm sorry it took such a long time to answer you.

[...]

PR #2091, opened 2024-02-06
Add pytorch-notebook image variants with cuda 11 and 12 (x86_64 versions only)

johanna-reiml-hpi:
## Describe your changes

This reworks the build pipeline to allow building CUDA variants for the `pytorch-notebook` image. Examples images: https://quay.io/repository/ai-hpi/pytorch-notebook

- **Commit** "feat: build cuda variants of pytorch" (43f2d320): minimal example for how to add cuda variants for the pytorch-notebook image.
- **Commit** "feat: build with variant tag" (6198575d): rework the pipeline to build these images under the tag of the `pytorch-notebook` instead (adds a "cuda-" and "cuda11-" prefix). This logic is implemented by adding a new "variant" argument for the GitHub Actions and the tagging scripts. I also refactored `docker.yaml` to get rid of the redundancy for the aarch64 vs x86_64 build.

In the [issue](https://github.com/jupyter/docker-stacks/issues/1557) there were discussions about using the [official CUDA images](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/cuda/tags) instead of `ubuntu:22.04` as the root container. This is not desirable in my opinion, as these images are not updated frequently and are based on older `ubuntu:22.04` images. Instead a simpler approach works: packages like pytorch and tensorflow bundle their own CUDA binaries via pip these days.

I looked into CUDA builds for TensorFlow and PyTorch (I am not sure if CUDA support is really a priority for other notebooks). In order to [support NVIDIA driver versions](https://docs.nvidia.com/deploy/cuda-compatibility/) older than 525, builds for CUDA 11.8 should also be included.

- For PyTorch (https://pytorch.org/get-started/locally/) everything works as expected. I also added NVIDIA's pip index for their most recent packages as an install argument `--extra-index-url=https://pypi.nvidia.com`, but it's likely not needed.
- For TensorFlow (https://www.tensorflow.org/install/source?hl=en#gpu) there were multiple issues:
    - CUDA 11: The most recent version (2.15.0) updated the CUDA dependencies from 11.8 to 12.2. So the last version which supports CUDA 11.8 is 2.14.0. However, that version is not functional under python 3.11, because a dependency (tensorrt==8.5.3.1) is not available for python 3.11: https://github.com/tensorflow/tensorflow/issues/61986
    - CUDA 12 (latest): Using `pip install --no-cache-dir --extra-index-url=https://pypi.nvidia.com tensorrt tensorflow[and-cuda]` didn't seem to work for me and tensorflow was [unable to find tensorrt](https://github.com/tensorflow/tensorflow/issues/61468). I assume the workaround in the linked issue would work, but it seems better to either manually build the package, try nightly builds, or wait for 2.16.0.

## Issue ticket if applicable

Fix (for pytorch): https://github.com/jupyter/docker-stacks/issues/1557

    - [x] Example build: https://github.com/johanna-reiml-hpi/docker-stacks/actions/runs/7788922093
    - [x] Example images: https://quay.io/repository/ai-hpi/pytorch-notebook?tab=tags
    - [x] No aarch64 tests yet (I assume you want to use your self-hosted runners)
    - [x] Not sure what parts need to be updated. I mentioned the existence of optional CUDA tags under the pytorch-notebook section.


johanna-reiml-hpi:
From what I can see the build for aarch64 seems to fail due to running out of disk space. Is there an issue with also enabling this part for the self-hosted aarch64 runners? Right now it only runs for x86_64.

```yaml
      # Image with CUDA needs extra disk space
      - name: Free disk space 🧹
        if: contains(inputs.variant, 'cuda') && inputs.platform == 'x86_64'
        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be
        with:
          tool-cache: false
          android: true
          dotnet: true
          haskell: true
          large-packages: false
          docker-images: false
          swap-storage: false
```

tagging/get_prefix.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def get_file_prefix_for_platform(platform: str, name: str) -> str:
+    return f"{platform}-{name}"
+
+
+def get_tag_prefix_for_platform(platform: str, name: str) -> str:
+    if name == DEFAULT_VARIANT:
+        return platform
+    return f"{platform}-{name}"
```
mathbunnyru, line 15:
```suggestion
def get_file_prefix_for_platform(platform: str, variant: str) -> str:
    return f"{platform}-{variant}"


def get_tag_prefix_for_platform(platform: str, variant: str) -> str:
    if variant == DEFAULT_VARIANT:
        return platform
    return f"{platform}-{variant}"
```

docs/using/selecting.md
```diff
@@ -195,7 +195,7 @@ It contains:
 `jupyter/pytorch-notebook` includes popular Python deep learning libraries.
 
 - Everything in `jupyter/scipy-notebook` and its ancestor images
-- [pytorch](https://pytorch.org/) machine learning library
+- [pytorch](https://pytorch.org/) machine learning library (optionally with CUDA)
```
mathbunnyru, line 198: I think we should add a note section, which says that you should add a `cuda-` or `cuda11-` prefix to the tag to get GPU-enabled versions and explain the difference.

Makefile
```diff
@@ -66,11 +67,11 @@ linkcheck-docs: ## check broken links
 	sphinx-build -W --keep-going --color -b linkcheck docs/ docs/_build/
 
 
-
+hook/%: VARIANT?=default
```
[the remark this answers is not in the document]
mathbunnyru: There are always 3 empty lines between sections, let's keep it this way for consistency.

.github/workflows/docker-build-test-upload-all.yml
```diff
@@ -0,0 +1,244 @@
+name: Build, test, and push Docker Images
```
mathbunnyru, line 1:
```suggestion
name: Build, test, and push Docker Images for one platform
```

.github/workflows/docker-build-test-upload-all.yml
```diff
@@ -0,0 +1,244 @@
...
+  OWNER: ${{ github.repository_owner }}
+
+on:
+  workflow_call:
+    inputs:
+      platform:
+        description: Image platform
+        required: true
+        type: string
+      runs-on-fast:
```
mathbunnyru, line 14:
```suggestion
      fast-runner-tag:
```

.github/workflows/docker-build-test-upload-all.yml
```diff
@@ -0,0 +1,244 @@
...
+        description: GitHub Actions Runner image (fast)
+        required: true
+        type: string
+      runs-on-slow:
```
mathbunnyru, line 18:
```suggestion
      slow-runner-tag:
```

.github/workflows/docker-build-test-upload-all.yml
mathbunnyru, line 1: I would have preferred having this refactoring in a separate PR first, but I'm ok keeping it here.
I have another concern though  - can we make left side of the run better?
With this PR, I need to click all build to find `datascience` build, for example.

Screenshot 2024-02-08 at 13 25 08

.github/workflows/docker.yml
```diff
@@ -1,4 +1,4 @@
-name: Build, test, and push Docker Images
+name: Build, merge tags, update wiki, test recipes
 
 # [FAST_BUILD] in the PR title makes this workflow only build
```
mathbunnyru, line 3: We need to move comment about `[FAST_BUILD]` to a new file.

.github/workflows/docker.yml
```diff
@@ -60,380 +62,86 @@ concurrency:
...
+            { image: all-spark-notebook, variant: default },
           ]
-    needs: [aarch64-images-tag-push, x86_64-images-tag-push]
-    if: github.repository_owner == 'jupyter' && !contains(github.event.pull_request.title, '[FAST_BUILD]')
+    needs: [build-aarch64, build-x86_64]
+    if: |
+      always() &&
+      needs.build-x86_64.result == 'success' &&
+      (needs.build-aarch64.result == 'success' || needs.build-aarch64.result == 'skipped') &&
+      !contains(github.event.pull_request.title, '[FAST_BUILD]')
```
mathbunnyru, line 117: Could you please explain why this need to this complicated?

mathbunnyru (changes requested):

mathbunnyru:
> there an issue with also enabling this part for the self-hosted aarch64 runners?

Self-hosted runners are not usually bloated with lots of software. I created our aarch64 runners using ubuntu provided by Google Cloud and running this on top: https://github.com/jupyter/docker-stacks/blob/main/aarch64-runner/setup.sh

Could you please tell me how much free disk is needed to build one GPU image?
I am not sure people need aarch64 GPU images, so let's not build them in this PR, though.
Also, this will increase the build time (we have limited amount of self-hosted runners)

.github/workflows/docker-build-test-upload-all.yml
mathbunnyru: I also think you can still create a separate PR for this change (having `docker-build-test-upload-all.yml` file), then this PR will look much more readable.

Makefile
```diff
@@ -36,8 +36,9 @@ help:
 
 
 build/%: DOCKER_BUILD_ARGS?=
+build/%: ROOT_CONTAINER?=ubuntu:22.04
```
mathbunnyru, line 39: Slightly unrelated I suppose (I would be happy to merge this in a separate PR).

images/pytorch-notebook/Dockerfile
```diff
@@ -11,7 +11,7 @@ LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
 # Fix: https://github.com/koalaman/shellcheck/wiki/SC3014
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
-# Install PyTorch with pip
+# Install PyTorch with pip (https://pytorch.org/get-started/locally/)
```
mathbunnyru, line 14: Thank you for adding a link

images/pytorch-notebook/cuda/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
+
+# Fix: https://github.com/hadolint/hadolint/wiki/DL4006
+# Fix: https://github.com/koalaman/shellcheck/wiki/SC3014
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+# Install PyTorch with pip (https://pytorch.org/get-started/locally/)
+# hadolint ignore=DL3013
+RUN pip install --no-cache-dir --extra-index-url=https://pypi.nvidia.com \
```
mathbunnyru, line 16: One more thought - let's call this image cuda12 and I think we can add something like `--index-url 'https://download.pytorch.org/whl/cu121'` to the installation command.

When cuda13 will be released in the future, we won't break anyone by silently updating their image from cuda12 to cuda13. We will just add a new prefix tag `cuda13` and everyone who can and wants to upgrade will be able to do that with a small change.

.github/workflows/docker-build-test-upload.yml
```diff
@@ -11,10 +11,18 @@ on:
         description: Parent image name
         required: true
         type: string
+      parent-variant:
+        description: Parent variant tag prefix
+        required: true
+        type: string
```
mathbunnyru, line 17: I suggest making `default` the default value here 😃.
Currently, parent image is always default, so we will add the option if someone needs it in the future, but `docker-build-test-upload-all.yml` will be slightly more simple.

mathbunnyru:
@johanna-reiml-hpi I think the quality of this PR is awesome 👍 
Great job.
I left a few comments of some things that may be improved, please, take a look.

We now have a [policy](https://jupyter-docker-stacks.readthedocs.io/en/latest/maintaining/new-images-and-packages-policy.html) for merging new images and adding new packages 🎉

The voting deadline is the 8th of March (a month since I posted this message).

.github/workflows/docker.yml
johanna-reiml-hpi: This is technically not needed for this repo. Should I remove it?

I added this for my own image builds, because skipped aarch64 builds do not count for `needs: [build-aarch64, build-x86_64]`. If this is skipped it will only push tags with the `x86_64_` prefix. So it will not have tags like `:hub-4.0.2` and only `x86_64-hub-4.0.2`.

mathbunnyru: I see. No, I would like my fork to conitnue to work properly (you can see that sometimes I run some things only for `jupyter` GitHub user).
And you might have noticed that your fork works as it should too 🙂

.github/workflows/docker-build-test-upload.yml
johanna-reiml-hpi: Added `default` as the default value.

images/pytorch-notebook/cuda/Dockerfile
johanna-reiml-hpi: I added `--index-url 'https://download.pytorch.org/whl/cu121'` and renamed the `cuda` variant to `cuda12`.

Makefile
johanna-reiml-hpi: Reverted.

.github/workflows/docker.yml
johanna-reiml-hpi: Added this adjusted check to both the `merge-tags-fast` and `merge-tags` (without the refactored ´-all.yaml`).

.github/workflows/docker.yml
johanna-reiml-hpi: Reverted to the original workflow, so let's do this in another PR.

.github/workflows/docker-build-test-upload.yml
```diff
@@ -52,59 +73,70 @@ jobs:
...
         run: docker build --rm --force-rm --tag ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ inputs.image }} images/${{ inputs.image }}/ --build-arg REGISTRY=${{ env.REGISTRY }} --build-arg OWNER=${{ env.OWNER }}
         env:
           DOCKER_BUILDKIT: 1
           # Full logs for CI build
           BUILDKIT_PROGRESS: plain
         shell: bash
 
+      - name: Build image (variant) 🛠
+        if: inputs.variant != 'default'
+        run: docker build --rm --force-rm --tag ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ inputs.image }} images/${{ inputs.image }}/${{ inputs.variant }}/ --build-arg REGISTRY=${{ env.REGISTRY }} --build-arg OWNER=${{ env.OWNER }}
```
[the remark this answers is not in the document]
johanna-reiml-hpi: Added the ternary operator.

.github/workflows/docker-build-test-upload-all.yml
johanna-reiml-hpi: I reverted the refactor.

.github/workflows/docker-build-test-upload-all.yml
johanna-reiml-hpi: Reverted the refactor.

.github/workflows/docker-build-test-upload-all.yml
johanna-reiml-hpi: Reverted the refactor.

.github/workflows/docker-build-test-upload-all.yml
```diff
@@ -0,0 +1,244 @@
...
+
+env:
+  REGISTRY: quay.io
+  OWNER: ${{ github.repository_owner }}
+
+on:
+  workflow_call:
+    inputs:
+      platform:
+        description: Image platform
```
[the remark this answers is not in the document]
johanna-reiml-hpi: Reverted the refactor.

.github/workflows/docker-build-test-upload-all.yml
johanna-reiml-hpi: Reverted the refactor.

Makefile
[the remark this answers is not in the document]
johanna-reiml-hpi: Fixed.

tagging/get_prefix.py
johanna-reiml-hpi: Adjusted.

docs/using/selecting.md
johanna-reiml-hpi: Added:

> **GPU Acceleration:** Append a CUDA version prefix (`cuda12-` for CUDA 12 or `cuda11-` for CUDA 11) to the image tag to allow PyTorch operations to use compatible NVIDIA GPUs for accelerated computation.

johanna-reiml-hpi:
> > there an issue with also enabling this part for the self-hosted aarch64 runners?
> 
> Self-hosted runners are not usually bloated with lots of software. I created our aarch64 runners using ubuntu provided by Google Cloud and running this on top: https://github.com/jupyter/docker-stacks/blob/main/aarch64-runner/setup.sh
> 
> Could you please tell me how much free disk is needed to build one GPU image? I am not sure people need aarch64 GPU images, so let's not build them in this PR, though. Also, this will increase the build time (we have limited amount of self-hosted runners)

I am not sure how much space is exactly needed, but the [14 GB](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-private-repositories) provided for `ubuntu-latest` weren't enough and resulted in failed builds. I assume probably like 20 GB in total would be sufficient.

Makefile
mathbunnyru: Added this to main as a separate commit.

.github/workflows/docker-build-test-upload.yml
```diff
@@ -11,10 +11,19 @@ on:
...
+        type: string
+        default: default
       image:
         description: Image name
         required: true
         type: string
+      variant:
+        description: Variant tag prefix
+        required: true
+        type: string
```
mathbunnyru, line 26: I suggest making default the default value here as well.

.github/workflows/docker.yml
```diff
@@ -65,6 +65,7 @@ jobs:
     with:
       parent-image: ""
       image: docker-stacks-foundation
+      variant: default
```
mathbunnyru, line 68: If you apply the suggestion above, then you won't need to specify `variant: default` here in most of the places.

.github/workflows/docker.yml
```diff
@@ -280,25 +346,28 @@ jobs:
...
+            { image: docker-stacks-foundation, variant: default },
+            { image: base-notebook, variant: default },
+            { image: minimal-notebook, variant: default },
+            { image: scipy-notebook, variant: default },
+            { image: r-notebook, variant: default },
+            { image: julia-notebook, variant: default },
+            { image: tensorflow-notebook, variant: default },
+            { image: pytorch-notebook, variant: default },
+            { image: pytorch-notebook, variant: cuda12 },
+            { image: pytorch-notebook, variant: cuda11 },
```
mathbunnyru, line 367: My personal preference is to have these three sorted: `default`, `cuda11`, `cuda12`.
In other places as well.

tagging/apply_tags.py
```diff
@@ -69,9 +72,20 @@ def apply_tags(
         required=True,
         help="Owner of the image",
     )
+    arg_parser.add_argument(
+        "--variant",
+        required=True,
+        type=str,
```
mathbunnyru, line 78:
```suggestion
```

Not needed here, it will be string automatically, matches `--owner`.

mathbunnyru (commented):
Only small comments. If you have some time, let's change these small details as well.

mathbunnyru:
> > > there an issue with also enabling this part for the self-hosted aarch64 runners?
> > 
> > 
> > Self-hosted runners are not usually bloated with lots of software. I created our aarch64 runners using ubuntu provided by Google Cloud and running this on top: https://github.com/jupyter/docker-stacks/blob/main/aarch64-runner/setup.sh
> > Could you please tell me how much free disk is needed to build one GPU image? I am not sure people need aarch64 GPU images, so let's not build them in this PR, though. Also, this will increase the build time (we have limited amount of self-hosted runners)
> 
> I am not sure how much space is exactly needed, but the [14 GB](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-private-repositories) provided for `ubuntu-latest` weren't enough and resulted in failed builds. I assume probably like 20 GB in total would be sufficient.

I cleaned up the space on our aarch64 runners.
Unfortunately, Docker has an issue, where cleaning up doesn't remove everything it should, so had to clean up manually.

mathbunnyru:
> > > > there an issue with also enabling this part for the self-hosted aarch64 runners?
> > > 
> > > 
> > > Self-hosted runners are not usually bloated with lots of software. I created our aarch64 runners using ubuntu provided by Google Cloud and running this on top: https://github.com/jupyter/docker-stacks/blob/main/aarch64-runner/setup.sh
> > > Could you please tell me how much free disk is needed to build one GPU image? I am not sure people need aarch64 GPU images, so let's not build them in this PR, though. Also, this will increase the build time (we have limited amount of self-hosted runners)
> > 
> > 
> > I am not sure how much space is exactly needed, but the [14 GB](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-private-repositories) provided for `ubuntu-latest` weren't enough and resulted in failed builds. I assume probably like 20 GB in total would be sufficient.
> 
> I cleaned up the space on our aarch64 runners. Unfortunately, Docker has an issue, where cleaning up doesn't remove everything it should, so had to clean up manually.

Seems to work - the build is green 🎉

mathbunnyru:
One more thought - I'm not sure about coda versioning, but it might be worth it to add more precise version of cuda as a tag prefix.
People might want to (or be forced to) stay on 12.1 even when 12.2 `pytorch-cuda` is released.

So, the prefix tag will be `cuda12-` (nothing to change here), but the final image can look something like `jupyter/pythorch-notebook:cuda12-cuda12.1`, so people will be to more precisely say they cuda version.

I don't consider this to be a blocker for this PR, but it would definitely be nice to add this tag later.

johanna-reiml-hpi:
I had another look at the official [pytorch Dockerfile](https://github.com/pytorch/pytorch/blob/main/Dockerfile), [CUDA Dockerfiles](https://gitlab.com/nvidia/container-images/cuda/-/tree/master/dist/12.1.1/ubuntu2204?ref_type=heads) and [nvidia documentation for some env variables](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/1.14.5/docker-specialized.html#dockerfiles), so here are a few extra points to consider:

- The following ENV should be added to the Dockerfile according to the linked NVIDIA documentation:
```
ENV NVIDIA_VISIBLE_DEVICES all
ENV NVIDIA_DRIVER_CAPABILITIES compute,utility
```
- I don't think the CUDA installation for aarch64 worked actually. The CUDA variant build times are identical to the CPU ones. I checked the [official install guide](https://pytorch.org/get-started/locally/) and when selecting MacOS you get: `# CUDA is not available on MacOS, please use default package`. The official pytorch Dockerfile also only installs the CPU version for aarch64. I will exclude aarch64 from this PR.
- The official pytorch Dockerfile installs the CUDA dependencies via conda. This adds CUDA binaries to conventional paths instead of having CUDA as a pip package (`ENV LD_LIBRARY_PATH /usr/local/nvidia/lib:/usr/local/nvidia/lib64`, `ENV PATH /usr/local/nvidia/bin:/usr/local/cuda/bin:$PATH`). I am not sure which approach is better. The current version might be nicer in terms of simplicity (and follows the official install guide).

johanna-reiml-hpi:
> One more thought - I don't know a lot about cuda versioning, but it might be worth it to add more precise version of cuda as a tag prefix (so, we need to add a new tagger to cuda flavored images). People might want to (or be forced to) stay on 12.1 even when 12.2 `pytorch-cuda` is released.
> 
> So, the prefix tag will stay `cuda12-` (nothing to change here), but the final image can look something like `jupyter/pytorch-notebook:cuda12-cuda12.1`, so people will be to more precisely say they cuda version.
> 
> What do you think? I don't consider this to be a blocker for this PR, but it would definitely be nice to add this tag later.

Seems okay to add. However, it's also possible that pytorch might add a ` --index-url 'https://download.pytorch.org/whl/cu122'` alongside the current ` --index-url 'https://download.pytorch.org/whl/cu121'`. In that case image variants would have to be built for both versions. I think it might be less confusing to do something like this instead:

-  `jupyter/pytorch-notebook:cuda12` (latest, so cuda12.2)
-  `jupyter/pytorch-notebook:cuda12.2`
-  `jupyter/pytorch-notebook:cuda12.1`

.github/workflows/docker.yml
johanna-reiml-hpi: Reordered it.

.github/workflows/docker.yml
johanna-reiml-hpi: Added the default.

.github/workflows/docker-build-test-upload.yml
johanna-reiml-hpi: Added the default.

tagging/apply_tags.py
johanna-reiml-hpi: Fixed.

mathbunnyru:
> > One more thought - I don't know a lot about cuda versioning, but it might be worth it to add more precise version of cuda as a tag prefix (so, we need to add a new tagger to cuda flavored images). People might want to (or be forced to) stay on 12.1 even when 12.2 `pytorch-cuda` is released.
> > So, the prefix tag will stay `cuda12-` (nothing to change here), but the final image can look something like `jupyter/pytorch-notebook:cuda12-cuda12.1`, so people will be to more precisely say they cuda version.
> > What do you think? I don't consider this to be a blocker for this PR, but it would definitely be nice to add this tag later.
> 
> Seems okay to add. However, it's also possible that pytorch might add a ` --index-url 'https://download.pytorch.org/whl/cu122'` alongside the current ` --index-url 'https://download.pytorch.org/whl/cu121'`. In that case image variants would have to be built for both versions. I think it might be less confusing to do something like this instead:
> 
> * `jupyter/pytorch-notebook:cuda12` (latest, so cuda12.2)
> * `jupyter/pytorch-notebook:cuda12.2`
> * `jupyter/pytorch-notebook:cuda12.1`

I think having a separate image for each `major.minor` version of cuda is too much.
We usually only support latest versions.
In case of cuda it makes sense to support both `cuda11` and `cuda12`, but having to support cuda12.Y is too much.
That's why I prefer a solution with a new tagger and a prefix tag as I mentioned above.

images/pytorch-notebook/cuda11/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+# Install PyTorch with pip (https://pytorch.org/get-started/locally/)
+# hadolint ignore=DL3013
+RUN pip install --no-cache-dir --extra-index-url=https://pypi.nvidia.com --index-url 'https://download.pytorch.org/whl/cu118' \
+    'torch' \
+    'torchvision' \
+    'torchaudio'  && \
+    fix-permissions "${CONDA_DIR}" && \
+    fix-permissions "/home/${NB_USER}"
+
+# # https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/1.14.5/docker-specialized.html#dockerfiles
```
mathbunnyru, line 23:
```suggestion
# https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/docker-specialized.html#dockerfiles
```

images/pytorch-notebook/cuda12/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+# Install PyTorch with pip (https://pytorch.org/get-started/locally/)
+# hadolint ignore=DL3013
+RUN pip install --no-cache-dir --extra-index-url=https://pypi.nvidia.com --index-url 'https://download.pytorch.org/whl/cu121' \
+    'torch' \
+    'torchvision' \
+    'torchaudio'  && \
+    fix-permissions "${CONDA_DIR}" && \
+    fix-permissions "/home/${NB_USER}"
+
+# https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/1.14.5/docker-specialized.html#dockerfiles
```
mathbunnyru, line 23:
```suggestion
# https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/docker-specialized.html#dockerfiles
```

mathbunnyru (approved):

mathbunnyru (approved):

mathbunnyru:
@consideRatio @manics please, vote here: https://github.com/jupyter/docker-stacks/pull/2091#issuecomment-1934654952

I am ready to merge this before the vote deadline if I receive one more positive vote.

mathbunnyru:
> Vote to drop CUDA ambitions for tensorflow-notebook

If someone manages to enable cuda-enabled tensorflow image for our docker stacks in a sane way (installing pip/mamba packages, no manual build), then it means that the ecosystem is mature enough for us to support it.
I won't be against merging such a change.
In any case, after merging this PR, it should be easy to experiment.

I mostly agree with @consideRatio's "Misc PR review" section - @johanna-reiml-hpi please, take a look.

> Supported architectures

👍 

> Variant description

I prefer not to change anything here (I really like all the `cuda11` image tags have a long common prefix like `jupyter/pytorch-notebook:cuda11-`)

> PR Title

👍

> Cuda policy

👍

consideRatio:
Ah, i thought it was a suffix seeing f"{platform}-{variant}" somewhere, and that it should be seen as a suffix being put last there. But I think i misunderstood this and that its really a high prio prefix in the tag, so it makes sense to reference it as a prefix still.

I think the description could be improved on still though as i struggled to piece together things. Now i think something like "Some images has variants, they will be published with the same image name, but get a prefix added to the tag." could have been helpful to read somewhere.

mathbunnyru:
I am merging this one - I will add a policy and some documentation.

mathbunnyru:
Done: https://github.com/jupyter/docker-stacks/pull/2097/files

mathbunnyru:
New images are pushed and ready to use 🎉 
https://quay.io/repository/jupyter/pytorch-notebook?tab=tags

mathbunnyru:
@johanna-reiml-hpi thank you for this PR.
It's nice to have GPU enabled image at least for pytorch and I think `variant` feature might be used in other circumstances.

Issue #2095, opened 2024-02-19
Add container images for the GPU version of TensorFlow and PyTorch Notebook


y-vectorfield:
### What docker image(s) is this feature applicable to?

pytorch-notebook, tensorflow-notebook

### What change(s) are you proposing?

We can not implement the GPU version of TensorFlow and PyTorch on the containers using these images.
Therefore, I'd like to propose the images for the GPU version of TensorFlow and PyTorch Notebook images.

### How does this affect the user?

We can implement the GPU version of TensorFlow and PyTorch on Jupyter Notebook using these images.

benz0li:
Cross references:

* https://github.com/jupyter/docker-stacks/issues/1557
* https://github.com/jupyter/docker-stacks/pull/2091

mathbunnyru:
New CUDA enabled `pytorch-notebook` images are pushed and ready to use 🎉 
https://quay.io/repository/jupyter/pytorch-notebook?tab=tags

y-vectorfield:
Excellent!! Thank you very much.
Now, there is no TF version or Jax version.
Can I contribute these versions?

mathbunnyru:
> Excellent!! Thank you very much.

Most of the work was done by @johanna-reiml-hpi in https://github.com/jupyter/docker-stacks/pull/2091 🙂 

> Can I contribute these versions?

Yes. Please, create 2 separate PRs.

We already have a regular tensorflow image, and your PR should look very similar to the PR mentioned above (except it won't have common parts making `variant` feature work).
This might not work though - `tensorflow` is difficult to install, especially with GPU involved.
It would be nice to have `cuda11` and `cuda12` versions (so it will be the same as pytorch-notebook).

For jax, we don't have an image, and I haven't used this package - so if there are separate CPU and GPU packages, you should add a regular image, and then GPU variants for cuda11 and cuda12 as well.

Note: we have a [policy on adding new images and packages](https://jupyter-docker-stacks.readthedocs.io/en/latest/maintaining/new-images-and-packages-policy.html), so I can't promise you your PRs will be merged. But it's definitely worth trying - at least we will know the current state of installing these packages on top of our images. If we won't merge these images, we can always merge new recipes.

mathbunnyru:
@y-vectorfield how is it going? Do you need any help?

ChristofKaufmann:
[...]

PR #2100, opened 2024-03-10
Add cuda12 variant of tensorflow-notebook

ChristofKaufmann:
## Describe your changes

- This adds a cuda12 variant of the tensorflow-notebook analog to eccda243c30a7d85a6313b99648a3e7bc39b215e with the pytorch-notebook.
- The CPU version uses the `tensorflow-cpu` wheel now (to reduce size of the image).
- Regarding cuda11 variant: The current version of TensorFlow is 2.16.1 and it seems the last compatible PyPI wheel with CUDA 11.8 is TensorFlow 2.14.1 (according to the officially [tested versions](https://www.tensorflow.org/install/source#gpu)). I still tried TensorFlow 2.16.1 with CUDA 11.8.0, but it didn't work. The current version of `tensorflow-gpu` on [conda-forge](https://anaconda.org/conda-forge/tensorflow-gpu/files) is 2.15.0 and has a CUDA 11.8 build. So if you want a cuda11 variant, I can try to use the conda-forge version for that, but the TensorFlow version is not up-to-date.

## Issue ticket if applicable

Fix: #2095, #1557.


docs/using/selecting.md
```diff
@@ -18,11 +18,11 @@ The following sections describe these images, including their contents, relation
 
 ## CUDA enabled variant
 
-We provide CUDA accelerated version of `pytorch-notebook` image.
-Prepend a CUDA version prefix (like `cuda12-`) to the image tag to allow PyTorch operations to use compatible NVIDIA GPUs for accelerated computation.
+We provide CUDA accelerated version of `pytorch-notebook` and `tensorflow-notebook` images.
+Prepend a CUDA version prefix (like `cuda12-`) to the image tag to allow PyTorch or TensorFlow operations to use compatible NVIDIA GPUs for accelerated computation.
 We only build images for 2 last major versions of CUDA.
```
mathbunnyru, line 23:
```suggestion
We only build images for 2 last major versions of CUDA.
Note: `tensorflow-notebook` image only supports cuda12.
```

images/tensorflow-notebook/cuda12/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# Fix: https://github.com/koalaman/shellcheck/wiki/SC3014
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+# Install CUDA libs and cuDNN with mamba
+RUN mamba install --yes -c nvidia/label/cuda-12.3.2 \
+    'cuda-libraries-dev' \
+    'cuda-nvcc' \
+    'cuda-nvtx' \
+    'cuda-cupti' \
+    'cudnn' && \
```
mathbunnyru, line 20:
```suggestion
    'cuda-cupti' \
    'cuda-libraries-dev' \
    'cuda-nvcc' \
    'cuda-nvtx' \
    'cudnn' && \
```

Let's keep the list sorted

images/tensorflow-notebook/cuda12/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+FROM $BASE_CONTAINER
+
+LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
+
+# Fix: https://github.com/hadolint/hadolint/wiki/DL4006
+# Fix: https://github.com/koalaman/shellcheck/wiki/SC3014
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+# Install CUDA libs and cuDNN with mamba
+RUN mamba install --yes -c nvidia/label/cuda-12.3.2 \
```
mathbunnyru, line 15: Is there a chance not to hardcode minor.patch version here?
Also, 12.4 was released.

images/tensorflow-notebook/cuda12/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    'cuda-libraries-dev' \
+    'cuda-nvcc' \
+    'cuda-nvtx' \
+    'cuda-cupti' \
+    'cudnn' && \
+    mamba clean --all -f -y && \
+    fix-permissions "${CONDA_DIR}" && \
+    fix-permissions "/home/${NB_USER}"
+
+ENV XLA_FLAGS="--xla_gpu_cuda_data_dir=${CONDA_DIR}" \
```
mathbunnyru, line 25: Could you please add a link where this flag is recommended?

images/tensorflow-notebook/cuda12/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    LD_LIBRARY_PATH="${CONDA_DIR}/lib/${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}" \
+    # https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/docker-specialized.html#dockerfiles
+    NVIDIA_VISIBLE_DEVICES="all" \
+    NVIDIA_DRIVER_CAPABILITIES="compute,utility"
+
+# Install Tensorflow with pip
+RUN pip install --no-cache-dir tensorflow && \
```
mathbunnyru, line 32: Maybe we can use tensorflow from mamba?
I think in such a case we won't even need to list the dependencies.

tests/docker-stacks-foundation/test_packages.py
```diff
@@ -71,6 +71,11 @@
     "bzip2",
     "ca-certificates",
     "conda-forge::blas[build=openblas]",
+    "cuda-libraries-dev",
+    "cuda-nvcc",
+    "cuda-nvtx",
+    "cuda-cupti",
+    "cudnn",
```
mathbunnyru, line 78:
```suggestion
    "cuda-cupti",
    "cuda-libraries-dev",
    "cuda-nvcc",
    "cuda-nvtx",
    "cudnn",
```

mathbunnyru (changes requested):

images/tensorflow-notebook/cuda12/Dockerfile
consideRatio: > * The current version of `tensorflow-gpu` on [conda-forge](https://anaconda.org/conda-forge/tensorflow-gpu/files) is 2.15.0 and has a CUDA 11.8 build. So if you want a cuda11 variant, I can try to use the conda-forge version for that, but the TensorFlow version is not up-to-date.

This is the kind of complexity I recall when getting a tensorflow gpu image working, where the conda-forge version is often outdated, and trying to install with `pip` for gpu support was complicated.

I think if we choose between install complexity or relying on something regularly outdated, the install complexity may be prefered - otherwise we introduce things we can't control. At the same time, the fact that its outdated etc relates to how complicated it may be to keep installing something that works over time, which we then may be taking on.

This PR will probably demonstrate the current maturity of tensorflow gpu stuff upstream, if its as bad as I experienced it was a while back, then I think its better to not try to support this.

images/tensorflow-notebook/cuda12/Dockerfile
ChristofKaufmann: I used 12.3, because it is listed in the [tested build configurations](https://www.tensorflow.org/install/source#gpu). There seems to be no label like `cuda-12`.  Without label, i. e. `-c nvidia` I got the latest CUDA version, which is 12.4, currently. It worked, but I thought it is risky. A new CUDA release might be incompatible (I guess not before 13.x) and it seems we do not have a unit test e. g. to check the output of `python -c "import tensorflow as tf; print(tf.config.list_physical_devices('GPU'))"`. Should we add a unit test like this?
Secondly, the [pytorch-notebook](https://github.com/jupyter/docker-stacks/blob/07dc2d08734b384ab7370956d47909bda9bc928b/images/pytorch-notebook/cuda12/Dockerfile#L16) fixes the minor version as well.

consideRatio:
I voted :eyes: for now, I'd like to see that this seems reasonable to install and maintain long term by tensorflow gpu packages upstream make it easy enough, because it has been a mess historically in my experience and I don't want this project to take on maintaining function if its too messy. If the implementation looks not-messy, I'd be :+1:, but I think this may be a very notable commitment if it is, and that we better then protect the project limited maintenance capacity from taking on such maintenance burden.

images/tensorflow-notebook/cuda12/Dockerfile
ChristofKaufmann: Yes, it was very complicated and the conda-forge `tensorflow-gpu` package helped by providing `cudatoolkit` and `cudnn` within the same toolchain. But I just tried to install `tensorflow-gpu` from conda-forge into scipy-notebook and it fails due to conflicts. So maybe it changed and nowadays the installation of `tensorflow` from PyPI and `cuda` and `cudnn` from nvidia's conda-channel is the easiest way. For maintenance I imagine using the cuda version from the [tested builds](https://www.tensorflow.org/install/source#gpu) for a new TensorFlow release should work.

images/tensorflow-notebook/cuda12/Dockerfile
ChristofKaufmann: I just tried and is seems obsolete now. I got it e. g. from [here](https://github.com/tensorflow/tensorflow/issues/23783#issuecomment-462771434) or [here](https://stackoverflow.com/a/64872998/2414411). It was related to a warning about not finding `ptxas` and maybe also `libdevice.10.bc`. Then XLA might not have worked (`jit_compile=True`). It seems it got fixed in a recent version. I will remove the line.

images/tensorflow-notebook/cuda12/Dockerfile
mathbunnyru: > I used 12.3, because it is listed in the tested build configurations.

There is a chance that the table might be slightly outdated.

> There seems to be no label like cuda-12. Without label, i. e. -c nvidia I got the latest CUDA version, which is 12.4, currently.

Can we use something like `nvidia/label/cuda-12.*`?

> Should we add a unit test like this?

I am not sure if this test will work with a regular GitHub-hosted ubuntu runner.
Also, we currently don't have a way to run a test for `variant` image (but it's not difficult to add something like this).

images/tensorflow-notebook/cuda12/Dockerfile
mathbunnyru: In my opinion, we can't tell if the current maturity of tensorflow packages is better until we merge this PR, have weekly builds, and have several releases of cuda/cudnn and the tensorflow package itself.

So, I am ok with how Dockerfile currently looks like, but we need to see if it's gonna be ok after a few releases.

mathbunnyru: I am ready to give it a try as a maintainer (and I can always disable the build just by changing a few lines in the `docker.yml` config-like file).

images/tensorflow-notebook/cuda12/Dockerfile
ChristofKaufmann: > There is a chance that the table might be slightly outdated.

I looked at the table just a few hours after the release and it was up to date. But it won't ever be tested against a newer version.

> Can we use something like nvidia/label/cuda-12.*?

No, there is no such label. [Here](https://anaconda.org/nvidia/cuda-nvcc/labels) is a list.

> I am not sure if this test will work with a regular GitHub-hosted ubuntu runner.

Right, that is always problematic. Sorry.

mathbunnyru: > > Can we use something like nvidia/label/cuda-12.*?
> 
> No, there is no such label. [Here](https://anaconda.org/nvidia/cuda-nvcc/labels) is a list.

I meant maybe mamba supports label regex (I have no idea if it does or not).
In that case we will be fine with existing tags and won’t need to hardcode some particular version.

ChristofKaufmann: Despite using labels, I noticed all package versions are also in the main label. The version in the [list of labels](https://anaconda.org/nvidia/cuda-nvcc/labels) is a bit strange. So the labels are formatted in major.minor.patch and the version in major.minor.build. Usually, if patch increments, the build number just continues:

- label: cuda-11.6.0, version: 11.6.55
- label: cuda-11.6.1, version: 11.6.112
- label: cuda-11.6.2, version: 11.6.124

There is one exception: labels cuda-11.4.3 and cuda-11.4.4 both have version 11.4.152.
Nevertheless, I just tried:

- `mamba install -c nvidia 'cuda-nvcc<13'` and got version 12.4.99 (which is also in label cuda-12.4.0)
- `mamba install -c nvidia 'cuda-nvcc<12'` and got version 11.8.89 (which is also in label cuda-11.8.0)
- `mamba install -c nvidia 'cuda-nvcc<11.5'` and got version 11.4.152 (which is also in labels cuda-11.4.3 and cuda-11.4.4)
- `mamba install -c nvidia 'cuda-nvcc=12.3'` and got version 12.3.107 (which is also in label cuda-12.3.2)

So, we could use `'cuda-nvcc<13'`, to reduce maintenance work (avoid updating the version for every TensorFlow release), but these are not officially tested by TensorFlow (not sure, if there can occur incompatibilities with new minor versions). Using something like `'cuda-nvcc=12.3'` is more work (still avoiding the patch version), but officially [tested](https://www.tensorflow.org/install/source#gpu).

mathbunnyru: I'm ok with `cuda-nvcc=12.3`, please add the comment why we choose this version though.

mathbunnyru:
I cleaned aarch64 machines, builds should work better - unfortunately, docker is the worst in cleaning its cache.

images/tensorflow-notebook/cuda12/Dockerfile
ChristofKaufmann: This did not work, since the dependencies have no limitation in their versions. So using `cuda-nvcc=12.3` with the nvidia channel resulted in a mixture of 12.3 and 12.4. NVIDIA is quite sloppy in their packaging.

Then I noticed, that [cudnn from the nvidia channel](https://anaconda.org/nvidia/cudnn) is outdated. Apparently they dropped the support 3 years ago. The [cudnn from conda-forge](https://anaconda.org/conda-forge/cudnn) is quite up-to-date, but there is no CUDA 12 build yet, only CUDA 11.8.

So, I would like to go with the new installation method [supported by TensorFlow](https://www.tensorflow.org/install/pip), which is basically just `pip install tensorflow[and-cuda]`. This also has the advantage, that the installed CUDA version is always the [officially tested](https://www.tensorflow.org/install/source#gpu) version – so less maintenance for you. Usually the path to the nvidia libs should be found automatically, but in 2.16.1 there seems to be a [bug](https://github.com/tensorflow/tensorflow/issues/63362), so we have to add them. I prepared an activation script for that. Also `LD_LIBRARY_PATH` is not polluted with this method, because the paths from the pip installation contain only the nvidia libs. Before we added `${CONDA_DIR}/lib/`, which contains quite a lot libraries.

manics:
I don't feel qualified to give a :+1: or :-1:, @consideRatio has already highlighted the main issues around long-term maintainability so I think the decision should be from those who have ultimate responsibility for maintaining it.

images/tensorflow-notebook/cuda12/Dockerfile
mathbunnyru: Ok, let’s try it

ChristofKaufmann:
I spend some time to find the best way regarding maintainability. Now it looks quite similar to the PyTorch cuda variant, except that for TensorFlow:
- You can't choose between CUDA 11 and 12 – it will use the officially [tested build configuration](https://www.tensorflow.org/install/source#gpu).
- There is a workaround for a [bug](https://github.com/tensorflow/tensorflow/issues/63362).

mathbunnyru:
> You can't choose between CUDA 11 and 12 – it will use the officially [tested build configuration](https://www.tensorflow.org/install/source#gpu).

Why can't you pin the version when using `pip`?

ChristofKaufmann:
The extra "and-cuda" is defined [here](https://github.com/tensorflow/tensorflow/blob/6fe64ce213166c3b4430568943e5d125e0bd372d/tensorflow/tools/pip_package/setup.py#L167). There the package versions of the dependencies are fixed to the ones listed in [tested build configuration](https://www.tensorflow.org/install/source#gpu). There is no additional "and-cuda11" extra. That's what I meant.

mathbunnyru:
> The extra "and-cuda" is defined [here](https://github.com/tensorflow/tensorflow/blob/6fe64ce213166c3b4430568943e5d125e0bd372d/tensorflow/tools/pip_package/setup.py#L167). There the package versions of the dependencies are fixed to the ones listed in [tested build configuration](https://www.tensorflow.org/install/source#gpu). There is no additional "and-cuda11" extra. That's what I meant.

I guess there are many libraries installed alongside with tensorflow.
They probably either have cu12 in their name or in their version.
We can limit the version of one of such libraries and that's why pip will have to choose a proper cuda version.

At least I think it is worth trying.

ChristofKaufmann:
I tried to use the `-cu11` packages (except for `nvidia-nvjitlink-cu12`, since it is [new in CUDA 12](https://developer.nvidia.com/blog/cuda-12-0-compiler-support-for-runtime-lto-using-nvjitlink-library/)) using:
`CUDA11_DEPS=$(wget -qO- https://pypi.org/pypi/tensorflow/json | grep -o -e '[a-z_-]\+-cu12' | sed 's/-cu12/-cu11/; s/nvidia-nvjitlink-cu11//' | xargs)`

However, it does not work. The `tensorflow` package is linked against CUDA 12 libraries. It expects e. g. `libcudart.so.12`, while there is a `libcudart.so.11`.

<details><summary>Full import errors with TF_CPP_MAX_VLOG_LEVEL=3</summary>

```output
2024-03-15 02:38:23.246671: I external/local_tsl/tsl/platform/default/dso_loader.cc:70] Could not load dynamic library 'libcudart.so.12'; dlerror: libcudart.so.12: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/conda/lib/python3.11/site-packages/nvidia/cublas/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_cupti/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvcc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvrtc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_runtime/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cudnn/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cufft/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/curand/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusolver/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusparse/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/nccl/lib/
2024-03-15 02:38:23.246811: I external/local_tsl/tsl/platform/default/dso_loader.cc:70] Could not load dynamic library 'libcublas.so.12'; dlerror: libcublas.so.12: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/conda/lib/python3.11/site-packages/nvidia/cublas/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_cupti/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvcc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvrtc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_runtime/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cudnn/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cufft/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/curand/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusolver/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusparse/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/nccl/lib/
2024-03-15 02:38:23.246910: I external/local_tsl/tsl/platform/default/dso_loader.cc:70] Could not load dynamic library 'libcublasLt.so.12'; dlerror: libcublasLt.so.12: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/conda/lib/python3.11/site-packages/nvidia/cublas/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_cupti/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvcc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvrtc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_runtime/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cudnn/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cufft/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/curand/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusolver/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusparse/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/nccl/lib/
2024-03-15 02:38:23.247014: I external/local_tsl/tsl/platform/default/dso_loader.cc:70] Could not load dynamic library 'libcufft.so.11'; dlerror: libcufft.so.11: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/conda/lib/python3.11/site-packages/nvidia/cublas/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_cupti/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvcc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvrtc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_runtime/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cudnn/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cufft/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/curand/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusolver/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusparse/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/nccl/lib/
2024-03-15 02:38:23.261102: I external/local_tsl/tsl/platform/default/dso_loader.cc:59] Successfully opened dynamic library libcusolver.so.11
2024-03-15 02:38:23.261261: I external/local_tsl/tsl/platform/default/dso_loader.cc:70] Could not load dynamic library 'libcusparse.so.12'; dlerror: libcusparse.so.12: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/conda/lib/python3.11/site-packages/nvidia/cublas/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_cupti/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvcc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvrtc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_runtime/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cudnn/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cufft/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/curand/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusolver/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusparse/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/nccl/lib/
2024-03-15 02:38:23.261470: I external/local_tsl/tsl/platform/default/dso_loader.cc:59] Successfully opened dynamic library libcudnn.so.8
2024-03-15 02:38:23.261483: W tensorflow/core/common_runtime/gpu/gpu_device.cc:2251] Cannot dlopen some GPU libraries. Please make sure the missing libraries mentioned above are installed properly if you would like to use GPU. Follow the guide at https://www.tensorflow.org/install/gpu for how to download and setup the required libraries for your platform.
Skipping registering GPU devices...
[]
```
</details>

If you look at the conda-forge `tensorflow-gpu` package [files](https://anaconda.org/conda-forge/tensorflow-gpu/files), there are e. g. `cuda120` (12.0) and `cuda118` (11.8) builds. So the library version is fixed at build time and the PyPI `tensorflow` package provides only a CUDA 12 build.

mathbunnyru:
> I tried to use the `-cu11` packages (except for `nvidia-nvjitlink-cu12`, since it is [new in CUDA 12](https://developer.nvidia.com/blog/cuda-12-0-compiler-support-for-runtime-lto-using-nvjitlink-library/)) using:
> 
> `CUDA11_DEPS=$(wget -qO- https://pypi.org/pypi/tensorflow/json | grep -o -e '[a-z_-]\+-cu12' | sed 's/-cu12/-cu11/; s/nvidia-nvjitlink-cu11//' | xargs)`
> 
> 
> 
> However, it does not work. The `tensorflow` package is linked against CUDA 12 libraries. It expects e. g. `libcudart.so.12`, while there is a `libcudart.so.11`.
> 
> 
> 
> <details><summary>Full import errors with TF_CPP_MAX_VLOG_LEVEL=3</summary>
> 
> 
> 
> ```output
> 
> 2024-03-15 02:38:23.246671: I external/local_tsl/tsl/platform/default/dso_loader.cc:70] Could not load dynamic library 'libcudart.so.12'; dlerror: libcudart.so.12: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/conda/lib/python3.11/site-packages/nvidia/cublas/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_cupti/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvcc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvrtc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_runtime/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cudnn/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cufft/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/curand/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusolver/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusparse/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/nccl/lib/
> 
> 2024-03-15 02:38:23.246811: I external/local_tsl/tsl/platform/default/dso_loader.cc:70] Could not load dynamic library 'libcublas.so.12'; dlerror: libcublas.so.12: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/conda/lib/python3.11/site-packages/nvidia/cublas/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_cupti/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvcc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvrtc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_runtime/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cudnn/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cufft/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/curand/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusolver/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusparse/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/nccl/lib/
> 
> 2024-03-15 02:38:23.246910: I external/local_tsl/tsl/platform/default/dso_loader.cc:70] Could not load dynamic library 'libcublasLt.so.12'; dlerror: libcublasLt.so.12: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/conda/lib/python3.11/site-packages/nvidia/cublas/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_cupti/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvcc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvrtc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_runtime/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cudnn/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cufft/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/curand/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusolver/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusparse/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/nccl/lib/
> 
> 2024-03-15 02:38:23.247014: I external/local_tsl/tsl/platform/default/dso_loader.cc:70] Could not load dynamic library 'libcufft.so.11'; dlerror: libcufft.so.11: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/conda/lib/python3.11/site-packages/nvidia/cublas/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_cupti/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvcc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvrtc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_runtime/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cudnn/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cufft/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/curand/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusolver/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusparse/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/nccl/lib/
> 
> 2024-03-15 02:38:23.261102: I external/local_tsl/tsl/platform/default/dso_loader.cc:59] Successfully opened dynamic library libcusolver.so.11
> 
> 2024-03-15 02:38:23.261261: I external/local_tsl/tsl/platform/default/dso_loader.cc:70] Could not load dynamic library 'libcusparse.so.12'; dlerror: libcusparse.so.12: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/conda/lib/python3.11/site-packages/nvidia/cublas/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_cupti/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvcc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_nvrtc/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cuda_runtime/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cudnn/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cufft/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/curand/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusolver/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/cusparse/lib/:/opt/conda/lib/python3.11/site-packages/nvidia/nccl/lib/
> 
> 2024-03-15 02:38:23.261470: I external/local_tsl/tsl/platform/default/dso_loader.cc:59] Successfully opened dynamic library libcudnn.so.8
> 
> 2024-03-15 02:38:23.261483: W tensorflow/core/common_runtime/gpu/gpu_device.cc:2251] Cannot dlopen some GPU libraries. Please make sure the missing libraries mentioned above are installed properly if you would like to use GPU. Follow the guide at https://www.tensorflow.org/install/gpu for how to download and setup the required libraries for your platform.
> 
> Skipping registering GPU devices...
> 
> []
> 
> ```
> 
> </details>
> 
> 
> 
> If you look at the conda-forge `tensorflow-gpu` package [files](https://anaconda.org/conda-forge/tensorflow-gpu/files), there are e. g. `cuda120` (12.0) and `cuda118` (11.8) builds. So the library version is fixed at build time and the PyPI `tensorflow` package provides only a CUDA 12 build.

Thanks. I guess in that case let's rename the variant to simple `cuda` because we don't have any control over the cuda version. And please update the docs to mention it.

mathbunnyru (approved):

mathbunnyru:
https://github.com/jupyter/docker-stacks/pull/2100#issuecomment-1988030167

@yuvipanda what do you think about this PR?

twalcari:
While I have no official vote here, I would like to express my full support for this PR. Given that TensorFlow is mainly used for GPU-accelerated applications, it makes a lot of sense to have a GPU-capable docker image available. 

The current non-GPU enabled images feel like a neutered alternative that are OK for doing some preliminary exploration on how TensorFlow works, but are ineffective to be used in any real-world applications.

yuvipanda:
Sorry for the delay, @mathbunnyru.

I'm +1 on this change because it's using the *upstream* supported way to install tensorflow - the `and-cuda` variant (as described in https://www.tensorflow.org/install/pip#linux). 

The only (**non-blocking**) concern I have is that it's based on the scipy-notebook image, which installs packages primarily [from conda-forge](https://github.com/jupyter/docker-stacks/blob/dd06b93abc41a5de327f8fbe3e597e4ed4c1bc0d/images/scipy-notebook/Dockerfile#L30). And `tensorflow` has some dependencies (particularly numpy) that are already in the base image. So the question is, what happens if a newer (or older) version of numpy is required by tensorflow than what we get from conda-forge? Would mixing `pip` and `conda` like this cause issues? In my experience, it mostly does not (I literally did this [with tensorflow](https://github.com/2i2c-org/utoronto-image/blob/8c7ba58b8bb0070b70416a7092c31a79f6c5d330/environment.yml#L63) in another project a few months ago). And I'd rather us do this if it means we can directly use the method maintained by upstream. It's also what we do for pytorch now.  

So overall, +1 from me. Thank you for this contribution, @ChristofKaufmann! And thanks for your stewardship, @mathbunnyru

mathbunnyru:
> So the question is, what happens if a newer (or older) version of numpy is required by tensorflow than what we get from conda-forge?

We can always pin versions in some images if we need to and there is no other choice.

I think `numpy` is so widely used by everyone, so the `conda-forge` team puts lots of effort into releasing new versions and we won't even have to wait long.
But we'll only see this in practice, when we merge, and gain some experience.

So, let's try to merge this one 🙂

yuvipanda:
>  But we'll only see this in practice, when we merge, and gain some experience.

Big big +1! Thank you :)

ChristofKaufmann:
Thank you for your helpful comments to improve the code @mathbunnyru

PR #2101, opened 2024-03-16
Added new image for tinygrad notebooks

kotleta2007:
## Checklist

- [x] I have tested the code (ran the image, tinygrad works in a new notebook)

### DIFF
--- a/images/tinygrad-notebook/.dockerignore
+++ b/images/tinygrad-notebook/.dockerignore
@@ -1,0 +1,2 @@
+# Documentation
+README.md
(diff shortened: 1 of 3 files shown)


consideRatio:
Hey @kotleta2007!

I don't think its sustainable this this project to create a dedicated image to add a single small package like this. With each image added comes a long term maintenance effort.

If you need something like this, I think an idea could be to publish such image yourself and automate it to be rebuilt weekly. This could be done with a simple Dockerfile + .github/workflows that does build and publish (see for example https://github.com/docker/build-push-action).

mathbunnyru:
Or you can use cookiecutter template from here: https://github.com/jupyter/cookiecutter-docker-stacks

PR #2108, opened 2024-04-15
Add PySpark installation to Dockerfile

amaranaitsaidi:
## Describe your changes

This pull request adds the installation of PySpark to the Dockerfile for the Jupyter Stack with Spark.
This will allow users to use PySpark in their Jupyter notebooks without having to install it separately.

### DIFF
--- a/images/pyspark-notebook/Dockerfile
+++ b/images/pyspark-notebook/Dockerfile
@@ -64,10 +64,11 @@ RUN mamba install --yes \
     'grpcio-status' \
     'grpcio' \
     'pandas=2.0.3' \
+    'pyspark' \
     'pyarrow' && \
     mamba clean --all -f -y && \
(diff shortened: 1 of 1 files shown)


images/pyspark-notebook/Dockerfile
```diff
@@ -68,6 +68,6 @@ RUN mamba install --yes \
     mamba clean --all -f -y && \
     fix-permissions "${CONDA_DIR}" && \
     fix-permissions "/home/${NB_USER}"
-
+RUN pip install pyspark
```
mathbunnyru, line 71: Can we use `mamba` here?

Also, this is not how we install things.
Please, copy-paste the installation method we're using from any other our image, because we don't want to pollute our images with caches and so on.

mathbunnyru (changes requested):

PR #2119, opened 2024-05-30
Updated header level to fit rest of the page

392781:
## Describe your changes

This is a minor change regarding the header levels. I believe the intent was to have levels:

Selecting an Image
- Core Stacks
  - CUDA Enabled Variant
  - jupyter/docker-stacks-foundation
  - jupyter/base-notebook
  - jupyter/minimal-notebook
  - ...
- ...

The current iteration seems to indicate that `jupyter/...` derives from the "CUDA Enabled Variant" header when really it is part of the core stack.

In general, I would propose a reorganization of this particular page for additional clarity (perhaps in a future pull request). Mainly I would suggest just moving certain things around and changing some of the header levels.

## Issue ticket if applicable

N/A


mathbunnyru:
Thanks, merged.

PR #2126, opened 2024-07-19
Improve handling of tensorboard

twalcari:
### Problem statement

Tensorboard is a utility that is frequently used by developers working with TensorFlow or PyTorch to visually inspect their work. It is included by default when installing TensorFlow, and is even part of the very first [Get Started](https://www.tensorflow.org/tensorboard/get_started) notebook of TensorFlow. It integrates into Jupyter notebooks via an extension (`%load_ext tensorboard`), after which you can request the TensorBoard UI via the magic command `%tensorboard` .  

TensorBoard works by launching an additional process which listens on port 6006 (typically). However, this means that you also need to expose that extra port when starting an `tensorflow-notebook` instance:

`docker run -p 10000:8888 -p 6006:6006 quay.io/jupyter/jupyter-notebook:latest` .

If you fail to do this, you are greeted with a "Connection refused" error:

tensorboard-problem

The situation is even more problematic when the image is started via JupyterHub: in many cases the user cannot open an extra port, and certainly not on port 6006.

### Proposed solution

After looking into how CoLab solved this problem, I [found](https://github.com/tensorflow/tensorboard/blob/31ef2f8a0e1e7ecc31983fe9a4bebb048b1298a8/tensorboard/notebook.py#L381) an environment variable `TENSORBOARD_PROXY_URL` that can be used to specify a custom target location of the `iframe` being injected into the notebook.

This PR does 2 things:

- it adds `jupyter-server-proxy` to the image
- it properly configures the `TENSORBOARD_PROXY_URL` environment variable to use jupyter-server-proxy. It's done in a way that works both in the standalone case and in the JupyterHub case.

tensorboard-fixed


images/tensorflow-notebook/20tensorboard-proxy-env.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+set -e
+
+# Initialize the TENSORBOARD_PROXY_URL with the appropriate path
+# to use jupyter-server-proxy .
```
mathbunnyru, line 8:
```suggestion
# to use jupyter-server-proxy.
```

images/tensorflow-notebook/Dockerfile
```diff
@@ -13,6 +13,9 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 # Install tensorflow with pip, on x86_64 tensorflow-cpu
 RUN [[ $(uname -m) = x86_64 ]] && TF_POSTFIX="-cpu" || TF_POSTFIX="" && \
-    pip install --no-cache-dir "tensorflow${TF_POSTFIX}" && \
+    pip install --no-cache-dir "tensorflow${TF_POSTFIX}" \
+    "jupyter-server-proxy" && \
```
mathbunnyru, line 17:
```suggestion
    pip install --no-cache-dir \
    "tensorflow${TF_POSTFIX}" \
    "jupyter-server-proxy" && \
```

images/tensorflow-notebook/20tensorboard-proxy-env.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
mathbunnyru, line 1: I suggest we put this file in `cuda` subdirectory - it is only used for this variant of the image

mathbunnyru (changes requested):

images/tensorflow-notebook/20tensorboard-proxy-env.sh
twalcari: As both variants (default and cuda) are published in the container repository I would make sure that this problem is fixed in both of them?

mathbunnyru: You're right, for some (wrong) reason I thought TensorBoard is only useful in cuda image

twalcari: The copy from the parent folder did not work, according to the [docs](https://docs.docker.com/reference/dockerfile/#copy:~:text=be%20used%20instead.-,COPY,obeys%20the%20following%20rules%3A,-The) of Docker:

> The `<src>` path is resolved relative to the build context. If you specify a relative path leading outside of the build context, such as COPY ../something /something, parent directory paths are stripped out automatically. The effective source path in this example becomes COPY something /something

So I did move it to the cuda folder to work around this, but now the non-cuda Dockerfile references to the file inside the cuda-folder. Do we leave it like that, or do we copy the file instead to keep it "clean" ?

mathbunnyru: Unfortunately, I don't think there is a good solution to this problem in our case
It's either "put the file in cuda subdir", or copy-paste the file in 2 places.

Let's have it in cuda subdir, I don't think that's gonna be a problem

mathbunnyru:
@twalcari sorry for not mentioning it earlier, please update the following doc: https://github.com/jupyter/docker-stacks/blob/main/docs/using/selecting.md#jupytertensorflow-notebook

mathbunnyru:

The voting deadline is the 1st of August (a week since I posted this message).

twalcari:
> @twalcari sorry for not mentioning it earlier, please update the following doc: https://github.com/jupyter/docker-stacks/blob/main/docs/using/selecting.md#jupytertensorflow-notebook

What do you want me to add there? Tensorboard is an utility that is distributed together with tensorflow, so I don't think we should mention it separately? Do I need to mention jupyter-server-proxy there?

mathbunnyru:
> > @twalcari sorry for not mentioning it earlier, please update the following doc: https://github.com/jupyter/docker-stacks/blob/main/docs/using/selecting.md#jupytertensorflow-notebook
> 
> What do you want me to add there? Tensorboard is an utility that is distributed together with tensorflow, so I don't think we should mention it separately? Do I need to mention jupyter-server-proxy there?

I would only mention `jupyter-server-proxy`.

PR #2129, opened 2024-07-30
Add netbase to docker-stacks-foundation image - fixes #2128

### LINKED ISSUE #2128: Include netbase Ubuntu package in images

base-notebook, docker-stacks-foundation, minimal-notebook

### What change(s) are you proposing?

Add the [netbase Ubuntu package](https://ubuntu.pkgs.org/22.04/ubuntu-main-amd64/netbase_6.3_all.deb.html) to images.

> This package provides the necessary infrastructure for basic TCP/IP based networking.
> In particular, it supplies common name-to-number mappings in /etc/services, /etc/rpc, /etc/protocols and /etc/ethertypes

[/etc/protocols](https://man7.org/linux/man-pages/man5/protocols.5.html) is part of the POSIX spec and required by the C functions getprotobyname and getprotobynumber. Some reading here: https://unix.stackexchange.com/questions/680494/what-is-the-significance-of-etc-protocols-in-linux

### Anything else?

If this proposal is accepted, it's as easy as adding `netbase` to an apt-get install somewhere in the chain of images.

AlexHill:
## Describe your changes

Add the [netbase Ubuntu package](https://ubuntu.pkgs.org/22.04/ubuntu-main-amd64/netbase_6.3_all.deb.html) to the docker-stacks-foundation image.

This could also go in base, or minimal, but as a part of POSIX it seems like this belongs in the foundation image. It also seems analogous to the locales package installed in the same spot.

## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/2128

### DIFF
--- a/images/docker-stacks-foundation/Dockerfile
+++ b/images/docker-stacks-foundation/Dockerfile
@@ -30,6 +30,7 @@ RUN apt-get update --yes && \
     bzip2 \
     ca-certificates \
     locales \
+    netbase \
     sudo \
     # - `tini` is installed as a helpful container entrypoint,
     #   that reaps zombie processes and such of the actual executable we want to start
--- a/images/docker-stacks-foundation/Dockerfile
+++ b/images/docker-stacks-foundation/Dockerfile
@@ -30,6 +30,9 @@ RUN apt-get update --yes && \
     bzip2 \
     ca-certificates \
     locales \
+    # - `netbase` provides /etc/{protocols,rpc,services}, part of POSIX
+    #   and required by various C functions like getservbyname and getprotobyname
+    #   https://github.com/jupyter/docker-stacks/pull/2129
     netbase \
     sudo \
     # - `tini` is installed as a helpful container entrypoint,


images/docker-stacks-foundation/Dockerfile
```diff
@@ -30,6 +30,7 @@ RUN apt-get update --yes && \
     bzip2 \
     ca-certificates \
     locales \
+    netbase \
```
mathbunnyru, line 33: It would be nice to add a comment why we're adding this package

mathbunnyru:

The voting deadline is the 6th of August (a week since I posted this message).

consideRatio:
Its very small (12kB download, 40kB installed) and its maintained by ubuntu people so I also expect this to be sufficiently secure and robust.

I'm looking to understand the value of this package on a bit higher level. I understand `/etc/protocols` shows up thanks to this for example, and that is required by some C functions - @AlexHill did you run into a limitation of not having `/etc/protocols` using Python/R/Julia or some CLI etc, or did you use these C functions directly?

images/docker-stacks-foundation/Dockerfile
benz0li: > Add netbase so that `/etc/{protocols,rpc,services}` exists

– https://github.com/docker-library/python/commit/b8c94a31a98a535477200482a32c95192f85af5b

benz0li: No reason is given *why* it was added to the Docker "Official Image" for `python`.

Maybe @yosifkit can give some insight.

yuvipanda:
`getprotocolbynumber` was rejected for inclusion in CPython as being 'rarely used' (https://github.com/python/cpython/issues/68997), but `getprotobyname` (https://github.com/python/cpython/blob/1cac0908fb6866c30b7fe106bc8d6cd03c7977f9/Modules/socketmodule.c#L6142) is part of cpython. That calls the glibc function, which as I understand requires these files to work (as they're part of POSIX).

This means if there is at least one package that calls `getprotobyname`, I'm happy for this to be included. I quickly checked the two obvious culprits (`requests` and `aiohttp`), and found no uses of `getprotobyname`. Do you have a link to any library that actually uses that call?

AlexHill:
Yes, I ran into this trying to use the [spead2](https://pypi.org/project/spead2/) Python package to analyse astronomical data, (though I’m not sure whether the calls in question are in Python code or in the C++ extension).

Took a bit of tracking down - my test case in base ubuntu:22.04 with just git and python added works, because netbase is a recommended package of git. But we install git with --no-install-recommends in the minimal image, so it doesn’t end up installed.

yuvipanda:
Good enough explanation for me :) given it is part of posix / glibc, i voted +1

images/docker-stacks-foundation/Dockerfile
yosifkit: It was to fix https://github.com/docker-library/python/issues/270, or to ensure that `socket.getservbyname()` works.

mathbunnyru: Thanks! @AlexHill could you please add a comment to the code so the next person who sees it will know why it was added. It would be nice to add a link to this PR there as well.

mathbunnyru:
Nice, thanks, everyone! Waiting for the comment to be added in the dockerfile and then I'll merge this

AlexHill:
Thanks - added a comment.

PR #2141, opened 2024-09-01
Recipe using bake to build with custom arguments

### LINKED ISSUE #2140: Example bake file for custom docker builds

docker-stacks-foundation

### What change(s) are you proposing?

I'm using these docker stacks built locally with python3.12, out of the box that's quite cumbersome to do. There are args for each dockerfile but these need to be built manually with custom arguments in the correct order.

To simplify this process, I've configured this repo as a submodule, I'm then using a [bake](https://docs.docker.com/build/bake/) file to automate the builds. So in my local repo users can run 2 commands and have a comparable config setup:
```
docker buildx bake
docker compose up
```
This use case is explicitly cited by the [bake docs](https://docs.docker.com/build/bake/reference/#use-another-target-as-base)

> You can Use this feature when you have multiple Dockerfiles that can't be easily merged into one.

Makes using customized builds easier.

### Anything else?

For my example use, see [the bake file](https://github.com/alwaysmpe/what2_dsp/blob/main/docker-bake.hcl) in [my repo](https://github.com/alwaysmpe/what2_dsp/)  (the rest of the code is pretty hacked together stuff, I'm using it to teach basics of DSP).

Happy to create a PR  but not sure if this would be better as a recipe/example or whether it's something that you'd want as part of your standard build process.

alwaysmpe:
## Describe your changes

Add example Bake file for building locally with custom arguments.

The recipe bake file currently just references head of main, I've not mentioned how/why they should change this. Didn't think it was necessary.

## Issue ticket if applicable

Fix #2140 

### DIFF
--- a/docs/using/recipe_code/docker-bake.python312.hcl
+++ b/docs/using/recipe_code/docker-bake.python312.hcl
@@ -33,7 +33,7 @@ target "minimal-notebook" {
     tags = ["minimal-notebook"]
 }
 
-target "custom-jupyter" {
+target "custom-notebook" {
     context = "."
     contexts = {
         minimal-notebook = "target:minimal-notebook"
--- a/docs/using/recipe_code/docker-bake.python312.hcl
+++ b/docs/using/recipe_code/docker-bake.python312.hcl
@@ -1,0 +1,45 @@
+group "default" {
+    targets = ["custom-notebook"]
+}
+
+target "foundation" {
+    context = "https://github.com/jupyter/docker-stacks.git#main:images/docker-stacks-foundation"
+    args = {
+        PYTHON_VERSION = "3.12"
+    }
+    tags = ["docker-stacks-foundation"]
+}
+
+
+target "base-notebook" {
+    context = "https://github.com/jupyter/docker-stacks.git#main:images/base-notebook"
+    contexts = {
+        docker-stacks-foundation = "target:foundation"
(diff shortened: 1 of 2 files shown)


docs/using/recipes.md
```diff
@@ -525,3 +525,57 @@ they may be explained in the "Installation instructions" section of the Download
...
+```{note}
+Custom arguments may result in build errors due to incompatibility.
+If so your use-case may require a fully customized stack.
+```
+
+As a basic example, if you want to build a custom image based on the `minimal-notebook` image using `Python 3.12`,
+then with a Dockerfile like:
+
+```{code-block} Dockerfile
+:caption: Dockerfile
```
mathbunnyru, line 551: this looks cool!

docs/using/recipe_code/docker-bake.python312.hcl
```diff
@@ -0,0 +1,45 @@
...
+    contexts = {
+        base-notebook = "target:base-notebook"
+    }
+    args = {
+        BASE_CONTAINER = "base-notebook"
+    }
+    tags = ["minimal-notebook"]
+}
+
+target "custom-jupyter" {
```
mathbunnyru, line 36:
```suggestion
target "custom-notebook" {
```

I suppose this should be `custom-notebook`, as this should be the default target

PR #2145, opened 2024-09-14
docs: Core Stacks: move CUDA docs to later in the page

manics:
https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#core-stacks starts with `CUDA enabled variant`, implying it's the first thing people should know about. It's not, since it only applies to a subset of the images.


mathbunnyru:
Thanks. I agree, it makes more sense.

Issue #2146, opened 2024-09-25
Docker stacks foundation image build fails with current mamba


vizeit:
### What docker image(s) are you using?

base-notebook, docker-stacks-foundation, minimal-notebook, pytorch-notebook, scipy-notebook

### Host OS system

Ubuntu 22.04

### Host architecture

x86_64

### What Docker command are you running?

Docker build

### How to Reproduce the problem?

Build docker stacks foundation image, the build throws an error

### Actual behavior

build results into the following error,
```
#15 ERROR: process "/bin/bash -o pipefail -c set -x &&     arch=$(uname -m) &&     if [ \"${arch}\" = \"x86_64\" ]; then         arch=\"64\";     fi &&     wget --progress=dot:giga -O -         \"[https://micro.mamba.pm/api/micromamba/linux-${arch}/latest\](https://micro.mamba.pm/api/micromamba/linux-$%7Barch%7D/latest/)" | tar -xvj bin/micromamba &&     PYTHON_SPECIFIER=\"python=${PYTHON_VERSION}\" &&     if [[ \"${PYTHON_VERSION}\" == \"default\" ]]; then PYTHON_SPECIFIER=\"python\"; fi &&     ./bin/micromamba install         --root-prefix=\"${CONDA_DIR}\"         --prefix=\"${CONDA_DIR}\"         --yes         \"${PYTHON_SPECIFIER}\"         'mamba'         'jupyter_core' &&     rm -rf /tmp/bin/ &&     mamba list --full-name 'python' | tail -1 | tr -s ' ' | cut -d ' ' -f 1,2 | sed 's/\\.[^.]*$/.*/' >> \"${CONDA_DIR}/conda-meta/pinned\" &&     mamba clean --all -f -y &&     fix-permissions \"${CONDA_DIR}\" &&     fix-permissions \"/home/${NB_USER}\"" did not complete successfully: exit code: 109
------
[...]

PR #2147, opened 2024-10-01
Install mamba 2.0

mathbunnyru:
## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/2146


mathbunnyru:
`The following argument was not expected: --no-builds`

We use it here: https://github.com/jupyter/docker-stacks/blob/55ea74f89f7a8e65f27cab87fd8340b25a5a14e5/tests/package_helper.py#L64

@JohanMabille is this expected?

mathbunnyru:
Also: `bash: line 1: conda: command not found`

It seems that conda command is missing

mathbunnyru:
> `The following argument was not expected: --no-builds`
> 
> We use it here:
> 
> https://github.com/jupyter/docker-stacks/blob/55ea74f89f7a8e65f27cab87fd8340b25a5a14e5/tests/package_helper.py#L64
> 
> @JohanMabille is this expected?

Created an upstream issue: https://github.com/mamba-org/mamba/issues/3492

mathbunnyru:
Fixed this issue with the output of `mamba list --full-name` (the output format has slightly changed, not a bug).
Also fixed an issue with `mamba env export` (reported to the upstream, but we no longer need it to be fixed).
And I did it in a way that works with both mamba v1 and v2.

The only remaining issue is with the missing `conda` executable.
We use it like this: `eval "$(conda shell.bash hook)"`
As far as I see, the `conda` package is no longer installed when installing `mamba`.
And if I try to install it manually, it downgrades mamba to v1.
I assume many our users have `conda install/activate` commands in the images based on our image, and I don't want to break them.
So, we should probably wait till we can install both conda and mamba v2 simultaneously.

flaviomartins:
> Hi Ayaz,
> 
> Regarding failing update due to conda, you might want to track [mamba-org/mamba#3503](https://github.com/mamba-org/mamba/issues/3503).

There is an RC now https://github.com/mamba-org/mamba/issues/3503#issuecomment-2455537329

mathbunnyru:
> > Hi Ayaz,
> > Regarding failing update due to conda, you might want to track [mamba-org/mamba#3503](https://github.com/mamba-org/mamba/issues/3503).
> 
> There is an RC now [mamba-org/mamba#3503 (comment)](https://github.com/mamba-org/mamba/issues/3503#issuecomment-2455537329)

Thanks, I saw that as well yesterday. Hope it won't be too long before release.

jjerphan:
The failure is due to conda being absent since mamba 2.0 does not depend on it.

For now conda and mamba 2.0 cannot be installed, but soon will be installable (see https://github.com/mamba-org/mamba/issues/3503#issuecomment-2397281920).

mathbunnyru:
Yes, waiting for the new `conda-libmamba-solver` to be released!

mathbunnyru:
> Yes, waiting for the new `conda-libmamba-solver` to be released!

It's been released, checking if everything works.

tests/docker-stacks-foundation/test_packages.py
```diff
@@ -76,10 +76,9 @@
     "hdf5",
     "jupyterhub-singleuser",
     "jupyterlab-git",
-    "mamba[version='<2.0.0']",
-    "notebook[version='>",
```
mathbunnyru, line 80: https://github.com/mamba-org/mamba/issues/3646

tests/docker-stacks-foundation/test_packages.py
```diff
@@ -76,10 +76,9 @@
...
+    "mamba",
+    "notebook>",
     "openssl",
-    "pandas[version='>",
```
mathbunnyru, line 82: I don't think we need this in both v1 and v2, will remove it as a separate commit today, if it's true.

mathbunnyru:
Everything seems to work fine. I will merge this tomorrow (because we already built some images today, and I don't want to change the meaning of today's date tag).

PR #2151, opened 2024-10-03
Use awk to pin python major.minor version

mathbunnyru:

### DIFF
--- a/images/docker-stacks-foundation/Dockerfile
+++ b/images/docker-stacks-foundation/Dockerfile
@@ -131,7 +131,7 @@ RUN set -x && \
     rm -rf /tmp/bin/ && \
     # Pin major.minor version of python
     # https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-pkgs.html#preventing-packages-from-updating-pinning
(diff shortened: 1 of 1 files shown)


images/docker-stacks-foundation/Dockerfile
```diff
@@ -131,7 +131,7 @@ RUN set -x && \
     rm -rf /tmp/bin/ && \
     # Pin major.minor version of python
     # https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-pkgs.html#preventing-packages-from-updating-pinning
-    mamba list --full-name 'python' | tail -1 | tr -s ' ' | cut -d ' ' -f 1,2 | sed 's/\.[^.]*$/.*/' >> "${CONDA_DIR}/conda-meta/pinned" && \
+    mamba list --full-name 'python' | awk 'END{sub("[^.]*$", "*", $2); print $1 " " $2}' >> "${CONDA_DIR}/conda-meta/pinned" && \
```
mathbunnyru, line 134: It is still quite clear what's going on if you know the basics of how `awk` works.
At the same time, this command works with both mamba v1 and v2 (the v2 has a prefix space, which is ignored by `awk`).

Issue #2154, opened 2024-10-09
Rename `ROOT_CONTAINER` and `BASE_CONTAINER` to `ROOT_IMAGE` and `BASE_IMAGE`


mathbunnyru:
These are not containers, but images.
In Docker, these 2 are different things, so we should use proper naming.

This change might be breaking for people, who build custom images, but it should only require a few lines change to fix it.

PR #2155, opened 2024-10-09
Rename: ROOT_CONTAINER->ROOT_IMAGE, BASE_CONTAINER->BASE_IMAGE

mathbunnyru:
## Describe your changes

There is no such thing as parent (base/root) container, it's about images, so let's use proper naming

## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/2154

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,9 @@
+# Changelog
+
+## 2024-10-09
+
+### Changed
+
+_This change might only breaking if you build your custom set of images._
+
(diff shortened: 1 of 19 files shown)


mathbunnyru:
This is not breaking if you simply run one of our images or use it as a parent image.
But it might be breaking for people who build custom set of images - I think they are advanced users and it's ok to break small things sometimes, they can fix the problem easily.

manics:
What might be nice is to have a (sort of) changelog page, but rather than being comprehensive it only has major/breaking changes along with the date? This could also be a nice place to highlight the planned update to Python 3.12.

mathbunnyru:
> What might be nice is to have a (sort of) changelog page, but rather than being comprehensive it only has major/breaking changes along with the date? This could also be a nice place to highlight the planned update to Python 3.12 when it's made.

Done, please, take a look

mathbunnyru:
https://github.com/jupyter/docker-stacks/blob/15f39abfed925a6ee89954ccd968777372ce346b/CHANGELOG.md

manics (approved):

PR #2157, opened 2024-10-09
Do not use shallow clones in Read the Docs to properly show last upda…

mathbunnyru:
…ted time

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -13,6 +13,9 @@ build:
     # nodejs: "20"
     # rust: "1.70"
     # golang: "1.20"
+  jobs:
+    post_checkout:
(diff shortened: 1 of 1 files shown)


mathbunnyru:
We use `sphinx-last-updated-by-git` which has a caveat described here: https://github.com/mgeier/sphinx-last-updated-by-git

I checked and the fix works.
Before: https://jupyter-docker-stacks.readthedocs.io/en/latest/maintaining/new-images-and-packages-policy.html

Screenshot 2024-10-09 at 15 51 31

After: https://jupyter-docker-stacks--2157.org.readthedocs.build/en/2157/maintaining/new-images-and-packages-policy.html
Screenshot 2024-10-09 at 15 51 53

PR #2162, opened 2024-10-23
Use mamba env export with `--json`

mathbunnyru:


mathbunnyru:
It seems that it was fixed in the upstream and let's use `json` to have less dependencies: https://github.com/mamba-org/mamba/pull/3559

When new mamba is released, I will check that it works now. If not I will let `mamba` team know.

mathbunnyru:
The upstream has released `mamba` v2 version where this issue is almost fixed.

Unfortunately, there is a bug in the case of using the `--from-history` flag.
When the issue is fixed in the upstream we'll be able to merge this PR.

https://github.com/mamba-org/mamba/issues/3582

mathbunnyru:
mamba 2.0.4 has been released, so this is fixed, and it won't prevent us from updating to mamba v2.
So, I'm merging this

PR #2167, opened 2024-11-04
[TMP] Pin min notebook version

mathbunnyru:

### DIFF
--- a/images/base-notebook/Dockerfile
+++ b/images/base-notebook/Dockerfile
@@ -42,7 +42,7 @@ RUN mamba install --yes \
     'jupyterhub' \
     'jupyterlab' \
     'nbclassic' \
-    'notebook' && \
+    'notebook>=7.2.2' && \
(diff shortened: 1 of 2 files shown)


mathbunnyru:
`jupyterlab` 4.3.0 has been released, but the new `notebook` which supports this version hasn't yet.

Unfortunately, previous versions of `notebook` (<v7) didn't have a version restriction on `jupyterlab`, so the solver installs jupyterlab `4.3.0` and notebook `6.5.7` (and obviously it doesn't work)

To fix this problem, I decided to pin the minimum `notebook` version.
We might need to keep this pin forever so we don't have the same problem with new releases of jupyterlab.

@krassowski @minrk @bollwyvl do you maybe know a better solution?

mathbunnyru:
Merging this, because currently, our builds are failing.
Please let me know, if there is a better solution.

Issue #2170, opened 2024-11-05
Nodejs has been dowgraded from 22.9.0 to 12.4.0 in latest images


manics:
### What docker image(s) are you using?

base-notebook

### Host OS system

Ubuntu 20.04

### Host architecture

x86_64

### What Docker command are you running?

```
docker run -it --rm quay.io/jupyter/base-notebook:2024-10-28 mamba list > 2024-10-28.txt
docker run -it --rm quay.io/jupyter/base-notebook:2024-11-04 mamba list > 2024-11-04.txt
diff -u 2024-10-28.txt 2024-11-04.txt
```

### How to Reproduce the problem?

See above

### Command output

```diff
--- 2024-10-28.txt      2024-11-05 12:18:56.990393826 +0000
+++ 2024-11-04.txt      2024-11-05 12:20:23.399905391 +0000
@@ -23,7 +23,7 @@
 attrs                     24.2.0             pyh71513ae_0    conda-forge
 babel                     2.14.0             pyhd8ed1ab_0    conda-forge
 beautifulsoup4            4.12.3             pyha770c72_0    conda-forge
-bleach                    6.1.0              pyhd8ed1ab_0    conda-forge
+bleach                    6.2.0              pyhd8ed1ab_0    conda-forge
 blinker                   1.8.2              pyhd8ed1ab_0    conda-forge
 boltons                   24.0.0             pyhd8ed1ab_0    conda-forge
 brotli-python             1.1.0           py312h2ec8cdc_2    conda-forge
@@ -42,7 +42,7 @@
 conda-libmamba-solver     24.9.0             pyhd8ed1ab_0    conda-forge
 conda-package-handling    2.4.0              pyh7900ff3_0    conda-forge
 conda-package-streaming   0.11.0             pyhd8ed1ab_0    conda-forge
-configurable-http-proxy   4.6.2                hbf95b10_1    conda-forge
+configurable-http-proxy   4.2.3           node12_hac6e442_0    conda-forge
 cryptography              43.0.3          py312hda17c39_0    conda-forge
 debugpy                   1.8.7           py312h2ec8cdc_0    conda-forge
 decorator                 5.1.1              pyhd8ed1ab_0    conda-forge
@@ -61,7 +61,6 @@
 httpcore                  1.0.6              pyhd8ed1ab_0    conda-forge
 httpx                     0.27.2             pyhd8ed1ab_0    conda-forge
[...]

PR #2171, opened 2024-11-05
Install jupyterhub-base instead of jupyterhub

consideRatio:
## Describe your changes

If you'd do `pip install jupyterhub`, you get the Python package jupyterhub. If you do `mamba install jupyterhub` however, you also get nodejs and configurable-http-proxy. Those are required if you are to run a jupyterhub server, but isn't required in order to start a jupyter server and establish contact with a jupyterhub server via the `jupyterhub-singleuser` command.

This PR makes us install `jupyterhub-base` instead, which is more like `pip install jupyterhub` as it doesn't bundle with other things to start jupyterhub as a server. To avoid a breaking change, nodejs is now explicitly installed as that could be a used dependency for people and could perhaps benefit from being removed in independent on this change.

## Related issue

`nodejs` was found to be downgraded from a modern version to the very old node 12 within the issue https://github.com/jupyter/docker-stacks/issues/2170. It seems that its only a dependency for `jupyterhub` (and via its dependency on `configurable-http-proxy`), so by switching to installing `jupyterhub-base`, nodejs may not get installed any more.

I think if nodejs is wanted, then we should install it directly and not rely on getting it via `jupyterhub` the conda-forge package.

## Checklist (especially for first-time contributors)

- [x] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests
- [x] I will try not to use force-push to make the review process easier for reviewers
- [ ] I have updated the documentation for significant changes

PR #2172, opened 2024-11-06
base-notebook: stop installing nodejs from conda-forge

consideRatio:
## Describe your changes

nodejs was a conda dependency of jupyterhub, but by installing
jupyterhub-base we no longer need it and could opt to remove it.

By doing this, building base-notebook led to a reported size reduction
from 974MB to 828MB, which is a 146MB / 15% size reduction.

## Issue ticket if applicable

- closes https://github.com/jupyter/docker-stacks/issues/2170 which is related if its still open

  I removed a test related to `npm`, dependency of `nodejs` conda-forge package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 This changelog only contains breaking and/or significant changes manually introduced to this repository (using Pull Requests).
 All image manifests can be found in [the wiki](https://github.com/jupyter/docker-stacks/wiki).
 
-## 2024-11-07
+## 2024-11-08
 
 Affected: all images except `docker-stacks-foundation`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 This changelog only contains breaking and/or significant changes manually introduced to this repository (using Pull Requests).
 All image manifests can be found in [the wiki](https://github.com/jupyter/docker-stacks/wiki).
 
-## 2024-11-??
+## 2024-11-07
 
 Affected: all images except `docker-stacks-foundation`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 This changelog only contains breaking and/or significant changes manually introduced to this repository (using Pull Requests).
 All image manifests can be found in [the wiki](https://github.com/jupyter/docker-stacks/wiki).
 
+## 2024-11-??
+
+Affected: all images except `docker-stacks-foundation`.
+
+- **Breaking:** base-notebook: stop installing nodejs from conda-forge ([#2172](https://github.com/jupyter/docker-stacks/pull/2172)).
+
+  Reason: It isn't a direct dependency on anything in the images any more, and increased the image size with ~150MB.
+
 ## 2024-11-06
 
 Affected: all images except `docker-stacks-foundation`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All image manifests can be found in [the wiki](https://github.com/jupyter/docker
 
 Affected: all images except `docker-stacks-foundation`.
 
-- **Breaking:** base-notebook: stop installing nodejs from conda-forge ([#2172](https://github.com/jupyter/docker-stacks/pull/2172)).
+- **Breaking:** `base-notebook`: stop installing `nodejs` from `conda-forge` ([#2172](https://github.com/jupyter/docker-stacks/pull/2172)).
 
   Reason: It isn't a direct dependency on anything in the images any more, and increased the image size with ~150MB.
 
--- a/docs/using/recipe_code/jupyterhub_version.dockerfile
+++ b/docs/using/recipe_code/jupyterhub_version.dockerfile
@@ -1,6 +1,6 @@
 FROM quay.io/jupyter/base-notebook
 
-RUN mamba install --yes 'jupyterhub-base==4.0.1' && \
+RUN mamba install --yes 'jupyterhub-singleuser==4.0.1' && \
     mamba clean --all -f -y && \
     fix-permissions "${CONDA_DIR}" && \
     fix-permissions "/home/${NB_USER}"
--- a/docs/using/selecting.md
+++ b/docs/using/selecting.md
@@ -54,7 +54,7 @@ It contains:
 
 - Everything in `jupyter/docker-stacks-foundation`
 - Minimally functional Server (e.g., no LaTeX support for saving notebooks as PDFs)
-- `notebook`, `jupyterhub-base`, and `jupyterlab` packages
+- `notebook`, `jupyterhub-singleuser`, and `jupyterlab` packages
 - A `start-notebook.py` script as the default command
 - A `start-singleuser.py` script useful for launching containers in JupyterHub
 - Options for a self-signed HTTPS certificate
--- a/docs/using/selecting.md
+++ b/docs/using/selecting.md
@@ -55,8 +55,6 @@ It contains:
 - Everything in `jupyter/docker-stacks-foundation`
 - Minimally functional Server (e.g., no LaTeX support for saving notebooks as PDFs)
 - `notebook`, `jupyterhub-base`, and `jupyterlab` packages
-  Note: we're also installing `nodejs` as it has historically been installed indirectly as a dependency of `jupyterhub` package, which was used before.
-  See more at: <https://github.com/jupyter/docker-stacks/pull/2171>
 - A `start-notebook.py` script as the default command
 - A `start-singleuser.py` script useful for launching containers in JupyterHub
 - Options for a self-signed HTTPS certificate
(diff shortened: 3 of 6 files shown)


mathbunnyru:
@mathbunnyru @consideRatio @yuvipanda @manics please vote here: 👍 or 👎 

We need 2 positive votes to accept this change.

mathbunnyru:
> By doing this, building base-notebook led to a reported size reduction from 974MB to 828MB, which is a 146MB / 15% size reduction.

That's great!

mathbunnyru:
I've checked and the latest published image is using the proper nodejs:

`x86_64` base (which had the problem): https://github.com/jupyter/docker-stacks/wiki/x86_64-default-base-notebook-28a0b25bcec1 
`nodejs 22.9.0`

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 This changelog only contains breaking and/or significant changes manually introduced to this repository (using Pull Requests).
 All image manifests can be found in [the wiki](https://github.com/jupyter/docker-stacks/wiki).
 
+## 2024-11-??
+
+Affected: all images except `docker-stacks-foundation`.
+
+- **Breaking:** base-notebook: stop installing nodejs from conda-forge ([#2172](https://github.com/jupyter/docker-stacks/pull/2172)).
```
mathbunnyru, line 10:
```suggestion
- **Breaking:** `base-notebook`: stop installing `nodejs` from `conda-forge` ([#2172](https://github.com/jupyter/docker-stacks/pull/2172)).
```

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 This changelog only contains breaking and/or significant changes manually introduced to this repository (using Pull Requests).
 All image manifests can be found in [the wiki](https://github.com/jupyter/docker-stacks/wiki).
 
+## 2024-11-??
```
mathbunnyru, line 6:
```suggestion
## 2024-11-07
```

I will probably merge this tomorrow

mathbunnyru:
@consideRatio why are we using `jupyterhub-base` here and not `jupyterhub-singleuser`?
And what's the difference between these 2?

consideRatio:
I understand the difference to also bundling with jupyterlab as a dependency when using jupyterhub-singleuser conda-forge package

mathbunnyru:
So, in our case, it will be the same thing in the end no matter what we use (because we're installing `jupyterlab` in the same place ourselves)?

consideRatio:
Yes i think so

mathbunnyru:
Then I think it's better to use `jupyterhub-singleuser`. It better shows our intent - we want this image to be used as a singleuser image.

We're using `jupyterhub-singleuser` executable in our images.
As mentioned by @ykazakov in https://github.com/jupyter/docker-stacks/pull/2171#issuecomment-2461893180 it is installed by `jupyterhub-base` package, but I think it's counterintuitive.
If `conda-forge` package managers decide to move the binary, we will still be safe.

@consideRatio @ykazakov what do you think?

flaviomartins:
> Then I think it's better to use `jupyterhub-singleuser`. It better shows our intent - we want our images to be used as a singleuser images.
> 
> We're using `jupyterhub-singleuser` executable in our images. As mentioned by @ykazakov in [#2171 (comment)](https://github.com/jupyter/docker-stacks/pull/2171#issuecomment-2461893180) it is installed by `jupyterhub-base` package, but I think it's counterintuitive. If `conda-forge` package managers decide to move the binary, we will still be safe.
> 
> @consideRatio @ykazakov what do you think?

I researched this a little bit and found it a little confusing that the `conda-forge` package maintainers decided to use jupyterhub-base as a dependency for the main jupyterhub package instead of going with jupyterhub-singleuser (here https://github.com/conda-forge/jupyterhub-feedstock/blob/970985288045a6f87239f2b662bccac3563b79cb/recipe/meta.yaml#L106).

But it now seems like you are right! `jupyterhub` is for server, thus `jupyterhub-singleuser` is the package for this docker image.

consideRatio:
@minrk do you have a recommendation here?

I have a guess that jupyterhub-singleuser was initially used, and then jupyterhub-base came when it wasnt a good assumption that jupyterlab was required or similar, so in that sense, it be more relevant to encourage use of jupyterhub-base than -singleuser?

ykazakov:
> Then I think it's better to use `jupyterhub-singleuser`. It better shows our intent - we want our images to be used as a singleuser images.

I agree. Looks like the package `jupyterhub-singleuser` does not install any files by itself. So it is mainly used for aggregating dependencies needed to run the user servers.

minrk:
The conda `jupyterhub` package is the _application_ JupyterHub, which requires the Python package and the proxy (i.e. if you `conda install jupyterhub`, you can run `jupyterhub`; this is not true of `jupyterhub-base`). `jupyterhub-base` is just the Python `jupyterhub` package, what you get from `pip install`. This matches a widespread pattern in conda of e.g. `matplotlib-base`, etc. where the `-base` package is the bare minimum, and may not be fully functional without other dependencies, while the default name gets you a more 'default' install with fuller dependencies. It is a bit analogous to apt's `recommends`, which are dependencies that you get by default, but can technically opt-out of if you want with `--no-install-recommends`. `-base` is the `--no-recommends` package naming convention.

`jupyterhub-singleuser` is an empty metapackage to get enough to be able to execute `jupyterhub-singleuser` (i.e. base + jupyterlab, though technically other combinations are still valid without jupyterlab - notebook, jupyter-server + custom UI, etc.).

I'd say `jupyterhub-base` is the right choice here, because we're already installing the single-user server components separately and explicitly, installing `jupyterhub-singleuser` doesn't get anything.

minrk:
There's no real downside to `jupyterhub-singleuser` either, if you want to go that way. If it every develops more precise constraints (e.g. jupyter-server>2), those constraints will be applied in `jupyterhub-singleuser` (it will never have any content, it will always only be a collection of dependencies, at least until/unless jupyterhub-singleuser becomes a separate Python package, which I have no plans to do)

mathbunnyru:
Thanks, everyone!

If there are no real downsides to using `-singleuser` let's have it.
@consideRatio please make it as part of this PR (no need to create a new one).

I'm 100% sure it will be more clear for people who don't know the nuances of these packages what's going on. And even for those who do, it still looks clear - we want to run singleuser instances, so we install `-singleuser` package.

minrk:
I'll be a little more precise: there are no real downsides to JupyterHub-singleuser _here_ because we install JupyterLab. The only real downside is that it excludes valid singleuser environments that don't want to include JupyterLab.

flaviomartins:
@consideRatio @mathbunnyru Would it make sense to keep the dependency on `jupyterhub-base` here, and add the `jupyterhub-singleuser` dependency on the `minimal-notebook` Dockerfile? Just an idea. What if the `jupyterlab` dependency where to move to `minimal-notebook` too?

mathbunnyru:
> > I'll be a little more precise: there are no real downsides to JupyterHub-singleuser _here_ because we install JupyterLab. The only real downside is that it excludes valid singleuser environments that don't want to include JupyterLab.
> 
> @consideRatio @mathbunnyru Just an idea. Would it make sense to add the dependency on `jupyterhub-base` here in the `base-notebook` image, and the `jupyterhub-singleuser` dependency on the `minimal-notebook` Dockerfile instead? Effectively, moving the `jupyterlab` dependency to `minimal-notebook` as well?
> 
> I think every image in this repository has `minimal-notebook` as `BASE_IMAGE`, so nothing changes there. Moving the `jupyterlab` dependencies to `minimal-notebook` would allow someone to use `base-notebook` as their `BASE_IMAGE` to roll with something other than `jupyterlab`, etc. (e.g., build `another-minimal-notebook` image)

It's a whole different discussion, to be honest.

> Effectively, moving the `jupyterlab` dependency to `minimal-notebook` as well?

This is likely to break so many users.
And if you want to move `jupyterlab` to `minimal-notebook`, you will have to move `notebook` as well, because it depends on `jupyterlab`: https://github.com/conda-forge/notebook-feedstock/blob/main/recipe/meta.yaml

> I think every image in this repository has `minimal-notebook` as `BASE_IMAGE`, so nothing changes there. Moving the `jupyterlab` dependencies to `minimal-notebook` would allow someone to use `base-notebook` as their `BASE_IMAGE` to roll with something other than `jupyterlab`, etc. (e.g., build `another-minimal-notebook` image)

That's not true. We already have `docker-stacks-foundation` for exactly this reason. It allows you to install anything you want and doesn't have `jupyterlab` included.

So I don't think this change would solve any problem, but will definitely break many installations.

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 This changelog only contains breaking and/or significant changes manually introduced to this repository (using Pull Requests).
 All image manifests can be found in [the wiki](https://github.com/jupyter/docker-stacks/wiki).
 
+## 2024-11-07
```
mathbunnyru, line 6:
```suggestion
## 2024-11-08
```

mathbunnyru:
Great, thanks. I will merge this when tests pass (should take around an hour).

PR #2176, opened 2024-12-02
Remove pandas from EXCLUDED_PACKAGES

mathbunnyru:


tests/docker-stacks-foundation/test_packages.py
```diff
@@ -79,7 +79,6 @@
     "mamba[version='<2.0.0']",
     "notebook[version='>",
     "openssl",
-    "pandas[version='>",
```
mathbunnyru, line 82: Don't think we need this line at all

PR #2178, opened 2024-12-03
Downgrade to Julia 1.11.1

mathbunnyru:
## Describe your changes

Julia `1.11.2` has just been released and it seems that it causes our self-hosted aarch64 runners to hang.

Previous build was using `1.11.1` and has been working fine: https://github.com/jupyter/docker-stacks/actions/runs/12123374006/job/33801780215

Let's try to downgrade to `1.11.1` and see if it helps.

### DIFF
--- a/images/minimal-notebook/setup-scripts/setup_julia.py
+++ b/images/minimal-notebook/setup-scripts/setup_julia.py
@@ -47,6 +47,11 @@ def get_latest_julia_url() -> tuple[str, str]:
     triplet = unify_aarch64(platform.machine()) + "-linux-gnu"
     file_info = [vf for vf in latest_version_files if vf["triplet"] == triplet][0]
     LOGGER.info(f"Latest version: {file_info['version']} url: {file_info['url']}")
+    if file_info["version"] == "1.11.2":
+        LOGGER.warning(
+            "Not using Julia 1.11.2, because it hangs in GitHub self-hosted runners"
+        )
+        return file_info["url"].replace("1.11.2", "1.11.1"), "1.11.1"
     return file_info["url"], file_info["version"]
 
 


benz0li:
> Julia `1.11.2` has just been released and it seems that it causes our self-hosted aarch64 runners to hang.

@mathbunnyru Let us see, if Julia 1.11.2 causes the same problem with b-data's/my GitLab runners.

mathbunnyru:
Although we merged mamba v2, I don't think it's related, because mamba v2 was build just fine in the branch.

Julia v1.11.2 has just been released, so I suppose it can be a problem somehow.
https://github.com/JuliaLang/julia/releases/tag/v1.11.2

mathbunnyru:
Unfortunately, GitHub Runner seems to lose the logs, crash the host machine, so it's not easy to debug.
I was able to take a look at the logs online:

Screenshot 2024-12-03 at 16 34 12

Note that for 7-8 minutes there were no logs.
In v1.11.1 it started outputting some logs after 2.5 minutes: https://github.com/jupyter/docker-stacks/actions/runs/12123374006/job/33801780215#step:8:65

mathbunnyru:
@benz0li do you have something similar to our `setup-julia-packages.bash`? If not, it will work for you fine.
It hangs (or crashes, or something) inside this script.

benz0li:
> @benz0li do you have something similar to our `setup-julia-packages.bash`?

Yes. In the `Dockerfile`: https://github.com/b-data/jupyterlab-julia-docker-stack/blob/a8b728c4e70ae460072cc22197bf632e5954197d/base/latest.Dockerfile#L307-L344

mathbunnyru:
v1.11.1 builds just fine for `julia-notebook`: https://github.com/jupyter/docker-stacks/actions/runs/12144295877/job/33863916105

mathbunnyru:
And `datascince-notebook` as well: https://github.com/jupyter/docker-stacks/actions/runs/12144295877/job/33864340318?pr=2178

mathbunnyru:
This seems to fix the builds, so I'm merging this. When I have time, I will report the problem to the upstream.

mathbunnyru:
> > > Julia `1.11.2` has just been released and it seems that it causes our self-hosted aarch64 runners to hang.
> > 
> > 
> > @mathbunnyru Let us see, if Julia 1.11.2 causes the same problem with b-data's/my GitLab runners.
> 
> No problem with b-data's/my GitLab runners: See https://gitlab.b-data.ch/jupyterlab/julia/base/-/jobs/135429

It seems that you're using `Apple Silicon M2 Max, 3.5 GHz, 8 cores, 48 GB RAM` - very powerful machine.
We're using `t2a-standard-1 ` machine https://cloud.google.com/compute/docs/general-purpose-machines#t2a_machine_types, which is a magnitude worse 😃
I think Julia started taking too much resources, and now the VM is hanging.

It might still be a performance bug in a new release.

giordano:
Is there any way to reproduce the hang you observed on a local machine? Or at least, a clear indication of what's the workload you're running.

mathbunnyru:
@giordano I created an issue, let's continue the discussion there: https://github.com/jupyter/docker-stacks/issues/2198

I also added the answer to your question

PR #2182, opened 2024-12-09
Try to use ubuntu-24.04 GitHub runners

mathbunnyru:
## Describe your changes

**I'm not planning to merge this, but I would like to check everything works.**

Ubuntu 24.04 will soon be used when someone uses `ubuntu-latest`: https://github.com/actions/runner-images/issues/10636

So, we need to verify it works well for our workflows.


mathbunnyru:
Everything seems to work except for script to setup our self-hosted runners.
I hope we won't need them anymore quite soon, so we should be fine, so I'm closing this PR.

PR #2193, opened 2024-12-29
Try to use latest Julia 1.11.2

mathbunnyru:


mathbunnyru:
I thought the problem with hanging during Julia installation was only happening on aarch64: https://github.com/jupyter/docker-stacks/pull/2193

But this doesn't seem to be true anymore: https://github.com/jupyter/docker-stacks/actions/runs/12531493732/job/34949338865

Maybe, it depends on the speed of the julia packages download or something like that and not the version?

mathbunnyru:
No, the new versions still works worse

mathbunnyru:
`1.11.3` also hangs the runners

mathbunnyru:
Closing this one, because switching to GitHub-hosted runners should resolve the issue

Issue #2201, opened 2025-01-16
Linux arm64 hosted runners now available for free 😃


manics:
> Now in public preview, Linux arm64 hosted runners are available for free in public repositories. Following the [release of arm64 larger hosted runners in June](https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/), this offering now extends to the open source-community. Powered by the Cobalt 100-based processors, these 4 vCPU runners can deliver up to a 40% performance boost compared to Microsoft Azure’s previous generation of Arm-based VMs, providing a power-efficient compute layer for your workloads. Arm-native developers can now build, test and deploy entirely within the arm64 architecture without the need for virtualization on your Actions runs.

https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

mathbunnyru:
@yuvipanda you will be really happy about this :)

mathbunnyru:
Took me 13 minutes to come up with a PR, you can see how much I've been waiting for this. Let's see if it works.

choldgraf:
amazing! it makes me happy when things we once paid for are now free 😅

Issue #2198, opened 2025-01-08
Julia 1.11.2 hangs the aarch64 builds


mathbunnyru:
More info: https://github.com/jupyter/docker-stacks/pull/2178

mathbunnyru:
@giordano

> Is there any way to reproduce the hang you observed on a local machine? Or at least, a clear indication of what's the workload you're running.

Hi,

1. The machine running the build is `t2a-standard-1` on google cloud: https://gcloud-compute.com/t2a-standard-1.html
2. The only thing running on the machine is the GitHub actions runner, which actually performs the build
3. It hangs here (https://github.com/jupyter/docker-stacks/pull/2178#issuecomment-2515095811): https://github.com/jupyter/docker-stacks/blob/main/images/minimal-notebook/setup-scripts/setup-julia-packages.bash#L39

I think to reproduce you need to run:
1. https://github.com/jupyter/docker-stacks/blob/main/images/minimal-notebook/setup-scripts/setup_julia.py
2. https://github.com/jupyter/docker-stacks/blob/main/images/minimal-notebook/setup-scripts/setup-julia-packages.bash

Don't forget to revert the patch I made to not use 1.11.2

giordano:
Sorry, I never came back to this because I don't have access to Google Cloud nor any aarch64-linux machine with Docker set up, but I just tried to run the following
```sh
JULIA_CPU_TARGET="generic;cortex-a57;thunderx2t99;carmel,clone_all;apple-m1,base(3);neoverse-512tvb,base(3)" julia --project=. -e '
import Pkg;
Pkg.update();
Pkg.add([
    "HDF5",
    "IJulia",
    "Pluto"
]);
Pkg.precompile();
'
```
with Julia v1.11.2 and v1.11.3 on Nvidia Grace, Apple M1 (with Asahi Linux), Ampere eMAG, and Ampere Altra Max M128-30 (the latter should be somewhat similar to the Google Cloud one) and all processes finished within 2-3 minutes for me.

mathbunnyru:
> Sorry, I never came back to this because I don't have access to Google Cloud nor any aarch64-linux machine with Docker set up, but I just tried to run the following
> 
> JULIA_CPU_TARGET="generic;cortex-a57;thunderx2t99;carmel,clone_all;apple-m1,base(3);neoverse-512tvb,base(3)" julia --project=. -e '
> import Pkg;
> Pkg.update();
> Pkg.add([
>     "HDF5",
[...]

PR #2202, opened 2025-01-16
Use GitHub-hosted aarch64 runners

mathbunnyru:
## Issue ticket if applicable

Fix: https://github.com/jupyter/docker-stacks/issues/2201
Fix: https://github.com/jupyter/docker-stacks/issues/2198


manics:
There are some bugs in the ubuntu-24.04-arm image- some environment variables are incorrect:
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3605

consideRatio:
Beautiful!!! :heart: :tada:

consideRatio:
A job failed in this PR, but not in the main branch - a difference between these jobs is the test runner's python version, which is 3.10 (main branch) vs 3.13 (in this PR) --- 

image

image

Then in the x86_64 test from main branch and this PR, the Python test runner version is 3.12:

image

---

Maybe the Python version in the aarch image is floating around a bit? Perhaps because the self-hosted aarch runners have a 3.x mapped to 3.10, while the new aarch runners have a 3.x mapped to 3.13, and the x86 runners have 3.x mapped to 3.12?

I figure the fix could be to pin to 3.12

consideRatio:
I pushed a commit to see if it helps to pin to 3.12!

mathbunnyru:
Thanks @consideRatio!
Let's try to restart it a few times to see if it works.

mathbunnyru:
`zstd: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?` - I don't think Docker is stable enough in these runners (and I suspect that might have been the problem with the previous failure as well).

consideRatio:
Oh, hmmm, i wonder if it could be a timing thing - were the CI scripts started faster than docker daemon in arm64 compared to in x64.

These things are messy to debug :/

mathbunnyru:
> Oh, hmmm, i wonder if it could be a timing thing - were the CI scripts started faster than docker daemon in arm64 compared to in x64.
> 
> These things are messy to debug :/

I think we have quite a nice example of this happening!

This only runs `docker build` and nothing else.
https://github.com/jupyter/docker-stacks/actions/runs/12999905803/job/36256210849?pr=2202

mathbunnyru:
I added a `sleep 10` - let's see if this eliminates the problem

mathbunnyru:
Wow, it seems that even `@actions/checkout` can fail (possibly for the same reason): https://github.com/jupyter/docker-stacks/actions/runs/13000122202/job/36256874950?pr=2202#step:3:20

And we can't have this workaround in `create-dev-env` (because we need to checkout first)

.github/workflows/docker-build-test-upload.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     steps:
       # Image with CUDA needs extra disk space
       - name: Free disk space 🧹
-        if: contains(inputs.variant, 'cuda') && inputs.platform == 'x86_64'
+        if: contains(inputs.variant, 'cuda') && runner.arch == 'X64'
```
mathbunnyru, line 44: Let's use `inputs.platform` when we want to name things
`runner.arch` is better, because it's standard when using GitHub runners

I will merge this change separately, so it's not part of this PR

mathbunnyru:
It seems that GitHub-hosted aarch64 runners aren't currently stable enough. So, let's wait for a while, I hope it will be better.

mathbunnyru:
The relevant discussions: https://github.com/orgs/community/discussions/148648

.github/workflows/docker.yml
```diff
@@ -66,8 +66,7 @@ jobs:
       parent-image: ""
       image: docker-stacks-foundation
       platform: aarch64
-      runs-on: ARM64_FAST
-    if: github.repository_owner == 'jupyter'
+      runs-on: ubuntu-24.04-arm
```
consideRatio, line 69: Github rep has said that they believe the ubuntu-22.04-arm runners may be stable @mathbunnyru - I think using ubuntu 22.04 as runner environment may be fine still - I'm not fully sure.

mathbunnyru: > Github rep has said that they believe the ubuntu-22.04-arm runners may be stable @mathbunnyru

Great, I'll try it out!

> I think using ubuntu 22.04 as runner environment may be fine still - I'm not fully sure.

Definitely still fine, because that's what self-hosted runners are still running, and I had 0 issues due to that 😄 
So, in a way, switching to `ubuntu-22.04-arm` is even better (we're changing less important things).

consideRatio: Ooooh looking better so far - not seen all tests pass but some

mathbunnyru: > Ooooh looking better so far - not seen all tests pass but some

Yes, and actually, the last run of `24.04` was fine as well.
But I'm completely  ok with switching to `ubuntu-22.04-arm` first

mathbunnyru:
Seems to work fine.

I updated the ChangeLog date (for tomorrow) and if this is successful once again, I will merge this PR.

I'm merging this a bit early (it's still testing phase for GitHub), but I'll keep Google Cloud self-hosted instances (they turn on on schedule and now don't cost that much), if something happens (probably until GitHub releases these runners as stable and then for one more month, to give it a full test by other contributors).
There is a reason to merge early - this fixes an issue where we were not able to install latest Julia (I don't know the reasons, and they become irrelevant for this project after merging this change).

And one more cool thing: currently the build takes around [31m 56s](https://github.com/jupyter/docker-stacks/actions/runs/13250510749)
This is almost as good, as we had it few years ago when we were building only x86_64 images and on a single runner, but now we do more things (and better): uploading artifacts (proven to be useful when something goes wrong), having many more images (foundation/julia/pytorch and pytorch/tensorflow cuda variants have all been added after multi-arch images), 2 architectures and running all tests for all images.
So, I would say, it's a great result, and I'm pretty happy with it.
Of course, it's not only due to using GitHub-hosted runners, but this was the biggest (or, smallest?) bottleneck in terms of build time for a few years.

mathbunnyru:
@consideRatio @yuvipanda @manics just so you know, I changed the permissions to run PRs as follows:

Screenshot 2025-02-12 at 14 54 03

Before, I had to manually run all commits of all external contributors. It was important, because our self-hosted runners were running arbitrary code from PR. Now we're using GitHub-hosted runners and there is no reason to be so restrictive.

Should make my life a bit easier - I had to click "Run" on all commits of all external contributors 😄

PR #2203, opened 2025-01-27
Don't override singleuser ip to 0.0.0.0

manics:
## Describe your changes

JupyterHub controls which interface(s) the singleuser server listens on.

The IP is set in the Spawner configuration, and passed to the singleuser server in the environment variable `JUPYTERHUB_SERVICE_URL`:
https://jupyterhub.readthedocs.io/en/latest/reference/spawners.html#note-on-ips-and-ports

Docker-stacks shouldn't override it. For example, if you configure a JupyterHub container spawner to use a docker-stacks image and use the default CMD it won't work with IPv6.

### DIFF
--- a/images/base-notebook/start-singleuser.py
+++ b/images/base-notebook/start-singleuser.py
@@ -8,9 +8,7 @@ import sys
 # Entrypoint is start.sh
 command = ["jupyterhub-singleuser"]
 
-# set default ip to 0.0.0.0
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

mathbunnyru:
@manics I'll merge this when tests pass, but I would be grateful if you added a test case both for this PR and previous PR.

PR #2208, opened 2025-02-10
Surface PYTHON_VERSION as a Makefile variable

mnp:
## Describe your changes

* Surface PYTHON_VERSION as a Makefile variable. This can be overridden like `make PYTHON_VERSION=3.10 build-all`.
* Wrap and indent a growing command line.

## Issue ticket if applicable
n/a


mathbunnyru (changes requested):
In the PR comment you say that you use it with `build-all`, but I don't see these args mentioned in `build-all`, does bash propagate them automatically?

I have a slight problem, that we'll have default python version in 2 separate places, but I think it's not a huge deal.

PR #2210, opened 2025-02-11
Use ubuntu-24.04 instead of ubuntu-latest

mathbunnyru:

### DIFF
--- a/.github/workflows/contributed-recipes.yml
+++ b/.github/workflows/contributed-recipes.yml
@@ -21,7 +21,7 @@ on:
 
 jobs:
   generate-matrix:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


mathbunnyru:
Why I want to use explicit tag:
- consistent with aarch64 runners (there is no `ubuntu-latest-arm`-like tag at the moment)
- we control when we switch, no bizarre situation when suddenly start building everything on new runners without even knowing about it
- no warnings like https://github.com/jupyter/docker-stacks/actions/runs/12788007366 (having 67 warnings makes me to ignore them completely)
- explicit is better than implicit

Downside:
- need to make one more commit every 2 years :)

I don't think is gonna affect images in any way, but it's nice to be in control of how we build our images.

PR #2212, opened 2025-02-17
Build contributed recipes in PRs

mathbunnyru:


mathbunnyru:
The idea behind this PR is to be able to build contributed recipes right in PRs if images change.
There is some work that needs to be done in `docker.yml` to make this work, so I'm going to merge this one as is, and then work on it.
And right now builds are failing, so I can't actually test this easily.

PR #2213, opened 2025-02-17
Enable building contributed recipes in PRs

mathbunnyru:


mathbunnyru:
With this change contributed recipes stage will run in parallel to pushing tags and merging tags, so we should decrease time it takes to build everything and then the contributed recipes.

mathbunnyru:
In this PR I change both `contributed-recipes` and `docker` workflows.

This means that `contributed-recipes` runs twice :
1. It is triggered because PR changes the workflow file. It runs using images already pushed to the registry: https://github.com/jupyter/docker-stacks/actions/runs/13377348557/job/37359262602?pr=2213
2. As part of `docker.yml` workflow. This doesn't use registry at all and uses images built during running `docker.yml`

PR #2214, opened 2025-02-17
First upload artifacts and then run tests to make sure we can easily …

mathbunnyru:
…debug broken images

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -458,11 +458,7 @@ jobs:
             { image: all-spark-notebook, variant: default },
           ]
     needs: [aarch64-images-tag-push, x86_64-images-tag-push]
-    if: |
-      always() &&
-      needs.x86_64-images-tag-push.result == 'success' &&
-      (needs.aarch64-images-tag-push.result == 'success' || needs.aarch64-images-tag-push.result == 'skipped') &&
-      !contains(github.event.pull_request.title, '[FAST_BUILD]')
+    if: (github.repository_owner == 'jupyter' || github.repository_owner == 'mathbunnyru') && !contains(github.event.pull_request.title, '[FAST_BUILD]')
 
   merge-tags-fast:
     uses: ./.github/workflows/docker-merge-tags.yml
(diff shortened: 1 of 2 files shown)


mathbunnyru:
Now, we're first running tests and if they fail, we don't have any debug information about the image.
I want to first upload manifests and so on (small files), then upload image (it's relatively reliable), and then run tests (this sometimes break).

.github/workflows/docker.yml
```diff
@@ -458,11 +458,7 @@ jobs:
             { image: all-spark-notebook, variant: default },
           ]
     needs: [aarch64-images-tag-push, x86_64-images-tag-push]
-    if: |
```
mathbunnyru, line 461: This is not needed in this repo: https://github.com/jupyter/docker-stacks/pull/2091#discussion_r1485652584

And it actually makes some harm:
https://github.com/jupyter/docker-stacks/actions/runs/13362035735

In this run, building aarch64 `r-notebook` failed.
In my opinion, we should have not even tried to merge tags.
Instead, we tried, but only failed because manifest file was not available (because we were uploading it after running tests).

Instead, we should keep this condition simple and in accordance with other places.

mathbunnyru:
Having artefacts is extremely useful when for example x86_64 tests are fine, but there is a problem with aarch64 image (it builds as well, but tests don't work).
This is the case we have right now.

mathbunnyru:
Seems to work, I'm able to download `r-notebook` aarch64 artefacts, even though the job fails.
This should help to debug the problems with the images due to incompatible packages.

mathbunnyru:
Difference between newly built R image and the last successful is the following:

```diff
<   r-cli                          3.6.3
---
>   r-cli                          3.6.4
36c36
<   r-dials                        1.3.0
---
>   r-dials                        1.4.0
121c121
<   r-parsnip                      1.2.1
---
>   r-parsnip                      1.3.0
145c145
<   r-r6                           2.5.1
---
>   r-r6                           2.6.1
156c156
<   r-recipes                      1.1.0
---
>   r-recipes                      1.1.1
189c189
<   r-testthat                     3.2.1.1
---
>   r-testthat                     3.2.3
```

PR #2215, opened 2025-02-17
Pin some packages to fix `r-notebook` and `datascience-notebook` under aarch64

mathbunnyru:

### DIFF
--- a/images/datascience-notebook/Dockerfile
+++ b/images/datascience-notebook/Dockerfile
@@ -57,6 +57,9 @@ RUN mamba install --yes \
     'r-tidymodels' \
     'r-tidyverse' \
     'rpy2' \
+    # Temporary fix
+    'r-parsnip<1.3.0' \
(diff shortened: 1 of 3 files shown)


mathbunnyru:
I was able to determine which packages changed since last build: https://github.com/jupyter/docker-stacks/pull/2214#issuecomment-2663604823

Changing these 2 packages back fixes tests.

mathbunnyru:
And here is explanation why this change is valid:

https://github.com/tidymodels/recipes/blob/main/DESCRIPTION
https://github.com/tidymodels/parsnip/blob/main/DESCRIPTION

In both these files you will find `sparsevctrs (>= 0.2.0),` has been added in the latest version.
And this package is not yet available for `aarch64`: https://github.com/conda-forge/r-sparsevctrs-feedstock/pull/4

So, let's not use latest versions until package we depend on becomes available.

PR #2218, opened 2025-02-17
Don't create extra free space in runners for cuda images

mathbunnyru:


mathbunnyru:
Runners have around 28G space, it might be enough for us.

https://github.com/jlumbroso/free-disk-space doesn't seem to be well supported, so if everything works without it, I would prefer not to use this action.

PR #2222, opened 2025-02-18
Use Python 3.12 for internal code

mathbunnyru:


mathbunnyru:
The images have been providing python 3.12 for almost 4 months.
There is no reason why our code should be stuck with python 3.9

.github/workflows/pre-commit.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Set Up Python 🐍
         uses: actions/setup-python@v5
         with:
-          python-version: 3.x
```
mathbunnyru, line 24: Let's be explicit here

PR #2227, opened 2025-02-20
Add timeouts to all GitHub jobs

mathbunnyru:


mathbunnyru:
I recently had a timeout issue in my fork: https://github.com/mathbunnyru/docker-stacks/actions/runs/13424787928/job/37505995405

And we actually hit 6h timeout many times when building aarch64 images on self-hosted runners.

So, let's add GitHub timeouts, to fail fast.
These are not super strict and I would rather give a few extra minutes for a job to finish than get a build failure.
And we might want to increase some of these in the future.

PR #2231, opened 2025-02-21
Better tests directory structure

mathbunnyru:


mathbunnyru:
I checked - number of tests for aarch64-base stays the same.

tests/conftest.py
```diff
@@ -52,95 +34,10 @@ def image_name() -> str:
...
-        return [line for line in logs.splitlines() if line.startswith(pattern)]
-
-    def remove(self) -> None:
-        """Kills and removes the tracked docker container."""
-        if self.container:
-            self.container.remove(force=True)
-
-
 @pytest.fixture(scope="function")
-def container(docker_client: docker.DockerClient, image_name: str) -> Container:
```
mathbunnyru, line 143: This was a bug, we don't return Container, but our TrackedContainer wrapper

PR #2235, opened 2025-02-22
Clarify container.remove() behaviour

mathbunnyru:

### DIFF
--- a/tagging/utils/docker_runner.py
+++ b/tagging/utils/docker_runner.py
@@ -39,9 +39,8 @@ class DockerRunner:
     ) -> None:
         assert self.container is not None
         LOGGER.info(f"Removing container {self.container.name} ...")
-        if self.container:
-            self.container.remove(force=True)
-            LOGGER.info(f"Container {self.container.name} removed")
+        self.container.remove(force=True)
+        LOGGER.info(f"Container {self.container.name} removed")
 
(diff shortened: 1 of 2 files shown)


tagging/utils/docker_runner.py
```diff
@@ -39,9 +39,8 @@ def __exit__(
     ) -> None:
         assert self.container is not None
         LOGGER.info(f"Removing container {self.container.name} ...")
-        if self.container:
```
mathbunnyru, line 42: `assert` above means that this `if` is always true

tests/utils/tracked_container.py
```diff
@@ -92,5 +92,9 @@ def _lines_starting_with(logs: str, pattern: str) -> list[str]:
 
     def remove(self) -> None:
         """Kills and removes the tracked docker container."""
-        if self.container:
+        if self.container is None:
+            LOGGER.info("No container to remove")
+        else:
+            LOGGER.info(f"Removing container {self.container.name} ...")
             self.container.remove(force=True)
+            LOGGER.info(f"Container {self.container.name} removed")
```
mathbunnyru, line 100: Just better logs.

There is some sort of code duplication between DockerRunner and TrackedContainer, but for now let's keep `tagging/` and `tests` dir completely independent of each other.

PR #2239, opened 2025-02-23
[FAST_BUILD] Refactor tagging: create functions, better logs and names, textwrap.d…

mathbunnyru:
…edent


tagging/apps/apply_tags.py
```diff
@@ -15,9 +15,6 @@
 
 
 def apply_tags(config: Config) -> None:
-    """
```
mathbunnyru, line 18: These docstrings are not helpful (especially in apps), and not consistent at all

tagging/apps/apply_tags.py
```diff
@@ -28,6 +25,8 @@ def apply_tags(config: Config) -> None:
         LOGGER.info(f"Applying tag: {tag}")
         docker["tag", config.full_image(), tag] & plumbum.FG
 
+    LOGGER.info(f"All tags applied to image: {config.image}")
```
mathbunnyru, line 28: Tried to add it everywhere so looks are more meaningful

tagging/apps/merge_tags.py
```diff
@@ -15,24 +15,31 @@
...
 
 
-def merge_tags(config: Config) -> None:
-    """
-    Merge tags for x86_64 and aarch64 images when possible.
-    """
-    LOGGER.info(f"Merging tags for image: {config.image}")
-
-    all_tags: set[str] = set()
+def read_tags_from_files(config: Config) -> set[str]:
```
mathbunnyru, line 18: Separate function makes it more understandable

tagging/apps/write_manifest.py
```diff
@@ -24,10 +22,12 @@
 MARKDOWN_LINE_BREAK = "<br />"
 
 
-def write_build_history_line(
-    config: Config, filename: str, all_tags: list[str]
-) -> None:
-    LOGGER.info("Appending build history line")
+def get_build_history_line(config: Config, filename: str, container: Container) -> str:
```
mathbunnyru, line 25: The idea is that get-like functions are more testable.
And we don't mix writing to files and creating file content

tagging/hierarchy/get_manifests.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+from tagging.hierarchy.images_hierarchy import ALL_IMAGES
+from tagging.manifests.manifest_interface import ManifestInterface
+
+
+def get_manifests(image: str | None) -> list[ManifestInterface]:
```
mathbunnyru, line 7: After this refactoring, there is no need to have both taggers and manifests in the same place.
So, it's better to have 2 functions.
Their implementation is similar, but that's ok

tagging/utils/quoted_output.py
```diff
@@ -1,5 +1,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
+import textwrap
```
mathbunnyru, line 3: Makes the code a bit easier to read

PR #2240, opened 2025-02-24
[FAST_BUILD] Use default permissions for GitHub workflows

mathbunnyru:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -523,8 +523,6 @@ jobs:
     uses: ./.github/workflows/docker-wiki-update.yml
     needs: [aarch64-images-tag-push-fast, x86_64-images-tag-push-fast]
     if: contains(github.event.pull_request.title, '[FAST_BUILD]')
-    permissions:
-      contents: write
 
   contributed-recipes:
     uses: ./.github/workflows/contributed-recipes.yml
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -523,8 +523,6 @@ jobs:
     uses: ./.github/workflows/docker-wiki-update.yml
     needs: [aarch64-images-tag-push-fast, x86_64-images-tag-push-fast]
     if: contains(github.event.pull_request.title, '[FAST_BUILD]')
-    permissions:
-      contents: write
```
mathbunnyru, line 527: I don't think this is needed, because we only change wiki in `main`, and this won't run in `main`

.github/workflows/pre-commit.yml
```diff
@@ -7,9 +7,6 @@ on:
       - main
   workflow_dispatch:
 
-permissions:
-  contents: read
```
mathbunnyru, line 11: I think defaults should be good enough

.github/workflows/sphinx.yml
```diff
@@ -27,8 +27,6 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 10
     if: github.repository_owner == 'jupyter' || github.repository_owner == 'mathbunnyru' || github.event_name != 'schedule'
-    permissions:
-      contents: write
```
mathbunnyru, line 31: This is wrong now - this was needed to support multi-lang docs and push strings to the main branch. We don't need that now.

PR #2241, opened 2025-02-24
Use tabulate in wiki to create better tables

mathbunnyru:

### DIFF
--- a/wiki/update_wiki.py
+++ b/wiki/update_wiki.py
@@ -5,9 +5,11 @@ import argparse
 import datetime
 import logging
 import shutil
+import textwrap
 from pathlib import Path
 
 import plumbum
+import tabulate
 from dateutil import relativedelta
 
 git = plumbum.local["git"]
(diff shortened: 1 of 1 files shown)


wiki/update_wiki.py
```diff
@@ -88,13 +86,21 @@ def generate_home_wiki_page(wiki_dir: Path, repository: str) -> None:
                 year_month_date,
                 year_month_date + relativedelta.relativedelta(day=31),
             )
-            monthly_line = f"| [`{year_month}`](./{year_month}) | {builds: <6} | {images: <6} | {commits_url: <95} |\n"
```
mathbunnyru, line 91: This hardcode won't look good in forks

wiki/update_wiki.py
```diff
@@ -103,12 +109,14 @@ def generate_home_wiki_page(wiki_dir: Path, repository: str) -> None:
 def update_monthly_wiki_page(
     wiki_dir: Path, year_month: str, build_history_line: str
 ) -> None:
-    MONTHLY_PAGE_HEADER = f"""\
-# Images built during {year_month}
+    MONTHLY_PAGE_HEADER = textwrap.dedent(
+        f"""\
+        # Images built during {year_month}
```
mathbunnyru, line 114: This table didn't and won't look nice, because we're appending lines to the existing file and the length is variable

PR #2249, opened 2025-03-10
Do not compress images in `actions/upload-artifact`

mathbunnyru:
## Describe your changes

We already use zst to compress our images, so it shouldn't be necessary to compress on upload.
https://github.com/actions/upload-artifact?tab=readme-ov-file#inputs
`For large files that are not easily compressed, a value of 0 is recommended for significantly faster uploads.`


mathbunnyru:
Without compression:
https://github.com/jupyter/docker-stacks/actions/runs/13756405308/job/38464739116?pr=2249
`Final size is 3187269994 bytes`

With compression:
https://github.com/jupyter/docker-stacks/actions/runs/13754574729/job/38460313324
`Final size is 3185148832 bytes`

The difference in size is around 0.1%, which means compression doesn't make sense at all.

And we get faster uploads.

PR #2252, opened 2025-03-12
Make taggers manifest functions

mathbunnyru:


mathbunnyru:
Free functions work as good as `Class.staticmethod`, but are much shorter.

mathbunnyru:
I checked locally and created files stayed the same, there was only a slight change in the command output (due to changes from CamelCase to snake_case).

PR #2253, opened 2025-03-13
Allow mamba ssl ca store to be set

jabbera:
## Describe your changes

mamba doesn't respect the normal SSL configuration environment variables.  This allows the configuration of this
at build time. This way someone building behind a corporate proxy doesn't need to intject and intermediate image
between the docker-stacks-foundation image and the base-notebook image.

This PR is not complete but before update docs/tests I wanted to see if it would be accepted as a feature.

- [X] I have performed a self-review of my code

### DIFF
--- a/images/base-notebook/Dockerfile
+++ b/images/base-notebook/Dockerfile
@@ -7,7 +7,7 @@ FROM $BASE_IMAGE
 
 # mamba doesn't respect and of the normal TLS environment variables REQUESTS_CA_BUNDLE, SSL_CERT_FILE, CURL_CA_BUNDLE
 # This causes a problem when building images behind a corporate proxy even when the images have the certificates installed
-# mamba config set ssl_verify "$SSL_CERT_FILE"; needs to be called. 
+# mamba config set ssl_verify "$SSL_CERT_FILE"; needs to be called.
 ARG SSL_CERT_FILE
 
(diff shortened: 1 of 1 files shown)


mathbunnyru:
Thank you for the PR.

May I ask you, are you building our images on your own or just build an image based on one of our images?

In the first case, I think since you're building your own set of images, having this small diff in your fork should be fine, I suppose?
In the second case you could set the option in the image you're building and it should work fine for you.

I honestly don't know, if the feature you're proposing is often and if it is worth having it by default.

jabbera:
I'm building my own images but using your dockerfiles. We pin versions of libraries and install a bunch of custom stuff (like certificates!) I have no issue maintaining the patch. I just thought others might want this because the solution is not obvious.

mathbunnyru:
> I'm building my own images but using your dockerfiles. We pin versions of libraries and install a bunch of custom stuff (like certificates!) I have no issue maintaining the patch. I just thought others might want this because the solution is not obvious.

Got it, thanks for the clarification.
To be honest, this is a first time I see someone having this problem .

I will close the PR for now, but if someone reads this and has the same/similar issue, please, let us know - if this is a common problem, we'll deal with it.

In any case, thanks for this PR, @jabbera, it will be useful to people having the same problem.

PR #2254, opened 2025-03-20
Refactor healthcheck tests to use one function

mathbunnyru:

### DIFF
--- a/tests/by_image/base-notebook/test_healthcheck.py
+++ b/tests/by_image/base-notebook/test_healthcheck.py
@@ -11,6 +11,32 @@ from tests.utils.tracked_container import TrackedContainer
 LOGGER = logging.getLogger(__name__)
 
 
+def get_healthy_status(
+    container: TrackedContainer,
+    env: list[str] | None,
+    cmd: list[str] | None,
+    user: str | None,
+) -> str:
+    running_container = container.run_detached(
+        tty=True,
+        environment=env,
+        command=cmd,
+        user=user,
+    )
+
+    # giving some time to let the server start
+    finish_time = time.time() + 10
+    sleep_time = 1
+    while time.time() < finish_time:
+        time.sleep(sleep_time)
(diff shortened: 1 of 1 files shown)


tests/by_image/base-notebook/test_healthcheck.py
```diff
@@ -11,6 +11,32 @@
...
+
+    # giving some time to let the server start
+    finish_time = time.time() + 10
+    sleep_time = 1
+    while time.time() < finish_time:
+        time.sleep(sleep_time)
+
+        status = get_health(running_container)
+        if status == "healthy":
+            return status
```
mathbunnyru, line 35: Fast exit in case get_health reports the status is healthy

tests/by_image/base-notebook/test_healthcheck.py
```diff
@@ -145,18 +141,6 @@ def test_not_healthy(
...
     cmd: list[str] | None,
 ) -> None:
-    running_container = container.run_detached(
-        tty=True,
-        environment=env,
-        command=cmd,
-    )
-
-    # giving some time to let the server start
-    finish_time = time.time() + 5
```
mathbunnyru, line 155: `5` is actually a bug here, because this way it tests nothing - if I put above, it's not enough to healthcheck to become green in tests

tests/by_image/base-notebook/test_healthcheck.py
```diff
@@ -11,6 +11,32 @@
...
+    running_container = container.run_detached(
+        tty=True,
+        environment=env,
+        command=cmd,
+        user=user,
+    )
+
+    # giving some time to let the server start
+    finish_time = time.time() + 10
+    sleep_time = 1
```
mathbunnyru, line 29: Sleeping 1 second instead of 0.1, because otherwise we produce too much logs when running `pytest` directly

PR #2206, opened 2025-01-29
Test server listening on IPv4/IPv6

manics:
## Describe your changes

Tests jupyter-server listen on IPv4 and IPv6 interfaces if available

## Issue ticket if applicable

This is a test for https://github.com/jupyter/docker-stacks/pull/2204

## Checklist (especially for first-time contributors)

- [ ] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests
- [ ] I will try not to use force-push to make the review process easier for reviewers
- [ ] I have updated the documentation for significant changes

PR #2255, opened 2025-03-20
Test server listening on IPv4/IPv6

mathbunnyru:

### DIFF
--- a/.github/actions/create-dev-env/action.yml
+++ b/.github/actions/create-dev-env/action.yml
@@ -16,10 +16,6 @@ runs:
       shell: bash
 
     - name: Set Up Docker 🐳
-      id: docker-setup
       uses: docker/setup-docker-action@b60f85385d03ac8acfca6d9996982511d8620a19 # v4.3.0
-
-outputs:
-  docker-sock:
-    description: "Docker socket path"
-    value: ${{ steps.docker-setup.outputs.sock }}
+      with:
+        set-host: true
--- a/.github/actions/create-dev-env/action.yml
+++ b/.github/actions/create-dev-env/action.yml
@@ -16,10 +16,10 @@ runs:
       shell: bash
 
     - name: Set Up Docker 🐳
-      id: docker
+      id: docker-setup
       uses: docker/setup-docker-action@b60f85385d03ac8acfca6d9996982511d8620a19 # v4.3.0
 
 outputs:
   sock:
     description: "Docker socket path"
-    value: ${{ steps.docker.outputs.sock }}
+    value: ${{ steps.docker-setup.outputs.sock }}
--- a/.github/actions/create-dev-env/action.yml
+++ b/.github/actions/create-dev-env/action.yml
@@ -17,3 +17,7 @@ runs:
 
     - name: Set Up Docker 🐳
       uses: docker/setup-docker-action@b60f85385d03ac8acfca6d9996982511d8620a19 # v4.3.0
+
+    - name: Output Docker Version
+      run: docker --version
+      shell: bash
--- a/.github/actions/create-dev-env/action.yml
+++ b/.github/actions/create-dev-env/action.yml
@@ -15,6 +15,10 @@ runs:
         pip install --upgrade -r requirements-dev.txt
       shell: bash
 
+    # We need to have a recent docker version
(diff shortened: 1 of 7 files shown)


mathbunnyru:
All the credit goes to @manics and https://github.com/jupyter/docker-stacks/pull/2206

Just wanted to test a few different things here.

mathbunnyru:
The error we get is `could not find an available, non-overlapping IPv6 address pool among the defaults to assign to the network `

It was reported here: https://github.com/moby/moby/issues/41438
Which was marked as duplicate of: https://github.com/moby/moby/pull/47768
Which was fixed in: https://github.com/moby/moby/releases/tag/v27.0.1, `released this Jun 25, 2024`

Ubuntu 24.04 uses `Docker Client 26`: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md
And `Docker Engine Version 26.* will be updated to Docker Engine Version 28* from 2025-05-09`: https://github.com/actions/runner-images/issues/11766

tests/by_image/base-notebook/data/check_listening.py
```diff
@@ -0,0 +1,51 @@
...
+    # Give some time for server to start
+    finish_time = time.time() + 10
+    sleep_time = 1
+    while time.time() < finish_time:
+        time.sleep(sleep_time)
+        try:
+            requests.get("http://localhost:8888/api")
+            break
+        except requests.RequestException:
+            pass
```
mathbunnyru, line 20: There is a bug here - we fall through no matter if requests was successful or not, will fix it

mathbunnyru:
When running from python still uses old version `'Name': 'Engine', 'Version': '26.1.3'`

manics:
Thanks for following this up, been busy with too many other things!

mathbunnyru:
`[gw1] [ 98%] PASSED tests/by_image/base-notebook/test_ips.py::test_ipv46`

Yep, made it work, now it's time to make it work for other tests as well 😃

tests/utils/get_container_health.py
```diff
@@ -4,7 +4,6 @@
 from docker.models.containers import Container
 
 
-def get_health(container: Container) -> str:
-    api_client = docker.APIClient()
```
mathbunnyru, line 8: This is funny coincidence.
I refactored code around health check earlier today 😃 

Didn't notice this bug.
We're using separate API client here, not the one from the `docker_client` fixture.

That's why it was impossible to make both healthcheck and ipv6 tests work.
Passing client as param and using it should fix the problem

.github/actions/create-dev-env/action.yml
```diff
@@ -14,3 +14,8 @@ runs:
         pip install --upgrade pip
         pip install --upgrade -r requirements-dev.txt
       shell: bash
+
+    - name: Set Up Docker 🐳
+      uses: docker/setup-docker-action@b60f85385d03ac8acfca6d9996982511d8620a19 # v4.3.0
+      with:
+        set-host: true
```
mathbunnyru, line 21: This magic option changes anything - I didn't actually need to change other code around docker client

mathbunnyru:
`x86_64-base` works 🎉

consideRatio:
Wieeee nice debugging effort into this @mathbunnyru!!!!

mathbunnyru:
@manics @consideRatio may I ask you to review this change?

Fun fact - I didn't even read what code by @manics about ipv6 does while doing all this 😆

@manics please, also tell me if you're ok if I just squash merge this PR as is (so the commit will be under my name).
If not, please, let me know what I should do - in any case, I don't want to take someone else's work and tell that it's mine.

mathbunnyru:
> Wieeee nice debugging effort into this @mathbunnyru!!!!

Thanks!
As you can see, it was quite a journey 😃 

It's nice that when GitHub runners update their Docker version, we can just remove setup action, and everything else will stay the same (no changes to code at all).

consideRatio (approved):
This seems right! Thank you @manics and @mathbunnyru!!

mathbunnyru:
Thanks!

All the images build and test just fine.
Rewrote check_listening.py a little bit to use one function for both ipv4 and ipv6, will merge as is.

PR #2256, opened 2025-03-21
Refactor TrackedContainer run_detached/exec_cmd functions

mathbunnyru:


tests/by_image/base-notebook/test_container_options.py
```diff
@@ -15,13 +15,13 @@
 def test_cli_args(container: TrackedContainer, http_client: requests.Session) -> None:
     """Image should respect command line args (e.g., disabling token security)"""
     host_port = find_free_port()
-    running_container = container.run_detached(
```
mathbunnyru, line 18: There is no need to return anything from `run_detached` - TrackedContainer has a container inside and it's better to provide specific functions we need (this way we're encapsulating docker functions we use)

tests/by_image/base-notebook/test_container_options.py
```diff
@@ -15,13 +15,13 @@
...
+    container.run_detached(
         command=["start-notebook.py", "--IdentityProvider.token=''"],
         ports={"8888/tcp": host_port},
     )
     resp = http_client.get(f"http://localhost:{host_port}")
     resp.raise_for_status()
-    logs = running_container.logs().decode()
+    logs = container.get_logs()
```
mathbunnyru, line 24: This is used often and it is cleaner to have a separate method

tests/by_image/base-notebook/test_container_options.py
```diff
@@ -47,8 +47,7 @@ def test_nb_user_change(container: TrackedContainer) -> None:
     )
     command = f'stat -c "%F %U %G" /home/{nb_user}/.jupyter'
     expected_output = f"directory {nb_user} users"
-    exec_result = running_container.exec_run(command, workdir=f"/home/{nb_user}")
-    output = exec_result.output.decode().strip("\n")
```
mathbunnyru, line 51: There were some `.strin("\n")` - I don't think they are actually needed, so let's try to remove them

tests/by_image/base-notebook/test_healthcheck.py
```diff
@@ -32,11 +32,11 @@ def get_healthy_status(
     while time.time() < finish_time:
         time.sleep(sleep_time)
 
-        status = get_health(running_container, docker_client)
+        status = get_health(container, docker_client)
         if status == "healthy":
             return status
 
-    return get_health(running_container, docker_client)
+    return status
```
mathbunnyru, line 39: Already have value above, no need to get it again

tests/by_image/base-notebook/test_ips.py
```diff
@@ -38,13 +38,9 @@ def test_ipv46(container: TrackedContainer, ipv6_network: str) -> None:
...
     cont_data_dir = "/home/jovyan/data"
     LOGGER.info("Testing that server is listening on IPv4 and IPv6 ...")
-    running_container = container.run_detached(
+    container.run_detached(
         network=ipv6_network,
         volumes={str(host_data_dir): {"bind": cont_data_dir, "mode": "ro,z"}},
         tty=True,
     )
-
-    command = ["python", f"{cont_data_dir}/check_listening.py"]
```
mathbunnyru, line 47: For simplicity made `cmd` always a string.
python docker allows both list or str, having str loogs good for us.

Additional improvement - if something goes wrong, copy-paste command will be easier

tests/by_image/base-notebook/test_ips.py
```diff
@@ -38,13 +38,9 @@ def test_ipv46(container: TrackedContainer, ipv6_network: str) -> None:
...
-    exec_result = running_container.exec_run(command)
-    LOGGER.info(exec_result.output.decode())
-    assert exec_result.exit_code == 0
```
mathbunnyru, line 50: No need to check any of this, because it's all part of new method

tests/by_image/docker-stacks-foundation/test_packages.py
```diff
@@ -109,25 +109,16 @@ def get_package_import_name(package: str) -> str:
     return PACKAGE_MAPPING.get(package, package)
 
 
-def _check_import_package(
```
mathbunnyru, line 112: With new method, this becomes one simple call, so removed some methods here

tests/utils/conda_package_helper.py
```diff
@@ -41,58 +40,36 @@ class CondaPackageHelper:
...
+        self.container = container
         LOGGER.info(f"Starting container {container.image_name} ...")
-        return container.run_detached(
+        container.run_detached(
             tty=True,
             command=["bash", "-c", "sleep infinity"],
         )
 
-    @staticmethod
-    def _conda_export_command(from_history: bool) -> list[str]:
```
mathbunnyru, line 62: In my opinion this adds complexity, but doesn't add too much value, so removed this method

tests/utils/conda_package_helper.py
```diff
@@ -41,58 +40,36 @@ class CondaPackageHelper:
...
-                )
+            env_export = self.container.exec_cmd(
+                "mamba env export --no-build --json --from-history"
             )
+            self.requested = CondaPackageHelper._parse_package_versions(env_export)
         return self.requested
 
-    def _execute_command(self, command: list[str]) -> str:
-        """Execute a command on a running container"""
-        exec_result = self.running_container.exec_run(command, stderr=False)
```
mathbunnyru, line 93: Removed stderr here, because mamba was updated recently and now the output is clean (as it was before)

tests/utils/get_container_health.py
```diff
@@ -1,9 +1,11 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import docker
-from docker.models.containers import Container
 
+from tests.utils.tracked_container import TrackedContainer
 
-def get_health(container: Container, client: docker.DockerClient) -> str:
```
mathbunnyru, line 7: Once again, let's not use `docker.models.containers` unless we really have to

tests/utils/tracked_container.py
```diff
@@ -19,22 +19,18 @@ class TrackedContainer:
         Docker client instance
     image_name: str
         Name of the docker image to launch
-    **kwargs: dict, optional
```
mathbunnyru, line 22: We only construct TrackedContainer in one place, it doesn't make sense to allow this customization, and we simply add `detach=True` in `run_detached`

docs/maintaining/tagging_examples/docker_runner.py
```diff
@@ -3,4 +3,4 @@
 from tagging.utils.docker_runner import DockerRunner
 
 with DockerRunner("ubuntu") as container:
-    DockerRunner.exec_cmd(container, cmd="env", print_output=True)
```
mathbunnyru, line 6: Easier to always print output

PR #2260, opened 2025-03-22
Use tty for running docker commands by default

mathbunnyru:

### DIFF
--- a/tests/by_image/docker-stacks-foundation/test_user_options.py
+++ b/tests/by_image/docker-stacks-foundation/test_user_options.py
@@ -175,6 +175,8 @@ def test_set_uid(container: TrackedContainer) -> None:
     Additionally, verify that "--group-add=users" is suggested in a warning to restore
     write access.
     """
+    # This test needs to have tty disabled, the reason is explained here:
+    # https://github.com/jupyter/docker-stacks/pull/2260#discussion_r2008821257
     logs = container.run_and_wait(
         timeout=5, no_warnings=False, user="1010", command=["id"], tty=False
     )
--- a/tests/by_image/docker-stacks-foundation/test_user_options.py
+++ b/tests/by_image/docker-stacks-foundation/test_user_options.py
@@ -176,10 +176,7 @@ def test_set_uid(container: TrackedContainer) -> None:
     write access.
     """
     logs = container.run_and_wait(
-        timeout=5,
-        no_warnings=False,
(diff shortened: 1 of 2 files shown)


mathbunnyru:
I first made our tests work fine without tty (and fixed a bug in our start script) and unified tests, but now I want to have tty by default:
- sometimes we have different results with tty and without it, so it's nice to have the tests which work fine in both cases and to control from one place if tty is enabled (unfortunately, before, TTY flag was quite random)
- by default I want to have tty enabled: the output is coloured, which makes debugging much easier when something fails

tests/by_image/docker-stacks-foundation/test_user_options.py
```diff
@@ -176,10 +176,7 @@ def test_set_uid(container: TrackedContainer) -> None:
     write access.
     """
     logs = container.run_and_wait(
-        timeout=5,
-        no_warnings=False,
-        user="1010",
-        command=["id"],
+        timeout=5, no_warnings=False, user="1010", command=["id"], tty=False
```
mathbunnyru, line 179: This is the only test which fails with tty enabled (I tested base-notebook locally).
It also fails when running manually:
`docker run -it --user 1010 quay.io/jupyter/base-notebook:latest id`
If I remove `-t`, then it gives expected output.

The source of /usr/local/bin/before-notebook.d/10activate-conda-env.sh unfortunately doesn't work

mathbunnyru: Got it, I removed the script in the image, and then ran it's content manually (`eval "$(conda shell.bash hook)"`) and in the end it gives:
```
`Would you like conda to send this report to the core maintainers? [y/N]:`

So in a non-interactive mode it just finishes, but in interactive it fails after timeout.

mathbunnyru: And it fails after 40 seconds which is more than our timeout, that's why test reports as failed

https://github.com/conda/conda/blob/4691c2826d5282637674de754b504c2f045e23d5/conda/exception_handler.py#L285

mathbunnyru: Setting report_errors in config won't work here, because in this test the `/home/jovyan` is not accessible.

So the best we can do is to put a link why we need to disable tty there, at least now I know the reason.

PR #2261, opened 2025-03-23
Improve logs around running docker

mathbunnyru:

### DIFF
--- a/tagging/utils/docker_runner.py
+++ b/tagging/utils/docker_runner.py
@@ -22,11 +22,10 @@ class DockerRunner:
         self.docker_client: docker.DockerClient = docker_client
 
     def __enter__(self) -> Container:
-        LOGGER.info(f"Creating container for image {self.image_name} ...")
+        LOGGER.info(f"Creating a container for the image: {self.image_name} ...")
+        default_kwargs = {"detach": True, "tty": True}
         self.container = self.docker_client.containers.run(
-            image=self.image_name,
-            command=self.command,
-            detach=True,
+            image=self.image_name, command=self.command, **default_kwargs
         )
         LOGGER.info(f"Container {self.container.name} created")
         return self.container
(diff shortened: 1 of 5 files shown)


tagging/utils/docker_runner.py
```diff
@@ -22,11 +22,10 @@ def __init__(
         self.docker_client: docker.DockerClient = docker_client
 
     def __enter__(self) -> Container:
-        LOGGER.info(f"Creating container for image {self.image_name} ...")
+        LOGGER.info(f"Creating a container for the image: {self.image_name} ...")
+        default_kwargs = {"detach": True, "tty": True}
```
mathbunnyru, line 26: This class is more-or-less like TrackedContainer, so let's have same options here

tagging/utils/docker_runner.py
```diff
@@ -48,6 +47,6 @@ def exec_cmd(container: Container, cmd: str) -> str:
         exec_result = container.exec_run(cmd)
         output = exec_result.output.decode().rstrip()
         assert isinstance(output, str)
-        LOGGER.info(f"Command output: {output}")
+        LOGGER.debug(f"Command output: {output}")
```
mathbunnyru, line 50: Sometimes the output is huge, and in most cases it is not useful

tests/pytest.ini
```diff
@@ -1,7 +1,7 @@
 [pytest]
 addopts = -ra --color=yes
 log_cli = 1
-log_cli_level = DEBUG
+log_cli_level = INFO
```
mathbunnyru, line 4: One can always use DEBUG if needed

tests/utils/conda_package_helper.py
```diff
@@ -41,8 +41,6 @@ class CondaPackageHelper:
 
     def __init__(self, container: TrackedContainer):
         self.container = container
-
-        LOGGER.info(f"Starting container {self.container.image_name} ...")
```
mathbunnyru, line 45: self.container.run_detached is going to log the same thing

tests/utils/conda_package_helper.py
```diff
@@ -101,7 +99,7 @@ def available_packages(self) -> dict[str, set[str]]:
             )
             # Keeping command line output since `mamba search --outdated --json` is way too long ...
             self.available = CondaPackageHelper._extract_available(
-                self.container.exec_cmd("mamba search --outdated --quiet")
+                self.container.exec_cmd("conda search --outdated --quiet")
```
mathbunnyru, line 102: `mamba search --outdated --quiet` doesn't work at all :(

PR #2265, opened 2025-03-25
Reduce indirection when running docker commands

mathbunnyru:


tests/by_image/datascience-notebook/test_julia_datascience.py
```diff
@@ -1,8 +1,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
-from tests.utils.run_command import run_command
 from tests.utils.tracked_container import TrackedContainer
 
 
 def test_julia(container: TrackedContainer) -> None:
-    run_command(container, "julia --version")
```
mathbunnyru, line 8: Now, `run_and_wait` is good enough to be used directly, having `run_command` doesn't help in any way.

PR #2267, opened 2025-03-25
Make free_host_port a fixture

mathbunnyru:


tests/utils/find_free_port.py
```diff
@@ -1,12 +0,0 @@
-# Copyright (c) Jupyter Development Team.
-# Distributed under the terms of the Modified BSD License.
-import socket
-from contextlib import closing
-
-
-def find_free_port() -> str:
-    """Returns the available host port. Can be called in multiple threads/processes."""
-    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
```
mathbunnyru, line 9: In theory, if there is a delay between running this function and using this port by docker, another thread might reuse the same port.

I would like to have a safer implementation, where we guarantee that until the end of the test no one else will get this port.

PR #2270, opened 2025-03-26
Robust registry operations: adjust timelimit and use retries

mathbunnyru:

### DIFF
--- a/.github/workflows/docker-merge-tags.yml
+++ b/.github/workflows/docker-merge-tags.yml
@@ -18,7 +18,7 @@ on:
       timeout-minutes:
         description: Timeout in minutes
         type: number
-        default: 25
+        default: 10
     secrets:
       REGISTRY_USERNAME:
         required: true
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-merge-tags.yml
```diff
@@ -18,7 +18,7 @@ on:
       timeout-minutes:
         description: Timeout in minutes
         type: number
-        default: 25
```
mathbunnyru, line 21: These timeouts were too big and almost always it should end in ~6 minutes.
With retry 10 should be enough.

.github/workflows/docker-merge-tags.yml
```diff
@@ -67,9 +67,13 @@ jobs:
 
       - name: Merge tags for the images 🔀
         if: env.PUSH_TO_REGISTRY == 'true'
-        run: >
-          python3 -m tagging.apps.merge_tags
-          --image ${{ inputs.image }}
-          --variant ${{ inputs.variant }}
-          --tags-dir /tmp/jupyter/tags/
+        run: |
+          python3 -m tagging.apps.merge_tags \
```
mathbunnyru, line 71: A bit of code duplication, but I don't think that's a huge deal

mathbunnyru:
I'm a bit tired of rerunning workflows manually, so decided to add retries to workflows.
Thankfully, the code is written in a way, that there is no harm to run it once again (no matter if previous attempt failed or not).

mathbunnyru:
Retries definitely help: https://github.com/jupyter/docker-stacks/actions/runs/14131791270/job/39597506278

PR #2274, opened 2025-04-01
[FAST_BUILD] Apply and merge tags in the same place

mathbunnyru:


mathbunnyru:
These are the advantages:
1. It should work faster (We don't need to download images twice, and we're not trying to pull if not needed)
2. Less load on the registry (we're only pushing and not pulling at all when not needed)
3. Warn if the tag for one platform doesn't exist for another (in this build or at all)-
4. We're testing more in PRs: now we can be sure that we can at least create the docker manifest
5. All the registry related things happen in one place, which should make it easier to debug if something goes wrong.

PR #2276, opened 2025-04-03
Update custom recipes versions

mathbunnyru:


docs/using/recipe_code/docker-bake.custom-python.hcl
```diff
@@ -5,7 +5,7 @@ group "default" {
 target "foundation" {
     context = "https://github.com/jupyter/docker-stacks.git#main:images/docker-stacks-foundation"
     args = {
-        PYTHON_VERSION = "3.12"
+        PYTHON_VERSION = "3.13"
```
mathbunnyru, line 8: We provide python 3.12, and let's build 3.13 foundation + base + minimal as part of our CI.
Also renamed the file, so in future we won't have to do it

PR #2277, opened 2025-04-03
Force using keyword-only arguments where types are the same

mathbunnyru:

### DIFF
--- a/tests/utils/tracked_container.py
+++ b/tests/utils/tracked_container.py
@@ -1,7 +1,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import logging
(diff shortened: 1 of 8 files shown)


tests/utils/tracked_container.py
```diff
@@ -123,7 +124,7 @@ def get_warnings(logs: str) -> list[str]:
         return TrackedContainer._lines_starting_with(logs, "WARNING")
 
     @staticmethod
-    def _lines_starting_with(logs: str, pattern: str) -> list[str]:
+    def _lines_starting_with(logs: str, pattern: LiteralString) -> list[str]:
```
mathbunnyru, line 127: This is a nice way to fix the problem

PR #2278, opened 2025-04-03
Improve some tests

mathbunnyru:


mathbunnyru:
The failed test is unrelated.
It is extremely simple, it's doing `apt --version`
5 seconds were not enough, I will increase all timeouts to at least 10 seconds.

I would like to have as little build failures as possible.

mathbunnyru:
The failed test in [x86_64-all-spark](https://github.com/jupyter/docker-stacks/pull/2278#logs) might be spurious, because:
1. Tested worked just fine in aarch64
2. it is not timeout related
3. it was marked as spurious before, it seems that the issue is back

I will update this branch with other recent changed - if the CI is green, I will merge this and will add pytest-retry back

PR #2279, opened 2025-04-04
Use pytest-rerunfailures for flaky tests

mathbunnyru:


mathbunnyru:
There is an official plugin for flaky tests, so let's use it!
https://github.com/pytest-dev/pytest-rerunfailures/tree/master

PR #2282, opened 2025-04-11
Make docker tag-push depend on contributed recipes in CI

mathbunnyru:


mathbunnyru:
We have a build where all the images were built and tested and therefore pushed to the registry: https://github.com/jupyter/docker-stacks/actions/runs/14375821119
But quite a simple recipe fails.

We should not push our images if recipes like this fail, so recipes can be viewed as tests for our images.

PR #2284, opened 2025-04-13
Use proper BASE_IMAGE in contributed recipes for forks

mathbunnyru:

### DIFF
--- a/.github/workflows/contributed-recipes.yml
+++ b/.github/workflows/contributed-recipes.yml
@@ -65,7 +65,7 @@ jobs:
           variant: default
 
       # Not pulling the image, because it might be loaded from previous step or will be downloaded automatically
-      - name: Build recipe 🛠
+      - name: Build recipe with parent image 🛠
         if: ${{ matrix.parent-image != '' }}
         run: |
           docker build \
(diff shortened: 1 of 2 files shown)


mathbunnyru:
What happened:
1. I pinned `libxml2` in both `jupyter/docker-stacks` and my fork (almost simultaneously).
2. My fork started running `docker.yml` workflow, which then runs `contributed-recipes.yml` workflow
3. Instead of using fork built images, contributed recipes in my fork used images pushed to the jupyter registry (images with libxml2 pin were not yet pushed, so previous images were used)
4. I got a failure: https://github.com/mathbunnyru/docker-stacks/actions/runs/14419023194/job/40439441899

.github/workflows/contributed-recipes.yml
```diff
@@ -62,7 +66,30 @@ jobs:
...
       # Not pulling the image, because it might be loaded from previous step or will be downloaded automatically
       - name: Build recipe 🛠
-        run: docker build --rm --force-rm --tag my-custom-image -f ./${{ matrix.dockerfile }} ./
+        if: ${{ matrix.parent-image != '' }}
+        run: |
+          docker build \
+            --rm --force-rm \
+            --tag my-custom-image \
+            -f ./${{ matrix.dockerfile }} \
+            --build-arg BASE_IMAGE=${{ env.REGISTRY }}/${{ env.OWNER }}/${{ matrix.parent-image }} \
```
mathbunnyru, line 75: I don't think it's possible to write a ternary here, but if someone knows how do it, it would be great

PR #2285, opened 2025-04-13
Use tenacity.retry for merge_tags

mathbunnyru:

### DIFF
--- a/tagging/apps/merge_tags.py
+++ b/tagging/apps/merge_tags.py
@@ -3,10 +3,9 @@
 # Distributed under the terms of the Modified BSD License.
 import logging
 import os
-import time
-from collections.abc import Callable
 
 import plumbum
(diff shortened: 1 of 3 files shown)


tagging/apps/merge_tags.py
```diff
@@ -103,16 +101,10 @@ def merge_tags(
 ) -> None:
     LOGGER.info(f"Trying to merge tag: {merged_tag}")
 
-    # This allows to rerun the script without having to remove the manifest manually
-    try:
-        docker["manifest", "rm", merged_tag] & plumbum.FG
```
mathbunnyru, line 108: This was wrong and it has to be done before every call to `docker manifest create`. Even if `docker manifest create` failed, it could have created local manifest, which we have to delete before next attempt.

PR #2286, opened 2025-04-13
Retry login operation in docker-tag-push.yml

mathbunnyru:


mathbunnyru:
I've recently seen login operation fail (unfortunately, I wasn't able to find the run).

This change completes a list of changes made to ensure that our workflows need as least manual retriggers as possible - now all docker registry related operations have some sort of retry (it might be simple bash `||` or retrying inside `merge_tags.py`).

PR #2293, opened 2025-04-28
log hook messages to stderr

minrk:
## Describe your changes

- use `_log` in run-hooks.sh, like start.sh (needs to redefine it for unnittests only, could split into a common utility file)
- `_log` writes to stderr instead of stdout for easier output capture (piping, etc.)

In general, it's a good practice to avoid logging to stdout, since it means you can't capture command output with pipes, e.g.

```
docker run --rm -it quay.io/jupyter/base-notebook conda env export > base-env.yml
```

because it will include the 'uid: 1000`, etc. messages.

I looked through the tests, and it's tricky because the tests run with `tty: true`, in which case docker merges stdout and stderr, so can't test the splitting is correct. Running with `tty: false`, e.g. `docker run --rm -i pwd > stdout.txt 2>stderr.txt` does the right thing in a manual test. I'm just not 100% sure of the best way to add a test case for that with the current infrastructure.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 88
 select = C, E, F, W, B, B950
-extend-ignore = E203, E501, W503
+extend-ignore = E203, E501, E704, W503
--- a/tests/utils/tracked_container.py
+++ b/tests/utils/tracked_container.py
@@ -1,7 +1,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import logging
-from typing import Any, LiteralString
+from typing import Any, Literal, LiteralString, overload
 
 import docker
 from docker.models.containers import Container
@@ -54,9 +54,9 @@ class TrackedContainer:
         )
         LOGGER.info(f"Container {self.container.name} created")
 
-    def get_logs(self) -> str:
+    def get_logs(self, **kwargs: Any) -> str:
         assert self.container is not None
-        logs = self.container.logs().decode()
+        logs = self.container.logs(**kwargs).decode()
         assert isinstance(logs, str)
         return logs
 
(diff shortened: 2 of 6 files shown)


minrk:
Thanks! I added a way to split stdout and stderr and some tests to exercise this.

tests/utils/tracked_container.py
```diff
@@ -54,9 +54,9 @@ def run_detached(self, **kwargs: Any) -> None:
         )
         LOGGER.info(f"Container {self.container.name} created")
 
-    def get_logs(self) -> str:
+    def get_logs(self, **kwargs: Any) -> str:
```
mathbunnyru, line 57: Let's add `stdout` and `stderr` params with their default values, to make it easy for IDEs give suggestions and show that we only care about these params.

.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 88
 select = C, E, F, W, B, B950
-extend-ignore = E203, E501, W503
+extend-ignore = E203, E501, E704, W503
```
mathbunnyru, line 4: Do we need this? If it's just in one-two places, let's ignore it inline, without changing the config.

mathbunnyru (changes requested):
Overall, looks great, asked to change 2 small things.

.flake8
minrk: it's disabled by default and seems to be a fundamentally wrong rule that's incompatible with type annotations, so I don't think it should be enabled.

mathbunnyru: Ah, got it, makes sense, thanks for the explanation 👍

tests/utils/tracked_container.py
minrk: Good call, done.

mathbunnyru (approved):
🚀
Please, squash-merge when CI is green

PR #2295, opened 2025-05-07
Do not update permissions for user cache

peterjasc:
## Describe your changes
When building images on macOS Version 15.4.1
`docker build ./images/pyspark-notebook --platform linux/amd64`
returns 
```
#5 28.91 chgrp: changing group of '/home/jovyan/.cache/rosetta': Operation not permitted
#5 28.93 chmod: changing permissions of '/home/jovyan/.cache/rosetta': Operation not permitted
```

Instead of chaging permissions for cache, we should ignore it.

## Issue ticket if applicable
https://github.com/jupyter/docker-stacks/issues/2296

### DIFF
--- a/docs/using/common.md
+++ b/docs/using/common.md
@@ -256,6 +256,7 @@ You can use either `mamba`, `pip`, or `conda` (`mamba` is recommended) to instal
 # the installation
 mamba install --yes some-package && \
     mamba clean --all -f -y && \
+    (rm -r /home/${NB_USER}/.cache/rosetta || true) && \
     fix-permissions "${CONDA_DIR}" && \
     fix-permissions "/home/${NB_USER}"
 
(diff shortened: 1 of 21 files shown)


peterjasc:
Maybe removing `/home/jovyan/.cache/rosetta` is better. These files there because the mac is emulating amd64 and therefore are not required during running the image. If they were needed for macOS run, the OS should create them

mathbunnyru (changes requested):
I don't like this implementation because it creates the impression that these files are created after `mamba install` and somehow related to mamba.
But that's not the case, as far as I understand.

I like the approach here: before switching from root to some user we delete the folder.
At least it will limit the number of places where you have to make this change (and our recipes won't change), and this is more clear if someone also has this problem: https://github.com/docker/for-mac/issues/7440#issuecomment-2463247231

mathbunnyru:
@peterjasc hi, could you please work on this?

PR #2300, opened 2025-05-24
Install Spark 4 release version

mathbunnyru:


tests/shared_checks/nbconvert_check.py
```diff
@@ -17,7 +17,7 @@ def check_nbconvert(
     no_warnings: bool = True,
 ) -> str:
     """Check if nbconvert is able to convert a notebook file"""
-    cont_data_file = "/home/jovyan/data/" + host_file.name
+    cont_data_file = "/home/jovyan/" + host_file.name
```
mathbunnyru, line 20: For some reason, this test requires the directory where the file is to be writeable now.
I don't have time to dig into why this happens, so I just changed the mounted file location

PR #2352, opened 2025-11-18
Enable CUDA build for ARM64

dictcp:
## Enable CUDA build for ARM64

Since there are ARM64 platform like DGX spark, we need CUDA build for ARM64.
(give DGX spark case, given it requires CUDA capacity of 12.1, so latest CUDA is fine)

### DIFF
--- a/.github/workflows/docker-tag-push.yml
+++ b/.github/workflows/docker-tag-push.yml
@@ -48,6 +48,7 @@ jobs:
           image: ${{ inputs.image }}
           variant: ${{ inputs.variant }}
           platform: aarch64
+        if: ${{ !contains(inputs.variant, 'cuda11') }}
 
       - name: Download x86_64 image tar and apply tags 🏷
         uses: ./.github/actions/apply-single-tags


.github/workflows/docker-tag-push.yml
```diff
@@ -48,7 +48,6 @@ jobs:
           image: ${{ inputs.image }}
           variant: ${{ inputs.variant }}
           platform: aarch64
-        if: ${{ !contains(inputs.variant, 'cuda') }}
```
mathbunnyru: You either have to make it work for pytorch-cuda11, or do not download it here, otherwise it will fail

mathbunnyru (changes requested):

mathbunnyru:
> @mathbunnyru I think you need to change the first item in the checklist from
> 
> `I have performed a self-review of my code`
> 
> to
> 
> `I, or Copilot on my behalf, have performed a review of the code`

I would probably say that I still want people to take a look at their PRs, and essentially, this check should be marked and done before submitting the PR (and Copilot can't do that).

mathbunnyru:
Failing CI is due to an attempt to merge commit-hash-based tags, which are not pushed from PRs. Will have to fix it later, but this change seems good to go. (this issue is unrelated to this PR at all)

mathbunnyru (approved):

PR #2357, opened 2025-11-24
Add recipe on running Jupyter Docker Stacks with Singularity

fbnrst:
## Describe your changes

This adds a recipe on starting the images using Singularity instead of Docker.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -11,7 +11,7 @@
     }
   },
 
-  "postCreateCommand": "pre-commit install",
+  "postCreateCommand": "pre-commit install --install-hooks",
 
   "customizations": {
     "vscode": {


docs/using/recipes.md
```diff
@@ -520,3 +520,17 @@ they may be explained in the "Installation instructions" section of the Download
...
+You can also start Jupyter Docker Stacks containers using **Singularity** instead of Docker. For example:
+
+```bash
+singularity run --bind "${PWD}:/home/${USER}/work" --containall docker://quay.io/jupyter/datascience-notebook:2025-03-14
+```
+
+- `--bind "${PWD}:/home/${USER}/work` binds your current directory into the container at `/home/$USER/work`. Here, $USER is your username.
+- `--containall` runs the container in a fully isolated environment, ignoring most of the host’s environment and filesystem except explicitly bound paths, ensuring that Python libraries in the
+  user’s home directory do not interfere with the container’s libraries.
+- Replace `quay.io/jupyter/datascience-notebook:2025-03-14` with the desired stack and tag.
```
mathbunnyru: I would remove this, it is quite obvious :)

docs/using/recipes.md
```diff
@@ -520,3 +520,17 @@ they may be explained in the "Installation instructions" section of the Download
...
+
+## Running Jupyter Docker Stacks with Singularity
+
+You can also start Jupyter Docker Stacks containers using **Singularity** instead of Docker. For example:
+
+```bash
+singularity run --bind "${PWD}:/home/${USER}/work" --containall docker://quay.io/jupyter/datascience-notebook:2025-03-14
+```
+
+- `--bind "${PWD}:/home/${USER}/work` binds your current directory into the container at `/home/$USER/work`. Here, $USER is your username.
```
fbnrst: add trailing `"`
```suggestion
- `--bind "${PWD}:/home/${USER}/work"` binds your current directory into the container at `/home/$USER/work`. Here, $USER is your username.
```

docs/using/recipes.md
mathbunnyru: > Maybe I should also mention in the recipe, that the default behavior of Singularity is to mount the users home directory.

I think this is a good idea though 🙂

PR #2366, opened 2025-11-25
Test_mergify

fbnrst:
test mergify setup

- add some nonsense command that will make tests fail
- send this to mergify queue to test it is working


fbnrst:
sorry, wrong target branch

PR #2399, opened 2026-01-21
Pin protobuf version for tensoflow

xcompass:
## Describe your changes

TensorFlow 2.20.0 was built using protobuf version 5.28.3. And currently protobuf 6.31.1 was installed in the image. Then running TF, the following UserWarning messages are printed to users:

UserWarning: Protobuf gencode version 5.28.3 is exactly one major version older than the runtime version 6.31.1 at
tensorflow/core/framework/attr_value.proto. Please update the gencode to avoid compatibility violations in the next runtime release.

Related issue from upstream: https://github.com/tensorflow/tensorflow/issues/98980


mathbunnyru (changes requested):

mathbunnyru:
I'm so sorry, @xcompass, it seems our test suite doesn't work well with complicated expressions for versions.
Please, either fix how the version is being parsed, or we can hardcode the latest version of protobuf, which works well.

In either case, please test locally as well (let me know if you need help to set it up).

xcompass:
No worries @mathbunnyru . Any pointers how to set up test locally? I'm on mac. Thanks

xcompass:
Thanks @mathbunnyru. The parsing is fixed and tests passed locally.

One note, I had to use `DOCKER_HOST=unix://$HOME/.docker/run/docker.sock make test-all` as docker desktop doesn't expose DOCKER_HOST automatically on Mac.

Issue #2395, opened 2026-01-03
Clarification request: when to build images locally vs pull from Docker Hub


CoderLink06:
Hello, I am a student exploring this repository as a potential GSoC contributor.

While reading the README, I was unsure about when a user should build the Docker images locally versus simply pulling pre-built images from Docker Hub.

It would be helpful to have a short clarification or example in the README explaining:
- common scenarios for using Docker Hub images
- situations where building locally is recommended
- any trade-offs involved

I believe this would help new users and contributors get started more confidently.
Thank you.

mathbunnyru:
Hi @CoderLink06, thank you for reporting the issue. It's always great to see new people who want to use our images.

First of all, it's never `Docker Hub`, we only push images to `Quay.io` and all examples in the documentation use this registry as well.

I would say the criteria are quite simple:
1. We have a [Selecting](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html) section in the docs. If some image matches more or less closely with your needs, choose it.
2. If for some reason it's not enough, we have a [Building a custom set of images](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/custom-images.html) section, which describes customisation points of our images and various ways of building your own image(s).

Hope this helps.

If you know how to describe it better, I would appreciate any suggestions/PRs.

PR #2404, opened 2026-01-26
docs: clarify when to pull Quay.io image vs build locally

CoderLink06:
This PR adds a short clarification to help new users decide when to pull an existing image from Quay.io versus when to build a custom image locally.

It is based on the discussion in issue #2395 and aims to make the image selection workflow clearer for first-time users.

Changes:
- Added a brief "Pull vs build" guidance block to docs/using/selecting.md

### DIFF
--- a/docs/using/selecting.md
+++ b/docs/using/selecting.md
@@ -4,9 +4,10 @@
 - [Image Relationships](#image-relationships)
 - [Community Stacks](#community-stacks)
 
-> **Pull vs build: quick guidance**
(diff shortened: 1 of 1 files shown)


CoderLink06:
I added a small follow-up clarification to `docs/using/running.md` explaining host vs container port mapping for the `-p 10000:8888` example, since this seems to be a common source of confusion for new users.

Happy to adjust or remove if you’d prefer this to live elsewhere in the docs.

PR #1677, opened 2022-04-04
Add section "Using the Podman CLI"

eriksjolund:
* Add exmaple 4 that demonstrates how to run the container
  image docker.io/jupyter/r-notebook:6b49f3337709 with rootless podman.

Signed-off-by: Erik Sjölund <erik.sjolund@gmail.com>

PR #2439, opened 2026-05-07
Use podman keep-id options for uid and gid

eriksjolund:
Podman 4.3.0 and later support the syntax

```
--userns keep-id:uid=${uid},gid=${gid}
```

Use this syntax because fewer command-line arguments are then needed for the 
`podman run` command. 

Mention the troubleshooting tip. The tip is relevant for older podman versions before version 4.3.0.

Remove the sudo warning because the command does not map the container user to the root user.

__Side note 1:__
I added the Podman example in https://github.com/jupyter/docker-stacks/pull/1677

### DIFF
--- a/docs/using/running.md
+++ b/docs/using/running.md
@@ -131,13 +131,6 @@ uid=1000
 gid=100
 ```
 
-Set the shell variables _subuidSize_ and _subgidSize_ to the number of subordinate UIDs and GIDs, respectively.
-
-```bash
-subuidSize=$(( $(podman info --format "{{ range .Host.IDMappings.UIDMap }}+{{.Size }}{{end }}" ) - 1 ))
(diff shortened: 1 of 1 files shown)


mathbunnyru:
@eriksjolund is this still a draft?

eriksjolund:
>  I will try not to use force-push to make the review process easier for reviewers

Oops, I missed this information. I did a force-push.

> @eriksjolund is this still a draft?

It's now ready for review.

mathbunnyru (changes requested):
> I was not able to test the new changes because I ran out of disk space.

Please test your changes, and if they work, update the PR description.

Unfortunately, I don't use podman at all, so I'm not able to verify this

eriksjolund:
I reinstalled the VM with more disk space so I was then able to verify that the changes are correct.
I've now updated the PR description.

mathbunnyru (approved):

mathbunnyru:
@eriksjolund Thanks for maintaining this part of the documentation 👍

Issue #2428, opened 2026-04-23
Work on OpenSSF Scorecard Report


mathbunnyru:
Right now it's 6.2 https://scorecard.dev/viewer/?uri=github.com%2Fjupyter%2Fdocker-stacks

### Are you a maintainer?

- [x] Yes, I am a maintainer

goelakash:
@mathbunnyru is there anything here that warrants special attention? I could maybe take a look.

goelakash:
I was going through the scorecard items with the lowest scores to prioritize work here. The CII-Best-Practices check is at zero because no badge has been registered yet.

I looked into it and it requires a maintainer to register the project at https://www.bestpractices.dev/ and fill out a self-assessment questionnaire. Most of the criteria (version control, automated tests, security policy, license, documentation) should already be met, so it's mainly a matter of going through the form.

@mathbunnyru would you be open to registering for this? Happy to help fill in details if needed.

mathbunnyru:
> I was going through the scorecard items with the lowest scores to prioritize work here. The CII-Best-Practices check is at zero because no badge has been registered yet.
> 
> I looked into it and it requires a maintainer to register the project at [bestpractices.dev](https://www.bestpractices.dev/) and fill out a self-assessment questionnaire. Most of the criteria (version control, automated tests, security policy, license, documentation) should already be met, so it's mainly a matter of going through the form.
> 
> [@mathbunnyru](https://github.com/mathbunnyru) would you be open to registering for this? Happy to help fill in details if needed.

Yes, I'll do it, thanks, but it will take a bit of time, since I'm pretty busy lately.
Thanks for working this issue, though!

goelakash:
@mathbunnyru 
A few branch protection settings that would improve the score for Branch Protection and Code-Review (Settings > Branches > main rule) could also lead to a quick impvements ("Include administrators", "Require review from Code Owners", "Require approval of the most recent reviewable push")

The score has moved from 6.2 --> 7.8 thanks to the recent PRs merged :D

goelakash:
Images are currently published unsigned, which is why the Signed-Releases check scores zero. Opened #2455 to track adding cosign keyless signing to the publish workflow.

mathbunnyru:
[...]

PR #2445, opened 2026-05-29
Pin devcontainer base image to digest

goelakash:
## Summary
- Pins the `.devcontainer/Dockerfile` base image to its sha256 digest
- Improves the Pinned-Dependencies check in the OpenSSF Scorecard

Ref: #2428

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/devcontainers/python:3.13
+FROM mcr.microsoft.com/devcontainers/python:3.13@sha256:e797f7e02a8adf1c6d375a4873591205cc867a2d68083506e7f3aa89822f2026
 
 COPY requirements-dev.txt /tmp/requirements-dev.txt
 COPY docs/requirements.txt /tmp/requirements-docs.txt
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -37,3 +37,9 @@ updates:
       interval: weekly
     commit-message:
       prefix: "[FAST_BUILD] "
+  - package-ecosystem: docker
+    directory: .devcontainer/
+    schedule:
+      interval: weekly
+    commit-message:
+      prefix: "[FAST_BUILD] "


goelakash:
@mathbunnyru made a minimal change to see if it improves the SSF scorecard number

consideRatio (commented):
What happens if they publish an updated version of the 3.13 tag, will this no longer work to build, or will it ignore the tag and use the digest, which may be around still?

Edit: i think the tag is practically ignored when pulling once there is a digest, you could do name:adfjfjekfk@digest and it would work also!

goelakash:
@consideRatio makes sense, maybe we would need it to be updated on a regular cadence then too. Added a dependabot config to update the digest. 
The downside I see here is to regularly merge those PRs.

consideRatio:
> maybe we would need it to be updated on a regular cadence then too. Added a dependabot config to update the digest.

:heart: yes - anything else is game over from a maintenance perspective - even weekly bumps may be too much work. @mathbunnyru is pulling a lot of weight to maintain this repo currently, it will be a judgement call on what to do here relating also to work effort - is this pinning worth the maintenance burden, is weekly bumps the right interval, etc.

Great to see you @goelakash, and thank you again @mathbunnyru for your great work in this project!!

goelakash:
Yeah, totally open to that discussion too. Lets keep this PR open and see if other people have any opinions regarding this. The main benefit i see is accidentally preventing shipping any particular dependency with known bugs and which is recalled - which happens maybe 2-3 times a year typically in open source, but mostly in JS land from experience.

mathbunnyru (approved):
Let's give it a try

@goelakash, your commits do not seem to be signed (in all your PRs). If you could fix that in all the PRs, that would be great.

mathbunnyru:
@goelakash I merged without signed commits, but in the future, it would be great if you set it up for your git config

goelakash:
@mathbunnyru thank you, i am setting up the signing key for all the repos today.

mathbunnyru:
@goelakash great!
Thanks for the contributions 🎉

PR #2446, opened 2026-05-29
Add top-level permissions: {} to all workflows

goelakash:
## Summary
- Adds `permissions: {}` at the workflow level to all 13 workflow files; jobs already declare their own so behavior is unchanged
- Addresses the Token-Permissions check in the OpenSSF Scorecard

#2428


goelakash:
@mathbunnyru if you could review. I think this would be non-intrusive.

mathbunnyru (approved):

PR #2451, opened 2026-05-31
Add CodeQL workflow for static analysis

goelakash:
## Summary
- Adds a CodeQL workflow that scans Python code on every push to main, on PRs, and weekly
- Actions pinned to the same hash already used in scorecard.yml
- Improves the SAST check in the OpenSSF Scorecard

Ref: #2428

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,37 @@
+name: CodeQL
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+  schedule:
+    - cron: "0 6 * * 1"
+
+permissions: {}
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      security-events: write
+    strategy:
+      fail-fast: false
+      matrix:
+        language: ["python"]
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Initialize CodeQL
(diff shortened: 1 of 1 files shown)


mathbunnyru:
> CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled

I already enabled default CodeQL analysis in this repo - this is not purely a python project. For example, I want github workflows to be checked as well

goelakash:
Makes sense, closing since default setup already covers this. Thanks for the context!

mathbunnyru:
I think automatic CodeQL checks are not working very well, because they might not be launched on PRs not touching relevant languages.
This gives warnings like:

> Code scanning is waiting for results from CodeQL for the commits

, whcih make me bypass it.

So, maybe let's make it explicit (as this PR does).
Could you please add a language for GitHub Actions/Workflows?
I'm not sure if Docker is supported and if it's worth having for us or would just create noise.

goelakash:
Reopened with `actions` added to the language matrix. Shell scripts and Dockerfiles are already covered by shellcheck/bashate and hadolint via pre-commit.

goelakash:
CI is failing with:

```
CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled
```

I think the default CodeQL setup needs to be disabled at the repo level before this workflow can upload results.  @mathbunnyru

mathbunnyru:
> CI is failing with:
> 
> ```
> CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled
> ```
> 
> I think the default CodeQL setup needs to be disabled at the repo level before this workflow can upload results. @mathbunnyru

Yes, sure, I was waiting for you to update the workflow for actions, so we have no CodeQL downtime :)

mathbunnyru (approved):

PR #2452, opened 2026-05-31
Improve logging in start.sh and run-hooks.sh

goelakash:
## Summary
- Refactors `_log` to accept level as first argument with timestamp output
- Adds convenience wrappers: `_log_info`, `_log_warn`, `_log_error`
- Output format: `LEVEL[YYYY-MM-DD HH:MM:SS] message`
- `_log` remains available for custom levels if needed
- Adds tests for timestamp format and `JUPYTER_DOCKER_STACKS_QUIET` behavior

Ref: #1532

### DIFF
--- a/images/docker-stacks-foundation/Dockerfile
+++ b/images/docker-stacks-foundation/Dockerfile
@@ -142,7 +142,7 @@ RUN set -x && \
     fix-permissions "/home/${NB_USER}"
 
 # Copy local files as late as possible to avoid cache busting
-COPY run-hooks.sh start.sh /usr/local/bin/
+COPY _docker_stacks_log.sh run-hooks.sh start.sh /usr/local/bin/
 
 # Configure container entrypoint
 ENTRYPOINT ["tini", "-g", "--", "start.sh"]
--- a/tests/by_image/base-notebook/test_start_container.py
+++ b/tests/by_image/base-notebook/test_start_container.py
@@ -42,24 +42,23 @@ def test_start_notebook(
         f"Test that the start-notebook.py launches the {expected_command} server from the env {env} ..."
     )
     container.run_detached(environment=env, ports={"8888/tcp": free_host_port})
-    # sleeping some time to let the server start
-    time.sleep(2)
+    # Wait for the server to be reachable (proves start-notebook.py finished launching)
+    if expected_start:
+        resp = http_client.get(f"http://localhost:{free_host_port}")
+        assert resp.status_code == 200, "Server is not listening"
+    else:
+        # For non-listening cases (e.g. jupyterhub-singleuser), give the script time to print
+        time.sleep(5)
     logs = container.get_logs()
     LOGGER.debug(logs)
-    # checking that the expected command is launched
     assert (
         f"Executing: {expected_command}" in logs
     ), f"Not the expected command ({expected_command}) was launched"
-    # checking errors and warnings in logs
     assert "ERROR" not in logs, "ERROR(s) found in logs"
     for exp_warning in expected_warnings:
         assert exp_warning in logs, f"Expected warning {exp_warning} not found in logs"
     warnings = TrackedContainer.get_warnings(logs)
     assert len(expected_warnings) == len(warnings)
-    # checking if the server is listening
-    if expected_start:
-        resp = http_client.get(f"http://localhost:{free_host_port}")
-        assert resp.status_code == 200, "Server is not listening"
 
 
 def test_tini_entrypoint(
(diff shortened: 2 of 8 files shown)


mathbunnyru (dismissed):

images/docker-stacks-foundation/Dockerfile
```diff
@@ -142,7 +142,7 @@ RUN set -x && \
     fix-permissions "/home/${NB_USER}"
 
 # Copy local files as late as possible to avoid cache busting
-COPY run-hooks.sh start.sh /usr/local/bin/
+COPY _log.sh run-hooks.sh start.sh /usr/local/bin/
```
mathbunnyru: Let's call this file `_docker_stacks_log.sh` (simply rename it everywhere) just in case someone already has `_log.sh` in `/usr/local/bin`?

goelakash: good callout - _log.sh is too generic.

mathbunnyru:
@goelakash, I would highly appreciate not using force push, as this makes it really difficult to see the incremental changes you made.

In the end, we always do a squash merge, so it'll be nice in the main branch.

goelakash:
@mathbunnyru makes sense, i was going to ask you if thats preferred or not

tests/by_image/base-notebook/test_start_container.py
```diff
@@ -42,24 +42,23 @@ def test_start_notebook(
         f"Test that the start-notebook.py launches the {expected_command} server from the env {env} ..."
     )
     container.run_detached(environment=env, ports={"8888/tcp": free_host_port})
-    # sleeping some time to let the server start
-    time.sleep(2)
+    # Wait for the server to be reachable (proves start-notebook.py finished launching)
+    if expected_start:
+        resp = http_client.get(f"http://localhost:{free_host_port}")
```
mathbunnyru, line 47: It's ran in `detached` just a line above, I don't think docker gives any guarantee that anything will happen in a reasonable time.
Or are you relying on `http_client.get` retries here?

tests/by_image/docker-stacks-foundation/test_logging.py
```diff
@@ -0,0 +1,66 @@
...
+def test_log_format_includes_timestamp(container: TrackedContainer) -> None:
+    """Log messages should include a timestamp and level."""
+    _, stderr = container.run_and_wait(
+        timeout=10,
+        user="root",
+        environment=["NB_USER=root", "NB_UID=0", "NB_GID=0"],
+        command=["echo", "done"],
+        split_stderr=True,
+    )
+    assert re.search(r"INFO\[\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\]", stderr)
```
mathbunnyru: Let's search for a bit more specific string, because this might happen accidentally from `JupyterLab` itself, for example

goelakash: yeah, good point

tests/by_image/base-notebook/test_start_container.py
goelakash: https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry

The `http_client` fixture in `tests/conftest.py` configures a `requests.Session` with `Retry(total=5, backoff_factor=1)` from `urllib3`, which gives delays of 0, 2, 4, 8, 16s between attempts (~30s total).

So `http_client.get()` waits for the server to come up before failing, which is what we rely on here.

goelakash:
@mathbunnyru Regarding the force-push - is the main concern usually that it breaks the review flow as inline-diffs are not shows by Github, or that we simply prefer to make incremental changes and maintain a clear view of past checkpoints?

if its the former, please +1 this suggestion I wrote:
https://github.com/orgs/community/discussions/197515#discussioncomment-17126870

tests/by_image/base-notebook/test_start_container.py
mathbunnyru: Nice, thanks 👍

mathbunnyru:
> @mathbunnyru Regarding the force-push - is the main concern usually that it breaks the review flow as inline-diffs are not shows by Github, or that we simply prefer to make incremental changes and maintain a clear view of past checkpoints?
> 
> if its the former, please +1 this suggestion I wrote: [github.com/orgs/community/discussions/197515#discussioncomment-17126870](https://github.com/orgs/community/discussions/197515#discussioncomment-17126870)

Both. Upvoted :)

mathbunnyru (approved):
I love this change, thanks!

Will merge when CI passes

goelakash:
@mathbunnyru looks like its blocked by codeql - not sure why

PR #2459, opened 2026-06-02
[FAST_BUILD] Add color to log output based on level

goelakash:
## Summary
- Colorizes ERROR (red), WARNING (yellow), and DEBUG (cyan) log messages automatically when stderr is a terminal
- No color when output is piped or redirected (standard tty detection via `-t 2`)
- Respects the `NO_COLOR` environment variable
- INFO messages remain uncolored to keep default output clean
- Updates the test utility to strip ANSI codes before matching level prefixes, so existing assertions work with or without color

Addresses #1532 (checkbox: "Based on the log level color can be added").

### DIFF
--- a/images/docker-stacks-foundation/_docker_stacks_log.sh
+++ b/images/docker-stacks-foundation/_docker_stacks_log.sh
@@ -5,8 +5,7 @@
 # The _log function is used for everything these scripts want to log.
 # It will always log errors and warnings but can be silenced for other messages
 # by setting the JUPYTER_DOCKER_STACKS_QUIET environment variable.
-# Colorizes WARNING and ERROR output when stderr is a terminal.
-# Respects NO_COLOR (https://no-color.org/).
+# Colorizes ERROR, WARNING, and DEBUG output when stderr is a terminal.
 _log () {
     local level="${1}"
     case "${level}" in
--- a/images/docker-stacks-foundation/_docker_stacks_log.sh
+++ b/images/docker-stacks-foundation/_docker_stacks_log.sh
@@ -5,6 +5,8 @@
 # The _log function is used for everything these scripts want to log.
 # It will always log errors and warnings but can be silenced for other messages
 # by setting the JUPYTER_DOCKER_STACKS_QUIET environment variable.
+# Colorizes WARNING and ERROR output when stderr is a terminal.
+# Respects NO_COLOR (https://no-color.org/).
 _log () {
     local level="${1}"
     case "${level}" in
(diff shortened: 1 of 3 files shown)


goelakash:
The x86_64-tensorflow timeout is unrelated to this PR. The job hits the 15-minute timeout for tensorflow on Python 3.13 (all other tensorflow variants have 25 minutes). Opened #2465 to fix it.

images/docker-stacks-foundation/_docker_stacks_log.sh
```diff
@@ -5,6 +5,8 @@
 # The _log function is used for everything these scripts want to log.
 # It will always log errors and warnings but can be silenced for other messages
 # by setting the JUPYTER_DOCKER_STACKS_QUIET environment variable.
+# Colorizes WARNING and ERROR output when stderr is a terminal.
```
mathbunnyru:
```suggestion
# Colorizes ERROR, WARNING, and DEBUG output when stderr is a terminal.
```

images/docker-stacks-foundation/_docker_stacks_log.sh
```diff
@@ -16,7 +18,20 @@ _log () {
             ;;
     esac
     if [[ "${level}" == "ERROR" ]] || [[ "${level}" == "WARNING" ]] || [[ "${JUPYTER_DOCKER_STACKS_QUIET}" == "" ]]; then
-        echo "${level}[$(date '+%Y-%m-%d %H:%M:%S')] $*" >&2
+        local msg
+        msg="${level}[$(date '+%Y-%m-%d %H:%M:%S')] $*"
```
mathbunnyru, line 21: Can't we write this in a single line?

tests/by_image/docker-stacks-foundation/test_logging.py
```diff
@@ -66,3 +66,58 @@ def test_log_without_level_defaults_to_info(container: TrackedContainer) -> None
...
+        command=[
+            "bash",
+            "-c",
+            'source /usr/local/bin/_docker_stacks_log.sh && _log_error "test error"',
+        ],
+    )
+    assert "\x1b[0;31m" in logs
+    assert "ERROR[" in logs
+    assert "test error" in logs
+    assert "\x1b[0m" in logs
```
mathbunnyru: What do you think about making these strings variables at the top of this module and reusing them everywhere in this file?
It would practically remove the hardcoded text

mathbunnyru (changes requested):

tests/by_image/docker-stacks-foundation/test_logging.py
goelakash: that makes sense, but with unit-tests - the trade off is always DRY vs reproducibility. A minor change to those strings could lead to several tests failing - which usually is a bad test pattern from a maintainability perspective. What do you say?

mathbunnyru: I think they won't be changed too much, once they are merged, but a mistake in a variable name, for example,e will be caught by mypy early.

images/docker-stacks-foundation/_docker_stacks_log.sh
goelakash: I guess thats fine too. The split ensures that `$()` related issues can be isolated to that assignment line - but i agree that this is redundant in this case.

goelakash: @mathbunnyru I tried merging the two - but shellcheck flags it as `SC2155: local x=$(...) masks the exit status of the command substitution, so the local declaration always returns 0 even if $(...) failed`.

mathbunnyru: Wow, I didn't know that, let's keep as is

mathbunnyru (approved):

PR #2461, opened 2026-06-02
Add development setup guide for local pre-PR checks

### LINKED ISSUE #2460: Add a unified local pre-PR checklist to CONTRIBUTING.md
## Problem

There is no single place a contributor can go to find the complete sequence of local checks that mirror what CI runs. The steps are scattered across:

- `CONTRIBUTING.md` (says "run the linters" and links to docs site)
- `docs/contributing/lint.md` (pre-commit commands)
- `docs/contributing/tests.md` (build/test commands)

This means contributors have to read multiple documents and piece the workflow together themselves, leading to unnecessary round-trips with GitHub Actions (~10 minutes per push) to discover failures that could have been caught locally in seconds.

## Proposal

Add a concise "Before you submit" checklist to `CONTRIBUTING.md` (or a single doc page) that consolidates the full local workflow in one place:

```bash
# 1. Install dev dependencies
pip install --upgrade -r requirements-dev.txt

# 2. Run all linters (including mypy)
pre-commit run --all-files --hook-stage manual

# 3. Build the image you changed
make build/<image>

# 4. Run tests for that image
make test/<image>
```

This doesn't replace the detailed docs, it just gives contributors a quick-reference so they can run the full CI-equivalent check locally before pushing.

goelakash:
## Summary
- Adds a new `docs/contributing/dev-setup.md` page that consolidates all prerequisites and the full local check sequence into one place
- Covers: Docker, Python, Make prerequisites; one-time setup (pip install, pre-commit install); pre-PR checklist (lint, build, test); common examples; troubleshooting
- Adds the page to the Contributor Guide in `docs/index.rst` (listed first)
- Updates `CONTRIBUTING.md` to link to the new page

Fix #2460

### DIFF
--- a/docs/contributing/dev-setup.md
+++ b/docs/contributing/dev-setup.md
@@ -4,9 +4,9 @@ This page covers everything you need to run the full CI-equivalent checks locall
 
 ## Prerequisites
 
-- [Docker](https://docs.docker.com/get-docker/)
+- [Docker](https://docs.docker.com/get-started/get-docker/)
 - Python 3.12+
-- [GNU Make](https://www.gnu.org/software/make/)
+- GNU Make
 - Git
 
 ## One-time setup
--- a/docs/contributing/dev-setup.md
+++ b/docs/contributing/dev-setup.md
@@ -49,10 +49,12 @@ This means changes to a parent image won't be reflected unless you also rebuild 
 Replace `<image-name>` with the image you modified (e.g., `docker-stacks-foundation`, `base-notebook`, `scipy-notebook`).
 
 ```{note}
-`make test/<image-name>` runs the tests defined under `tests/by_image/<image-name>/`
-against the named image only. CI additionally re-runs that same test set against every
-downstream image, so a change in `docker-stacks-foundation` is verified across all images
-in CI even though locally you'd only run `make test/docker-stacks-foundation`.
+`make test/<image-name>` runs the tests for that image plus every parent image's
+tests against it (e.g., testing `scipy-notebook` also runs `base-notebook` and
+`docker-stacks-foundation` test files against the `scipy-notebook` image). CI
+additionally runs the same test set against every downstream image, so a change
+in `docker-stacks-foundation` is verified across all images in CI even though
+locally you'd only run `make test/docker-stacks-foundation`.
 ```
 
 ## Common examples
--- a/docs/contributing/dev-setup.md
+++ b/docs/contributing/dev-setup.md
@@ -12,10 +12,10 @@ This page covers everything you need to run the full CI-equivalent checks locall
 ## One-time setup
 
 ```bash
-# Clone the repository (HTTPS)
-git clone https://github.com/jupyter/docker-stacks.git
-# or via SSH
+# Clone the repository (SSH, recommended)
 git clone git@github.com:jupyter/docker-stacks.git
+# or via HTTPS
+git clone https://github.com/jupyter/docker-stacks.git
 
 cd docker-stacks
 
@@ -28,7 +28,7 @@ pre-commit install --install-hooks
 
 ## Pre-PR checklist
 
-Run these steps locally before pushing. They mirror what CI does.
+Run these steps locally before pushing.
 
 ```bash
 # 1. Run all linters including mypy
@@ -41,12 +41,18 @@ make build/<image-name>
 make test/<image-name>
 ```
 
+```{note}
+If the parent image isn't built locally, Docker pulls it from the registry.
+This means changes to a parent image won't be reflected unless you also rebuild it locally.
+```
+
 Replace `<image-name>` with the image you modified (e.g., `docker-stacks-foundation`, `base-notebook`, `scipy-notebook`).
 
 ```{note}
-Tests run against all images that inherit from the one you specify.
-If you modified `docker-stacks-foundation`, you only need to build and test that image,
-but your changes will also be tested against downstream images in CI.
+`make test/<image-name>` runs the tests defined under `tests/by_image/<image-name>/`
+against the named image only. CI additionally re-runs that same test set against every
+downstream image, so a change in `docker-stacks-foundation` is verified across all images
+in CI even though locally you'd only run `make test/docker-stacks-foundation`.
 ```
 
 ## Common examples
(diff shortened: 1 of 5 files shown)


docs/using/selecting.md
```diff
@@ -127,7 +127,7 @@ It contains:
   [rsqlite](https://cran.r-project.org/web/packages/RSQLite/index.html),
   [shiny](https://shiny.posit.co),
   [tidymodels](https://www.tidymodels.org/),
-  [unixodbc](https://www.unixodbc.org)
+  unixODBC
```
goelakash: @mathbunnyru this link doesn't load - causes sphinx doc build to fail (ran locally and faced the same issue). https://www.unixodbc.org/ is not reachable

docs/contributing/dev-setup.md
```diff
@@ -0,0 +1,65 @@
...
+- GNU Make
+- Git
+
+## One-time setup
+
+```bash
+# Clone the repository (HTTPS)
+git clone https://github.com/jupyter/docker-stacks.git
+# or via SSH
+git clone git@github.com:jupyter/docker-stacks.git
```
mathbunnyru, line 16: Let's make this one first, I think GitHub prefers ssh most of the time

docs/contributing/dev-setup.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Pre-PR checklist
+
+Run these steps locally before pushing. They mirror what CI does.
+
+```bash
+# 1. Run all linters including mypy
+pre-commit run --all-files --hook-stage manual
+
+# 2. Build the image you changed
```
mathbunnyru, line 37: We need to add a comment that if parent image is not available locally, it will be pulled from registry

docs/contributing/dev-setup.md
```diff
@@ -0,0 +1,65 @@
...
+make build/<image-name>
+
+# 3. Run tests for that image
+make test/<image-name>
+```
+
+Replace `<image-name>` with the image you modified (e.g., `docker-stacks-foundation`, `base-notebook`, `scipy-notebook`).
+
+```{note}
+Tests run against all images that inherit from the one you specify.
```
mathbunnyru: This is not clear, because it sounds like `make test/docker-stacks-foundation` will run `scipy-notebook` tests. Maybe we can formulate it differently?

docs/using/selecting.md
mathbunnyru: The link is correct, but sometimes this website is down.
Let's add it to ignore list of linkchecker in conf.py and not change this file

docs/contributing/dev-setup.md
```diff
@@ -0,0 +1,65 @@
...
+# Install Python development dependencies
+pip install --upgrade -r requirements-dev.txt
+
+# Install pre-commit hooks (runs linters automatically on git commit)
+pre-commit install --install-hooks
+```
+
+## Pre-PR checklist
+
+Run these steps locally before pushing. They mirror what CI does.
```
mathbunnyru: CI actually never uses Makefile. But, it mostly does the same thing (this is just a comment, probably no action required)

docs/contributing/dev-setup.md
```diff
@@ -0,0 +1,65 @@
...
+pre-commit run --all-files --hook-stage manual
+make build/docker-stacks-foundation
+make test/docker-stacks-foundation
+
+# Working on the base notebook image
+pre-commit run --all-files --hook-stage manual
+make build/docker-stacks-foundation
+make build/base-notebook
+make test/base-notebook
+```
```
mathbunnyru, line 78: Let's maybe mention `make build-all` and `make test-all`?

mathbunnyru (changes requested):

mathbunnyru:
@goelakash I also updated the description to say `Fix #2460`. I highly recomment using it, because it saves time for maintainers and makes the history more clear. There are some special words, which would close an issue when PR gets merged (addresses is not one of them: https://graphite.com/guides/closing-issues-with-pull-requests-in-github)

docs/using/selecting.md
goelakash: Cool. i didn't know that was possible - will look into it.

docs/contributing/dev-setup.md
goelakash: Ok, usually https is used by scripts because ssh clone may not be setup for them. But since these are dev docs - we can prioritize ssh here.

docs/contributing/dev-setup.md
goelakash: Ok, for latency concerns in local testing?

docs/contributing/dev-setup.md
goelakash: Let me look at it again - this message is basically assuming that the contributor understands the Dockerfile - but maybe that assumption is too casual.

docs/contributing/dev-setup.md
goelakash: Good call.

docs/contributing/dev-setup.md
mathbunnyru: Mostly because someone might change something in one of the parent images, and expect it to be magically propagated during building/testing

docs/contributing/dev-setup.md
goelakash: ok, i think thats fine, i'll remove the second statement as it adds minimal information.

goelakash:
@mathbunnyru thanks, i thought i did mention it, but thanks for adding the issue link if it was missing.

docs/contributing/dev-setup.md
```diff
@@ -0,0 +1,76 @@
...
+This means changes to a parent image won't be reflected unless you also rebuild it locally.
+```
+
+Replace `<image-name>` with the image you modified (e.g., `docker-stacks-foundation`, `base-notebook`, `scipy-notebook`).
+
+```{note}
+`make test/<image-name>` runs the tests defined under `tests/by_image/<image-name>/`
+against the named image only. CI additionally re-runs that same test set against every
+downstream image, so a change in `docker-stacks-foundation` is verified across all images
+in CI even though locally you'd only run `make test/docker-stacks-foundation`.
```
mathbunnyru: That's not true, even locally, `make test/<image-name>` runs parents test files

mathbunnyru (changes requested):

docs/contributing/dev-setup.md
goelakash: Corrected

mathbunnyru (approved):
Great job, thanks!

PR #2472, opened 2026-06-05
chore: improve docker-stacks maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in tests/conftest.py, tests/run_tests.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/tests/by_image/base-notebook/test_notebook.py
+++ b/tests/by_image/base-notebook/test_notebook.py
@@ -9,6 +9,7 @@ def test_secured_server(
     container: TrackedContainer, http_client: requests.Session, free_host_port: int
 ) -> None:
     """Jupyter Server should eventually request user login."""
+    assert isinstance(free_host_port, int) and 0 < free_host_port <= 65535
     container.run_detached(ports={"8888/tcp": free_host_port})
     resp = http_client.get(f"http://localhost:{free_host_port}")
     resp.raise_for_status()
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -69,7 +69,7 @@ def option(name: str) -> str:
 @pytest.fixture(scope="function")
 def container(
     docker_client: docker.DockerClient, image_name: str
-) -> Generator[TrackedContainer]:
+) -> Generator[TrackedContainer, None, None]:
```
goelakash, line 72: I think is an overloaded function - so this doesn't improve the current test. lmk

mathbunnyru: https://docs.python.org/3/library/typing.html#annotating-generators-and-coroutines

> The SendType and ReturnType parameters default to None

Agree with you, I don't think we need this change

tests/by_image/base-notebook/test_notebook.py
```diff
@@ -9,6 +9,7 @@ def test_secured_server(
     container: TrackedContainer, http_client: requests.Session, free_host_port: int
 ) -> None:
     """Jupyter Server should eventually request user login."""
+    assert isinstance(free_host_port, int) and 0 < free_host_port <= 65535
```
mathbunnyru, line 12: What's the point of checking if it's int? We already have type annotation.
And port is always at max 65535, also don't see the point

mathbunnyru:
I don't see a value which this PR brings, please, let me know if I missed something, closing for now

Issue #2471, opened 2026-06-05
Recipe request: JAX with CUDA


goelakash:
@mathbunnyru 
Would a JAX-with-CUDA recipe in `docs/using/recipes.md` be welcome? Same shape as the `xgboost` one.

JAX is widely used (~21M monthly PyPI downloads), and "JAX install with CUDA falls back to CPU" is a recurring issue, e.g. https://stackoverflow.com/q/79817557. The modern install path is simple, so the recipe would be small.

If yes, any preference on base image or CUDA version? Otherwise I'll mirror the existing recipes and JAX's upstream recommendation.

goelakash:
On CUDA version: JAX upstream now recommends CUDA 13. From https://docs.jax.dev/en/latest/installation.html:

> We recommend migrating to the CUDA 13 wheels; at some point in the future we will drop CUDA 12 support.

So I'd default the recipe to `jax[cuda13]` unless preferred otherwise.

mathbunnyru:
Yes, sure, feel free to submit a PR. You might have seen it already, but I have CI set up to build recipes too, so it should be easy to verify if it’s working or not

PR #2473, opened 2026-06-06
Add JAX with CUDA recipe

goelakash:
Issue: #2471.

Adds a small recipe to `docs/using/recipes.md` that installs JAX with CUDA 13 wheels into the base notebook image. Same shape as the `xgboost` recipe.


goelakash:
Closing in favor of a clean version with proper commit history.

PR #2474, opened 2026-06-06
Add JAX with CUDA recipe

### LINKED ISSUE #2471: Recipe request: JAX with CUDA

goelakash:
Closes #2471.

Adds a small recipe to `docs/using/recipes.md` that installs JAX with CUDA 13 wheels into the base notebook image. Same shape as the `xgboost` recipe.

### DIFF
--- a/docs/using/recipe_code/jax_cuda.dockerfile
+++ b/docs/using/recipe_code/jax_cuda.dockerfile
@@ -1,0 +1,6 @@
+ARG BASE_IMAGE=quay.io/jupyter/base-notebook
+FROM $BASE_IMAGE
+
+RUN pip install -v --no-cache-dir 'jax[cuda13]' && \
+    fix-permissions "${CONDA_DIR}" && \
+    fix-permissions "/home/${NB_USER}"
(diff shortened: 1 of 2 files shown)


mathbunnyru (changes requested):
Please update the description to use one of the special phrases, which would close an issue when PR gets merged: https://graphite.com/guides/closing-issues-with-pull-requests-in-github

`Issue #NUM` won't do anything with an issue on PR merge, `Closes #NUM` will close it when PR is merged.

docs/using/recipe_code/jax_cuda.dockerfile
```diff
@@ -0,0 +1,6 @@
+ARG BASE_IMAGE=quay.io/jupyter/base-notebook
```
mathbunnyru, line 1: Thanks for putting recipe code in a separate Dockerfile, it's now tested in CI, so I know this works, and we'll know if it stops 🙂

goelakash:
Nice, let me update it

mathbunnyru (approved):

PR #2484, opened 2026-06-21
ci: sign container images with cosign for supply-chain security

### LINKED ISSUE #2455: Sign container images with cosign for supply-chain security
The OpenSSF Scorecard flags the Signed-Releases check because published Docker images are not cryptographically signed.

Signing images with cosign (keyless, via GitHub OIDC) would allow consumers to verify image provenance (e.g. Chainguard Images uses `sigstore/cosign-installer` in their release workflow).

Related to #2428

theamitmehra:
## What

Signs all published Docker images with cosign (keyless, via GitHub OIDC) 
after they're pushed to Quay.io, addressing the OpenSSF Scorecard 
Signed-Releases check flagged in #2455.

## How it works

- Adds `id-token: write` permission for GitHub OIDC authentication
- Captures the image digest immediately after `docker push`
- Signs by digest (not tag) so the signature is tied to the immutable image content
- Uses `sigstore/cosign-installer` pinned to a specific commit hash

## Layer inheritance is unaffected

The `.sig` signature is attached as a separate OCI artifact on Quay.io —
the image and its layers are completely untouched, so the existing layer
sharing between stacked images (base-notebook → minimal-notebook → scipy-notebook)
works exactly as before.

## Verification

Once merged, users can verify any image with:
cosign verify quay.io/jupyter/base-notebook@sha256:... \
  --certificate-identity-regexp="https://github.com/jupyter/docker-stacks" \
  --certificate-oidc-issuer="https://token.actions.githubusercontent.com"

Note: Cosign is now installed on every run (including PRs) for better CI coverage. The actual signing step is still gated behind PUSH_TO_REGISTRY == 'true' so it only runs on main.

Fixes #2455

### DIFF
--- a/.github/workflows/docker-tag-merge.yml
+++ b/.github/workflows/docker-tag-merge.yml
@@ -68,7 +68,6 @@ jobs:
         id: login
 
       - name: Merge tags for the images 🔀
-        if: env.PUSH_TO_REGISTRY == 'true'
         run: |
           python3 -m tagging.apps.merge_tags \
             --image ${{ inputs.image }} \
@@ -77,7 +76,7 @@ jobs:
         shell: bash
 
       - name: Install Cosign 🔏
-        uses: sigstore/cosign-installer@v4.1.0
+        uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
 
       - name: Sign image with Cosign 🔏
         if: env.PUSH_TO_REGISTRY == 'true'
--- a/.github/workflows/docker-tag-merge.yml
+++ b/.github/workflows/docker-tag-merge.yml
@@ -32,6 +32,7 @@ jobs:
     runs-on: ubuntu-24.04
     permissions:
       contents: read
+      id-token: write
     timeout-minutes: ${{ inputs.timeout-minutes }}
 
     steps:
@@ -67,11 +68,28 @@ jobs:
         id: login
 
       - name: Merge tags for the images 🔀
+        if: env.PUSH_TO_REGISTRY == 'true'
+        id: merge
         run: |
           python3 -m tagging.apps.merge_tags \
             --image ${{ inputs.image }} \
             --variant ${{ inputs.variant }} \
             --tags-dir /tmp/jupyter/tags/
+          IMAGE_DIGEST=$(docker manifest inspect ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ inputs.image }} --verbose | python3 -c "import sys,json; data=json.load(sys.stdin); print(data[0]['Descriptor']['digest']) if isinstance(data, list) else print(data['Descriptor']['digest'])")
+          echo "digest=${IMAGE_DIGEST}" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: Install Cosign 🔏
+        if: env.PUSH_TO_REGISTRY == 'true'
+        uses: sigstore/cosign-installer@3454372f43399081ed03b604cb2d021dabca52bb # v3.8.2
+
+      - name: Sign image with Cosign 🔏
+        if: env.PUSH_TO_REGISTRY == 'true'
+        env:
+          DIGEST: ${{ steps.merge.outputs.digest }}
+        run: |
+          cosign sign --yes \
+            ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ inputs.image }}@${DIGEST}
         shell: bash
 
       - name: Logout from Registry 🔐
--- a/.github/workflows/docker-tag-merge.yml
+++ b/.github/workflows/docker-tag-merge.yml
@@ -69,27 +69,22 @@ jobs:
 
       - name: Merge tags for the images 🔀
         if: env.PUSH_TO_REGISTRY == 'true'
-        id: merge
         run: |
           python3 -m tagging.apps.merge_tags \
             --image ${{ inputs.image }} \
             --variant ${{ inputs.variant }} \
             --tags-dir /tmp/jupyter/tags/
-          IMAGE_DIGEST=$(docker manifest inspect ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ inputs.image }} --verbose | python3 -c "import sys,json; data=json.load(sys.stdin); print(data[0]['Descriptor']['digest']) if isinstance(data, list) else print(data['Descriptor']['digest'])")
-          echo "digest=${IMAGE_DIGEST}" >> $GITHUB_OUTPUT
         shell: bash
 
       - name: Install Cosign 🔏
-        if: env.PUSH_TO_REGISTRY == 'true'
-        uses: sigstore/cosign-installer@3454372f43399081ed03b604cb2d021dabca52bb # v3.8.2
+        uses: sigstore/cosign-installer@v4.1.0
 
       - name: Sign image with Cosign 🔏
         if: env.PUSH_TO_REGISTRY == 'true'
-        env:
-          DIGEST: ${{ steps.merge.outputs.digest }}
         run: |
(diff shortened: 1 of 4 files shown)


mathbunnyru (commented):
1. If I read the description correctly, this shouldn't be a problem: https://github.com/jupyter/docker-stacks/issues/2455#issuecomment-4670374806
2. When merging tags, do we also need to sign? https://github.com/jupyter/docker-stacks/blob/main/.github/workflows/docker-tag-push-merge.yml

.github/workflows/docker-tag-push.yml
```diff
@@ -69,9 +70,25 @@ jobs:
...
+        id: push
         run: |
           docker push --all-tags ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ inputs.image }} || \
           docker push --all-tags ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ inputs.image }}
+          IMAGE_DIGEST=$(docker inspect --format='{{index .RepoDigests 0}}' ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ inputs.image }} | cut -d'@' -f2)
+          echo "digest=${IMAGE_DIGEST}" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: Install Cosign 🔏
+        if: env.PUSH_TO_REGISTRY == 'true'
```
mathbunnyru: Also, I think nothing is preventing you from installing this all the time - it's better to do everything possible in PR

.github/workflows/docker-tag-push.yml
```diff
@@ -69,9 +70,25 @@ jobs:
 
       - name: Push single platform images to Registry 📤
         if: env.PUSH_TO_REGISTRY == 'true'
+        id: push
         run: |
           docker push --all-tags ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ inputs.image }} || \
           docker push --all-tags ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ inputs.image }}
+          IMAGE_DIGEST=$(docker inspect --format='{{index .RepoDigests 0}}' ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ inputs.image }} | cut -d'@' -f2)
```
mathbunnyru: We only need this digest in the sign step, maybe let's calculate it there?
The this step won't need to be changed

mathbunnyru (changes requested):

.github/workflows/docker-tag-push.yml
```diff
@@ -69,9 +70,25 @@ jobs:
...
+          echo "digest=${IMAGE_DIGEST}" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: Install Cosign 🔏
+        if: env.PUSH_TO_REGISTRY == 'true'
+        uses: sigstore/cosign-installer@3454372f43399081ed03b604cb2d021dabca52bb # v3.8.2
```
mathbunnyru: Please, use latest version

mathbunnyru:
All my comments apply to both files

theamitmehra:
yes, i am on it. updating my PR

theamitmehra:
Done! Updated both files in the latest commit. installed cosign unconditionally, moved digest calculation into the sign step, and updated to v4.1.0. Applied the same changes to docker-tag-merge.yml as well. kindly check

.github/workflows/docker-tag-merge.yml
```diff
@@ -67,13 +68,25 @@ jobs:
         id: login
 
       - name: Merge tags for the images 🔀
+        if: env.PUSH_TO_REGISTRY == 'true'
```
mathbunnyru: This script handles this internally, so adding it here makes it less checked on PRs, so please, remove this change

.github/workflows/docker-tag-merge.yml
```diff
@@ -67,13 +68,25 @@ jobs:
...
         run: |
           python3 -m tagging.apps.merge_tags \
             --image ${{ inputs.image }} \
             --variant ${{ inputs.variant }} \
             --tags-dir /tmp/jupyter/tags/
         shell: bash
 
+      - name: Install Cosign 🔏
+        uses: sigstore/cosign-installer@v4.1.0
```
mathbunnyru: Please, use the hash-pinned versions of actions, because using an action like this is actually a security issue (in case the used action gets compromised)

(This also needs to be fixed below as well)

Also, as I asked before, please use the latest version, it's 4.1.2 now

mathbunnyru (changes requested):

.github/workflows/docker-tag-merge.yml
theamitmehra: Done, I have addressed all review comments. Kindly check

mathbunnyru (dismissed):
LGTM, thanks 👍

mathbunnyru:
When I approved the workflow to run, I got a notification from GitHub:
```
The workflow is not valid. jupyter/docker-stacks/.github/workflows/docker-tag-push-merge.yml@ec2b9cc666fba987e3efe595acf03596a2b2a064 (Line: 23, Col: 3): Error calling workflow 'jupyter/docker-stacks/.github/workflows/docker-tag-push.yml@ec2b9cc666fba987e3efe595acf03596a2b2a064'. The nested job 'tag-push' is requesting 'id-token: write', but is only allowed 'id-token: none'.
```

You need to carefully pass the required permissions

theamitmehra:
Hi @mathbunnyru, I understand the issue. `id-token: write` needs to be granted by the caller workflow (`docker-tag-push-merge.yml`) for the signing to work in the called workflows.

I see two places that need updating:
1. The top-level `permissions: {}` in `docker-tag-push-merge.yml` needs `id-token: write` added
2. Both `tag-push` and `tag-merge` jobs need `id-token: write` added to their permissions

Is this the right approach, or would you prefer a different way to handle this?

mathbunnyru:
> Hi @mathbunnyru, I understand the issue. `id-token: write` needs to be granted by the caller workflow (`docker-tag-push-merge.yml`) for the signing to work in the called workflows.
> 
> I see two places that need updating:
> 
> 1. The top-level `permissions: {}` in `docker-tag-push-merge.yml` needs `id-token: write` added
> 2. Both `tag-push` and `tag-merge` jobs need `id-token: write` added to their permissions
> 
> Is this the right approach, or would you prefer a different way to handle this?

1. is not needed - we always keep it empty on the workflow level.
2. is correct, but you need to do it on the job level

Also, closing and reopening PR won't work, because you're first time contributor, so it requires me to approve running the workflows.
So, please, push a fix, and I'll run workflows.

theamitmehra:
I've added `id-token: write` at the job level for both `tag-push` and `tag-merge` in `docker-tag-push-merge.yml`, keeping `permissions: {}` at the workflow level unchanged. Could you please approve the workflow run?

.github/workflows/docker-tag-push-merge.yml
mathbunnyru, line 4: This is also a callable worklfow, so you need to pass this permission all to the top of where it's being called from

theamitmehra: ok got it

theamitmehra: I have added `id-token:` write to the caller workflows (tag-push-merge and tag-push-merge-fast) as suggested.

mathbunnyru: Seems to work now, thanks. I'll merge when CI passes.

mathbunnyru (approved):

mathbunnyru:
@theamitmehra, the `docker inspect` fails.
The reason is that we remove "latest" tag, when pushing single-platform image.
This is correct, because latest should mean "multi-arch".

But you're trying to use it for cosign, and it fails
Could you please send a PR with a fix?

Also, please set up the build configuration in your fork, so we're sure it works.
https://github.com/jupyter/docker-stacks/actions/runs/27990857759/job/82847190754

theamitmehra:
Hi @mathbunnyru, I've set up the build in my fork and opened a fix PR here: https://github.com/jupyter/docker-stacks/pull/2494

The CI run passed successfully with no cosign errors: https://github.com/jupyter/docker-stacks/actions/runs/28002419059

All 42 tag-push-merge jobs completed and 140 artifacts were produced.

Issue #2486, opened 2026-06-22
Build reliability: flaky docker.yml failures on main


goelakash:
Several `docker.yml` runs on `main` fail for reasons unrelated to the change under test. Actions to minimize these flaky build failures:

- [x] Sleep 10s between registry login retries (#2479)
- [x] Sleep 10s between registry push retries (#2485). `docker push` retries back-to-back (~0.25s apart) with no backoff, so a brief quay.io blip fails both attempts. Seen in [docker-stacks-foundation tag-push, 2026-06-01](https://github.com/jupyter/docker-stacks/actions/runs/26743436702) and [pytorch-notebook/cuda12 tag-push, scheduled 2026-06-08](https://github.com/jupyter/docker-stacks/actions/runs/27116151552)
- [ ] Root-cause the sparklyr/dbplyr `[PARSE_SYNTAX_ERROR] ... at or near '"id"'` failure in `test_spark_r_nbconvert` instead of relying on flaky-retry. See [all-spark build-test-upload, 2026-06-20](https://github.com/jupyter/docker-stacks/actions/runs/27866515947)
- [x] Investigate the `10activate-conda-env.sh` startup stall that fails `test_start_container` in `all-spark` (container enters `start.sh`, runs hooks, never reaches "Executing: jupyter lab"). See [scheduled 2026-05-25](https://github.com/jupyter/docker-stacks/actions/runs/26383287219), [scheduled 2026-05-18](https://github.com/jupyter/docker-stacks/actions/runs/26013548599)

mathbunnyru:
I haven't seen any flakiness lately.
@goelakash, I appreciate you working on this.

If something new arises, that would probably be single problems; let's create separate issues, because I think the most common reasons for failures have been fixed.

PR #2487, opened 2026-06-22
Enable double-quoted identifiers in sparklyr test notebook

goelakash:
Part of #2486

The `local_sparklyr` test notebook runs `sdf_len(sc, 100) %>% spark_apply(...)`. To execute that pipeline, sparklyr/dbplyr generate Spark SQL that refers to the `id` column `sdf_len` creates, quoting it as a double-quoted identifier (`"id"`).

This worked under Spark 3, but the all-spark image installs the latest Spark when the version is unpinned, which is now Spark 4. Spark 4 enables ANSI mode by default, and `spark.sql.ansi.doubleQuotedIdentifiers` defaults to `false`, so Spark 4 reads `"id"` as a string literal rather than a column name and rejects the query:

```
[PARSE_SYNTAX_ERROR] Syntax error at or near '"id"'. SQLSTATE: 42601
```

The test fails on a few runs (it exhausts its flaky reruns, then fails), so it is a real Spark-4 incompatibility, not flakiness.

Fix: set `spark.sql.ansi.doubleQuotedIdentifiers = true` in the notebook's `spark_config()` so Spark again reads the double-quoted names sparklyr emits as identifiers.

References:
- All-spark installs the latest Spark when unpinned: https://github.com/jupyter/docker-stacks/blob/91aceb0b50cd35c8c914ce6cb56ed33eb10abf9a/images/pyspark-notebook/Dockerfile#L27
- Spark 4 enables ANSI mode by default (Spark SQL migration guide, "Upgrading from Spark SQL 3.5 to 4.0"): https://spark.apache.org/docs/latest/sql-migration-guide.html#upgrading-from-spark-sql-35-to-40
- `spark.sql.ansi.doubleQuotedIdentifiers` (default `false`; true reads `"..."` as identifiers, false as string literals): https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L5301
- Failing run: https://github.com/jupyter/docker-stacks/actions/runs/27866515947

Reproduced locally on pyspark 4.1.2 (exact error) and confirmed the config fixes it.

### DIFF
--- a/tests/by_image/all-spark-notebook/data/local_sparklyr.ipynb
+++ b/tests/by_image/all-spark-notebook/data/local_sparklyr.ipynb
@@ -12,6 +12,8 @@
     "conf <- spark_config()\n",
     "# Set the catalog implementation in-memory\n",
     "conf$spark.sql.catalogImplementation <- \"in-memory\"\n",
+    "# Spark 4 reads double-quoted names as string literals by default, breaking sparklyr-generated SQL\n",
+    "conf$spark.sql.ansi.doubleQuotedIdentifiers <- \"true\"\n",
     "\n",
     "# Spark session & context\n",
     "sc <- spark_connect(master = \"local\", config = conf)\n",


mathbunnyru (commented):
You say that the test fails on every run, but that's not true

Here is a very recent run, and it worked fine:
https://github.com/jupyter/docker-stacks/actions/runs/27910857330

If it was failing every time, I would have to fix it immediately, and not ignore it for a while 😂 

Could you please investigate a little bit more?

goelakash:
Sorry. that was a typo. Its sporadic - but let me take a look again.

mathbunnyru:
> Sorry. that was a typo. Its sporadic - but let me take a look again.

@goelakash any luck here? 🙂

goelakash:
Hey @mathbunnyru let me look into this later, currently busy on another project. Thanks for the reminder though!

mathbunnyru:
(verified using Claude, but this looks very reasonable to me)

I dug into the original failure ([this run](https://github.com/jupyter/docker-stacks/actions/runs/27866515947)): the error didn't come from the notebook's SQL, but from dbplyr's schema probe (`SELECT 0L AS "id"`) during `sdf_len()`. dbplyr 2.6.0 changed its Hive backend to quote identifiers with ANSI double quotes, which broke sparklyr <= 1.9.4 — the failing build installed exactly `r-sparklyr 1.9.4` + `r-dbplyr 2.6.0`. So it wasn't Spark 4 ANSI mode: it failed deterministically (on both architectures) only during the window before the fix reached conda-forge.

This was fixed upstream in [sparklyr 1.9.5](https://github.com/sparklyr/sparklyr/blob/main/NEWS.md#sparklyr-195):

> Restores compatibility with dbplyr (>= 2.6.0), which changed the Hive backend to quote identifiers with ANSI double-quotes

Latest images already ship sparklyr 1.9.5 — the test passes without this workaround (I also reran the notebook pipeline many times on the latest image and couldn't reproduce), and all scheduled builds since 2026-06-29 are green.

PR #2488, opened 2026-06-22
Fix flaky test_start_notebook log race with a wait_until helper

goelakash:
Part of #2486

Fixes the flaky `test_start_notebook` failure on `all-spark`, where the log check failed even though the server had started (the HTTP 200 assert before it passed). `get_logs()` reads stdout once, and the container's buffered output can flush just after the server is reachable, so the read misses `Executing: jupyter lab`.

Now it waits for the line instead of reading once, via a new `wait_until` helper in `tests/utils/wait.py` (also reused in `test_healthcheck`).

Failing runs: [1](https://github.com/jupyter/docker-stacks/actions/runs/26383287219), [2](https://github.com/jupyter/docker-stacks/actions/runs/26013548599)

### DIFF
--- a/tests/by_image/base-notebook/test_healthcheck.py
+++ b/tests/by_image/base-notebook/test_healthcheck.py
@@ -1,11 +1,11 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import logging
-import time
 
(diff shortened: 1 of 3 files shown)


mathbunnyru (approved):
Nice, thanks!

And I appreciate adding a common functionality to a new small utility

PR #2496, opened 2026-06-25
Use shfmt instead of bashate

mathbunnyru:
## Describe your changes

`bashate` seems to be abandoned, and it also doesn't auto-format files, so they end up being slightly different.

`sh` with very little setup formats existing files nicely


mathbunnyru:
@consideRatio do you mind approving the PR? 🙂 
I saw your like, but would be great to see explicit approval

consideRatio (approved):
Absolutely!

PR #2530, opened 2026-07-28
Fix GitHub redirect links

mathbunnyru:


mathbunnyru:
Locally, it works cleanly. There are some rate-limited requests, but I'm not sure we can solve those without a quality drop. Merging this, as it fixes current issues.

PR #2531, opened 2026-07-28
Images review fixes

mathbunnyru:
Was done using Claude (and by leveraging https://claude.com/contact-sales/claude-for-oss)

### DIFF
--- a/images/docker-stacks-foundation/Dockerfile
+++ b/images/docker-stacks-foundation/Dockerfile
@@ -88,11 +88,12 @@ RUN echo "auth requisite pam_deny.so" >> /etc/pam.d/su && \
     fix-permissions "${CONDA_DIR}" && \
     fix-permissions "/home/${NB_USER}"
 
-# macOS Rosetta virtualization creates junk directory which gets owned by root further up.
-# It'll get re-created, but as USER runner after the next directive so hopefully should not cause permission issues.
-#
-# More info: https://github.com/jupyter/docker-stacks/issues/2296
-RUN rm -rf "/home/${NB_USER}/.cache/"
+# Pre-create the cache dir owned by the user, macOS Rosetta junk can then only appear inside it
+RUN rm -rf "/home/${NB_USER}/.cache/" && \
+    mkdir -p "/home/${NB_USER}/.cache" && \
+    chown "${NB_USER}:${NB_GID}" "/home/${NB_USER}/.cache"
+
+RUN rm -rf "/home/${NB_USER}/.cache/rosetta"
 
 USER ${NB_UID}
 
@@ -149,11 +150,7 @@ RUN mkdir /usr/local/bin/start-notebook.d && \
 
 COPY 10activate-conda-env.sh /usr/local/bin/before-notebook.d/
 
-# macOS Rosetta virtualization creates junk directory which gets owned by root further up.
-# It'll get re-created, but as USER runner after the next directive so hopefully should not cause permission issues.
-#
-# More info: https://github.com/jupyter/docker-stacks/issues/2296
-RUN rm -rf "/home/${NB_USER}/.cache/"
+RUN rm -rf "/home/${NB_USER}/.cache/rosetta"
 
 # Switch back to jovyan to avoid accidental container runs as root
 USER ${NB_UID}
--- a/images/pyspark-notebook/setup_spark.py
+++ b/images/pyspark-notebook/setup_spark.py
@@ -42,16 +42,12 @@ def get_latest_spark_version() -> str:
     LOGGER.info(f"Available versions: {versions}")
 
     # Compare versions semantically
-    def version_array(ver: str) -> tuple[int, int, int, str]:
-        # 3.5.3 -> [3, 5, 3, ""]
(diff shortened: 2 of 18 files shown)


images/tensorflow-notebook/cuda/Dockerfile
```diff
@@ -42,4 +43,4 @@ ENV NVIDIA_VISIBLE_DEVICES="all" \
 # Puts the nvidia-smi binary (system management interface) on path
 # with associated library files to execute it
 ENV PATH="${PATH}:/usr/local/nvidia/bin" \
-    LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/usr/local/nvidia/lib64"
+    LD_LIBRARY_PATH="${LD_LIBRARY_PATH:+${LD_LIBRARY_PATH}:}/usr/local/nvidia/lib64"
```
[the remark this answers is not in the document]
mathbunnyru: Seems to work fine

images/pyspark-notebook/setup_spark.py
```diff
@@ -42,20 +42,34 @@ def get_latest_spark_version() -> str:
...
+    taken from the dev/infra/Dockerfile of the Spark release
+    """
+    url = (
+        "https://raw.githubusercontent.com/apache/spark/"
+        f"v{spark_version}/dev/infra/Dockerfile"
+    )
+    LOGGER.info(f"Downloading pandas version from: {url}")
+    resp = requests.get(url, timeout=60)
+    resp.raise_for_status()
+    if match := re.search(r"pandas==([\d.]+)", resp.text):
```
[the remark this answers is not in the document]
mathbunnyru: Let's see how it works, should be fine

mathbunnyru:
@consideRatio one more PR for you 🙂

images/r-notebook/Dockerfile
```diff
@@ -24,10 +24,6 @@ RUN apt-get update --yes && \
     gcc && \
     apt-get clean && rm -rf /var/lib/apt/lists/*
 
-# macOS Rosetta virtualization creates junk directory which gets owned by root further up.
-# It'll get re-created, but as USER runner after the next directive so hopefully should not cause permission issues.
-#
-# More info: https://github.com/jupyter/docker-stacks/issues/2296
 RUN rm -rf "/home/${NB_USER}/.cache/"
```
consideRatio: Rm .cache is better than a specific subgolder, as otherwise the .cache folder may stay around - potentially created and owned by the wrong user in the future etc also.

This PR mixes back and forth between .cache and .cache/rosetta

mathbunnyru: We have to do it, because in the images below scipy-notebook, there is no .cache dir, and we have to remove wronly-created root-owned dir created by rosetta.
But since we populate the matplotlib cache, we should only delete the rosetta subdir

mathbunnyru: Thinking about it more, we can probably create the .cache dir in the foundation image, it shouldn't be then re-created by rosetta, and we'll only have to remove rosetta dir.

mathbunnyru: @consideRatio I think it worked, please, take a look.
It won't be automatically tested in CI, because it requires building x86_64 images under an Apple Silicon Mac, but I tested locally.
If all is good, please approve, I will rebase-merge.

mathbunnyru: I also verified that this rosetta subdirectory is created both under root and non-root executions.
So, removing it at the end of each section is correct.

images/docker-stacks-foundation/Dockerfile
```diff
@@ -88,11 +88,12 @@ RUN echo "auth requisite pam_deny.so" >> /etc/pam.d/su && \
...
-# It'll get re-created, but as USER runner after the next directive so hopefully should not cause permission issues.
-#
-# More info: https://github.com/jupyter/docker-stacks/issues/2296
-RUN rm -rf "/home/${NB_USER}/.cache/"
+# Pre-create the cache dir owned by the user, macOS Rosetta junk can then only appear inside it
+RUN rm -rf "/home/${NB_USER}/.cache/" && \
+    mkdir -p "/home/${NB_USER}/.cache" && \
+    chown "${NB_USER}:${NB_GID}" "/home/${NB_USER}/.cache"
+
+RUN rm -rf "/home/${NB_USER}/.cache/rosetta"
```
consideRatio, line 96: Overall, when the removal is its own dedicated step, there will be a layer crested, which is then removed - the image size grows to reflect it.

If the remval is done in the same step that created content, it wont expand the image size -- it never became part of s layer

Is this a "final cleanup" of all previous steps, it makes sense, but ideally they are cleaned up in each step creating unwanted content

consideRatio (approved):

images/docker-stacks-foundation/Dockerfile
mathbunnyru: @consideRatio

1. I think you forgot that it only happens when you build x86_64 images under Apple Silicon Mac. So, this doesn't make any difference for images we push, only when someone does it locally
2. I checked, and `rosetta` is actually an empty dir.

So, I think we should prefer simplicity (only the final cleanup)

PR #2534, opened 2026-07-31
Sign pushed images with cosign

### LINKED ISSUE #2455: Sign container images with cosign for supply-chain security

mathbunnyru:
The digest to sign is calculated by a new tagging app, which resolves it from the first tag in the tags file, because the image can't be referenced by bare name: the "latest" tag is deliberately removed before pushing single-platform images.

All tags of an image point to the same digest, so a single signature covers all of them.
The digest calculation runs in PRs as well, tolerating images missing from the registry,
and only the signing itself is gated behind PUSH_TO_REGISTRY.

I also merged these commits to my fork and verified my fork images:
```
cosign verify quay.io/mathbunnyru/base-notebook:latest \
    --certificate-identity-regexp='^https://github\.com/mathbunnyru/docker-stacks/' \
    --certificate-oidc-issuer=https://token.actions.githubusercontent.com

Verification for quay.io/mathbunnyru/base-notebook:latest --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - The code-signing certificate was verified using trusted certificate authority certificates

[{"critical":{"identity":{"docker-reference":"quay.io/mathbunnyru/base-notebook:latest"},"image":{"docker-manifest-digest":"sha256:750fd765951630e08ed8206116f94917cdb461b60da070d12a8f8b2091b0c183"},"type":"https://sigstore.dev/cosign/sign/v1"},"optional":{}}]
```

Fixes #2455

### DIFF
--- a/.github/workflows/docker-tag-merge.yml
+++ b/.github/workflows/docker-tag-merge.yml
@@ -90,9 +90,10 @@ jobs:
           echo "IMAGE_REF=${IMAGE_REF}" >> "${GITHUB_ENV}"
         shell: bash
 
+      # The bundle format stores the signature using OCI 1.1 referrers and requires cosign v3+ to verify
       - name: Sign multi-platform image with Cosign 🔏
         if: env.PUSH_TO_REGISTRY == 'true'
-        run: cosign sign --yes "${IMAGE_REF}"
+        run: cosign sign --yes --new-bundle-format=true "${IMAGE_REF}"
         shell: bash
 
       - name: Logout from Registry 🔐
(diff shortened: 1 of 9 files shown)


mathbunnyru:
@consideRatio could you please take a look here?

mathbunnyru:
I also verified that the layer-inheritance property is intact on the signed images - meaning if image A is inherited from B, and B was pulled, only new layers of A will be needed to pull.

